PR #50, opened 2019-10-22
Fix minor README file extension inconsistency

adamaltman:
When the default configuration file extension was renamed from yml to yaml
it wasn't renamed through the entire README.


knidarkness:
Thanks, merged into master.

PR #55, opened 2019-10-24
Types extension

knidarkness:
This PR introduces a concept of extending and redefining types and rules.

**Description**

From this version, user can add custom fields into default objects in all levels, which will then validate using either existing validators (for existing types, either with custom rules, created by user).

To extend or define a type, user should create a file and pass it as a `typeExtension` parameter in the config. It should follow such pattern (quite similar to usual reducers):

```js
module.exports = (types) => ({
  ...types,
  OpenAPIParameter: {
    ...(types.OpenAPIParameter && types.OpenAPIParameter),
    properties: {
      ...(types.OpenAPIParameter && types.OpenAPIParameter.properties),
      allOf: null,
    },
    resolveType: (node) => (node.allOf ? 'OpenAPIParameterWithAllOf' : 'OpenAPIParameter'),
  },
  OpenAPIParameterWithAllOf: {
    name: 'OpenAPIParameterWithAllOf',
    properties: {
      allOf: 'OpenAPIParameter',
    },
  },
});

``` 
Here, user adds field `allOf` to `Parameter` type.

`null` value in the `properties` object is allowed and means that given property is a leaf in the type-tree.

To extend ruleset user can create another file and set it as `customRules` param of config with such contents:

```js
class ValidateOpenAPIParameterWithAllOf {
  static get rule() {
    return 'parameterWithAllOf';
  }

  OpenAPIParameter() {
    return {
      onEnter: (node, definition, ctx) => {
        const result = [];
        let validators = {};

        if (Object.keys(node).length === 1) {
          validators = {
            description: ctx.getRule('oas3-schema/parameter').validators.description,
          };
        } else {
          validators = {
            ...ctx.getRule('oas3-schema/parameter').validators,
            description: null,
          };
        }

        const fieldMessages = ctx.validateFieldsHelper(validators);
        result.push(...fieldMessages);

        return result;
      },
    };
  }
}

module.exports = [
  ValidateOpenAPIParameterWithAllOf,
];
```


src/resolver.js
```diff
@@ -10,6 +10,9 @@ import { isUrl } from './utils';
  * for it. If failed (e.g. because of undefined value) -- return null, to indicate that such
  * reference does not exist.
  *
+ * TODO: we might need a feature to support validation of "URL" based definitions if future, so
```
adamaltman:
```suggestion
 * TODO: we might need a feature to support validation of "URL" based definitions in the future, so
```

src/loader/index.js
```diff
@@ -47,13 +52,36 @@ function loadRuleset(config) {
...
     } else {
-      const ruleInstance = new Rule();
+      const ruleInstance = new Rule({});
       ruleSet.push(ruleInstance);
+      // console.log(ruleInstance.validators ? ruleInstance.validators() : null);
     }
   });
 
+  const ruleRedefinitiojnsPath = getRuleRedefinePath(config);
+  const ruleRedefs = require(ruleRedefinitiojnsPath);
```
RomanHotsiy:
```suggestion
  const ruleRedefs = require(ruleRedefinitionsPath);
```

src/loader/index.js
```diff
@@ -47,13 +52,36 @@ function loadRuleset(config) {
...
       }
     } else {
-      const ruleInstance = new Rule();
+      const ruleInstance = new Rule({});
       ruleSet.push(ruleInstance);
+      // console.log(ruleInstance.validators ? ruleInstance.validators() : null);
     }
   });
 
+  const ruleRedefinitiojnsPath = getRuleRedefinePath(config);
```
RomanHotsiy:
```suggestion
  const ruleRedefinitionsPath = getRuleRedefinePath(config);
```

src/typesExtention/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const extension = ctx.config.typeExtension[definitionName];
+  if (!extension) {
+    return definition;
+  }
+  const alteredFields = Object.keys(extension.properties);
+
+  for (let i = 0; i < alteredFields.length; i++) {
+    const pathToDefaultType = `${__dirname}/../types/${extension.properties[alteredFields[i]]}.js`;
+    let alteredType;
```
RomanHotsiy: I think you can assign the value directly on this line

src/visitors/rules/custom/parameterAllOf.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable class-methods-use-this */
+
+class parameterAllOf {
+  constructor(config) {
+    this.config = { ...config };
+    switch (this.config.level) {
```
RomanHotsiy: Do we need every custom rule to add this code? Maybe we can provide some base class with this behaviour handled. Also, where is this level used later? Is it propagated down the visitor invocations?

Also, I thing we should export constants for those 1, 2, 3, 4 numbers :)

src/visitors/rules/custom/parameterAllOf.js
```diff
@@ -0,0 +1,41 @@
...
+        break;
+      case 'error':
+      default:
+        this.config.level = 4;
+        break;
+    }
+  }
+
+  get rule() {
+    return 'parameterAllOf';
```
RomanHotsiy: Maybe use static field instead of a getter everywhere? Is there any reason go use getter?

src/visitors/rules/custom/parameterAllOf.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  OpenAPIParameter() {
+    return {
+      onEnter: (node, _, ctx, unresolved, traverseTools) => {
```
RomanHotsiy: how to add an error from this validator? Could you add an example of validating the description field to be string (using custom code)

src/visitors/rules/custom/parameterAllOf.js
knidarkness: no reason, was used initially to get it dynamically based on path, now completely redundant

RomanHotsiy: ok, let's update it via separate PR later

src/typesExtention/index.js
knidarkness: agree

src/visitors/rules/custom/parameterAllOf.js
knidarkness: I wanted to make custom rules completely 0-dependency. We can just inherit from AbstractVisitor  class and don't do this manually

src/visitors/rules/custom/parameterAllOf.js
RomanHotsiy: https://github.com/Redocly/openapi-cli/issues/62

src/visitors/rules/custom/parameterAllOf.js
knidarkness: Also, numbers are in constants elsewhere

RomanHotsiy: Hmm. I think 0-dep makes sense. Maybe, instead we can just keep string values and adjust other related code to accept them instead of/ additionally to number values?

src/visitors/rules/custom/parameterAllOf.js
knidarkness: for now we allow types (re)definition, not so much custom validators. we can (not sure if already) load additional rules (with are the same syntax/structure) as our own ones and run them on types we want

src/visitors/rules/custom/parameterAllOf.js
knidarkness: Hmmm, yes, that might work fine

PR #58, opened 2019-10-29
docs: Improve the docs about config options

adamaltman:
This has more detailed examples of configuration options.

It is pending the change/rename to `excludedPaths` for one config option.  Merge after that change is ready.


knidarkness:
Thank you, Adam!

zeke:
Nice!

PR #60, opened 2019-11-04
feat: add minor semantic rules

knidarkness:


knidarkness:
@adamaltman, no, just usual rules.

zeke:
👍

PR #65, opened 2019-12-13
feat: allow running custom rules before default ones

knidarkness:
This PR allows running mutating visitors before any checks are run.

To do so, you can set the `transformers` parameter in the config in the following manner:

```yaml
transformers: ./customTransformers.js
```

Format of the file is identical to `customRules`. See https://github.com/Redocly/openapi-cli/pull/65/files#diff-10ed1367325550f13d1c4e4424f3b527


adamaltman:
This is missing any corresponding changes to the README.  Also, I'm unclear about if someone were to want to run one of the system visitors early (for example, operation description).  How does configuring the executionOrder work in conjunction with the off/on and levels? 

Another question is why not be able to specify the concrete execution order?

```
rules:
  overlaysMerger:
    executionOrder: early
  mergeChecker:
    executionOrder: late
```
What happens when you want more than one early or late and you want to control that order?

RomanHotsiy:
@adamaltman this is not final API. We're trying different variants.

PR #71, opened 2020-01-02
fix: checkPathParamExists rule

rkrauskopf:
Rule was assuming that all of our paths were stored in a dereferenced openapi file rather than being a $ref. This should cover both cases now.

### DIFF
--- a/definitions/syntetic/syntetic-1.yaml
+++ b/definitions/syntetic/syntetic-1.yaml
@@ -39,6 +39,24 @@ paths:
             application/json:
               schema:
                 type: object
+  '/user/{id}':
+    get:
+      parameters:
+        - in: path
+          name: test
+          description: User id
+          required: true
+          schema:
+            type: string
+      operationId: withPathParam
+      description: Get user by id
+      responses:
+        '200':
+          description: example description
+          content:
+            application/json:
+              schema:
+                type: object
 externalDocs:
   description:
     $ref: inc/docs-description.md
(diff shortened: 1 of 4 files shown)


rkrauskopf:
@RomanHotsiy, I made some updates to this rule based on recommendations from @jamesmessinger. One of the assumptions that we made was that the visitedNodes will always have the OpenAPI object as the root. Please let me know if that won't always be the case and we'll rethink our fix.

RomanHotsiy:
Hey @rkrauskopf,

Thanks for the PR!

Yes. I see what's the issue is. Your assumption is correct.

I would though simplify the whole rule! This is guaranteed that IF the first item of `visitedNodes` is `"paths"` then the second item is the actual path. There is no need to looks for path through all the stack.

Would you be open to update this PR? We can do it from our side too if you don't have time. Just let us know.

rkrauskopf:
@RomanHotsiy I made the update based on your recommendations. Let me know if that's what you had in mind and feel free to make any additional changes you feel are necessary.

RomanHotsiy:
@rkrauskopf we will update this PR a bit next week. We didn't manage to squeeze it into todays release.

rkrauskopf:
Thank you!

RomanHotsiy:
Updated it and released in `1.8.1` (@rkrauskopf, not that the config format has changed in `0.8.0`. Old config format is still supported though. To migrate you need to move all your config under `lint` key, see example here https://github.com/Redocly/openapi-cli#rules

PR #79, opened 2020-01-06
fix: support transitive refs

RomanHotsiy:
You can check commit by commit.

Notable changes:

- Support transitive refs
- Fixed URL paths resolution

@knidarkness, please, review carefully and test a bit if you can! Thanks!

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -12,7 +12,7 @@ const urlPattern = new RegExp('^(https?:\\/\\/)?' // protocol
 + '(\\?[;&a-z\\d%_.~+=-]*)?' // query string
 + '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
 /* eslint-disable import/prefer-default-export */
-const globalUrlPattern = new RegExp('^(https?:\\/\\/)' // protocol
+const fullyQualifiedUrlPattern = new RegExp('^(https?:\\/\\/)' // protocol
 + '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' // domain name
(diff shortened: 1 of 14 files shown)


src/utils/index.js
```diff
@@ -11,8 +11,16 @@ const urlPattern = new RegExp('^(https?:\\/\\/)?' // protocol
...
+/* eslint-disable import/prefer-default-export */
+const globalUrlPattern = new RegExp('^(https?:\\/\\/)' // protocol
++ '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' // domain name
++ '((\\d{1,3}\\.){3}\\d{1,3}))' // OR ip (v4) address
++ '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' // port and path
++ '(\\?[;&a-z\\d%_.~+=-]*)?' // query string
++ '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
 
 export const isUrl = (string) => urlPattern.test(string);
+export const isGlobalUrl = (string) => globalUrlPattern.test(string);
```
adamaltman: I think there is a better name for this such as isFullyQualifiedUrl.

https://www.google.com/search?q=fully+qualified+url&oq=fully+qualified+url&aqs=chrome..69i57j0l7.9247j0j7&sourceid=chrome&ie=UTF-8

Also, when looking at the actual pattern difference between urlPattern and this pattern, is the only difference that the protocol is not required for the urlPattern?  (And how is it useful without the protocol?)

RomanHotsiy: Agree, updated.

> And how is it useful without the protocol?)

To check url fields in user definitions (sometimes they do not specify protocol).

PR #83, opened 2020-01-07
feat: allow non-yaml $refs in scalars.

knidarkness:
For now, only in 'description' field. However, this can be updated by adding 'resolvableScalars' field to the type definition for which the user wants to allow such scalars. 'resolvableScalars' is an array which must contain a list of scalars of the type for which $refs are allowed.

Now, we allow only file system references. However, it might make sense to allow also document-based $refs, like 'some/other/file.yaml#/some/node', although it will increase complexity of this module significantly.

### DIFF
--- a/definitions/syntetic/inc/docs-description.md
+++ b/definitions/syntetic/inc/docs-description.md
@@ -1,0 +1,1 @@
+some test
--- a/definitions/syntetic/inc/docs-description.md
+++ b/definitions/syntetic/inc/docs-description.md
@@ -1,1 +1,9 @@
-some test
+# Test markdown file
+
+This is a test markdown file.
+
+Include it in your OpenAPI definition description like this: 
+
+
+    description:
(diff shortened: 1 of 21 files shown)


definitions/syntetic/inc/docs-description.md
```diff
@@ -0,0 +1 @@
+some test
```
adamaltman:
```suggestion
# Test markdown file

This is a test markdown file.

Include it in your OpenAPI definition description like this: 


    description:
      $ref: path/to/file.md

```
I know it's just a test... but I think this spices it up a bit.  And if someone opens the file, they can see more about the behavior too.

RomanHotsiy:
This PR has to be rewritten so it uses generic `resolveNode` function which has support for http and filesystem refs as well as transitive or local ref.

I have other notes about this PR but merging it for now as it solves the main use case.

PR #86, opened 2020-01-13
feat: add --force to bundle command

RomanHotsiy:


adamaltman:
The error messages seem fine to me.

.eslintrc.yml
```diff
@@ -13,6 +13,7 @@ parserOptions:
   sourceType: module
 rules:
   class-methods-use-this: off
+  max-len: [1, 120]
```
knidarkness: @RomanHotsiy , how is max-len used? Could not find it anywhere else in the code. Also, if it's for identifiers' lengths should not it be an integer instead of an array? (which I suppose indicates MIN and MAX lengths)

src/visitors/bundler.js
```diff
@@ -177,7 +177,7 @@ class Bundler {
           node.components = {};
         }
 
-        if (ctx.result.some((e) => e.severity === messageLevels.ERROR)) {
+        if (!this.config.ignoreErrors && ctx.result.some((e) => e.severity === messageLevels.ERROR)) {
```
knidarkness: Also, here we will create bundle even if there are unresolved `$ref`s, that is not exactly "validation error", rather a structural/syntax ones.

Is that intended behavior?

.eslintrc.yml
RomanHotsiy: It's .eslintrc, it's for max code line length in our code (the default from airbnb-base is 100 which I find quite outdated for modern screens)

src/visitors/bundler.js
RomanHotsiy: Yes. I think we are OK with that. These refs will be in-place so next tool can catch the error.
I will need --force for docs preview. We will show there are errors but we will show docs preview too...

.eslintrc.yml
knidarkness: Ouch, missed the filename :(

PR #91, opened 2020-01-15
feat: add basic preview-docs command

RomanHotsiy:


RomanHotsiy:
I merged it straight away, but I would like your review anyway @knidarkness

PR #93, opened 2020-01-16
fix: typo in error message

adamaltman:


src/visitors/bundler.js
```diff
@@ -30,7 +30,7 @@ const getComponentName = (refString, components, componentType, node, ctx) => {
 
     errors.push(
       ctx.createError(
-        `Two schemas with the same name but different content are referneced. Renamed ${prevName} to ${name}`,
+        `Two schemas with the same name but different content are referenced. Renamed ${prevName} to ${name}`,
```
adamaltman: I think this is better: 

```suggestion
        `Two schemas are referenced with the same name but different content. Renamed ${prevName} to ${name}`,
```

PR #101, opened 2020-01-24
feat: support aliases in entryPoints argument

knidarkness:
Allow to call `openapi` command in following ways:

```bash
openapi { validate | bundle } alias anotherAlias ./some_path/definition.yaml  # will lookup for aliases resolutions in the apiDefinitions field of config and then validate both aliased files and direct file path
```

## Important note
For now collision is possible, e.g.:
```yaml
apiDefinitions:
  public: ./definition.yaml
  "private.yaml": ./another/file.yaml
```
Will cause `openapi validate private.yaml` to lookup definition in the `./another/file.yaml` even if `private.yaml` exists in current dir.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -86,9 +86,7 @@ export function getFallbackEntryPointsOrExit(argsEntrypoints, config = getConfig
   ) {
     res = Object.values(config.apiDefinitions);
   } else if (argsEntrypoints && argsEntrypoints.length && config.apiDefinitions) {
-    for (let i = 0; i < argsEntrypoints.length; i++) {
-      res[i] = config.apiDefinitions[res[i]] || res[i];
-    }
+    res = res.map(aliasOrPath => config.apiDefinitions[aliasOrPath] || aliasOrPath);
   }
 
   if (!res || !res.length) {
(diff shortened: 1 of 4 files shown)


RomanHotsiy:
Collision is OK

src/config.js
```diff
@@ -85,6 +85,11 @@ export function getFallbackEntryPointsOrExit(argsEntrypoints, config = getConfig
     && Object.keys(config.apiDefinitions).length > 0
   ) {
     res = Object.values(config.apiDefinitions);
+  } else if (argsEntrypoints && argsEntrypoints.length && config.apiDefinitions) {
+    const aliases = Object.keys(config.apiDefinitions);
+    for (let i = 0; i < argsEntrypoints.length; i++) {
+      res[i] = aliases.indexOf(argsEntrypoints[i]) !== -1 ? config.apiDefinitions[res[i]] : res[i];
+    }
```
RomanHotsiy:
```
for (let entrypoint of argsEntrypoints) {
  res[i] = config.apiDefinitions[entrypoint] || res[i];
}
```

PR #107, opened 2020-03-09
feat: async API & headers support & registry integration

knidarkness:
TODO: figure out way to store variables, such as URLs and remove testing localhost values.

For now, headers are set as following

```
  headers:
    - name: Client-Header
      value: sometestvalue
      regexp: api.redoc.ly/.*
    - name: Custom
      envVariable: TEST_VAR
      regexp: api.redoc.ly/.*
```

Also, this PR introduces commands:

`openapi-cli login` and `openapi-cli logout`, see `registry-dependencies` branches on internal repos


definitions/.redocly.yaml
```diff
@@ -0,0 +1,64 @@
...
+  sample: ./syntetic/to_bundle/bundle.yaml
+  simpleDefinition: ./syntetic/syntetic.yaml
+  rebilly: ./openapi-directory/rebilly-full.yaml
+
+lint:
+  codeframes: on
+  headers:
+    - name: Client-Header
+      value: sometestvalue
+      regexp: api.redoc.ly/.*
```
RomanHotsiy: Regexp may be confusing to some users and not always required, could we add `hostname` syntax sugar ?

Also this specific regexp is insecure, see bellow for details.

definitions/.redocly.yaml
```diff
@@ -0,0 +1,64 @@
...
+    debug-info: off
+
+    parameterPartial: warning
+    parameterWithAllOf: warning
+
+    oas3-schema/parameter: off
+    oas3-schema/external-docs:
+      url: off
+
+    reg-rule:
```
RomanHotsiy: We need to rename it. Let's discuss the name in Slack

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@redocly/openapi-cli",
-  "version": "0.8.4",
+  "version": "0.8.5",
```
RomanHotsiy: we need minor bump to `0.9.0` not patch. But let's add a separate commit for that.

src/bundle.js
```diff
@@ -31,13 +32,14 @@ export const bundleToFile = (fName, outputFile, force) => {
     },
   };
 
-  const ctx = createContext(document, doc, resolvedFileName, lintConfig);
+  const redoclyClient = new RedoclyClient();
```
RomanHotsiy: Why we need passing `redoclyClient` to `createContext`? Can't `createContext` create it itself?

src/cli/index.js
```diff
@@ -45,14 +51,41 @@ const cli = () => {
   program
     .version(version, '-v, --version', 'Output current version of the OpenAPI CLI.');
 
+  program
+    .command('login')
```
RomanHotsiy: I think we wanted `registry login` not just `login` ?

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+      writeFileSync(credentialsPath, JSON.stringify(credentials));
+      process.stdout.write('Authorization confirmed.\n');
+    } else {
+      process.stdout.write('Authorization failed. Please check if you entered a valid token.\n');
+    }
+  }
+
+  logout() {
+    const credentialsPath = resolve(homedir(), '.redocly.token.json');
+    if (existsSync(credentialsPath)) {
```
RomanHotsiy: Add a console message here too.

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+  login(accessToken) {
+    const credentialsPath = resolve(homedir(), '.redocly.token.json');
+    const authDetails = RedoclyClient.authorize(accessToken);
+    if (authDetails) {
+      this.accessToken = accessToken;
+      const credentials = {
+        token: accessToken,
+      };
+      writeFileSync(credentialsPath, JSON.stringify(credentials));
+      process.stdout.write('Authorization confirmed.\n');
```
RomanHotsiy: Make it green + maybe some emoji ?

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  getAuthorizationHeader() {
+    return this.accessToken;
+  }
+
+  login(accessToken) {
+    const credentialsPath = resolve(homedir(), '.redocly.token.json');
+    const authDetails = RedoclyClient.authorize(accessToken);
```
RomanHotsiy: Add silver log `Logging in...`

src/cli/index.js
```diff
@@ -45,14 +51,41 @@ const cli = () => {
...
+      const client = new RedoclyClient();
+      client.logout();
+    });
+
+  program
+    .command('ls')
+    .description('Get a list of available definitions')
+    .action(async () => {
+      const client = new RedoclyClient();
+      await client.authorize('aaa', 'bbb');
```
RomanHotsiy: ? Maybe remove this ls at all.

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+          }
+        }
+      `);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
+
+  listDefinitions() {
```
RomanHotsiy: If you remove `ls`, remove this too

src/config.js
```diff
@@ -70,6 +69,17 @@ export function getConfig(options) {
   const transformingVisitors = require(lintConfig.transformers);
   lintConfig.transformingVisitors = transformingVisitors;
 
+  if (!resolvedConfig.lint) {
+    resolvedConfig.lint = {};
+  }
+
+  resolvedConfig.lint.headers = (resolvedConfig.lint && resolvedConfig.lint.headers) ? (resolvedConfig.lint.headers.map((header) => ({
```
RomanHotsiy: `(resolvedConfig.lint && resolvedConfig.lint.headers || []).map` instead of ternary operator would be less verbose

src/config.js
```diff
@@ -70,6 +69,17 @@ export function getConfig(options) {
...
+    ...header,
+    value: header.envVariable ? process.env[header.envVariable] : header.value,
+  }))) : [];
+
+  // console.log(resolvedConfig.lint.headers);
```
RomanHotsiy: remove this comment pls

src/resolver.js
```diff
@@ -56,7 +56,7 @@ function resolve(link, ctx, visited = []) {
   pushPath(ctx, resolvedFilePath, []);
 
   const resolvedLink = `${resolvedFilePath}#/${docPath}`;
-
+  // console.log(linkSplitted);
```
RomanHotsiy: remove comment pls

src/context.js
```diff
@@ -32,24 +33,39 @@ const getRule = (ctx, ruleName) => {
   return result ? result[0] : null;
 };
 
-function createContext(node, sourceFile, filePath, config) {
+function createContext(node, sourceFile, filePath, config, redoclyClient) {
   const [enabledRules, allRules] = loadRuleset(config);
+  config.headers = config.headers || [];
+  config.headers = [...(config.headers || []), {
```
RomanHotsiy: no need to `(config.headers || [])`, see line above

src/context.js
```diff
@@ -32,24 +33,39 @@ const getRule = (ctx, ruleName) => {
...
+    regexp: `https://api.${process.env.REDOCLY_DOMAIN || 'redoc.ly'}/registry.*`,
```
RomanHotsiy: this regexp is insecure, dot is treated as any symbol.

```
new RegExp(`https://api.redoc.ly/registry.*`).test('https://apisredoc.ly/registry-test') === true
```

We either should ask users to specify true regexps here which I think will be complicated to users (considering even we as developers make insecure regexps) or we should come up with a better idea

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+import query from './query';
+
+export default class RedoclyClient {
+  constructor() {
+    const credentialsPath = resolve(homedir(), '.redocly.token.json');
+    let credentials = {};
+    if (existsSync(credentialsPath)) {
+      credentials = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
+    }
+    const authDetails = RedoclyClient.authorize(credentials.token);
```
RomanHotsiy: I don't think we should check if token is authorized each time + definitely not in constructor.

I would move it to `getAuthorizationHeader`. Also why we check it if we don't show any warning/error ?

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+      branchId: parseInt(process.env.BRANCH, 10),
+    }, {
+      Authorization: authorizationToken,
+    });
+    return r;
+  }
+
+  processRegistryDependency(link, ctx) {
+    const domain = process.env.REDOCLY_DOMAIN || 'redoc.ly';
+    if (link.indexOf(`https://api.${domain}/registry/`) !== 0) return;
```
RomanHotsiy: `startsWith`, `indexOf` is insecure, e.g. `link = "https://evil.com?url=https://api.${domain}/registry/"`

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+      definitionId: parseInt(process.env.DEFINITION, 10),
+      versionId: parseInt(process.env.VERSION, 10),
+      branchId: parseInt(process.env.BRANCH, 10),
+    }, {
+      Authorization: authorizationToken,
+    });
+    return r;
+  }
+
+  processRegistryDependency(link, ctx) {
```
RomanHotsiy: I would rename this method. The name implies that link is registry dependency but we check this in this method. And we do not process.

I would rename this method to `isRegistryURL` and push to context not in this method. This class should better not know about ctx. 

Also, make this method `static`

src/redocly/index.js
```diff
@@ -0,0 +1,161 @@
...
+
+    // we can be sure, that there is job UUID present
+    // (org, definition, version, bundle, branch, job)
+    // so skip this link.
+    if (pathParts.length === 6) return;
+
+    ctx.dependencies.push(link);
+  }
+
+  getLintConfig(organization, definitionName, versionName) {
```
RomanHotsiy: We don't need it ?

src/redocly/query.js
```diff
@@ -0,0 +1,36 @@
+import { XMLHttpRequest } from 'xmlhttprequest';
```
RomanHotsiy: Please, remove `xmlhttprequest` from here and from dependencies 🔥 and NEVER use it anymore. It uses some strange file system tricks to achieve synchronous execution. I constantly see some files appearing and disappearing in my IDE which is super weird and confusing.

Use node-fetch and update where needed to async/await

src/resolver.js
```diff
@@ -70,19 +70,33 @@ function resolve(link, ctx, visited = []) {
...
+            headers[ctx.headers[i].name] = ctx.headers[i].value;
+          }
+        }
+
+        const req = await fetch(resolvedFilePath, { headers });
 
-        if (xhr.status !== 200) {
+        if (!req.ok) {
+          // FIXME: lost file read errors here
+          // ctx.result.push(...);
```
RomanHotsiy: Could you add an error to result here?
Right now it will just silently fail e.g. for 404 links.

src/resolver.js
```diff
@@ -70,19 +70,33 @@ function resolve(link, ctx, visited = []) {
         ctx.result.push(createYAMLParseError(e, ctx, resolvedFilePath));
         return { node: undefined };
       }
-      // FIXME: lost yaml parsing and file read errors here
     } else if (isFullyQualifiedUrl(resolvedFilePath)) {
       try {
-        const xhr = new XMLHttpRequest();
-        xhr.open('GET', resolvedFilePath, false);
-        xhr.send();
+        ctx.redoclyClient.processRegistryDependency(resolvedFilePath, ctx);
```
RomanHotsiy: also, can we do it in custom rule vs here in resolver? e.g. in `semantic/registryDependencies.js`

src/visitors/rules/semantic/registryDependencies.js
```diff
@@ -0,0 +1,18 @@
+class RegsitryDependencies {
```
RomanHotsiy: I would put it in separate folder, not in `semantic`, maybe `registry` ?

src/visitors/rules/custom/regRule.js
```diff
@@ -0,0 +1,75 @@
...
+class RegRule {
+  static get rule() {
+    return 'reg-rule';
+  }
+
+  constructor(config) {
+    this.ruleSets = {};
+
+    for (const ruleName of Object.keys(config.rules || {})) {
+      const [treeLevel, field] = config.rules[ruleName].on.split('.');
```
RomanHotsiy: add basic check here in case `on` is undefined. Crash with some meaningful error

RomanHotsiy: `treeLevel` maybe rename to `typeName`?

src/visitors/rules/custom/regRule.js
```diff
@@ -0,0 +1,75 @@
...
+      const rule = {
+        not: false,
+        ...config.rules[ruleName],
+        treeLevel,
+        field,
+        name: ruleName,
+      };
+
+      const regexp = (!!rule.regexp && new RegExp(rule.regexp)) || null;
+      const startsWith = (!!rule.startsWith && new RegExp(`^${rule.startsWith}`)) || null;
```
RomanHotsiy: I would better check startsWith and endsWith with function vs regexp. Or we need to escape e.g. `.` and other symbols here (see regexp based discussion above)

package.json
knidarkness: ok

src/bundle.js
knidarkness: agree, that makes sense

src/visitors/rules/custom/regRule.js
knidarkness: ✅

src/visitors/rules/custom/regRule.js
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable no-underscore-dangle */
+
+class RegRule {
```
RomanHotsiy: StringMatcher

src/visitors/rules/custom/regRule.js
```diff
@@ -0,0 +1,111 @@
...
+  static get rule() {
+    return 'reg-rule';
```
RomanHotsiy: string-matcher

PR #108, opened 2020-03-09
fix: use fully qualified url for XHR instead of filePath

### LINKED ISSUE #105: Reference from external reference is broken
When I'm trying to use a reference like
```yaml
$ref: "https://rebilly.github.io/api-standards/schema/json-schema-draft-04/problem-details/documented-problem.json"
```
Which contains another
```json
{
    "$ref": "blank-problem.json"
}
```
The `blank-problem.json` reference should resolve to https://rebilly.github.io/api-standards/schema/json-schema-draft-04/problem-details/blank-problem.json, but it fails to resolve. 
According to debugger XRH is sent with `blank-problem.json` as url

justnero:
This was causing request to be sent with a filename instead of the URL it should go

Fixes #105

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -74,7 +74,7 @@ function resolve(link, ctx, visited = []) {
     } else if (isFullyQualifiedUrl(resolvedFilePath)) {
       try {
         const xhr = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
Published a new version: `0.8.4`

PR #114, opened 2020-04-03
feat: readable error message, when definition or file not found

knidarkness:


src/validate.js
```diff
@@ -56,7 +57,21 @@ export const validateFromUrl = async (link, options) => {
...
   const resolvedFileName = fName; // path.resolve(fName);
-  const doc = fs.readFileSync(resolvedFileName, 'utf-8');
+  let doc;
+  try {
+    doc = fs.readFileSync(resolvedFileName, 'utf-8');
+  } catch (error) {
+    let errorMessage = `Can't find definition or file with name: ${chalk.red(fName)}.\n`;
+    const definitions = getDefinitionNames();
+    if (definitions) {
+      errorMessage = `${errorMessage}\nThese definitions are present in the .redocly.yaml file:\n`;
```
RomanHotsiy, line 67: simplify the error message to something:

```js
errorMessage = `${errorMessage}\nAvailable definitions:\n${definitions.join('\n- )}\n`;

File: src/bundle.js [hunk 2]
```diff
@@ -1,16 +1,26 @@
...
 import { OpenAPIRoot } from './types/OAS3';
 import { OAS2Root } from './types/OAS2';
 
+import { fileNotFoundError } from './utils';
+
 export const bundleToFile = async (fName, outputFile, force) => {
   const resolvedFileName = fName; // path.resolve(fName);
-  const doc = fs.readFileSync(resolvedFileName, 'utf-8');
+  let doc;
+  try {
```
RomanHotsiy, line 16: @knidarkness Could you reuse entire code for reading the file?

knidarkness: agree

src/utils/index.js
```diff
@@ -209,3 +212,23 @@ export function match(url, pattern) {
   }
   return minimatch(url, pattern);
 }
+
+export function readFile(fileName) {
```
RomanHotsiy, line 216: Better rename it to `readYaml` and handle yaml errors here too vs extra try-catch in every place.

Also, make sure to supply filename to yaml.safeLoad so it prints a nicer error.

knidarkness: done

PR #118, opened 2020-04-09
fix: allow more complex logic for property types & fix additionalProperties type

### LINKED ISSUE #117: Schemas only referenced in additionalProperties tags are flagged as unused
### Bug with the rule: no-unused-schemas

Schemas that are only referenced within an additionalProperties tag are flagged as being unused.

**Steps to reproduce:**

1. Create a schema.
2. Create only one `$ref` usage within another schema's additional properties tag.
3. Try to validate with the no-unused-schemas rule on.

knidarkness:
Resolves #117 issue

### DIFF
--- a/src/visitors/rules/oas2-structural/validateOAS2Schema.js
+++ b/src/visitors/rules/oas2-structural/validateOAS2Schema.js
@@ -119,7 +119,10 @@ class ValidateOAS2Schema {
         if (node && node.items && Array.isArray(node.items)) return ctx.createError('Value of items must not be an array. It must be a Schema object', 'value');
         return null;
       },
-      additionalProperties: () => null,
+      additionalProperties: (node, ctx) => {
+        if (node && node.additionalProperties && typeof node.additionalProperties !== 'boolean') return ctx.createError(ctx.messageHelpers.fieldTypeMismatchMessageHelper('boolean'), 'value');
+        return null;
+      },
       description(node, ctx) {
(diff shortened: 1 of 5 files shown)


src/visitors/rules/structural/validateOpenAPISchema.js
```diff
@@ -120,7 +120,15 @@ class ValidateOpenAPISchema {
         if (node && node.items && Array.isArray(node.items)) return ctx.createError('Value of items must not be an array. It must be a Schema object', 'value');
         return null;
       },
-      additionalProperties: () => null,
+      additionalProperties: (node, ctx) => {
+        if (node
+          && node.additionalProperties
+          && (['boolean', 'object'].indexOf(typeof node.additionalProperties) === -1)) {
+          console.log(typeof node.additionalProperties);
```
RomanHotsiy, line 127: console.log

src/visitors/rules/oas2-structural/validateOAS2Schema.js
```diff
@@ -119,7 +119,10 @@ class ValidateOAS2Schema {
         if (node && node.items && Array.isArray(node.items)) return ctx.createError('Value of items must not be an array. It must be a Schema object', 'value');
         return null;
       },
-      additionalProperties: () => null,
+      additionalProperties: (node, ctx) => {
+        if (node && node.additionalProperties && typeof node.additionalProperties !== 'boolean') return ctx.createError(ctx.messageHelpers.fieldTypeMismatchMessageHelper('boolean'), 'value');
```
RomanHotsiy, line 123: can we have 'object' here? I see it is checked in OAS3 but not here.

PR #119, opened 2020-04-16
feat: additional tags semantic rule & enable provide-contact default

knidarkness:


src/visitors/rules/semantic/operation-tags-defined.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  _finishValidation() {
+    return {
+      onExit: (_node, _, ctx) => {
+        const errors = [];
+        for (const tag of this.operationTags) {
+          if (this.globalTagNames.indexOf(tag.name) === -1) {
+            const tmpPath = ctx.path;
+            ctx.path = tag.path;
```
RomanHotsiy, line 19: I believe you may lose `file` context here.

Could you modify code to create errors immediately vs postpone it Root exit?

knidarkness: 👍, Updated to create errors in-place

PR #143, opened 2020-06-08
feat: Validate operation now has glob support

kf6kjg:
Very helpful when working with multiple APIs in subfolders.


adamaltman:
Thanks for your contribution. It looks like this functionality has made it into the main branch already. I'm going to close this PR.

PR #146, opened 2020-06-19
Full rewrite the core engine

### LINKED ISSUE #142: Rewrite the core engine using typescript
We should rewrite `openapi-cli` with these considerations:

1. Use typescript (because it is consistent with our other projects and it is less error-prone).
1. Have a robust test suite.
1. Maintain same or better than current performance.
1. Improve and document the custom rules API.

RomanHotsiy:
Fixes #142

TBD


src/config/recommended.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    'parameter-description': 'off',
+    'no-path-trailing-slash': 'error',
+    'no-ambiguous-paths': 'warn',
+    'path-declaration-must-exist': 'error',
+    'path-not-include-query': 'error',
+    'path-parameters-defined': 'error',
+    'operation-description': 'off',
+    'operation-2xx-response': 'warn',
+    'operation-operationId': 'off',
```
adamaltman, line 21: I think this should be error or warning. Having an `operationId` improves so much about the OpenAPI definition's usability.

adamaltman: And because this is "recommended" I would actually probably make it an `error`.

docs/rules.md
```diff
@@ -0,0 +1,80 @@
...
+
+Some tooling forgets to strip trailing slashes off when it's joining the servers.url with paths, and you can get awkward URLs like `https://example.com/api//pets`. Best to just strip them off yourself.
+
+###  tags-alphabetical
+OpenAPI object should have alphabetical `tags`. This will be sorted by the `name` property.
+
+### operation-2xx-response
+Operation must have at least one `2xx` response. Any API operation (endpoint) can fail but presumably it is also meant to do something constructive at some point. If you forget to write out a success case for this API, then this rule will let you know.
+
+### operation-operationId-unique
```
adamaltman, line 38: Missing `#operation-operationId`

README.md
```diff
@@ -86,57 +88,68 @@ Given this command, it will start a local development server and display the add
...
+    no-empty-servers: warn
+
+    parameter-description: off
+    no-path-trailing-slash: error
+    path-declaration-must-exist: error
+    path-not-include-query: error
+    path-parameters-defined: error
+    operation-description: off
+    operation-2xx-response: warn
+    operation-operationId-unique: error
```
adamaltman, line 137: What about `operation-operationId`? I see it is not in the recommended currently. I commented below that I think it should be.

README.md
```diff
@@ -86,57 +88,68 @@ Given this command, it will start a local development server and display the add
...
+You can modify (or create) the `.redocly.yaml` file in the directory from which you are going to run the linter. Also, you can provide the path to the configuration file name other than `.redocly.yaml` by using `--config` option when running the @redocly/openapi-cli.
 
-From a high-level, there are two configurable features: codeframes and rules.
-```yaml
-lint:
-  codeframes: on
-  rules:
-  ...
-```
+From a high-level, there are two configurable features: plugins and rules.
```
adamaltman, line 93: what is a plugin?

RomanHotsiy: A plugin is a unified way to extend openapi-cli behavior (add custom rules, decorators, preprocessors, type extensions)

RomanHotsiy: Plugins can be local (local JS file) or npm package (we don't support it now but we will later).

docs/plugins.md
```diff
@@ -0,0 +1,125 @@
+# Working with Plugins
```
adamaltman, line 1: Let's move this to the redoc.ly/docs. I can do that.

RomanHotsiy: Yes. 👍

PR #147, opened 2020-06-22
feat: no-empty-string rule

knidarkness:


RomanHotsiy:
We have to rethink the idea of such generic rules. Closing this PR for now.

PR #148, opened 2020-06-22
docs: start updating docs after full-rewrite

knidarkness:


docs/TRANSFORMERS.md
```diff
@@ -10,4 +10,38 @@ You can do it with `transformers` visitors.
 
 ## Usage
 
-TBD
\ No newline at end of file
+First of all you will need to create a file with the custom rules and transformers. For example,
+
+```js
+export const id = 'customTransformers';
```
adamaltman, line 16: Does the file need to have an `id`? What is the meaning of it? How it is used elsewhere?

RomanHotsiy: Yes. It must to have `id`. plugin id, it is used to prefix all the rules of this plugin so they won't clash with built-in in rules

PR #149, opened 2020-06-22
feat: add checks for schema object fields

knidarkness:
Schema object fields checks  for correct **minimum** value.
_Here is the types list:_
```
multipleOf
maxLength
minLength
maxItems
minItems
maxProperties
minProperties
```


RomanHotsiy:
@andriyl would be great to support that but I'd get rid of the `constraints` wrapper and just add `minimum`/`maximum`/etc directly on the "schema".

Also, we just need `minimum` and `maximum`, `minLength and `maxLenght` are not used anywhere so let's not implement it.

RomanHotsiy (approved):
LGTM

PR #156, opened 2020-07-06
Full rewrite e2e tests

prooles:


__tests__/no-unresolved-refs-rule/.redocly.yaml
```diff
@@ -0,0 +1,42 @@
+apiDefinitions:
```
prooles, line 1: this test seems to be similar to previous

__tests__/no-unresolved-refs-error/openapi.yaml
```diff
@@ -0,0 +1,31 @@
...
+      operationId: getPetsByStatus
+      tags:
+        - pets
+      responses:
+        '200':
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApiResponse'
```
prooles, line 19: I think the idea of __tests__/no-unresolved-refs-error/ test is to check if the error will be returned in case ref is not resolved. So I think we should edit 'no-unresolved-refs-error' test case to not resolve ref and check an error.

__tests__/parameter-description-warning/openapi.yaml
```diff
@@ -0,0 +1,48 @@
...
+components:
+  parameters:
+    offsetParam:
+      in: query
+      name: offset
+      required: false
+      schema:
+        type: integer
+        minimum: 0
+      #description: The number of items to skip before starting to collect the result set.
```
prooles, line 15: I'd rather clean up the example and leave a single parameter without description to make the snapshot cleaner.

prooles (commented):
good job Alex!

AlexVarchuk:
> good job Alex!

Thanks, @prooles for the comments. I fixed it.

Marshevskyy (approved):

__tests__/info-contact-not-present-warning/.redocly.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+    no-example-value-and-externalValue: off
+
+    no-unused-components: off
+    no-unresolved-refs: off
+    no-enum-type-mismatch: off
+
+    boolean-parameter-prefixes: off
+    paths-kebab-case: off
+    spec: off
```
RomanHotsiy, line 41: Can we remove those `off` everywhere where we have `extends: []`

RomanHotsiy (changes requested):
It looks great!

Good job both @prooles and @AlexVarchuk.

I think we should remove the duplicated rules: 'off' when we have `extends: []`.

If this is an atomic test for a single rule we don't want to explicitly disable all the other rules. It doesn't make sense to me and makes tests harder to read (you have to spot which rule is on actually) and maintain (what if we add another rule, do we have to add it to every test?).

__tests__/info-contact-not-present-warning/.redocly.yaml
AlexVarchuk: Good point. Thanks. Fixed.

RomanHotsiy (approved):
Amazing!

PR #162, opened 2020-07-09
fix: trailing slash rule wrong location & '/' path

knidarkness:


src/rules/common/no-path-trailing-slash.ts
```diff
@@ -3,10 +3,11 @@ import { UserContext } from '../../walk';
 
 export const NoPathTrailingSlash: Oas3Rule | Oas2Rule = () => {
   return {
-    PathItem(_path: any, { report, key }: UserContext) {
-      if ((key as string).endsWith('/')) {
+    PathItem(_path: any, { report, key, location }: UserContext) {
+      if ((key as string).endsWith('/') && (key as string).length !== 1) {
```
RomanHotsiy, line 7: check exactly for `/`, if the path is `a` it should still report

knidarkness: > if the path is `a` it should still report
could you clarify plz? although, updated as exact check is definitely more readable

PR #169, opened 2020-07-20
feat: handle no openapi case in openapi

knidarkness:
I think we can safely exit(1) right when we know that document is not openapi, so that we don't need additional handling logic, but might be wrong here.

### DIFF
--- a/src/validate.ts
+++ b/src/validate.ts
@@ -102,8 +102,7 @@ export function detectOpenAPI(root: any): OasVersion {
   }
 
   if (!(root.openapi || root.swagger)) {
-    process.stderr.write('❗️❗️❗️ This doesn’t look like an OpenAPI document.\n');
-    process.exit(1);
+    throw new Error('This doesn’t look like an OpenAPI document.\n');
   }
 
   if (root.openapi && root.openapi.startsWith('3.0')) {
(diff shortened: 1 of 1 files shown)


src/validate.ts
```diff
@@ -101,6 +101,11 @@ export function detectOpenAPI(root: any): OasVersion {
     throw new Error(`Document must be JSON object, got ${typeof root}`);
   }
 
+  if (!(root.openapi || root.swagger)) {
+    process.stderr.write('❗️❗️❗️ This doesn’t look like an OpenAPI document.\n');
```
RomanHotsiy, line 105: Other code around throws, your code `process.exit(1)`. Make it consistent please

knidarkness: which way you think is better? as for ux i'd say that we can use exit(1) cause otherwise it outputs stacktrace which is not very useful for this case

RomanHotsiy: I would vote for throw Error. We can handle this error separately later.

There may be other definitions to be processed or it may be run in VS Code extension where `process.exit` doesn't make sense

knidarkness: fair enough, didn't think of vs code or lib-like usage

PR #175, opened 2020-08-02
Credit OAS-kit for original Speccy/Spectral ruleset

MikeRalphson:
Speccy (was, it is now unmaintained) a thin wrapper around OAS-Kit's `oas-linter`. These rules then went on to inspire the original set in Spectral.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,4 +37,4 @@ Unlike other OpenAPI linters, `@redocly/openapi-cli` defines the possible type t
 
 ## Credits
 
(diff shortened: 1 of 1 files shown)


MikeRalphson:
`oas-linter` also uses a visitor pattern to traverse the OAS document, unlike Spectral which uses JSON Path expressions to locate each rule's target.

RomanHotsiy:
Thanks, @MikeRalphson 👍

PR #176, opened 2020-08-08
feat: add 'json' option for --format

knidarkness:


package.json
```diff
@@ -58,8 +63,10 @@
     "js-yaml": "^3.14.0",
     "lodash.isequal": "^4.5.0",
     "minimatch": "^3.0.4",
+    "net": "^1.0.2",
```
RomanHotsiy, line 66: is it used?

package.json
```diff
@@ -58,8 +63,10 @@
...
     "node-fetch": "^2.6.0",
     "portfinder": "^1.0.26",
+    "readline": "^1.3.0",
```
RomanHotsiy, line 69: is it used?

src/cli/preview-server/server.ts
```diff
@@ -2,7 +2,7 @@ import * as http from 'http';
 import * as zlib from 'zlib';
 import { ReadStream } from 'fs';
 
-const SocketServer = require('simple-websocket/server');
+const SocketServer = !process.env.UPDATE_REGISTRY ? require('simple-websocket/server') : undefined as any; // need this for correct bundling, UPDATE_REGISTRY is used when run in our cloud
```
RomanHotsiy, line 5: Let's discuss it

knidarkness: removed this and used `null-loader` for webpack instead

package.json
knidarkness: removed

package.json
knidarkness: apparently yes, and that's weiiird

src/format/format.ts
```diff
@@ -57,14 +57,32 @@ export function formatProblems(
...
   problems = problems
-    .sort((a, b) => severityToNumber(a.severity) - severityToNumber(b.severity))
-    .slice(0, maxProblems);
-
-  if (!totalProblems) return;
-
-  if (format === 'codeframe') {
+    .sort((a, b) => severityToNumber(a.severity) - severityToNumber(b.severity));
+
+  problems = !formatJSON ? problems.slice(0, maxProblems) : problems;
```
RomanHotsiy, line 65: I would keep this maxProblems slice too. We can set --max-problems to a huge number in workflows build instead of adding different behavior here.

knidarkness: if you are sure, we can do that way, but i don't like the idea that we can have invalid JSON output with <...message...> in the workflows if somehow our number is not huge enough

PR #177, opened 2020-08-11
feat: add bundling opportunity

knidarkness:


prepareBundle.js
```diff
@@ -0,0 +1,12 @@
+const { execSync } = require('child_process');
+const package = require('./package.json');
+
+const version = package.version;
+const fileName = `openapi-cli.${version}.1111tar.gz`;
```
RomanHotsiy, line 5: ??

prepareBundle.js
```diff
@@ -0,0 +1,12 @@
...
+const fileNameLatest = `openapi-cli.latest.tar.gz`;
+
+execSync(`tar -zcvf ${fileName} dist`);
+execSync(`tar -zcvf ${fileNameLatest} dist`);
+
+execSync(`aws s3 cp ${fileName} s3://openapi-cli-dist --profile redoc.ly`);
```
RomanHotsiy, line 11: remove `--profile`, profile names can be different on different devices.

webpack.config.ts
```diff
@@ -0,0 +1,51 @@
...
+  entry: './src/cli.ts',
+  module: {
+    rules: [
+      {
+        test: /\.tsx?$/,
+        use: 'ts-loader',
+        exclude: /node_modules/,
+      },
+      {
+        test: /\.node$/,
```
RomanHotsiy, line 14: what is it used for? just curious

webpack.config.ts
```diff
@@ -0,0 +1,51 @@
...
+        loader: 'node-loader',
+      },
+      {
+        test: path.resolve(__dirname, 'node_modules/simple-websocket/server.js'),
+        use: 'null-loader',
```
RomanHotsiy, line 19: add a comment here explaining the bundle is not going to be used for live-preview?

webpack.config.ts
```diff
@@ -0,0 +1,51 @@
...
+  resolve: {
+    extensions: [ '.tsx', '.ts', '.js' ],
+  },
+  
+  node: {
+    __dirname: false,
+    fs: 'empty',
+  },
+
+  externals: [
```
RomanHotsiy, line 32: do we need externals considering we have null-loader?

prepareBundle.js
knidarkness: shit happened

webpack.config.ts
knidarkness: we have a .node compiled package which needs to be loaded

webpack.config.ts
knidarkness: need to 2-check

webpack.config.ts
knidarkness: :+1

webpack.config.ts
RomanHotsiy: yes, but what package it is? fsevents?

prepareBundle.js
knidarkness: okay, but i'll add cli arg check for that, as there might be profiles instead of only 1 set of creds

prepareBundle.js
knidarkness: fixed

prepareBundle.js
knidarkness: fixed

webpack.config.ts
knidarkness: works fine without

webpack.config.ts
knidarkness: done

webpack.config.ts
knidarkness: yep

RomanHotsiy: can it be null-loaded too?

package.json
```diff
@@ -43,10 +45,16 @@
     "@types/node-fetch": "^2.5.7",
     "@types/yargs": "^15.0.5",
     "jest": "^26.0.1",
+    "node-loader": "^1.0.1",
```
RomanHotsiy, line 48: looks like not needed anymore

PR #181, opened 2020-08-15
feat: use all parts of pointer to generate uniq component name if needed

jirutka:
Currently only the last part of the pointer (e.g. #/foo/bar -> bar) is
used for naming the component. If it's already taken, the base name of
the file reference is prepended. If it's still not unique, a sequence
number is appended to the name and warning is printed (`Two schemas are
referenced with the same name but different content.`).

The warning message is quite misleading and unnecessary because it's not
the user's fault that we used only the last part of the pointer to
generate unique component name.

To show some real-world example, let's consider file `examples.yml` with
examples for multiple types where each type may have multiple variants:

- #/Car/basic
- #/Car/electric
- #/Bike/basic
- #/Scooter/basic

The function `getComponentName()` will generate: basic, electric,
examples_basic, examples_basic-2, and warn about conflicting name of the
last. That's quite irritating, especially since the user cannot hide
these messages.

This commit solves this problem by changing the name generation
algorithm in `getComponentName()` to be gradually adding all the
components of the pointer from left to right before prepending the base
of the file reference and appending a serial.

Now for the example above the function will generate: basic, electric,
bike-basic, scooter-basic.

### DIFF
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -8,7 +8,7 @@ import { Oas2Types } from './types/oas2';
 import { NormalizedNodeType, normalizeTypes, NodeType } from './types';
 import { WalkContext, walkDocument, UserContext } from './walk';
 import { detectOpenAPI, openAPIMajor, OasMajorVersion } from './validate';
-import { Location, pointerBaseName, refBaseName } from './ref-utils';
+import { Location, refBaseName } from './ref-utils';
 import { Config, LintConfig } from './config/config';
 import { initRules } from './config/rules';
 import { reportUnresolvedRef } from './rules/no-unresolved-refs';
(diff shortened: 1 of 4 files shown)


src/bundle.ts
```diff
@@ -232,19 +232,21 @@ function makeBundleVisitor(version: OasMajorVersion, dereference: boolean, rootD
...
   ) {
     const [fileRef, pointer] = [target.location.source.absoluteRef, target.location.pointer];
+    const componentsGroup = components[componentType];
 
-    const pointerBase = pointerBaseName(pointer);
-    const refBase = refBaseName(fileRef);
+    let name = '';
 
-    let name = pointerBase || refBase;
+    const refParts = pointer.slice(2).split('/').filter(Boolean);  // slice(2) removes "#/"
```
adamaltman, line 239: What if there is no `pointer`? For example, in the test file `openapi-with-external-refs.yaml` above line 9:

`- $ref: ./param-b.yaml`

Is it the case where `name ? _${name}  : ''` will be an empty string?

jirutka: Yes, exactly.

RomanHotsiy (approved):
LGTM 👍 

Great job @jirutka

knidarkness (approved):
LGTM!

@jirutka Thanks!

PR #182, opened 2020-08-15
fix: remove extra blank line printed by printLintTotals()

jirutka:
`ignored` already ends with a newline, so when no rules are ignored, two
empty lines are printed after the green validation result message.
It's quite irritating when you run multiple linters in sequence...


RomanHotsiy:
@jirutka can you add this removed `\n` to the non-empty `ignored` message itself?

RomanHotsiy:
Thanks @jirutka !

PR #191, opened 2020-08-31
fix: crash with backrefs

### LINKED ISSUE #184: Bundle doesn't support OpenAPI yaml with $ref's
**Describe the bug**
Trying to run `openapi bundle -o openapi.json openapi.yaml` returns a heap of `cannot resolve $ref` errors. <br>
Running `openapi preview-docs` works fine though.

**To Reproduce**
Create any  YAML file with component refs (like schemas or routes) and try to bundle it.

**Expected behavior**
A bundled file

**Logs**
Exerpt:
```
[100] openapi.yaml:177:7 at #/components/schemas/Foo

Can't resolve $ref: ENOENT: no such file or directory, open '/home/user/MYPROJECT/docs/components/responses/components/schemas/foo.yaml'

175 |   $ref: "./components/schemas/object.yaml"
176 | Foo:
177 |   $ref: "./components/schemas/foo.yaml"
178 | Bar:
179 |   $ref: "./components/schemas/bar.yaml"

Error was generated by the bundler rule.
< ... 22 more problems hidden > increase with `--max-problems N`
```

**OpenAPI definition**
not possible

**Version(s)**
1.0.0-beta.8

**Additional context**
I'm looking for an easy way to run a yaml to json bundling through docker.

knidarkness:
Closes #184 

The problem was that when we went down the ref, we don't change the absolute path in rootNodeDocAbsoluteRef and that results in failures to resolve next $refs.

### DIFF
--- a/src/__tests__/fixtures/resolve/openapi-with-back.yaml
+++ b/src/__tests__/fixtures/resolve/openapi-with-back.yaml
@@ -1,0 +1,13 @@
+openapi: "3.0.1"
+info:
+  title: Open API
+  description: Open API
+  version: v0
+  contact:
+    email: some@body.com
+components:
+  schemas:
+    TypeA:
+      $ref: ./schemas/type-a.yaml#/
+    TypeB:
+      $ref: ./schemas/type-b.yaml#/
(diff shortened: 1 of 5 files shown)


RomanHotsiy (approved):

deniszgonjanin:
Thank you for the fix!

PR #200, opened 2020-09-08
fix: check node for null before running visitor

### LINKED ISSUE #195: Crash with "cannot read property ‘enum’ of null." in some cases
**Describe the bug**

```
Something went wrong when processing spec.yaml:

  - Cannot read property ‘enum’ of null.
```

**To Reproduce**
We have to figure it out, spotted in our prod logs.

**Expected behavior**
No crash.

**Logs**

```
npx @redocly/openapi-cli@latest lint spec.yaml --extends minimal --format json

validating spec.yaml...
Something went wrong when processing spec.yaml:

  - Cannot read property 'enum' of null.


TypeError: Cannot read property 'enum' of null
    at Schema (/tmp/openapi-cli/latest/dist/bundle.js:2:187024)
    at R (/tmp/openapi-cli/latest/dist/bundle.js:2:145849)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:144484)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at e (/tmp/openapi-cli/latest/dist/bundle.js:2:145102)
    at Object.t.walkDocument (/tmp/openapi-cli/latest/dist/bundle.js:2:146805)
```

**OpenAPI definition**
n/a

**Version(s)**
1.0.0-beta.11

knidarkness:
Closes #195

### DIFF
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -329,6 +329,9 @@ export function walkDocument<T>(opts: {
       ruleId: string,
       severity: ProblemSeverity,
     ) {
+      if (!node) {
+        return;
+      }
       const report = reportFn.bind(undefined, ruleId, severity);
       visit(
         node,
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -329,7 +329,7 @@ export function walkDocument<T>(opts: {
       ruleId: string,
       severity: ProblemSeverity,
     ) {
-      if (node === null || node === undefined && context.type.name !== 'scalar') {
+      if ((node === null || node === undefined) && context.type.name !== 'scalar') {
         // if node is not a scalar then visitor may crash when null or undefined is passed
         // instead of a node when it tries to access any fields
         return;
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -329,7 +329,9 @@ export function walkDocument<T>(opts: {
       ruleId: string,
       severity: ProblemSeverity,
     ) {
-      if (!node) {
+      if (node === null || node === undefined && context.type.name !== 'scalar') {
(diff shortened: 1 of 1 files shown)


src/walk.ts
```diff
@@ -317,6 +317,9 @@ export function walkDocument<T>(opts: {
       ruleId: string,
       severity: ProblemSeverity,
     ) {
+      if (!node) {
```
RomanHotsiy, line 320: check for `null` or `undefined` specifically, we may want to visit `0` or the empty string `""`.

Also, please, add some comment explaining this return (this is not obvious that type mismatch errors will be spotted by `spec` rule on the parent object.

Also, please check for `'scalar'` type here and visit scalar types regardless of value:

https://github.com/Redocly/openapi-cli/blob/master/src/walk.ts#L242

packages/core/src/walk.ts
```diff
@@ -329,6 +329,11 @@ export function walkDocument<T>(opts: {
       ruleId: string,
       severity: ProblemSeverity,
     ) {
+      if ((node === null || node === undefined) && (context.type?.name !== 'scalar')) {
+        // if node is not a scalar then visitor may crash when null or undefined is passed
+        // instead of a node when it tries to access any fields
+        return;
```
RomanHotsiy, line 335: If we do just this, we miss visitor run for null values so we won't be able to show validation errors:
Try changing any field (e.g. `Info.descrioption`) in the valid openapi to null, you won't get an error.

PR #204, opened 2020-09-09
feat: add oasVersion into the --format json output

knidarkness:
For now, it adds an `oasVersion` field to the top of the resulting JSON object. However, in the near future, we might consider adding smth like `meta` to store an increasing number of additional info.

```jsonc
{
  "oasVersion": "3.0.0",
  // ...
}
```

### DIFF
--- a/src/__tests__/detect-oas-version.test.ts
+++ b/src/__tests__/detect-oas-version.test.ts
@@ -1,0 +1,33 @@
+import { parseYamlToDocument, makeConfigForRuleset } from "./utils";
+import { validateDocument } from "../validate";
+import { BaseResolver } from "../resolve";
+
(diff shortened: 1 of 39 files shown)


knidarkness:
as an idea: add `meta` (think about naming, but semantically it'll be for metadata), and split `validateDocument` interface into smth like:

```jsonc
{
	problems: [], // problems go here,
	meta: { // for now just version, but will grow in future. also, rules will be able to write here smth that is not a problem
		oasVersion: "some"
	}
}
```

adamaltman:
I personally like the idea of `meta`.

RomanHotsiy:
Maintenance clean-up.
Outdated PR, implemented another way.

PR #205, opened 2020-09-09
feat: stats command

andriyl:


adamaltman:
Do you have a screenshot with the output? If so, add it to the PR description... I wonder if all of those colors will be readable in the terminal.

andriyl:
> Do you have a screenshot with the output? If so, add it to the PR description... I wonder if all of those colors will be readable in the terminal.

@adamaltman 
Please have a look at how it looks right now:

image

I'm really curious about how to make the output better.

adamaltman:
How can path items be more than operations?

adamaltman:
Nevermind, a path item can be empty (without operations defined) — unusual and revealed by this stats command. 👍 

The order looks strange. What if we order it like this:

References
External Documents (is this 
Schema (Can you add this?)
Parameters (can you add this?)
Links
Path Items
Operations
Tags

andriyl:
@adamaltman
Lists of stats are updated with the right order:
image
Also, the counting logic was improved as well.

adamaltman:
Our other commands, like lint, also show how long it took to run. And they have other display formats. Is there a JSON display format available here too?

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,112 @@
...
+    PathMap: {
+      leave() {
+        statsCount.parameters.total = statsCount.parameters.items.size;
+        statsCount.refs.total = statsCount.refs.items.size;
+        statsCount.links.total = statsCount.links.items.size;
+      },
+      PathItem: {
+        leave() { statsCount.pathItems.total++; },
+        Operation: { leave() { statsCount.operations.total++; }},
+        Parameter: { leave(parameter: Oas2Parameter | Oas3Parameter) {
```
adamaltman, line 85: Parameter can be defined on the PathItem and on the Operation. I’m not clear if this catches both of those?

RomanHotsiy: It catches both.

adamaltman:
🎉 

Can you please share the output for both `stylish` and `json` so that I can review?

andriyl:
> 🎉
> 
> Can you please share the output for both `stylish` and `json` so that I can review?
yes, sure.

stylish:

image

json:

image

src/cli.ts
```diff
@@ -12,14 +12,15 @@ import {
...
   promptUser,
   CircularJSONNotSupportedError,
+  getExecutionTime,
 } from './utils';
 import { formatProblems, OutputFormat } from './format/format';
 import { ResolveError, YamlParseError } from './resolve';
 import { loadConfig, Config, LintConfig } from './config/config';
 import { NormalizedProblem } from './walk';
 import { previewDocs } from './cli/preview-docs';
+import { handleStats } from './cli/stats-docs'
```
RomanHotsiy, line 22: rename file to just `stats`

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,141 @@
...
+import { normalizeTypes } from '../types';
+import { Oas3Types } from '../types/oas3';
+import { Oas2Types } from '../types/oas2';
+import { WalkContext, walkDocument } from '../walk';
+import { normalizeVisitors, Oas2Visitor, Oas3Visitor } from '../visitors';
+import * as colors from 'colorette';
+import { getFallbackEntryPointsOrExit } from '../cli';
+import { Oas2Parameter } from '../typings/swagger';
+import { Oas3Parameter, OasRef } from '../typings/openapi';
+import { performance } from 'perf_hooks';
```
RomanHotsiy, line 13: move external packages imports to the top

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,141 @@
...
+import { getExecutionTime } from '../utils';
+
+interface IStatsRow {
```
RomanHotsiy, line 16: No need to prefix interfaces with `I`. This is not common in TS

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,141 @@
...
+  metric: string;
+  total: number;
```
RomanHotsiy, line 18: What if some metric is not a number but e.g. a list of uniqute things or something more advanced?

Let's keep it simple for now but keep in mind we'll need other type of stats later.

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,141 @@
...
+function printStatsStylish(statsCount: IStatsCount) {
+  for (const node in statsCount) {
+    const { metric, total, color } = statsCount[node as StatsName];
+    process.stderr.write(colors[color](`${metric}: ${total} \n`));
+  }
+}
+
+function printStatsJson(statsCount: IStatsCount) {
+  const json: any = {};
+  Object.keys(statsCount).forEach((key) => {
```
RomanHotsiy, line 45: let's use `for of` for consistency

src/cli/stats-docs.ts
```diff
@@ -0,0 +1,141 @@
...
+      PathItem: {
+        leave() { statsCount.pathItems.total++; },
+        Operation: { leave() { statsCount.operations.total++; }},
+        Parameter: { leave(parameter: Oas2Parameter | Oas3Parameter) {
+          statsCount.parameters.items!.add(parameter.name)
+        }}
+      }
+    },
+    Components: {
+      Schema: { leave() { statsCount.schemas.total++; }}
```
RomanHotsiy, line 119: It won't catch OAS2 schemas: https://github.com/Redocly/openapi-cli/blob/master/src/types/oas2.ts#L16

Also, I would just use `NamedSchema` for OAS3 too. It should count all the top-level named schemas but can be defined on the top-level (no need to use nested visitor)

RomanHotsiy (commented):
Looks really good 👍 

I would just make a Stats Visitor as a separate module somewhere instead of defining it inline.

I understand it relies on global stats var. Then maybe define it in global module scope instead of inline.

Also left some small comments.

RomanHotsiy (approved):

RomanHotsiy:
Great job @andriyl 👏

PR #220, opened 2020-10-19
Fix typo in README.md

lucasbraune:
I took the liberty to fix a typo in README.md in order to familiarize myself with GitHub pull requests. Please feel free to close the request and fix the issue directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Currently, @redocly/openapi-cli supports these features:
 - [x] Bundle a multifile definition into a single file.
 - [x] Decorators to modify a validated definition during bundling.
 - [x] Preview reference docs for local development.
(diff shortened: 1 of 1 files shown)


RomanHotsiy:
Hey @lucasbraune, thanks for your PR.

Unfortunately `fka` = `formally known as` which is a correct and more appropriate usage here.

Would you just replace `fka` to `formally known as` so it is more inclusive for non-native speakers?

lucasbraune:
I think "formerly known as" is clearer than "fka". This acronym isn't nearly as common as "aka". MacOS, for example, puts a red squiggle under "fka", but not under "aka". That being said, after being reminded of the meaning of "fka", I think your README is perfectly fine: there is no need to change it.

Thank you for taking the time to review my request!

PR #241, opened 2021-01-11
fix: incorrect refs

andriyl:


packages/core/src/types/index.ts
```diff
@@ -57,6 +57,7 @@ export function mapOf(typeName: string) {
 
 export function normalizeTypes(
   types: Record<string, NodeType>,
+  incorrectRefs: object | null
```
RomanHotsiy, line 60: just add a default value `= {}`

packages/core/src/config/config.ts
```diff
@@ -346,6 +347,18 @@ export class LintConfig {
     }
   }
 
+  getIncorrectRefs() {
+    const { resolveIncorrectRefs } = this.rawConfig;
```
RomanHotsiy, line 351: move this to the constructor vs getter

packages/core/src/types/index.ts
```diff
@@ -67,12 +68,18 @@ export function normalizeTypes(
     } as any;
   }
 
-  for (const type of Object.values(normalizedTypes)) {
-    normalizeType(type);
+  if (incorrectRefs) {
+    for (const type of Object.values(normalizedTypes)) { normalizeIncorrectRefs(type, incorrectRefs); }
   }
-
+  for (const type of Object.values(normalizedTypes)) { normalizeType(type); }
```
RomanHotsiy, line 74: keep the original loop and just add normalizeIncorrectRefs call inside it.
No need for two loops.

RomanHotsiy (changes requested):
LGTM so far.

Do not forget about:

- the default value for resolveIncorrectRefs
- support for resolving all references using `*`

RomanHotsiy (changes requested):
You forgot to set a default value for incorrect refs: `Info.description`, `Tag.description`, `Operation.description`, `XCodeSample.source` (it used to work like this before).

Also, add `MediaType.example` and `Example.value` please.

PR #244, opened 2021-01-17
chore: getting document in bundle

andriyl:


packages/core/src/bundle.ts
```diff
@@ -14,16 +14,21 @@ import { reportUnresolvedRef } from './rules/no-unresolved-refs';
 export type Oas3RuleSet = Record<string, Oas3Rule>;
 
 export async function bundle(opts: {
-  ref: string;
+  ref: string | Document;
```
RomanHotsiy, line 17: better to have a separate prop for this: `ref` and `document`.

`ref` means `reference`, it doesn't make sense to pass the document here.

You won't need to do funky isDocument check then.

RomanHotsiy (changes requested):

RomanHotsiy (approved):

PR #255, opened 2021-02-25
Fix/hide api key

jacobator:
fix https://github.com/Redocly/openapi-cli/issues/252

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -6,6 +6,7 @@ import * as yaml from 'js-yaml';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as readline from 'readline';
+import { Writable } from 'stream';
 import {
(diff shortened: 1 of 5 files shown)


Marshevskyy (approved):

packages/cli/src/utils.ts
```diff
@@ -119,17 +120,35 @@ export function saveBundle(filename: string, output: string) {
...
+        },
+      });
+    }
+
     const rl = readline.createInterface({
       input: process.stdin,
-      output: process.stdout,
+      output,
+      terminal: true,
+      historySize: hideUserInput ? 0 : 30,
```
RomanHotsiy, line 143: what is 30?

jacobator: 30 is the default for historySize if omitted

RomanHotsiy (approved):

PR #267, opened 2021-03-15
fix: add error handling for failed file upload on push command

jacobator:
fix https://github.com/Redocly/openapi-cli/issues/248


Marshevskyy (commented):
push command does not have try/catch. I think it would be good to add

Marshevskyy:
Also it would be great to add at least some unit tests

packages/cli/src/commands/push.ts
```diff
@@ -217,15 +224,17 @@ function getDestinationProps(destination: string) {
...
+  const response = await fetch(url, {
     method: 'PUT',
     headers: {
       'Content-Length': fileSizeInBytes.toString()
     },
     body: readStream
-  })
+  });
+
+  return response.ok;
```
Marshevskyy, line 239: return just response

packages/cli/src/commands/push.ts
```diff
@@ -224,17 +224,15 @@ function getDestinationProps(destination: string) {
   return destination.substring(1).split(/[@\/]/);
 }
 
-async function uploadFileToS3(url: string, filePathOrBuffer: string | Buffer): Promise<boolean> {
+function uploadFileToS3(url: string, filePathOrBuffer: string | Buffer) {
```
Marshevskyy: why did you remove async?

jacobator: because I removed await inside the function

Marshevskyy (approved):
good job on tests!

PR #275, opened 2021-04-06
fix: fixes for fs lib to be able to use core on client

andriyl:
This change is about having the ability to use openapi-core on the client.

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -398,10 +398,6 @@ export class Config {
   }
 }
 
-export async function createConfig(configPath?: string, rawConfig: RawConfig = {}): Promise<Config> {
-  return new Config(rawConfig, configPath);
-}
-
 export async function loadConfig(configPath?: string, customExtends?: string[]): Promise<Config> {
   if (configPath === undefined) {
     configPath = findConfig();
@@ -435,7 +431,7 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
       ...(rawConfig.resolve.http.headers ?? []),
     ];
   }
-  return createConfig(configPath, rawConfig);
+  return new Config(rawConfig, configPath);
 }
 
 function findConfig() {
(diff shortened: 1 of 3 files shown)


packages/core/src/config/config.ts
```diff
@@ -418,6 +415,20 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
...
+  const rawConfig = await getRawConfig();
+  return new Config(rawConfig);
+}
+
+export async function loadConfig(configPath?: string, customExtends?: string[]): Promise<Config> {
+  if (configPath === undefined) {
+    configPath = findConfig();
+  }
+  const rawConfig = await getRawConfig(configPath, customExtends);
+
```
RomanHotsiy, line 431:
```suggestion
export async function createConfig(configPath?: string, rawConfig: RawConfig = {}): Promise<Config> {
```

packages/core/src/config/config.ts
```diff
@@ -397,11 +398,7 @@ export class Config {
   }
 }
 
-export async function loadConfig(configPath?: string, customExtends?: string[]): Promise<Config> {
-  if (configPath === undefined) {
-    configPath = findConfig();
-  }
-
+async function getRawConfig(configPath?: string, customExtends?: string[]) {
```
RomanHotsiy, line 401:
```suggestion
export async function loadConfig(configPath?: string, customExtends?: string[]): Promise<Config> {
  if (configPath === undefined) {
    configPath = findConfig();
  }
```

packages/core/src/config/config.ts
```diff
@@ -418,6 +415,20 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
     rawConfig.lint.extends = customExtends;
   }
 
+  return rawConfig
```
RomanHotsiy, line 418:
```suggestion
  return createConfig(configPath, rawConfig);
```

RomanHotsiy (changes requested):
I would simplify it.

RomanHotsiy (approved):

PR #283, opened 2021-04-17
Update the year to 2021 and Add License Header

pawelborkar:
I have updated the year from 2019 to 2021 and also added **MIT License**  header to the license to make more clear and updated.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,6 @@
-Copyright 2019 Redocly LLC
+MIT License
+
+Copyright (c) 2021 Redocly LLC. All rights reserved.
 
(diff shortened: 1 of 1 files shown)


adamaltman:
We were advised the license should reflect the year it was originally written.

andriyl:
@pawelborkar could you change the year from 2021 to 2019, please?

pawelborkar:
@andriyl  This PR wasn't accepted so the main license file has no change in it and reflects the year 2019.

>We were advised the license should reflect the year it was originally written.

I think it's better to close this PR.

andriyl:
@pawelborkar thank you

PR #288, opened 2021-04-29
Clone plugin structures rather than change in-place

### LINKED ISSUE #287: resolvePlugins() can't be called more than once
**Describe the bug**
The preview-docs server calls `reloadConfig()` each time the configuration file changes, which means `resolvePlugins()` is called repeatedly. Unfortunately, `resolvePlugins()` updates the `Plugin` type exported by a plugin _in place_, and does this each time it is called.

This then leads to the plugin id being prefixed to all rules, preprocessors and decorators multiple times.

**To Reproduce**

1. define a plugin with a preprocessor; this example is taken from the `local-plugin.js` example; put it in `plugins/repro.js`:
    ```js
    const id = 'repro';

    const preprocessors = {
      oas3: {
        'duplicate-description': () => {
          return {
            Info(info) {
              if (info.description) {
                info.description = info.description + '\n' + info.description;
              }
            },
          };
        },
      },
    };

    module.exports = {
      id,
      preprocessors,
    }
    ```

2. Create a configuration file that enables the preprocessor; `repro.yaml`:

    ```yaml
    lint:
      plugins:
        - './plugins/repro.js'
      preprocessors:
        repro/duplicate-description: on
    ```
3. run `npx openapi preview-docs --config repro.yaml <some-openapi.yaml>`, using an openapi definition file of your choice.
4. open `http://127.0.0.1:8080` and observe that the description has been doubled
4. make a change to the `repro.yaml` config file; adding a space suffices.
5. make a change to the openapi.yaml file being previewed.
5. observe that the description is no longer being doubled.

What has happened is that the plugin module has been updated in-place, and the `preprocessors.oas3` object now has a `repro/repro/duplicate-description` key, with as many extra `repro/` prefixes as there have been configuration reloads.

**Expected behavior**
When loading plugin configuration, either the require cache for the module should be cleared to avoid re-using the already loaded module, and / or the configuration should not be altered in-place.

**`openapi-cli` Version(s)**
1.0.0-beta.44

**`Node.js` Version(s)**
v14.0.0

**Additional context**
I actually hit this issue sooner; I created a proxy script that uses `loadConfig()` before it calls `previewDocs`, and so `previewDocs()` is the second caller to `loadConfig()`. I was scratching my head as to why my plugin code ended up with extra plugin id prefixes before I found that I could reproduce it with using `openapi preview-docs` alone.

mjpieters:
## What/Why/How?

Clone plugin structures that are being changed while importing.

`loadConfig()` can be called more than once, which results in preprocessors, rules and  decorator rules in plugins being prefixed with the plugin id more than once.

This happens when running `openapi preview-docs` and then changing the config file, for example.

## Reference

This fixes #287.

## Testing

I've included a test that demonstrates the issue, as a separate commit. Without the other commit in this PR, the test fails with:

```
 FAIL  packages/core/src/config/__tests__/resolve-plugins.test.ts
  ● resolving a plugin › should prefix preprocessor names with the plugin id

    expect(received).toHaveProperty(path)

    Expected path: "test-plugin/description-preprocessor"
    Received path: []

    Received value: {"test-plugin/test-plugin/description-preprocessor": [Function description-preprocessor]}

      17 |     const plugin = config.lint.plugins[0];
      18 |
    > 19 |     expect(plugin.preprocessors?.oas2).toHaveProperty('test-plugin/description-preprocessor');
         |                                        ^
      20 |   });
      21 |
      22 |   it('should prefix decorator names with the plugin id', async () => {

      at packages/core/src/config/__tests__/resolve-plugins.test.ts:19:40
      at fulfilled (packages/core/src/config/__tests__/resolve-plugins.test.ts:5:58)

  ● resolving a plugin › should prefix decorator names with the plugin id

    expect(received).toHaveProperty(path)

    Expected path: "test-plugin/inject-x-stats"
    Received path: []

    Received value: {"test-plugin/test-plugin/test-plugin/inject-x-stats": [Function inject-x-stats]}

      24 |     const plugin = config.lint.plugins[0];
      25 |
    > 26 |     expect(plugin.decorators?.oas3).toHaveProperty('test-plugin/inject-x-stats');
         |                                     ^
      27 |   });
      28 | });
      29 |

      at packages/core/src/config/__tests__/resolve-plugins.test.ts:26:37
      at fulfilled (packages/core/src/config/__tests__/resolve-plugins.test.ts:5:58)
```

(exact output can vary depending on test execution order).

## Check yourself

- [X] Code is linted

I didn't think this change affects workflows.

### DIFF
--- a/packages/core/src/config/__tests__/fixtures/plugin-config.yaml
+++ b/packages/core/src/config/__tests__/fixtures/plugin-config.yaml
@@ -1,0 +1,3 @@
+lint:
+  plugins:
+    - './plugin.js'
--- a/packages/core/src/config/__tests__/fixtures/plugin.js
+++ b/packages/core/src/config/__tests__/fixtures/plugin.js
@@ -1,0 +1,56 @@
+const id = 'test-plugin';
+
+/** @type {import('../../config').PreprocessorsConfig} */
+const preprocessors = {
+  oas2: {
+    'description-preprocessor': () => {
+      return {
+        Info(info) {
+          const title = info.title || 'API title';
+          info.description = `# ${title}\n\n${info.description || ''}`;
+        },
+      };
+    },
+  },
+};
+
+/** @type {import('../../config').CustomRulesConfig} */
+const rules = {
+  oas3: {
+    'openid-connect-url-well-known': () => {
+      return {
+        SecurityScheme(scheme, { location, report }) {
+          if (scheme.type === 'openIdConnect') {
+            if (!scheme.openIdConnectUrl?.endsWith('/.well-known/openid-configuration')) {
+              report({
+                message:
+                  'openIdConnectUrl must be a URL that ends with /.well-known/openid-configuration',
+                location: location.child('openIdConnectUrl'),
+              });
+            }
+          }
+        },
+      };
+    },
+  },
+};
+
+/** @type {import('../../config').DecoratorsConfig} */
+const decorators = {
+  oas3: {
+    'inject-x-stats': () => {
+      return {
+        Info(info) {
+          info['x-stats'] = { test: 1 };
+        },
+      };
+    },
+  },
+};
+
+module.exports = {
+  id,
+  preprocessors,
+  rules,
+  decorators,
+};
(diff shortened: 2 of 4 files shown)


RomanHotsiy (approved):
LGTM 👍

AlexVarchuk (approved):

RomanHotsiy:
Great job @mjpieters! 👏

PR #289, opened 2021-05-04
[Update] enum item validation with nullable property 

dinonuggies1:
## What/Why/How?

The linter will report an error when an enum item doesn't match the corresponding schema type. However, I believe there should be an exception when an item has a value of `- null` and `nullable: true` is set. This PR reflects this exception -- adding the logic necessary to evaluate the nullable property for both `Oas2Schema` and `Oas3Schema`

## Reference

https://swagger.io/docs/specification/data-models/enums/#nullable

## Testing

Test was included to show that the linter should not report an error when `nullable: true` and enum items contain a `null`

## Check yourself

- [X] Code is linted

### DIFF
--- a/packages/core/src/rules/common/__tests__/no-enum-type-mismatch.test.ts
+++ b/packages/core/src/rules/common/__tests__/no-enum-type-mismatch.test.ts
@@ -36,6 +36,39 @@ describe('Oas3 typed enum', () => {
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`Array []`);
   });
 
+  it('should not report on enum object if all items match type and enum is nullable', async () => {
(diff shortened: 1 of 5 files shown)


RomanHotsiy (approved):
LGTM 👍 

Great job! Thanks

RomanHotsiy:
We'll release it tomorrow.

dinonuggies1:
👋🏽  Hello!

I wanted to follow up to see if there were still plans to release a version with this change included? I noticed the latest publish on the npm registry was about a month ago. If not, no worries - just wanted to plan accordingly 😃  🙏🏽

RomanHotsiy:
Oops, I forgot about that. Sorry.

PR #293, opened 2021-05-07
fix: do not overwrite 'x-displayName' of a tag on join

rjelierse:
## What
When tags already specify `x-displayName`, setting it with the tag's name will probably result in setting a machine-readable value instead of the intended display name.

### Example
When joining documents that have `x-displayName` defined, it takes the `name` property of the tag regardless:

Source document:
```yaml
tags:
  - name: authentication
    x-displayName: Authentication
    description: Endpoints dealing with user authentication.
```

Resulting document:
```yaml
tags:
  - name: accounts_authentication
    x-displayName: authentication
    description: Endpoints dealing with user authentication.
```

Expected result:
```yaml
tags:
  - name: accounts_authentication
    x-displayName: Authentication
    description: Endpoints dealing with user authentication.
```

## Check yourself

- [x] Code is linted

### DIFF
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -147,7 +147,7 @@ packageVersion: string
         tag.description = addComponentsPrefix(tag.description, componentsPrefix!);
       }
       if (!joinedDef.tags.find((t: any) => t.name === entrypointTagName)) {
-        tag['x-displayName'] = tag.name;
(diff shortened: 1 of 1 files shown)


adamaltman (commented):
Looks good to me.

RomanHotsiy (approved):

RomanHotsiy:
Thanks!

Issue #297, opened 2021-05-18
paths-kebab-case rule isn't checking for snake_case


AntonKozachuk:
**Describe the bug**
The `paths-kebab-case` rule doesn't forbid snake_case paths, it only seems to detect camelCase.

**To Reproduce**
1. Add snake_case paths. 
2. Lint the spec.

**Expected behavior**
The `paths-kebab-case` rule detects the snake_case paths and shows an error or warning. 

**Logs**
N/a

**OpenAPI definition**
N/a

**`openapi-cli` Version(s)**
1.0.0-beta.39

**`Node.js` Version(s)**
v12.19.0

**Additional context**
None.

PR #298, opened 2021-05-18
fix: Improve path kebab case message and error on snake_case

adamaltman:
## What/Why/How?

The issues were reported by a user. The message didn't make sense, so it is fixed. And it didn't catch `snake_case` but did catch camelCase`.

## Testing

I added tests to the PR.

## Check yourself

- [ ] Code is linted (no, when I run prettier, it makes a lot of changes)
- [ ] Tested with redoc/reference-docs/workflows (shouldn't impact those)


RomanHotsiy (approved):

AlexVarchuk (approved):

AntonKozachuk:
Fixes #297.

Issue #286, opened 2021-04-27
Missing     "js-yaml": "^3.14.0" dependency in cli package


jandppw:
**Describe the bug**
In a project that is using js-yaml 4, `openapi bundle` fails

**To Reproduce**

- Create a project that has a dependency on `"js-yaml": "^4.0.0"`, and a dependency on `"@redocly/openapi-cli": "^1.0.0-beta.44",`
- run `npx openapi bundle`

```
bundling […]/openapi/index.yaml...
Something went wrong when processing […]/openapi/index.yaml:

  - Function yaml.safeDump is removed in js-yaml 4. Use yaml.dump instead, which is now safe by default..

(node:88437) UnhandledPromiseRejectionWarning: Error: Function yaml.safeDump is removed in js-yaml 4. Use yaml.dump instead, which is now safe by default.
    at Object.safeDump ([…]/node_modules/js-yaml/index.js:10:11)
    at Object.dumpBundle ([…]/node_modules/@redocly/openapi-cli/lib/utils.js:98:21)
    at Object.<anonymous> ([…]/node_modules/@redocly/openapi-cli/lib/commands/bundle.js:38:48)
    at Generator.next (<anonymous>)
    at fulfilled […]/node_modules/@redocly/openapi-cli/lib/commands/bundle.js:5:58)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:88437) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:88437) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

**Expected behavior**
No crash

**Logs**
NA

**OpenAPI definition**
NA

**`openapi-cli` Version(s)**
`1.0.0-beta.44`

**`Node.js` Version(s)**
`v14.16.1`

**Additional context**

The problem is in the `package.json` of `@redocly/openapi-cli`. Although `js-yaml` is used in that distribution (in `node_modules/@redocly/openapi-cli/lib/utils.js`, it is not mentioned in the `package.json`.

[...]

PR #306, opened 2021-05-31
chore: migrated to js-yaml@4

andriyl:
## What/Why/How?

fix: #286 

## Check yourself

- [x] Code is linted


andriyl:
fixed here: https://github.com/Redocly/openapi-cli/pull/403

PR #312, opened 2021-06-18
chore: update to ajv@8

andriyl:
## What/Why/How?
Migrate @Redoc/ajv from v6.12.3 to v8.6.0, it supports OpenAPI 3.1.0

## Check yourself

- [x] Code is linted
- [x] Tested with redoc/reference-docs


packages/core/src/rules/ajv.ts
```diff
@@ -14,20 +11,17 @@ export function releaseAjvInstance() {
 function getAjv(resolve: ResolveFn<any>, disallowAdditionalProperties: boolean) {
   if (!ajvInstance) {
     ajvInstance = new Ajv({
-      schemaId: 'auto',
+      schemaId: '$id',
```
RomanHotsiy, line 14: @andriyl review all the new and deprecated options and adjust it.

E.g. we do need `discriminator: true`

RomanHotsiy (changes requested):
Left a comment

RomanHotsiy (approved):

Issue #311, opened 2021-06-14
 RangeError: Maximum call stack size exceeded


lesyk-lesyk:
**Describe the bug**
`Maximum call stack size exceeded` error when schema fields point to themself.

![image](https://user-images.githubusercontent.com/16646570/121892085-f487f680-cd24-11eb-8487-071cf7b771b6.png)

**To Reproduce**
1. Given this `.redocly.yaml` file:
```yaml
lint:
  extends:
    - recommended
  rules:
    no-invalid-media-type-examples: error

``` 

2. And this `definition.json` OpenAPI file: 
```JSON
{
  "openapi": "3.0.1",
  "info": {
    "title": "TEST",
    "description": "TEST",
    "version": "v1"
  },
  "paths": {
    "/test": {
      "get": {
        "operationId": "test",
        "responses": {
          "200": {
            "description": "Test",
            "content": {
              "application/json": {
                "schema": { "$ref": "#/components/schemas/Pet" },
                "example": {}
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Pet": {
        "type": "object",
        "oneOf": [{ "$ref": "#/components/schemas/Pet" }]
      }
    }
  }
}

```
3. Run this command with these arguments:
`npx @redocly/openapi-cli lint definition.json`

`UnhandledPromiseRejectionWarning: RangeError: Maximum call stack size exceeded`

**Expected behavior**
No Unhandled Promise Rejection

**`openapi-cli` Version(s)**
```
$npx @redocly/openapi-cli --version
1.0.0-beta.50
```

**`Node.js` Version(s)**
```
$ node --version
v12.13.0
```

**Additional context**
Seems like error in package: [@redocly/ajv](https://github.com/Redocly/ajv)

RomanHotsiy:
@AntonKozachuk let the prospect/customer know, see https://github.com/Redocly/reference-docs/issues/208

PR #316, opened 2021-06-30
fix: Maximum call stack size exceeded issue in ajv lib

andriyl:
## What/Why/How?
fix: #311 

## Check yourself
- [x] Code is linted
- [x] Tested with redoc/reference-docs


__tests__/no-ref-recursion-error/snapshot.js
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
RomanHotsiy, line 1: rename to `no-invalid-media-type-examples-recursion-oneOf`

RomanHotsiy (approved):

PR #337, opened 2021-08-12
Docs: added a new installation page for the openapi-cli tool

bandantonio:
## What/Why/How?

added a new installation page for the openapi-cli tool

## Check yourself

- [ ] Code is linted

## Security


docs/installation.md
```diff
@@ -0,0 +1,187 @@
...
+[3/4] 🔗  Linking dependencies...
+[4/4] 🔨  Building fresh packages...
+
+success Installed "@redocly/openapi-cli@1.0.0-beta.54" with binaries:
+      - openapi
+✨  Done in 6.90s.
+```
+
+### Verify global installation
+
```
swapnilogale, line 41: I reckon this could be a H2 (## Verify global installation)

docs/installation.md
```diff
@@ -0,0 +1,187 @@
...
+validating petstore.yaml...
+petstore.yaml: validated in 43ms
+
+Woohoo! Your OpenAPI definition is valid. 🎉
+```
+
+### Docker
+
+To give the Docker container access to the OpenAPI definition files, you need to mount the containing directory as a volume. Assuming the OAS definition is rooted in the current working directory, the commnd will be the following:
+
```
swapnilogale, line 116: Minor typo: `command`

swapnilogale (commented):
This looks good Anton. Well done.

skadinna (dismissed):
This is good. I think the only issue is with the old admonitions syntax - instead of HTML tags, we now use `:::` as described [here in the docs](https://redoc.ly/docs/developer-portal/markdown/#admonitions). But we can easily fix that after the merge.

docs/installation.md
```diff
@@ -0,0 +1,187 @@
+# Installing OpenAPI-CLI
```
skadinna, line 1:
```suggestion
# Installing OpenAPI CLI
```

bandantonio:
> This is good. I think the only issue is with the old admonitions syntax - instead of HTML tags, we now use `:::` as described [here in the docs](https://redoc.ly/docs/developer-portal/markdown/#admonitions). But we can easily fix that after the merge.

@skadinna I fixed it. My intention to use the HTML version was the title that can be omitted. Sometimes it looks too verbose

skadinna:
> > This is good. I think the only issue is with the old admonitions syntax - instead of HTML tags, we now use `:::` as described [here in the docs](https://redoc.ly/docs/developer-portal/markdown/#admonitions). But we can easily fix that after the merge.
> 
> @skadinna I fixed it. My intention to use the HTML version was the title that can be omitted. Sometimes it looks too verbose

Yeah, the problem is that the portal will stop supporting the old syntaxt at some point, so all those admonitions that use it will be broken. :(

As for the admonitions title, you don't have to make it anything special, just put one word like "Tip", "Info" or "Important". That's how we usually do it in the rest of the docs. ;)

skadinna (approved):

swapnilogale (approved):

PR #338, opened 2021-08-13
Docs: added information about openapi-cli benefits and advantages

bandantonio:
## What/Why/How?

Added information about openapi-cli benefits and advantages

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -33,7 +33,7 @@ Currently, Redocly OpenAPI CLI supports these features:
 
 ## What makes Redocly OpenAPI CLI different
 
-OpenAPI CLI is your all-in-one Swiss army knife that can fulfill your needs when designing, defining, and working with OpenAPI definitions.
+As you can see from the previous section, OpenAPI CLI is your all-in-one swiss army knife that can fulfill all your needs when designing, defining, and working with OpenAPI definitions.
 
 There are four pillars that make Redocly OpenAPI CLI great - **performance**, **bundling**, **linting**, and **extensibility**. To better understand why you need to
 look at the OpenAPI authoring process from a higher level.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -33,29 +33,27 @@ Currently, Redocly OpenAPI CLI supports these features:
 
 ## What makes Redocly OpenAPI CLI different
 
-Unlike other OpenAPI validators, Redocly OpenAPI CLI defines the possible type tree of a valid OpenAPI definition and then traverses it. This approach is very similar to how compilers work, and results in major performance benefits over other approaches.
+As you can see from the previous section, OpenAPI CLI is your all-in-one swiss army knife that can fulfill all your needs when designing, defining, and working with OpenAPI definitions.
```
adamaltman, line 36:
```suggestion
OpenAPI CLI is your all-in-one Swiss army knife that can fulfill your needs when designing, defining, and working with OpenAPI definitions.
```
I think the transition can be perceived negatively. Also, I removed "all" because there is still more to do...

adamaltman (commented):
This is a great improvement!!! 🎉

docs/index.md
bandantonio: Thanks @adamaltman. No objections. Fixed.

docs/index.md
```diff
@@ -33,29 +33,27 @@ Currently, Redocly OpenAPI CLI supports these features:
...
-graph LR
-    preprocessors[Preprocessors]
-    rules[Rules]
-    decorators[Decorators]
-    preprocessors --> rules --> decorators
-```
+That's why people proceed with the multi-file approach, where you define the main structure of the API in the root definition file and everything else that can be reused or segmented into smaller units is located in separate files.
+
+The problem with the multi-file approach is that many existing tools offer multi-file support as the only feature, meaning that you will have yet another tool to install and maintain. OpenAPI CLI has a strong advantage here over other tools, as it bundles files automatically and it's just one of the powerful features it provides you with.
+
```
swapnilogale, line 53: This is good Anton, however, I am wondering if we can simplify this further via a diagram or a table to make it easier. If you have worked with [Mermaid diagrams](https://mermaid-js.github.io/mermaid/#/) before, maybe you can try and put this in a diagram?

swapnilogale (commented):
Good text overall, just a couple of suggestions and changes.

docs/index.md
```diff
@@ -33,29 +33,27 @@ Currently, Redocly OpenAPI CLI supports these features:
 
 ## What makes Redocly OpenAPI CLI different
 
-Unlike other OpenAPI validators, Redocly OpenAPI CLI defines the possible type tree of a valid OpenAPI definition and then traverses it. This approach is very similar to how compilers work, and results in major performance benefits over other approaches.
+OpenAPI CLI is your all-in-one Swiss army knife that can fulfill your needs when designing, defining, and working with OpenAPI definitions.
```
swapnilogale, line 37: Minor nitpick - Can probably leave out `that can fulfill your needs` from this sentence.

swapnilogale (commented):
I like your idea of adding a `What makes Redocly OpenAPI CLI different` section here. I reckon maybe adding a visual element (table, diagram) will also make it stand out for people trying this out for the first time.

docs/index.md
bandantonio: I don't mind adding some visuals, however, I tried with mermaid, and it was really hard to translate the information about features we are talking about into a diagram. A diagram is a specific tool and when it is used in a non-expected way, it may confuse even more.

bandantonio: Tried to resolve this case by adding two code block sections showing openapi definitions in single-file and multi-file approaches

swapnilogale: This works great Anton. I like this approach.

skadinna (changes requested):
Some phrasing adjustments still needed, but otherwise this looks great!

skadinna (approved):

swapnilogale (approved):

adamaltman (approved):

PR #343, opened 2021-08-19
feat: return type of reference in bundleDocument

AlexVarchuk:
## What/Why/How?
on vs code extension, we need the types of file reference.  We use it in the autocomplete.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -80,6 +80,7 @@ export async function bundleDocument(opts: {
   const ctx: BundleContext = {
     problems: [],
     oasVersion: oasVersion,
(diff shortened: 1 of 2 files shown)


packages/core/src/bundle.ts
```diff
@@ -80,6 +80,7 @@ export async function bundleDocument(opts: {
   const ctx: BundleContext = {
     problems: [],
     oasVersion: oasVersion,
+    refTypes: new Map(),
```
jacobator, line 83:
```suggestion
    refTypes: new Map<string, NormalizedNodeType>(),
```

andriyl (approved):

PR #345, opened 2021-08-25
docs: OpenAPI CLI commands home page

bandantonio:
## What/Why/How?

Removed redundancy from the OpenAPI CLI commands home page. Made the page more concise.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/commands/index.md
+++ b/docs/commands/index.md
@@ -22,8 +22,10 @@ To define the behavior of the commands, you can use configuration files:
 
 For more information, refer to the [configuration section](../configuration/index.mdx).
 
-:::warning
-The CLI tool looks for configuration files in the current working directory. If it detects them, it will use the options set in those configuration files for the commands. Learn more about the [configuration file structure and options](../configuration/index.mdx).
+:::attention Note
+
+The CLI tool looks for configuration files in the current working directory. If it detects them, it will use the options set in those configuration files for the commands. Learn more about the [configuration file structure and options](../configuration/configuration-file.mdx).
 
 When executing any of the commands, you can override the default configuration file by providing a path to another configuration file with the `--config` option.
+
 :::
--- a/docs/commands/index.md
+++ b/docs/commands/index.md
@@ -5,22 +5,15 @@ tocMaxDepth: 2
 
 Redocly OpenAPI CLI currently supports the following commands:
 
-- [bundle](bundle.md)
-- [join](join.md)
-- [lint](lint.md)
-- [login](login.md)
-- [logout](logout.md)
-- [preview-docs](preview-docs.md)
-- [push](push.md)
-- [split](split.md)
-- [stats](stats.md)
+commands            |                                 | 
+--------------------|---------------------------------|------------------
+[bundle](bundle.md) | [split](split.md)               | [join](join.md)
+[lint](lint.md)     | [preview-docs](preview-docs.md) |  [stats](stats.md)
+[login](login.md)   | [logout](logout.md)             | [push](push.md)
 
+To define the behavior of the commands, you can use configuration files:
 
-The following configuration files define the behavior of the CLI tool:
+- `.redocly.yaml` - define the location of your root files, linting rules, and reference docs configuration information.
+- `.redocly.lint-ignore.yaml` - ignore specific lint messages.
 
-- `.redocly.yaml` - used to define the location of your root files, linting rules, and reference docs configuration information.
-- `.redocly.lint-ignore.yaml` - used to ignore specific lint messages.
-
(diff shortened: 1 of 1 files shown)


docs/commands/index.md
```diff
@@ -5,22 +5,15 @@ tocMaxDepth: 2
...
-- [logout](logout.md)
-- [preview-docs](preview-docs.md)
-- [push](push.md)
-- [split](split.md)
-- [stats](stats.md)
+commands            |                                 | 
+--------------------|---------------------------------|------------------
+[bundle](bundle.md) | [split](split.md)               | [join](join.md)
+[lint](lint.md)     | [preview-docs](preview-docs.md) |  [stats](stats.md)
+[login](login.md)   | [logout](logout.md)             | [push](push.md)
```
adamaltman, line 12: The more I look at this, the more I wonder about why not leave it as a bullet list?

RomanHotsiy: I agree here, the table doesn't make too much sense I believe as there is just a single dimension.

I think it just saves some vertical space.

skadinna: Seconded (or thirded :smile: ). To the readers, there's no clear or obvious reason why these commands are grouped in the table the way they are. 

If the columns had titles, that _might_ help a little...if the titles indicated the grouping criteria or some shared property of the commands listed in each column. 

However, that means that every new command we develop needs to fit into the table, and we'll have to spend extra time thinking where to put each new command so that the table still makes sense. That seems like too much cost for little benefit, so it would be easier to just leave it as a plain alphabetically sorted list.

adamaltman (dismissed):

docs/commands/index.md
```diff
@@ -5,22 +5,19 @@ tocMaxDepth: 2
...
-The following configuration files define the behavior of the CLI tool:
+- `.redocly.yaml` - define the location of your root files, linting rules, and reference docs configuration information.
+- `.redocly.lint-ignore.yaml` - ignore specific lint messages.
 
-- `.redocly.yaml` - used to define the location of your root files, linting rules, and reference docs configuration information.
-- `.redocly.lint-ignore.yaml` - used to ignore specific lint messages.
-
-The CLI tool looks for configuration files in the current working directory. If it detects them, it will use the options set in those configuration files for the commands. Learn more about the [configuration file structure and options](../configuration/index.mdx).
-
-When executing any of the commands, you can override the default configuration file by providing a path to another configuration file with the `--config` option.
```
skadinna, line 26: suggestion: The part about overriding default conf and using custom configuration files with commands could be useful to keep here since it applies to all commands. Maybe we can add it as an admonition (tip or note), if you don't want to keep it in the main text?

docs/commands/index.md
```diff
@@ -5,22 +5,19 @@ tocMaxDepth: 2
...
+For more information about configuration, please refer to the [corresponding](../configuration/index.mdx) section.
```
skadinna, line 23:
```suggestion
For more information, refer to the [configuration section](../configuration/index.mdx).
```

skadinna: This suggestion is just to simplify the sentence a little bit, and to make the anchor text more relevant. In our docs, we try to avoid "empty", generic words and phrases like "here" , "this",  "related" as anchor text, and instead make it more descriptive whenever possible. This is supposedly good for SEO :) but it also helps the readers know where exactly the link will take them.

skadinna (approved):
This looks good! :+1:  I left just a few small suggestions.

bandantonio:
Seems like this pr is ready to be reviewed and merged

docs/commands/index.md
```diff
@@ -5,22 +5,25 @@ tocMaxDepth: 2
...
 
-- `.redocly.yaml` - used to define the location of your root files, linting rules, and reference docs configuration information.
-- `.redocly.lint-ignore.yaml` - used to ignore specific lint messages.
+For more information, refer to the [configuration section](../configuration/index.mdx).
 
+:::warning
 The CLI tool looks for configuration files in the current working directory. If it detects them, it will use the options set in those configuration files for the commands. Learn more about the [configuration file structure and options](../configuration/index.mdx).
 
 When executing any of the commands, you can override the default configuration file by providing a path to another configuration file with the `--config` option.
+:::
```
skadinna, line 29:
```suggestion
:::attention Note

The CLI tool looks for configuration files in the current working directory. If it detects them, it will use the options set in those configuration files for the commands. Learn more about the [configuration file structure and options](../configuration/configuration-file.mdx).

When executing any of the commands, you can override the default configuration file by providing a path to another configuration file with the `--config` option.

:::
```

skadinna (approved):

PR #347, opened 2021-08-26
Strip query parameters from request URL in preview server

edkelly-ovo:
## What/Why/How?

The file `packages/cli/src/commands/preview-docs/preview-server/preview-server.ts` contains a security defect that allows directory traversal of the complete file path on a server and the download of files via query parameters on the request URI.

Example: 

http://localhost:8080/jsp/help-sb-download.jsp?sbFileName=../../../../.redocly.yaml

Applying a `request.url.split('?')[0]` at the start of the preview server function has stopped this from occurring.

## Reference

https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts#L99

The usage of `path.resolve()` on this line can result in arbitrary download of any file on the server if query params are exploited.

## Testing

Fix tested manually

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/__tests__/utils.test.ts
+++ b/packages/cli/src/__tests__/utils.test.ts
@@ -1,4 +1,7 @@
-import { slash } from '../utils';
+import { isSubdir, slash } from '../utils';
+
+jest.mock("os");
+
 
 describe('slash path', () => {
   it('can correctly slash path', () => {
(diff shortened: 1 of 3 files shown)


RomanHotsiy:
@edkelly-ovo I can't actually reproduce the bug you describe. And I can't understand how it can happen from code. `path.resolve` will keep the `?` in place so it will just return 404. 

Can you provide more details?

edkelly-ovo:
@RomanHotsiy if I navigate to http://localhost:8080/jsp/help-sb-download.jsp?sbFileName=../../../../../../../../../../../../etc/passwd in Google Chrome on OSX 11.5, the server triggers a download of my /etc/passwd file.

I think this is happening because `path.resolve` is being provided the following arguments: 

* ./static
* ./jsp/help-sb-download.jsp?sbFileName=../../../../../../../../../../../../etc/passwd
which is yielding the path `/etc/passwd` when run.

`path.resolve` is interpreting (correctly) the second argument as a path to a file (which exists on disk). This file is then being sent in response via the `respondWithGzip` function.

If I strip off the query params from `request.url`, this stops the directory traversal from being possible

edkelly-ovo:
@RomanHotsiy any update on this please?

RomanHotsiy:
So I was able to reproduce it, it looks that in order to exploit it the file with the question mark must actually exists on the filesystem: `help-sb-download.jsp?sbFileName=..`.

I would implement a more robust solution by checking if the file path is within the current working dir. I pushed changes to this PR. @edkelly-ovo could you please verify it resolved the issue?

edkelly-ovo:
Hi @RomanHotsiy yes that works, thanks! I've reverted my original code which was a bit heavy handed, your solution is much more elegant.

I spotted a typo in `preview-server.ts` on the `isSubdir` import which I've also fixed :)

RomanHotsiy (approved):

edkelly-ovo:
Thanks @RomanHotsiy when is the next release going out please?

RomanHotsiy:
https://github.com/Redocly/openapi-cli/pull/366

PR #350, opened 2021-08-26
Docs: revised openapi-cli bundle command

bandantonio:
## Check yourself

- [ ] Code is linted

## Security


docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
-  --ext                Bundle file extension.   [choices: "json", "yaml", "yml"]
-  --skip-rule          Ignore certain rules.                             [array]
-  --skip-preprocessor  Ignore certain preprocessors.                     [array]
-  --skip-decorator     Ignore certain decorators.                        [array]
-  --dereferenced, -d   Produce fully dereferenced bundle.              [boolean]
-  --force, -f          Produce bundle output even when errors occur.   [boolean]
-  --config             Specify path to the config file.                 [string]
-```
+:::success Tip
+To learn more about preprocessors, rules, and decorators, please refer to [this](../custom-rules.md) document.
```
skadinna, line 12: suggestion: Can you please change the link anchor text to be more descriptive instead of just "this"? Something like "refer to the [custom rules page](link)" would be fine.

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+openapi bundle <entrypoints>...
+openapi bundle <entrypoints> [--max-problems=<n>]
+openapi bundle <entrypoints> [--lint] [--config=<path>]
+openapi bundle <entrypoints>... -o <outputName> --ext <ext>
+openapi bundle --version
+```
+
+## Options
+
+option                 | type      | required?    | default     | description
```
skadinna, line 27: I like this solution with the table a lot! :clap: :clap: 

Can you just please capitalize column titles? Option, Type, Required (you don't need the question mark), Default, Description. The reason for requesting this change is because we format column titles like that in the rest of our docs, so it would be good to keep it consistent.

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+-----------------------|:---------:|:------------:|:-----------:|------------
+`entrypoints`          | `array`   | yes          | `[]`        | Array of API root definition filenames that need to be bundled. Instead of full paths, you can use aliases assigned in your `apiDefinitions` within your `.redocly.yaml` configuration file as entrypoints.
```
skadinna, line 29:
```suggestion
`entrypoints`          | `array`   | yes          | `[]`        | Array of API root definition filenames that need to be bundled. Instead of full paths, you can use aliases assigned in the `apiDefinitions` section within your `.redocly.yaml` configuration file as entrypoints.
```

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--config`             | `string`  | no           | -           | Specify path to the config file
```
skadinna, line 30: question: Would it be possible to provide some more details about this path - perhaps add an example? Does the path have to be absolute or relative; can it be both?

docs/commands/bundle.md
skadinna: suggestion: This first description has a fullstop at the end, but other descriptions don't. If we're going to write the descriptions as full sentences, we should probably use fullstops everywhere for consistency.

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--dereferenced`, `-d` | `boolean` | no           | -           | Generate fully dereferenced bundle
```
skadinna, line 31: suggestion: You asked a really good question in Slack about what "fully dereferenced" means. I think it would be useful to explain that here because new users who are not familiar with this terminology will likely be confused about it, too. So it would help to have it explained; a sentence or two should be just enough.

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--ext`                | `string`  | no           | `yaml`      | Specify bundled file extension.<br />**Possible values:** `json`, `yaml`, `yml`
+`--force`, `-f`        | `boolean` | no           | -           | Generate bundle output even when errors occur
+`format`               | `string`  | no           | `codeframe` | Format for the output.<br />**Possible values:** `codeframe`, `stylish`
```
skadinna, line 34: question: What does `codeframe` do? What does `stylish` do? What's the difference between them; when to use one, and when to use the other?

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--help`               | `boolean` | no           | -           | Show help
+`--lint`               | `boolean` | no           | `false`     | Lint definition files
```
skadinna, line 36: question: Which criteria are used for linting if this option is set to `true`? Does it use the lint settings from the config file?

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--max-problems`       | `number`  | no           | 100         | Truncate output to display the specified maximum number of problems
```
skadinna, line 37: question: What's the allowed range of values here (min to max)?

docs/commands/bundle.md
```diff
@@ -1,117 +1,91 @@
...
+`--output`, `-o`       | `string`  | no           | -           | Name or folder for the bundle file. For example, `-o bundle.yaml` or `-o ./openapi`.<li>If you don't specify the extension, `.yaml` will be used by default.</li><li>If the specified folder doesn't exist, it will be created automatically.</li><br />**If the file specified as the bundler's output already exists, it will be overwritten**
+`--skip-decorator`     | `array`   | no           | -           | Ignore certain decorators
+`--skip-preprocessor`  | `array`   | no           | -           | Ignore certain preprocessors
+`--skip-rule`          | `array`   | no           | -           | Ignore certain rules
```
skadinna, line 41: question: Can the user specify which decorators/preprocessors/rules will be ignored? If yes, could we show an example how to do that?

skadinna (changes requested):
This is good work - the table really improves the readability of supported options :+1: 

However, I think we should provide more details around each option to help users understand how to work with them. I left some questions and suggestions for that.

docs/commands/bundle.md
bandantonio: I believe you can find answers for that in the PR for the [lint command](https://github.com/Redocly/openapi-cli/pull/360)

skadinna: So the idea is that the users should look there to find that info, too? If that's the case, the least we can do is provide a link.

docs/commands/bundle.md
bandantonio: Please check the content starting from the line [#77](https://github.com/Redocly/openapi-cli/pull/350/files#diff-00c97ff1398f245dc4a49f3ea92c5827a829e8332501f762fd1720ea05069402R77). If believe this is a short, yet descriptive example.

skadinna: Circling back on this - yeah, that example does exactly what I was looking for :+1: Could you add a sentence to the table with a link to that section? Similar to how we have those "see section below" links in tables for other commands.

docs/commands/bundle.md
skadinna: If we don't have a conclusive answer to this, I think you can just leave the text as it is, without extra details about min-max. In that case, we should open a GitHub issue to investigate and document this at some later point, outside of the GSOD project. Does that sound OK?

docs/commands/bundle.md
skadinna: Update: we can treat this the same way as the max-problems values. If you were not able to get an answer to this from the developers, please open an issue for documenting it at some later point.

docs/commands/bundle.md
bandantonio: Added missing example about custom config file

docs/commands/bundle.md
bandantonio: Added information about undocumented `--format=json` option. `codeframe` and `stylish` left blank. If it doesn't look good, we can leave only `json` atm

docs/commands/bundle.md
bandantonio: Left as is atm. I will create an issue with undocumented/need clarification parts

docs/commands/bundle.md
bandantonio: Yup, totally OK. I will create an issue with undocumented/need clarification parts. However, I hope to receive some clarifications from developers during the GSOD project :)

docs/commands/bundle.md
bandantonio: Added links to the section

docs/commands/bundle.md
```diff
@@ -1,117 +1,163 @@
...
-graph LR
-    preprocessors[Preprocessors]
-    rules[Rules]
-    decorators[Decorators]
-    preprocessors --> rules --> decorators
-```
-
-### `bundle` usage
+:::success Tip
+To learn more about preprocessors, rules, and decorators, please refer to the [custom rules page](../custom-rules.md).
```
skadinna, line 12:
```suggestion
To learn more about preprocessors, rules, and decorators, please refer to the [custom rules page](../resources/custom-rules.md).
```

docs/commands/bundle.md
```diff
@@ -1,117 +1,163 @@
...
+openapi bundle --skip-rule=no-sibling-refs,no-parent-tags
 ```
+
+```bash Skip decorators
 openapi bundle --skip-decorator=generate-code-samples,remove-internal-operations
 ```
 
-Learn more about [decorators](../resources/custom-rules.md).
+:::success Tip
+To learn more about preprocessors, rules, and decorators, please refer to the [custom rules page](../custom-rules.md).
```
skadinna, line 162:
```suggestion
To learn more about preprocessors, rules, and decorators, please refer to the [custom rules page](../resources/custom-rules.md).
```

skadinna (dismissed):

docs/commands/bundle.md
```diff
@@ -1,117 +1,163 @@
...
-Learn more about [preprocessors](../resources/custom-rules.md).
+```bash output
+...
+```
 
-#### Skip rule
+In this format, `bundle` shows the file name, line number, and column where the problem occurred. However, the output is compressed and omits other contexts and suggestions.
 
-You may want to skip specific rules upon running the command.
+#### Json
```
adamaltman, line 114: question: I wonder if it should be JSON (all caps)?

docs/commands/bundle.md
```diff
@@ -1,117 +1,163 @@
...
+openapi bundle --skip-rule=no-sibling-refs,no-parent-tags
 ```
+
+```bash Skip decorators
 openapi bundle --skip-decorator=generate-code-samples,remove-internal-operations
 ```
 
-Learn more about [decorators](../resources/custom-rules.md).
+:::success Tip
+To learn more about preprocessors, rules, and decorators, please refer to the [custom rules page](../resources/custom-rules.md).
```
adamaltman, line 162: question: This same sentence is near the start of the page. Do we need it both places?

docs/commands/bundle.md
skadinna, line 114:
```suggestion
#### JSON
```

skadinna: Absolutely. Thanks

docs/commands/bundle.md
skadinna: I would remove it from here, but l'm interested in what @bandantonio thinks about this

swapnilogale (commented):
Just made a couple of style fixes (essentially we avoid using words like "please", "kindly" in our docs). Looks good otherwise. @adamaltman and @skadinna have left 1 comment to address, but otherwise this is good to go.

docs/commands/bundle.md
bandantonio: @adamaltman The link appears here because if you want to know more information about custom rules, you have to scroll to the top and search for that link somewhere in the text. When you can immediately proceed to the extended section right from the place where some basics are described, it is actually a good idea. Anyway, if you or @skadinna think that it is redundant, I will remove one of them.

bandantonio: I think we can leave both tips as is - they look pretty organic.

I believe we can merge this PR.

adamaltman (approved):

PR #352, opened 2021-08-30
feat: add function for lint redocly config file

AlexVarchuk:
## What/Why/How?
Add function for lint .redocly config file.
In the near future, we should add the CLI command for it. And improve types.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/__tests__/lint.test.ts
+++ b/packages/core/src/__tests__/lint.test.ts
@@ -1,8 +1,8 @@
 import { outdent } from 'outdent';
 
-import { lintFromString } from '../lint';
+import { lintFromString, LintDocumentConfig } from '../lint';
 import { loadConfig } from '../config/load';
-import { replaceSourceWithRef } from '../../__tests__/utils';
+import { parseYamlToDocument, replaceSourceWithRef } from '../../__tests__/utils';
 
 describe('lint', () => {
   it('lintFromString should work', async () => {
@@ -34,6 +34,54 @@ describe('lint', () => {
             },
           ],
           "message": "Expected type \`License\` (object) but got \`string\`",
+          "ruleId": "spec",
+          "severity": "error",
+          "suggest": Array [],
+        },
+      ]
+    `);
+  });
+
+  it('lintDocumentConfig should work', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+      lint:
+        plugins:
+          - './local-plugin.js'
+        extends:
+          - recommended
+          - local/all
+        rules:
+          operation-2xx-response: warn
+          no-invalid-media-type-examples: error
+          path-http-verbs-order: error
+          boolean-parameter-prefixes: off
+      referenceDocs:
+        showConsole: true
+        layout:
+          scope: section
+        routingStrategy: browser
+        theme:
+          rightPanel:
+            backgroundColor: '#263238'
+          links:
+            color: '#6CC496'
+      `,
+      '',
+    );
+    const results = await LintDocumentConfig({document});
+
+    expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "location": Array [
+            Object {
+              "pointer": "#/",
+              "reportOnKey": true,
+              "source": "",
+            },
+          ],
+          "message": "The field \`apiDefinitions\` must be present on this level.",
           "ruleId": "spec",
           "severity": "error",
           "suggest": Array [],
--- a/packages/core/src/__tests__/lint.test.ts
+++ b/packages/core/src/__tests__/lint.test.ts
@@ -1,6 +1,6 @@
 import { outdent } from 'outdent';
 
-import { lintFromString, LintDocumentConfig } from '../lint';
+import { lintFromString, lintConfig } from '../lint';
 import { loadConfig } from '../config/load';
 import { parseYamlToDocument, replaceSourceWithRef } from '../../__tests__/utils';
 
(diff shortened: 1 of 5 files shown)


packages/core/src/__tests__/lint.test.ts
```diff
@@ -41,4 +41,52 @@ describe('lint', () => {
...
+      Array [
+        Object {
+          "location": Array [
+            Object {
+              "pointer": "#/",
+              "reportOnKey": true,
+              "source": "",
+            },
+          ],
+          "message": "The field \`apiDefinitions\` must be present on this level.",
```
RomanHotsiy, line 84: `apiDefinitions` is not required.

packages/core/src/lint.ts
```diff
@@ -89,3 +91,46 @@ export async function lintDocument(opts: {
   });
   return ctx.problems.map((problem) => config.addProblemToIgnore(problem));
 }
+
+export async function LintDocumentConfig(opts: {
```
RomanHotsiy, line 95: should start from lowercase, also, rename to simply `lintConfig`

packages/core/src/lint.ts
```diff
@@ -89,3 +91,46 @@ export async function lintDocument(opts: {
...
+
+  // TODO: maybe need deprecate this part with preprocessors and regularRules.
+  const oasMajorVersion = openAPIMajor(oasVersion);
+  const rules = config.getRulesForOasVersion(oasMajorVersion);
+  const types = normalizeTypes(configTypes, config);
+  const preprocessors = initRules(rules as any, config, 'preprocessors', oasVersion);
+  const regularRules = initRules(rules as any, config, 'rules', oasVersion);
+  const normalizedVisitors = normalizeVisitors([...preprocessors, ...regularRules], types);
+
+  const resolvedRefMap = await resolveDocument({
```
RomanHotsiy, line 121: I believe there is no need to resolve here, we dont' have $ref in config.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -0,0 +1,128 @@
+import { NodeType } from '.';
+
+const ConfigRoot: NodeType = {
+  properties: {
+    apiDefinitions: { type: 'object', additionalProperties: { type: 'string' } },
+    lint: 'Lint',
+    referenceDocs: 'ReferenceDocs',
+  },
+  required: ['apiDefinitions'],
```
RomanHotsiy, line 9: this is not required

packages/core/src/types/redocly-yaml.ts
```diff
@@ -0,0 +1,128 @@
...
+    oas3_1Rules: { type: 'object'},
+    preprocessors: { type: 'object'},
+    oas2Preprocessors: { type: 'object'},
+    oas3_0Preprocessors: { type: 'object'},
+    oas3_1Preprocessors: { type: 'object'},
+    decorators: { type: 'object'},
+    oas2Decorators: { type: 'object'},
+    oas3_0Decorators: { type: 'object'},
+    oas3_1Decorators: { type: 'object'},
+    resolve: {
```
RomanHotsiy, line 37: Don't do deep types (deeper than 2 for arrays), it's not supported by the spec rule I believe (not 100% sure). Instead have named types.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -0,0 +1,128 @@
...
+const ConfigRoot: NodeType = {
+  properties: {
+    apiDefinitions: { type: 'object', additionalProperties: { type: 'string' } },
+    lint: 'Lint',
+    referenceDocs: 'ReferenceDocs',
+  },
+  required: ['apiDefinitions'],
+};
+
+const Lint: NodeType = {
```
RomanHotsiy, line 12: I would name it as `ConfigLint`, same for rest

packages/core/src/lint.ts
```diff
@@ -89,3 +91,46 @@ export async function lintDocument(opts: {
...
+    oasVersion: oasVersion,
+  };
+
+  // TODO: maybe need deprecate this part with preprocessors and regularRules.
+  const oasMajorVersion = openAPIMajor(oasVersion);
+  const rules = config.getRulesForOasVersion(oasMajorVersion);
+  const types = normalizeTypes(configTypes, config);
+  const preprocessors = initRules(rules as any, config, 'preprocessors', oasVersion);
+  const regularRules = initRules(rules as any, config, 'rules', oasVersion);
+  const normalizedVisitors = normalizeVisitors([...preprocessors, ...regularRules], types);
```
RomanHotsiy, line 119: This is too verbose for simple case and also would be great to skip oasVersion hack.

Try this:

```suggestion
  // import specRule somewhere
  const rules = [ { severity: 'error', ruleId: 'spec', visitor: specRule({severity: 'error'}) } ];
  const normalizedVisitors = normalizeVisitors(rules, types);
```

Looks like you won't need config at all too.

RomanHotsiy (changes requested):
Left a few comments.

packages/core/src/lint.ts
AlexVarchuk: I should be more accurate. Thanks

packages/core/src/lint.ts
AlexVarchuk: Thanks, it very helped me. I thought about it but couldn't implement it.

andriyl (approved):

RomanHotsiy (dismissed):

RomanHotsiy (approved):

PR #353, opened 2021-08-31
Docs: revised join command

bandantonio:
## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/commands/join.md
+++ b/docs/commands/join.md
@@ -38,7 +38,7 @@ openapi join --version
 
 Option                               | Type               | Required     | Default     | Description
 -------------------------------------|:------------------:|:------------:|:-----------:|------------
-`entrypoints`                        | `array`, wildcard  | yes          | `[]`, -     | 1. Array of paths to API definition files that you want to join. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder
+`entrypoints`                        | `array`, wildcard  | yes          | -           | 1. Array of paths to API definition files that you want to join. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder
 `--help`                             | `boolean`          | no           | -           | Show help
 `--lint`                             | `boolean`          | no           | `false`     | Lint definition files
 `--prefix-components-with-info-prop` | `string`           | no           | -           | Prefix components with property value from info object. See the [prefix-components-with-info-prop section](#prefix-components-with-info-prop) below
--- a/docs/commands/join.md
+++ b/docs/commands/join.md
@@ -11,7 +11,9 @@ The `join` command is considered an experimental feature. This means it's still 
 The `join` command supports OpenAPI 3.x definitions only.
 :::
 
-With the `join` command, Redocly OpenAPI CLI can combine two or more API definition files into one so that maintainers of multiple API definitions can benefit from storing each endpoint as a standalone API definition file. However, this approach is not supported by the majority of OpenAPI tools, as they require a single API definition file.
+Maintainers of multiple API definitions can benefit from storing each endpoint as a standalone API definition file. However, this approach is not supported by the majority of OpenAPI tools, as they require a single API definition file.
+
+With Redocly OpenAPI CLI, you can solve this problem by using the `join` command that can combine two or more API definition files into a single one.
 
 To easily distinguish the origin of OpenAPI objects and properties, you can optionally instruct the `join` command to append custom prefixes to them.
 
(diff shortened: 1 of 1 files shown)


docs/commands/join.md
```diff
@@ -1,59 +1,68 @@
...
 
+:::warning Important
 The `join` command is considered an experimental feature. This means it's still a work in progress and may go through major changes.
-</div>
 
+The `join` command supports OpenAPI 3.x definitions only.
+:::
 
-Maintainers of multiple API definitions can benefit from storing each endpoint as a standalone API definition file. However, this approach is not supported by the majority of OpenAPI tools, as they require a single API definition file.
+With the `join` command, Redocly OpenAPI CLI can combine two or more API definition files into one so that maintainers of multiple API definitions can benefit from storing each endpoint as a standalone API definition file. However, this approach is not supported by the majority of OpenAPI tools, as they require a single API definition file.
```
skadinna, line 14: These two sentences don't connect logically. In the original text, the meaning was something like this:

1) Maintainers sometimes split their endpoints into standalone files instead of keeping them in the root file.
2) Most OpenAPI tools don't work with standalone files like that.
3) The join command helps with this problem because it can combine those standalone files into one. 

In this new version, these things are out of order. The second sentence talks about the approach not being supported because other tools need a single file...but you already explained how to get the single file in the previous sentence? So this second sentence is just confusing.

You may want to rewrite this part so that you start with the problem first, and then explain how the join command can resolve it. Or you can start by describing what the join command does, and then explain in which use-cases it can be the most helpful.

docs/commands/join.md
```diff
@@ -1,59 +1,68 @@
...
+openapi join first-api.yaml second-api.yaml
+openapi join first-api.yaml second-api.json
+openapi join ./*.yaml
+openapi join --version
 ```
 
+## Options
 
-The command:
+option                               | type               | required?    | default     | description
```
skadinna, line 37:
```suggestion
Option                               | Type               | Required    | Default     | Description
```

docs/commands/join.md
```diff
@@ -1,59 +1,68 @@
...
+-------------------------------------|:------------------:|:------------:|:-----------:|------------
+`entrypoints`                        | `array`, wildcard  | yes          | `[]`, -     | 1. Array of paths to API definition files that need to be joined. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder
```
skadinna, line 39:
```suggestion
`entrypoints`                        | `array`, wildcard  | yes          | `[]`, -     | 1. Array of paths to API definition files that you want to join. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder
```

docs/commands/join.md
```diff
@@ -1,59 +1,68 @@
...
+`--help`                             | `boolean`          | no           | -           | Show help
+`--lint`                             | `boolean`          | no           | `false`     | Lint definition files
+`--prefix-components-with-info-prop` | `string`           | no           | -           | Prefix components with property value from info object. See the [section](#) below
+`--prefix-tags-with-info-prop`       | `string`           | no           | -           | Prefix tags with property value from info object. See the [section](#) below
+`--prefix-tags-with-filename`        | `boolean`          | no           | `false`     | Prefix tags with property value from file name. See the [section](#) below
```
skadinna, line 44: These links don't lead anywhere - what should they link to? Also, please try using a more descriptive anchor text - if all sentences just say "See the section below", it could be difficult for the readers to distinguish which section you're talking about.

docs/commands/join.md
```diff
@@ -82,98 +89,89 @@ x-tagGroups:
...
+* If any of the input files contain the `x-tagGroups` object, the content of this object is ignored by the `join` command and not included in the output file.
+:::
 
 The `servers` object combines the content from all input files, starting with the content from the first file. Commented lines are not included in the output file.
 
 Path names and component names must be unique in all input files, but their content doesn't have to be unique for the `join` command to produce the output file. For each path item object, only the operation ID must be unique.
 
 If the `join` command detects any conflicting content while trying to combine the input files, it displays informative messages about the conflicts and exits without creating an output file. To prevent this, use optional parameters to add prefixes to tags and components.
 
+```bash example of conflicts
```
skadinna, line 103: I really like that you included this example! Good thinking - it will help users recognize this type of error output, so they know what to expect. :brain:

skadinna (changes requested):

docs/commands/join.md
```diff
@@ -1,59 +1,70 @@
...
+openapi join ./*.yaml
+openapi join --version
 ```
 
+## Options
 
-The command:
+Option                               | Type               | Required     | Default     | Description
+-------------------------------------|:------------------:|:------------:|:-----------:|------------
+`entrypoints`                        | `array`, wildcard  | yes          | `[]`, -     | 1. Array of paths to API definition files that you want to join. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder
```
skadinna, line 41: Question about the Default column - does this `[], - ` mean that there's no default value?

bandantonio: Yes. It's better to leave a hyphen here

skadinna: Got it, thanks!

bandantonio:
This pr seems to be ready as well

skadinna (approved):

Issue #272, opened 2021-04-01
Unable to preview-docs from Docker


PeeZu:
**Describe the bug**
I would like to use the Docker container to preview the documentation instead of using directly npx

**To Reproduce**
1. Assuming you have a valid `openapi.yaml` in the same directory
2. Run this command: `docker container run -t --rm -v ${PWD}:/spec -p 8080:8080 redocly/openapi-cli preview-docs /spec/openapi.yaml`
3. See error

**Expected behavior**
The Docker container exposing the preview documention

**Logs**
```
TypeError: o is not a constructor
    at Object.t.startWsServer (/bin/openapi-cli:43:129218)
    at Object.<anonymous> (/bin/openapi-cli:43:21717)
    at Generator.next (<anonymous>)
    at s (/bin/openapi-cli:43:19069)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
```

**`openapi-cli` Version(s)**
1.0.0-beta.40

**`Node.js` Version(s)**
v15.12.0

**Additional context**
Using the provided Dockerfile in this repo, works as expected outside Docker

Juneezee:
This bug is likely introduced in PR #177.

The `/bin/openapi-cli` file in the Dockerfile is compiled by Webpack
https://github.com/Redocly/openapi-cli/blob/360c865d23668c46cad9ee63677fce54925ce238/Dockerfile#L10-L15

However, the `simple-websocket` is configured to be null-loaded
https://github.com/Redocly/openapi-cli/blob/360c865d23668c46cad9ee63677fce54925ce238/webpack.config.ts#L20-L30

Therefore, the `SocketServer` will become `{}` when compiled by Webpack, hence the error `TypeError: o is not a constructor`.
https://github.com/Redocly/openapi-cli/blob/360c865d23668c46cad9ee63677fce54925ce238/packages/cli/src/commands/preview-docs/preview-server/server.ts#L5
https://github.com/Redocly/openapi-cli/blob/360c865d23668c46cad9ee63677fce54925ce238/packages/cli/src/commands/preview-docs/preview-server/server.ts#L65-L66

Pinging @RomanHotsiy and @knidarkness to raise awareness of this issue.

Juneezee:
Fixed in #364 (merged into `master` in commit 128aea923bf03437945f29392cd3cd2e94963d59)

PeeZu:
Great job guys !

PR #364, opened 2021-09-07
build: fix preview-docs command in Docker

Juneezee:
## What/Why/How?

As described in #272, running `preview-docs` in a `redocly/openapi-cli` Docker container will cause the container to crash instantly. This is not an issue with the Docker image. The problem lies in the Webpack build process that is causing the following `TypeError`. 

Running the following command `npm run bundle && node dist/bundle.js preview-docs resources/pets.yaml` will result in the following error as well.

```
TypeError: s is not a constructor
    at Object.t.startWsServer (/bin/openapi-cli:49:129227)
    at Object.<anonymous> (/bin/openapi-cli:49:21717)
    at Generator.next (<anonymous>)
    at o (/bin/openapi-cli:49:19069)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

---
Currently Webpack will load `simple-websocket` as an empty module. This is introduced in PR #177, which was merged into the `master` branch before the Dockerfile was first created in commit ceaa205497a6cee6a02876a75b8ddcd4e37aa23c. 
https://github.com/Redocly/openapi-cli/blob/360c865d23668c46cad9ee63677fce54925ce238/webpack.config.ts#L20-L26

---
However, `preview-docs` requires the websocket server to be started. If `simple-websocket` is loaded as an empty module, then `SocketServer` will have the value `{}` and hence `new SocketServer` will throw a `TypeError`.

https://github.com/Redocly/openapi-cli/blob/38139e51393d7589217362986b36fb48771ab703/packages/cli/src/commands/preview-docs/preview-server/server.ts#L5

https://github.com/Redocly/openapi-cli/blob/38139e51393d7589217362986b36fb48771ab703/packages/cli/src/commands/preview-docs/preview-server/server.ts#L65-L66

---

To fix this, we would need to configure Webpack to load `simple-websocket`. The steps are:

1. Use `babel-loader` to load the `simple-websocket` module. We will need to use the `@babel/plugin-proposal-class-properties` plugin as well, otherwise the following error will occur.
```
ERROR in ./node_modules/simple-websocket/server.js 45:19
Module parse failed: Unexpected token (45:19)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|   }
| 
>   _handleListening = () => {
|     this.emit('listening')
|   }
 @ ./packages/cli/src/commands/preview-docs/preview-server/server.ts 6:21-58
 @ ./packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
 @ ./packages/cli/src/commands/preview-docs/index.ts
 @ ./packages/cli/src/index.ts
```

2. Configure Webpack to use named instead of number for Module ID. `require.resolve` returns a number instead of a string by Webpack for optimization purposes (see [Webpack 4 require.resolve](https://v4.webpack.js.org/api/module-methods/#requireresolve) and [Webpack 4 Module IDs](https://v4.webpack.js.org/configuration/optimization/#optimizationmoduleids)). This is required otherwise line 107 will fail because `filePath` is a number, and `path.extname` expects a string.
https://github.com/Redocly/openapi-cli/blob/38139e51393d7589217362986b36fb48771ab703/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts#L99-L107

3. Copy the `packages/cli/src/commands/preview-docs/preview-server/default.hbs` file into `dist` folder, and include it into the Docker image too.

## Reference

This pull request fixes Issue #272.

### Before
before

### After
after

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+*
+!package.json
+!tsconfig.build.json
+!tsconfig.json
+!package-lock.json
+!packages/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,20 +2,26 @@ FROM node:alpine
 
 WORKDIR /build
 
-COPY tsconfig.json tsconfig.build.json /build/
+# Copy files required for npm install only
 COPY package.json package-lock.json /build/
-COPY packages /build/packages
+COPY packages/cli/package.json /build/packages/cli/
+COPY packages/core/package.json   /build/packages/core/
+COPY packages/cli/bin/ /build/packages/cli/bin/
 
-RUN npm ci --no-optional
+RUN npm ci --no-optional --ignore-scripts
+
+# Copy rest of the files
+COPY . /build/
+RUN npm run prepare
 
 # Install openapi-cli globally, similar to npm install --global @redocly/openapi-cli
 # but the local package is used here
 RUN mv -- "$(npm pack packages/cli/)" redocly-openapi-cli.tgz && \
-	npm install --global redocly-openapi-cli.tgz
+	  npm install --global redocly-openapi-cli.tgz
 
 # npm pack in the previous RUN command does not include these assets
 RUN cp packages/cli/src/commands/preview-docs/preview-server/default.hbs /usr/local/lib/node_modules/@redocly/openapi-cli/lib/commands/preview-docs/preview-server/default.hbs && \
-	cp packages/cli/src/commands/preview-docs/preview-server/hot.js /usr/local/lib/node_modules/@redocly/openapi-cli/lib/commands/preview-docs/preview-server/hot.js
+	  cp packages/cli/src/commands/preview-docs/preview-server/hot.js /usr/local/lib/node_modules/@redocly/openapi-cli/lib/commands/preview-docs/preview-server/hot.js
 
 # Clean up to reduce image size
 RUN npm cache clean --force && rm -rf /build
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,24 @@ FROM node:alpine
 
 WORKDIR /build
 
-COPY webpack.config.ts tsconfig.json tsconfig.build.json /build/
+COPY tsconfig.json tsconfig.build.json /build/
 COPY package.json package-lock.json /build/
 COPY packages /build/packages
 
 RUN npm ci --no-optional
-RUN npm run webpack-bundle
 
-FROM node:alpine
+# Install openapi-cli globally, similar to npm install --global @redocly/openapi-cli
+# but the local package is used here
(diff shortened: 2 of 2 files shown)


RomanHotsiy:
Great job on investigation here. I'm wandering if it's not easier to just copy unbundled sources into the docker container vs introducing more build steps.

@Juneezee what do you think? Would you be open to help with this?

Juneezee:
@RomanHotsiy Thank you for taking your time to review this PR :heart: !

> Great job on investigation here. I'm wandering if it's not easier to just copy unbundled sources into the docker container vs introducing more build steps.
> 
> @Juneezee what do you think? Would you be open to help with this?

Yes, the easiest way to fix #272 is actually building the Docker image using `npm install -g @redocly/openapi-cli`, and we don't need to add those `babel` into `package.json`. However, the final Docker image will be around 10 ~ 20 MB larger than the image built with `npm run webpack-bundle`.

If this is acceptable for the Redoc team, then I am happy to modify this PR to use the easy method mentioned above :smiley: .

RomanHotsiy:
> building the Docker image using `npm install -g @redocly/openapi-cli`

That will require package to be published on NPM first. I've seen npm may be lagging behind because of the cache sometimes. Please, keep it in mind.

> If this is acceptable for the Redoc team, then I am happy to modify this PR to use the easy method mentioned above 😃 .

This is acceptable. Please, modify the PR 🙏

Juneezee:
@RomanHotsiy Please take a look again. Thank you.

Juneezee:
@RomanHotsiy I'm sorry for the back and forth changes. I have just figured out a better way to build the Docker image without relying on `npmjs.com`. Now the Docker image can be built without having to wait package being published first to `npmjs.com`.

Please take another look again. Thanks.

RomanHotsiy:
Looks much better now, the only thing left is to split RUN into multiple commands to get proper caching, see this: https://deploy-docs.aptible.com/docs/dockerfile-caching#npm

Juneezee:
I have split the `RUN` commands and added some comments. I did some quick googling for Docker image build cache and found some example best practices demonstrated in

1. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
2. https://stackoverflow.com/a/50097098/7902371

I am not experienced enough with Docker build cache :cry: so I still couldn't find the correct boundary of whether to join or split multiple `RUN` commands

RomanHotsiy:
Thanks! I will update it slightly later this week.

RomanHotsiy:
Hey @Juneezee, take a look at my changes here b75b22c

If they look okay, let me know and I'll merge it.

RomanHotsiy (approved):

Juneezee:
@RomanHotsiy Looks good to me. Thanks

PR #367, opened 2021-09-13
fix: join command crash

ohorbachevskyi:
fixes: https://github.com/Redocly/openapi-cli/issues/339

#### Fixes:
- ~~add components prefix only for string description~~ (removed as unnecessary)
- ~~fix `collectPath` crash when we have `$ref` in path~~ (removed as unnecessary)
- resolve $ref and inline it into the resulting definition (this change fixes the whole issue)

#### Tests improvements:
- refactor snapshot tests to support other than `lint` command testing
- setup couple `join` command tests
- decrease e2e snapshot tests execution time by ~20 seconds 🚀 

> before
> Screenshot 2021-09-15 at 17 57 07
> 
> after
> Screenshot 2021-09-15 at 18 00 40

#### Note: 
We refactored e2e snapshot tests to support other than `lint` command testing (IMHO it is not perfect but it is the start). Here we added couple tests for `join` command, but other commands could be covered with tests in the same way. 
During refactoring we have a lot of "changed files", which basically was move from `__test__` folder into `__test__/lint` folder. 
If you want to review the PR easily, do it commit by commit:
:bookmark: tests refactoring https://github.com/Redocly/openapi-cli/pull/367/commits/b1c3b15ff390fb3aecd3c074e1a07c662103e9d3
:bookmark: resolve $ref before join https://github.com/Redocly/openapi-cli/pull/367/commits/c0fb44d8515e6dcb5ad405b2291e451b858e1ed6

## Check yourself

- [x] Code is linted

## Security


Marshevskyy (changes requested):
looks good, could you add a test that would crash before the fix, and with the fix, it will pass?

ohorbachevskyi:
> looks good, could you add a test that would crash before the fix, and with the fix, it will pass?

definitely yes! working on it.

packages/cli/src/commands/join.ts
```diff
@@ -231,6 +232,9 @@ packageVersion: string
           const pathOperation = paths[path][operation];
           joinedDef.paths[path][operation] = pathOperation;
           potentialConflicts.paths[path][operation] = [...(potentialConflicts.paths[path][operation] || []), entrypoint];
+          if (operation === '$ref') {
+            break; // do not proceed with reference, it's a string path
```
RomanHotsiy, line 236: Don't we have to resolve the `$ref`?

packages/cli/src/commands/join.ts
```diff
@@ -322,8 +326,8 @@ packageVersion: string
     const componentsPrefix = getInfoPrefix(openapi.info, prefixComponentsWithInfoProp, COMPONENTS);
     if (!openapi.openapi) exitWithError('Version of specification is not found in. \n');
     if (!openapi.info) exitWithError('Info section is not found in specification. \n');
-    if (openapi.info?.description) {
-      openapi.info.description = addComponentsPrefix(openapi.info.description, componentsPrefix);
+    if (typeof openapi.info?.description === 'string') {
```
RomanHotsiy, line 329: I think we should resolve $ref and inline it into the resulting definition.

RomanHotsiy (commented):
Left a few comments.

packages/cli/src/commands/join.ts
ohorbachevskyi: right now we don't resolve it, but I'll add this feature within this PR

packages/cli/src/commands/join.ts
ohorbachevskyi: yeah, I'll add this feature within this PR

RomanHotsiy:
image


Good job 😓

RomanHotsiy (approved):

andriyl (dismissed):

Marshevskyy (commented):
well done!
all your tests covers openapi: 3.0.0
could you add OAS 3.1 as well?

ohorbachevskyi:
> well done!
> all your tests covers openapi: 3.0.0
> could you add OAS 3.1 as well?

@Marshevskyy I can update version for `lint` tests to OAS 3.1.0, they will also pass as well (just checked it).
About `join` command - it supports only OAS 3.0.0 version ([here](https://github.com/Redocly/openapi-cli/blob/edef39afde6666757909dc19d1e567ca9f2cfe92/packages/cli/src/commands/join.ts#L86) the code which checks that), but I think it should support also OAS 3.1.0, need to investigate it separately. Propose to leave `join` tests as it is within current PR.

Marshevskyy (approved):

andriyl (approved):

ohorbachevskyi:
@RomanHotsiy I made two additional commits with some minor changes since you approved this PR (update package-json.lock and openapi versions for tests files). Github wants to re-approve it and guys did that, so I'm gonna squash this PR into the master branch.

packages/cli/src/commands/join.ts
```diff
@@ -58,6 +59,27 @@ packageVersion: string
     entrypoints.map(ref => externalRefResolver.resolveDocument(null, ref, true) as Promise<Document>)
   );
 
+  const bundleResults = await Promise.all(
```
acote-coveo, line 62: @ohorbachevskyi Hi! Is it possible that introducing a bundle inside the join command now runs all the decorators/preprocessor on my files before joining them? I noticed this regression after bump my version to the latest

PR #368, opened 2021-09-13
chore:add optional parameter to auth

AlexVarchuk:
## What/Why/How?
add an optional parameter for the set name of the query in authorize function.
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/redocly/index.ts
+++ b/packages/core/src/redocly/index.ts
@@ -36,7 +36,7 @@ export class RedoclyClient {
 
   async verifyToken(accessToken: string, verbose: boolean = false): Promise<boolean> {
     if (!accessToken) return false;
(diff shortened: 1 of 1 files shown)


andriyl (dismissed):

packages/core/src/redocly/index.ts
```diff
@@ -91,9 +91,9 @@ export class RedoclyClient {
     });
   }
 
-  static async authorize(accessToken: string, verbose: boolean = false) {
+  static async authorize(accessToken: string, verbose: boolean = false, queryName = '') {
```
RomanHotsiy, line 94: keep verbose last arg or convert last two to object

RomanHotsiy (approved):

Issue #333, opened 2021-08-07
[Bug] Example with schema array not validated


mugx-fc:
### Describe the bug

False positive validation when an example relates to a schema array

### To Reproduce

1. Given this OpenAPI file:

```
---
openapi: '3.0.3'
info:
  version: v0
  title: my_Api
  description: my_api
  contact:
    name: my_api
  license:
    name: Proprietary
    url: https://my_api.com
servers:
  - url: https://my_api.com

paths:
  /my_post:
    post:
      operationId: my_post
      description: my_post
      summary: my_post
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              additionalProperties: false
              properties:
                my_list:
                  type: array
                  uniqueItems: true
                  items:
                    type: string
                  example:
                    test
      responses:
        '200':
          description: My 200 response

```

2. observe the wrong example related to `my_list` which misses the `-`, mandatory when the schema is an array.
3. Run the lint command, for instance: `npx @redocly/openapi-cli lint my_test.yaml`
4. See that the validation wrongly passes:

`Woohoo! Your OpenAPI definition is valid. 🎉`

### Expected behavior

The given example is wrong:
 
```
example:
    test
```

a correct example should be:

```
example:
    - test
```

however the `openapi-cli lint` outputs a **false positive**.

### Additional context

- I've tried to augment the severity of the checks using `.redocly.yaml` with:

```
lint:
  extends:
    - all
```

more errors and warnings appears, however the wrong example still gets a false positive.

- I've tried to use another tool (`spectral lint`) which can validate the wrong example, for instance:

<img width="700" alt="Screenshot 2021-08-07 at 10 59 17" src="https://user-images.githubusercontent.com/54804348/128596487-003d5591-3845-48ed-95a5-7c164d7b3a58.png">

see that it correctly suggests: **the example property type should be an array**.

PR #375, opened 2021-09-20
fix: schema property:example should be validated

andriyl:
## What/Why/How?
fix: #333 
## Check yourself
- [x] Code is linted
~Tested with redoc/reference-docs/workflows~

## Security


adamaltman (approved):
Looks good to me.

packages/core/src/rules/common/spec.ts
```diff
@@ -110,6 +110,15 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
             }
           }
         }
+
+        if (propName === 'example' && node.type === 'array') {
```
RomanHotsiy, line 114: Why do we have it only for `array` type? Doesn't it make sense to validate all types?

packages/core/src/rules/common/spec.ts
```diff
@@ -110,6 +110,15 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
...
+          if (!matchesJsonSchemaType(propValue, node.type, false)) {
+            report({
+              message: `Expected type \`${propName}\` (array) but got \`${propValueType}\`.`,
```
RomanHotsiy, line 117: This message is invalid, change it to:

```suggestion
              message: `Expected type \`${node.type}\` but got \`${propValueType}\`.`,
```

RomanHotsiy (changes requested):
This is handling just a single case, need to make it generic.

BUT I would make it a separate rule so it can be disabled/enabled separatelly.

@andriyl can you change it to a separate rule?

packages/core/src/rules/common/spec.ts
adamaltman: I think this suggestion is good.

packages/core/src/rules/common/spec.ts
adamaltman: Good question -- we should validate all example types.

andriyl: will do for all types in one rule

andriyl:
@adamaltman @RomanHotsiy 
Added a separate rule and support version: 3.1.0 : 

```
properties:
  one:
    type:
      - integer
      - string
    example:
      - test
```

packages/core/src/config/all.ts
```diff
@@ -30,6 +30,7 @@ export default {
     'boolean-parameter-prefixes': 'error',
     'paths-kebab-case': 'error',
     spec: 'error',
+    'schema-example': 'error',
```
adamaltman, line 33: If I understand, this only checks the schema example type, and maybe we should name the rule `schema-example-type`. For example, I could have a schema:
```
type: string
maxLength: 5
example: Wonderful
```
The type is a string, but the example is still invalid because it violates the maxLength constraint.

andriyl: @adamaltman done

adamaltman (commented):
LGTM. Let's have someone else review the code in more detail. 🎉

adamaltman:
It should be added to the appropriate docs page too.

RomanHotsiy (changes requested):
It should not use Schema properties, just Schema. Otherwise it will miss all the schemas without any properties.

Also, rule name seems inconsistent with other rules, I believe we have something like no-invalid-examples or something like that (check it yourself plz, I'm on the phone)

__tests__/lint/no-invalid-schema-examples-nested-error/openapi.yaml
```diff
@@ -0,0 +1,79 @@
...
+          type: string
+          example: 50
+        nested:
+          allOf:
+            - $ref: '#/components/schemas/Dog'
+            - type: object
+              example: dog
+              properties:
+                huntingSkill:
+                  type: string
```
RomanHotsiy, line 61:
```suggestion
                  type: number
                  min: 200
```

packages/core/src/rules/common/no-invalid-schema-examples.ts
```diff
@@ -0,0 +1,54 @@
...
+              schema,
+              location.child('schema'),
+              dataLoc.pointer,
+              resolve,
+              false
+            );
+            if (!valid) {
+              for (let error of errors) {
+                report({
+                  message: `Example value must conform to the schema: ${error.message} but got \`${propValueType}\`.`,
```
RomanHotsiy, line 34: This message is invalid, check out this https://github.com/Redocly/openapi-cli/blob/0eee0fb1f69f1790566209f5449ac84265084d53/packages/core/src/rules/oas3/no-invalid-media-type-examples.ts#L43-L51

packages/core/src/rules/common/no-invalid-parameter-examples.ts
```diff
@@ -0,0 +1,55 @@
...
+          validateExample(parameter.example, location.child('example'));
+        }
+
+        if (parameter.examples) {
+          for (const [_key, example] of Object.entries(parameter.examples)) {
+            validateExample(example, location.child('examples'));
+          }
+        }
+
+        function validateExample(example: any, dataLoc: Location) {
```
RomanHotsiy, line 21: check possible duplication

packages/core/src/rules/common/no-invalid-parameter-examples.ts
```diff
@@ -0,0 +1,55 @@
...
+  return {
+    Parameter: {
+      leave(parameter: Oas3Parameter, { report, location, resolve }: UserContext) {
+        if (parameter.example) {
+          validateExample(parameter.example, location.child('example'));
+        }
+
+        if (parameter.examples) {
+          for (const [_key, example] of Object.entries(parameter.examples)) {
+            validateExample(example, location.child('examples'));
```
RomanHotsiy, line 17:
```suggestion
            validateExample(example, location.child(['examples', _key]));
```

packages/core/src/rules/common/no-invalid-schema-examples.ts
```diff
@@ -0,0 +1,54 @@
...
+import { Oas3_1Schema } from '../../typings/openapi';
+import { Location } from '../../ref-utils';
+
+export const NoInvalidSchemaExamples: any = () => {
+  return {
+    Schema: {
+      leave(schema: Oas3_1Schema, { report, location, resolve }: UserContext) {
+        if (schema.examples) {
+          for (const example of schema.examples) {
+            validateExample(example, location.child('examples'));
```
RomanHotsiy, line 13: fix location idx

packages/core/src/rules/oas3/index.ts
```diff
@@ -50,6 +52,8 @@ import { PathExcludesPatterns } from '../common/path-excludes-patterns';
 
 export const rules = {
   spec: OasSpec,
+  'no-invalid-schema-examples': NoInvalidSchemaExamples,
+  'no-invalid-parameter-examples': NoInvalidParameterExamples,
```
RomanHotsiy, line 56: move to the end

RomanHotsiy (changes requested):
Fix reporting please.

packages/core/src/rules/common/no-invalid-parameter-examples.ts
```diff
@@ -1,55 +1,35 @@
...
-      leave(parameter: Oas3Parameter, { report, location, resolve }: UserContext) {
+      leave(parameter: Oas3Parameter, ctx: UserContext) {
         if (parameter.example) {
-          validateExample(parameter.example, location.child('example'));
+          validateExample(
+            parameter.example,
+            parameter.schema!,
+            ctx.location.child('example'),
+            ctx,
+            false,
```
RomanHotsiy, line 15: let's add `disallowAdditionalProperties` option to all new rules too (true by default similar to no-invlaid-media-examples) rule.

RomanHotsiy (commented):
Looks pretty good!

I think we just need to add `disallowAdditionalProperties` to all the rules to make it consistent with no-ivalid-media-examples and to prevent future breaking change.

RomanHotsiy (changes requested):
I just realized this PR is missing docs also.

@andriyl could you add a draft into docs/ folder?

docs/resources/built-in-rules.md
```diff
@@ -242,6 +242,29 @@ lint:
       - application/json
 ```
 
+### no-invalid-schema-examples
+
+Verifies that example value conforms to the schema type. Disallows additional properties by default.
```
RomanHotsiy, line 247:
```suggestion
Verifies that schema example value conforms to the schema. Disallows additional properties by default.
```

docs/resources/built-in-rules.md
```diff
@@ -242,6 +242,29 @@ lint:
...
+lint:
+  no-invalid-schema-examples:
+    severity: error
+    disallowAdditionalProperties: false
+```
+
+
+### no-invalid-parameter-examples
+
+Verifies that parameter example value conforms to the schema type. Disallows additional properties by default.
```
RomanHotsiy, line 259:
```suggestion
Verifies that parameter example value conforms to the schema. Disallows additional properties by default.
```

RomanHotsiy (approved):

Issue #382, opened 2021-09-24
support --region for login command


AntonKozachuk:
**Is your feature request related to a problem? Please describe.**
We have two regions: us and eu.

They differ by `REDOCLY_DOMAIN` url:

us: api.redoc.ly
eu: api.eu.redocly.com

See here: https://github.com/Redocly/openapi-cli/blob/master/packages/core/src/redocly/query.ts#L3

**Describe the solution you'd like**
Add support `--region` for the login command.

**Describe alternatives you've considered**
N/a

**Additional context**
N/a

adamaltman:
Good one. Default to `US`, and additional option is `EU`. But maybe there are other ways to do this? What if the API URL was an env variable with a default value of the US URL? This may give more flexibility to be able to target local/dev/and other environments too.

RomanHotsiy:
@adamaltman it is like that right now. We have an env variable 😀

The cli flag would add alternative more user friendly approach.

adamaltman:
I like this option. Some users won't even know their region, so... this may be the easiest for them.

PR #427, opened 2021-10-29
support --region for login command #382

Antarus66:
## What/Why/How?

We have two regions: us and eu.

They differ by REDOCLY_DOMAIN url:

us: api.redoc.ly
eu: api.eu.redocly.com

- Add support --region for the login command.
- Add a region option to the config

## Reference

Closes: https://github.com/Redocly/openapi-cli/issues/382

## Check yourself

- [x] Code is linted

## Security


packages/core/src/redocly/index.ts
```diff
@@ -9,8 +9,11 @@ const TOKEN_FILENAME = '.redocly-config.json';
 export class RedoclyClient {
   private accessToken: string | undefined;
 
-  constructor() {
+  private readonly redoclyDomain: string;
+
+  constructor(redoclyDomain: string) {
```
Antarus66, line 14: Breaking change:

Using there: https://github.com/Redocly/redocly-portal/blob/main/engine/plugins/gatsby-source-redocly-filesystem/oas/openapi-cli-bundle.js#L30

packages/core/src/config/load.ts
```diff
@@ -2,35 +2,43 @@ import * as fs from 'fs';
...
+} from './config';
 
 import { defaultPlugin } from './builtIn';
 
-export async function loadConfig(configPath?: string, customExtends?: string[]): Promise<Config> {
-  if (configPath === undefined) {
-    configPath = findConfig();
-  }
-  let rawConfig: RawConfig = {};
+export async function loadConfig({
```
Antarus66, line 14: Breaking change:

Using in redocly-portal there: https://github.com/Redocly/redocly-portal/blob/main/engine/plugins/gatsby-source-redocly-filesystem/oas/openapi-cli-bundle.js#L33

Antarus66:
@RomanHotsiy Basing on the abovementioned "technically breaking" changes, should we merge this to the `general-release`?

packages/core/src/redocly/index.ts
slavikbez: What do you think about providing a default value for `redoclyDomain` here?

packages/cli/src/__tests__/commands/login.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    await handleLogin({});
+
+    expect(promptUser).toHaveBeenCalledWith(
+      expect.stringContaining(redoclyDomain),
+      expect.anything(),
+    );
+  });
+
+  it('log user it to domain taken from config', async () => {
```
slavikbez, line 35:
```suggestion
  it('log user in to domain taken from config', async () => {
```

packages/core/src/redocly/index.ts
slavikbez: I reviewed the remaining PR and it doesn't make sense to provide a default value here. Sorry for bothering

packages/core/src/rules/common/registry-dependencies.ts
```diff
@@ -9,16 +8,27 @@ export const RegistryDependencies: Oas3Decorator | Oas2Decorator = () => {
   return {
     DefinitionRoot: {
       leave() {
-        redoclyClient = new RedoclyClient();
+        if (!process.env.REDOCLY_DOMAIN) {
+          throw new Error('Redocly domain is not set');
```
slavikbez, line 12: Just wondering if it will throw if I bundle something on my local machine, like: `npx @redocly/openapi bundle my-api.yaml` ?

slavikbez (approved):
Left some comments, but looks good overall!

packages/core/src/rules/common/registry-dependencies.ts
Antarus66: It shouldn't - the env var is setting in code.

andriyl:
It was decided to use another approach and rewrite the flow from scratch.
PR: https://github.com/Redocly/openapi-cli/pull/459

Issue #391, opened 2021-10-07
`operation-4xx-response` built-in rule


adamaltman:
**Is your feature request related to a problem? Please describe.**

The spec indicates there should be a successful request and we have this rule already: `operation-2xx-response`.

However, it's likely any API may return an error. Someone may want to enforce this: `operation-4xx-response`

**Describe the solution you'd like**

I want to have a built-in rule which someone could enable to enforce `operation-4xx-response`.

**Describe alternatives you've considered**

None.

**Other info**

Related would be enforcing a specific shape of the response, such as following the `application/problem+json` https://datatracker.ietf.org/doc/html/rfc7807 format.

PR #430, opened 2021-11-01
feat: operation-4xx-response built-in rule #391

Antarus66:
## What/Why/How?

The spec indicates there should be a successful request and we have this rule already: operation-2xx-response.

However, it's likely any API may return an error. Someone may want to enforce this: operation-4xx-response

## Reference

Closes: #391

## Check yourself

- [x] Code is linted

## Security


Antarus66:
There are quite a lot of updated e2e test datasets and snapshots because the `operation-4xx-response` rule was added to the `recommended` ruleset with the `warning` severity.

RomanHotsiy:
@Antarus66 how about disabling it in specific tests? Except those that test recommended ruleset explicitly.

Antarus66:
> @Antarus66 how about disabling it in specific tests? Except those that test recommended ruleset explicitly.

There might be 2 options to avoid warnings for this rule is tests:

1. Provide a simplest valid definition - add a simple 400 response to data 
2. Explicitly disable it in every test where any response is present in definitions.

As for me, the 1st option seems more consistent, because in this case we keep our test data in accordance with our own recommendations and don't do any exceptions for specific rules overriding defaults. 

At the other hand, it might be unexpected to see the warning in case the only some other rule is specified. Let's discuss that.

RomanHotsiy:
@Antarus66 go with approach 1

Antarus66:
@RomanHotsiy Checked all the affected tests for isolation from the `recommended` ruleset. 

- If an isolation is necessary and was skipped, added the isolation (`extends: []`), removed 4xx responses from test data. 
- If the `recommended` ruleset is testing, made it clear, left added in this PR 4xx responses in test data.
- Fixed occasional typos.

RomanHotsiy (approved):
LGTM 👍

Issue #424, opened 2021-10-28
`preview-docs` hot reload not working


adamaltman:
**Describe the bug**

Hot reload is not working for `preview-docs`.

**To Reproduce**
1. Given any `.redocly.yaml` file
2. And any OpenAPI file(s)
3. Run this command with these arguments... `openapi preview-docs`
4. Open browser
5. Make change and save file
6. See browser doesn't reload (this is the error)

**Expected behavior**

**Logs**

Created a bundle for openapi/auth-service.yaml successfully
GET /: 25.841ms
GET /simplewebsocket.min.js: 2.664ms
GET /hot.js: 0.524ms
GET /openapi.json: 2.461ms
GET /favicon.png: 1.826ms
GET /copper_logo_gradient.png: 0.926ms
watch changed /Users/adam/.redacted.md

Bundling...

Created a bundle for openapi/auth-service.yaml successfully
```

**`openapi-cli` Version(s)**

1.0.0-beta.65

**`Node.js` Version(s)**
What version of `node.js` are you using?

- v16.1.0
- Also tried 14 LTS.

slavikbez:
@adamaltman I wasn't able to reproduce this problem. The page reloads when I make changes in the definition file.

https://user-images.githubusercontent.com/9461457/139729968-4f1ffde5-228d-49f2-9e8c-727f6108ac57.mov

adamaltman:
@slavikbez this repo: https://github.com/Rebilly/api-definitions and I'm previewing with the premium reference docs (requires you to login first).

slavikbez:
Located the problem: https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts#L108

**NOTE** this problem exists only with `pagination` set to `item` or `section`

adamaltman:
I have had the problem without pagination though too. I'm happy to check again after your fix.

Issue #423, opened 2021-10-28
`preview-docs` page refresh breaks with `pagination` section or item


adamaltman:
**Describe the bug**

When the reference docs config has:

```yaml
referenceDocs:
  pagination: section
  # or pagination: item
```

And you browse to an operation...

Then refresh the browser and voila!

**To Reproduce**
1. Given this `.redocly.yaml` file excerpt above (You can use rebilly/api-definitions core API)
2. And this OpenAPI file(s) (Use rebilly/api-definitions core API)
3. Run this command with these arguments... `openapi preview-docs core`
4. Browse to **Customers > Create a customer**.
5. Refresh the browser.
6. See error.

**Expected behavior**

It should work.

**Logs**

n/a

**`openapi-cli` Version(s)**

Latest as of Oct 28, 2021.

**`Node.js` Version(s)**
14 LTS.

slavikbez:
@adamaltman rebilly's core api seems too heavy to run locally. In very rare cases I get page initially loaded but when I reload browser I see the loading spinner forever. Is this a problem you're talking about or it's something different?

P.S. I was trying to run `preview-docs` for storefront api and browser reloading works good there.

slavikbez:
I was able to reproduce it. I ran it using reference-docs and when I reload the page I get: **404 Not Found**

slavikbez:
Located the problem: https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts#L108

**NOTE** this problem exists only with `pagination` set to `item` or `section`

PR #436, opened 2021-11-02
fix: hot reloading + page refresh for section pagination

slavikbez:
## What/Why/How?

Fixes [#423](https://github.com/Redocly/openapi-cli/issues/423) [#424](https://github.com/Redocly/openapi-cli/issues/424) 

## Screenshots (optional)

https://user-images.githubusercontent.com/9461457/139879901-84f3e223-94c1-4023-b029-1ff9f3a973d6.mov

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
+++ b/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
@@ -64,7 +64,7 @@ export default async function startPreviewServer(
     console.time(colorette.dim(`GET ${request.url}`));
     const { htmlTemplate } = getOptions() || {};
 
-    if (request.url === '/') {
+    function serveIndexPage() {
       respondWithGzip(
         getPageHTML(htmlTemplate || defaultTemplate, getOptions(), useRedocPro, wsPort),
         request,
@@ -73,7 +73,11 @@ export default async function startPreviewServer(
           'Content-Type': 'text/html',
         },
       );
-    } else if (request.url === '/openapi.json') {
+    }
+
+    if (request.url === '/') {
+      serveIndexPage()
+    } else if (request.url?.endsWith('/openapi.json')) {
       const bundle = await getBundle();
       if (bundle === undefined) {
         respondWithGzip(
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -96,39 +100,45 @@ export default async function startPreviewServer(
...
+              `Something went wrong: ${e.code || e.message}...\n`,
+              request,
+              response,
+              {},
+              500,
+            );
+          }
+        }
+    } else {
+      const filePath = path.resolve(htmlTemplate ? path.dirname(htmlTemplate) : process.cwd(), `.${request.url}`);
```
RomanHotsiy, line 132: We lost functionality that serves any other files from the dir.

`filePath` variable is not used anywhere down.

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -96,39 +100,45 @@ export default async function startPreviewServer(
...
+            respondWithGzip(
+              `Something went wrong: ${e.code || e.message}...\n`,
+              request,
+              response,
+              {},
+              500,
+            );
+          }
+        }
+    } else {
```
RomanHotsiy, line 131: I would add the following check for serving index pages and would keep else code for serving other files unchanged.

```suggestion
    } else if (request.url.endsWith('/') ||  path.extname(request.url) === '') {
```

RomanHotsiy (changes requested):
We lost functionality. Other files from `htmlTemplate` dir (or `cwd` if there is no `htmlTemplate`) should be served too (e.g. favicon.png, logo.png, etc).

slavikbez:
> We lost functionality. Other files from `htmlTemplate` dir (or `cwd` if there is no `htmlTemplate`) should be served too (e.g. favicon.png, logo.png, etc).

Can there be files other than just static assets? For example if there is a `some-file.md` in the root, should it be accessible by `/some-file` in the browser, like we do it in portal?

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
slavikbez: I think I got what you mean.

RomanHotsiy:
> Can there be files other than just static assets? 

@slavikbez No, we do not support this as part of this project. Just static assets.

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
slavikbez: Fixed

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -64,7 +64,7 @@ export default async function startPreviewServer(
     console.time(colorette.dim(`GET ${request.url}`));
     const { htmlTemplate } = getOptions() || {};
 
-    if (request.url === '/') {
+    if (request.url === '/' || request.url?.endsWith('/') || path.extname(request.url!) === '') {
```
RomanHotsiy, line 67:
```suggestion
    if (request.url?.endsWith('/') || path.extname(request.url!) === '') {
```

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -73,7 +73,7 @@ export default async function startPreviewServer(
           'Content-Type': 'text/html',
         },
       );
-    } else if (request.url === '/openapi.json') {
+    } else if (request.url?.endsWith('/openapi.json')) {
```
RomanHotsiy, line 76: Is it always requested from the root?

RomanHotsiy (commented):
Left a tiny suggestion and a question

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
RomanHotsiy: I think we should change `openapi.json` to `/openapi.json` here: https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts#L49

slavikbez: yes, we can fix it there.

RomanHotsiy (approved):

PR #438, opened 2021-11-03
chore: made the dockerhub workflow a consequtive step after publishing

Antarus66:
## What/Why/How?

The **CD to Docker Hub** workflow often fails on the first run. 

**Possible reason:** a race condition with the **CD to NPM & AWS S3** workflow, deployment part:
In case `packages/cli/package.json` contains the just releasing version of the `@redocly/openapi-core` as dependency,
it tries to load it from npm before it is published. 
**Solution:** Make the **CD to Docker Hub** call as the last job of the **CD to NPM & AWS S3** workflow.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/.github/workflows/dockerhub.yaml
+++ b/.github/workflows/dockerhub.yaml
@@ -2,11 +2,7 @@
 
 name: CD to Docker Hub
 
-on:
-  push:
-    tags:
-      - v[0-9]+.[0-9]+.[0-9]+
-      - v[0-9]+.[0-9]+.[0-9]+-[a-z]+.[0-9]+ # beta/rc releases
+on: workflow_call
 
 jobs:
   docker:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - v[0-9]*.[0-9]*.[0-9]*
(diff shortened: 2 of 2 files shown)


RomanHotsiy (approved):

.github/workflows/dockerhub.yaml
```diff
@@ -2,11 +2,7 @@
 
 name: CD to Docker Hub
 
-on:
-  push:
-    tags:
-      - v[0-9]+.[0-9]+.[0-9]+
-      - v[0-9]+.[0-9]+.[0-9]+-[a-z]+.[0-9]+ # beta/rc releases
+on: workflow_call
```
Marshevskyy, line 5: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows
says it should be 
```
on: 
  workflow_call:
```

Antarus66: It should be a case when you have a nested parameters there.
This [part of documentation](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-2) says that syntax on 1 line might be valid too.

Antarus66: We should test it anyway to know for sure.

PR #439, opened 2021-11-03
fix: inline dereferenced schema when original schema name matches resolved schema name

slavikbez:
## What/Why/How?

This is a fix for a `openapi bundle` command when there are external references.

**Prerequisites:**

`definition.yaml`
```yaml
openapi: 3.0.0
info:
  title: My API
  description: It ain't so wonderful, but at least it's mine.
  version: '1.0'
  contact:
    email: me@theintenet.com
    name: me
tags:
  - name: tag1
servers:
  - url: 'https://dev01.internet.com/api/v0'
    description: Development server.

components:
  schemas:
    vendor:
      $ref: ./vendor.schema.yaml
    myvendor:
      $ref: ./vendor.schema.yaml
```

`vendor.schema.yaml`
```yaml
title: vendor
type: object
description: Vendors
properties:
  key:
    type: string
    description: System-assigned key for the vendor.
    readOnly: true
  id:
    type: string
    description: |
      Unique identifier of the vendor.
      You must specify a unique vendor ID when creating a vendor unless document sequencing is configured, in which case the ID is auto-generated.
  name:
    type: string
    description: Name of the vendor.
  isOneTimeUse:
    type: boolean
    description: One-time use
    default: false

```

After running `openapi bundle definition.yaml`

<details>
   <summary>Original Result</summary>

```yaml
   openapi: 3.0.0
info:
  title: My API
  description: It ain't so wonderful, but at least it's mine.
  version: '1.0'
  contact:
    email: me@theintenet.com
    name: me
tags:
  - name: tag1
servers:
  - url: 'https://dev01.internet.com/api/v0'
    description: Development server.
components:
  schemas:
    vendor:
      $ref: '#/components/schemas/vendor-2'
    myvendor:
       $ref: '#/components/schemas/vendor-2'
    vendor-2:
      title: vendor
      type: object
      description: Vendors
```

</details>

<details>
   <summary>Result after fix</summary>
   ```yaml
openapi: 3.0.0
info:
  title: My API
  description: It ain't so wonderful, but at least it's mine.
  version: '1.0'
  contact:
    email: me@theintenet.com
    name: me
tags:
  - name: tag1
servers:
  - url: https://dev01.internet.com/api/v0
    description: Development server.
components:
  schemas:
    vendor:
      title: vendor
      type: object
      description: Vendors
      properties:
        key:
          type: string
          description: System-assigned key for the vendor.
          readOnly: true
        id:
          type: string
          description: >
            Unique identifier of the vendor.
            You must specify a unique vendor ID when creating a vendor unless
            document sequencing is configured, in which case the ID is
            auto-generated.
        name:
          type: string
          description: Name of the vendor.
        isOneTimeUse:
          type: boolean
          description: One-time use
          default: false
    myvendor:
      $ref: '#/components/schemas/vendor'
```
</details>

## Reference

Fixes [#439](https://github.com/Redocly/openapi-cli/issues/349)

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -40,7 +40,8 @@ export async function bundle(opts: {
     throw new Error('Document or reference is required.\n');
   }
 
-  const document = doc !== undefined ? doc : await externalRefResolver.resolveDocument(base, ref!, true);
+  const document =
+    doc !== undefined ? doc : await externalRefResolver.resolveDocument(base, ref!, true);
 
   if (document instanceof Error) {
     throw document;
@@ -69,7 +70,11 @@ export async function bundleDocument(opts: {
   const rules = config.getRulesForOasVersion(oasMajorVersion);
   const types = normalizeTypes(
     config.extendTypes(
-      customTypes ?? oasMajorVersion === OasMajorVersion.Version3 ? (oasVersion === OasVersion.Version3_1 ? Oas3_1Types : Oas3Types) : Oas2Types,
+      customTypes ?? oasMajorVersion === OasMajorVersion.Version3
+        ? oasVersion === OasVersion.Version3_1
+          ? Oas3_1Types
+          : Oas3Types
+        : Oas2Types,
       oasVersion,
     ),
     config,
@@ -182,7 +187,13 @@ function makeBundleVisitor(version: OasMajorVersion, dereference: boolean, rootD
         if (!componentType) {
           replaceRef(node, resolved, ctx);
         } else {
-          if (dereference) {
+          if (resolved.node.title === ctx.key && componentType === 'schemas') {
+            // if the schema name matches referenced file name we will
+            // just inline referenced schema
+            // w/o producing new schema name like (customer-2)
+            // for more details see: https://github.com/Redocly/openapi-cli/issues/349
+            replaceRef(node, resolved, ctx);
+          } else if (dereference) {
             saveComponent(componentType, resolved, ctx);
             replaceRef(node, resolved, ctx);
           } else {
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -187,10 +187,11 @@ function makeBundleVisitor(version: OasMajorVersion, dereference: boolean, rootD
         if (!componentType) {
           replaceRef(node, resolved, ctx);
         } else {
-          if (resolved.node.title === ctx.key && componentType === 'schemas') {
+          if (shouldPlaceSchemaReferenceInline(componentType, resolved, ctx)) {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: packages/core/src/bundle.ts [hunk 1]
```diff
@@ -182,7 +187,13 @@ function makeBundleVisitor(version: OasMajorVersion, dereference: boolean, rootD
         if (!componentType) {
           replaceRef(node, resolved, ctx);
         } else {
-          if (dereference) {
+          if (resolved.node.title === ctx.key && componentType === 'schemas') {
```
RomanHotsiy, line 190: Using schema.title is not a good idea. What if you have two different schemas with the same title? There may be no title at all too.

You should be using inferred schema name: https://github.com/Redocly/openapi-cli/blob/02ab3766ad45987beb630c72324e8cb9a623a0bb/packages/core/src/bundle.ts#L243

Also, there is no constraint here that this schema is referenced within components section.

The original issue is about:

```
schemas:
  A:
    $ref: ./A.yaml
```

This code will also work in this case:

```
/operation:
  put:
    requestBody:
      application/json:
        schema:
          type: object
          properties:
            A:
              $ref: ./A.yaml
```

RomanHotsiy (changes requested):
Almost there 👏

packages/core/src/bundle.ts
slavikbez: > You should be using inferred schema name:

before I check whether schema should be placed inline I need to know original name of the reference so I cannot use `getComponentName` in this case. I came up with[ this solution](https://github.com/Redocly/openapi-cli/pull/439/files#diff-0bad609a2903ba228646e389eaf5a4de07f9e9917b9811df3814180433e88585R309)

> This code will also work in this case:
```yaml
/operation:
  put:
    requestBody:
      application/json:
        schema:
          type: object
          properties:
            A:
              $ref: ./A.yaml
```

It won't. We're checking the component type [here](https://github.com/Redocly/openapi-cli/pull/439/files#diff-0bad609a2903ba228646e389eaf5a4de07f9e9917b9811df3814180433e88585R316). Also we have a [test](https://github.com/Redocly/openapi-cli/pull/439/files#diff-a813ab9a158e235b8602a38c1fd939472bd18641d11166b53bb6a4083ae4e9e0R91) that covers this case

RomanHotsiy: It will though. The test you linked is not for schemas but my example is. Try my example.

Also, I see a few cases that won't work correctly. I'll add a few tests to this PR (failing) tomorrow morning.

RomanHotsiy (approved):

Issue #281, opened 2021-04-14
openapi-cli bundle on Windows produces invalid schema names for types defined in externally referenced json schema files


austpaul:
**Describe the bug**
When running openapi bundle on Windows when externally referenced schema files are used, the schema name in the resulting API definition file contains parts of the external URL instead of the expected schema name.
This works as expected on Mac, and might be a Windows-only issue.

**To Reproduce**
1. Given an external json schema file
location: https://examplehostname.com/schemas/Account.yaml
```
type: object
title: Account
description: Resource representing an account
properties:
  id:
    type: string
    example: '123456789'
  name:
    type: string
    example: Contoso
```
2. And this local schema file in the API repository:
location: openapi/components/schemas/Customer.yaml
```
type: object
properties:
  customerName:
    type: string
  accounts:
    type: array
    items:
      $ref: 'https://examplehostname.com/schemas/Account.yaml'
```

3. Run openapi bundle

**Expected behavior**
Expected a valid OpenAPI definition with proper schema name for Account.
```
...
components:
  schemas:
    Customer:
      type: object
      properties:
        customerName:
          type: string
        accounts:
          type: array
          items:
            $ref: '#/components/schemas/Account'
    Account:
      type: object
      title: Account
      description: Resource representing an account
      properties:
        id:
          type: string
          example: '123456789'
        name:
          type: string
          example: Contoso
```

**Actual behavior**
The Account schema is named after the partial URL address. Including/omitting `title: Account` in the schema file has no impact on the behavior.
```
...
components:
  schemas:
    Customer:
      type: object
      properties:
        customerName:
          type: string
        accounts:
          type: array
          items:
            $ref: '#/components/schemas/https://examplehostname'
    'https://examplehostname':
      type: object
      title: Account
      description: Resource representing an account
      properties:
        id:
          type: string
          example: '123456789'
        name:
          type: string
          example: Contoso
```

**OpenAPI definition**
Specification version 3.0.1

**`openapi-cli` Version(s)**
1.0.0-beta.44

**`Node.js` Version(s)**
v12.14.1
(npm 6.14.4)

**Additional context**
Windows 10 pro, version 1909

PR #441, opened 2021-11-06
fix: resolve fetching of hosted schema on windows

slavikbez:
## What/Why/How?

Fixes [#281](https://github.com/Redocly/openapi-cli/issues/281) 

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -1,5 +1,4 @@
 import isEqual = require('lodash.isequal');
-import { sep as platformDependentSeparator } from 'path';
 import { BaseResolver, resolveDocument, Document } from './resolve';
 import { Oas3Rule, normalizeVisitors, Oas3Visitor, Oas2Visitor } from './visitors';
 import { Oas3Types } from './types/oas3';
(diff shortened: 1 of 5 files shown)


RomanHotsiy (approved):
Can we add test also?

packages/core/src/bundle.ts
```diff
@@ -271,7 +272,8 @@ function makeBundleVisitor(version: OasMajorVersion, dereference: boolean, rootD
       }
     }
 
-    name = refBaseName(fileRef) + (name ? `_${name}` : '');
+    const separator = isAbsoluteUrl(fileRef) ? '/' : platformDependentSeparator;
```
RomanHotsiy, line 275: Better to not use `platformDependentSeparator`... We should expect all paths in yaml file to have linux styled separators. Otherwise it won't work on our workflows runner.

Linux slashes work on windows too but not other way.

RomanHotsiy (changes requested):
Actually I found an issue.

packages/core/src/bundle.ts
slavikbez: In this case we're overriding separator only when absolute url is met. It won't affect relative references. Or I'm missing something ?

slavikbez:
> Can we add test also?

I was trying to add a test but later I realised that we mock entire node-fetch library [here](https://github.com/Redocly/openapi-cli/blob/master/__mocks__/node-fetch.ts) and I don't know how to override this mock using real node-fetch for my test

RomanHotsiy:
> and I don't know how to override this mock using real node-fetch for my test

@slavikbez I think you can use the mock and check if the function was called with right arguments. Or do I miss something?

packages/core/src/bundle.ts
RomanHotsiy: Oh, I see it was using path.sep before... I think this is a mistake. We should use `/[\/\\]/` regexp in windows, windows accepts forward slashes in most of its API at this point: both `C:\test/a/b/c` and `C:\test\a\b\c` are valid in Windows.

slavikbez:
> > and I don't know how to override this mock using real node-fetch for my test
> 
> @slavikbez I think you can use the mock and check if the function was called with right arguments. Or do I miss something?

Initially I wanted to fetch real hosted schema. But I followed your suggestion.

packages/core/src/bundle.ts
slavikbez: `/[\/\\]/` works good on mac. let me test it on windows

slavikbez: Works good on Windows as well 🎉

RomanHotsiy (approved):

Issue #396, opened 2021-10-07
`hide-x-internal` built-in decorators


adamaltman:
**Is your feature request related to a problem? Please describe.**

Hiding internal APIs is a common scenario.

**Describe the solution you'd like**

We have an article: https://redoc.ly/docs/cli/guides/hide-apis/ which shows how to make a decorator using `x-internal` to hide operations, paths, and schema properties. This could be improved to support parameters, etc. Also, it doesn't clean unused components (which should probably be its own built-in decorator).

**Describe alternatives you've considered**

None

**Additional context**

Related decorator ideas:

- Hide or select based on tags (for example, bundle only the customers and payments tags; or remove everything with the billing tag).

slavikbez:
Hiding parameters: https://github.com/Redocly/openapi-cli/issues/376#issuecomment-946861306

slavikbez:
@adamaltman should this be split to multiple decorators, like:
- hide-internal-paths
- hide-internal-operations
- hide-internal-properties
- hide-internal-parameters

adamaltman:
I don't think so, because people will want to hide the internal things or not hide them.

adamaltman:
However, I don't have a strong opinion either way... it just feels like too much to document.

RomanHotsiy:
I believe a single decorator makes more sense.

slavikbez:
I got stuck with tests here https://github.com/Redocly/openapi-cli/pull/443 and I need to switch to other more prioritised work. Maybe somebody will take over when time allows.

bojanz:
I have just posted a related issue detailing bugs when trying to make hide-internal-parameters work: #453.

bojanz:
Great job, everyone!

I have opened a followup bug report based on my testing of the new feature: #501.

PR #443, opened 2021-11-08
feat: remove-x-internal decorator

slavikbez:
## What/Why/How?
Closes [#396](https://github.com/Redocly/openapi-cli/issues/396) 

## Check yourself
- [x] Code is linted

## Security


packages/core/src/rules/common/hide-internals.ts
```diff
@@ -0,0 +1,95 @@
+import { Oas3Rule, Oas2Rule } from '../../visitors';
```
adamaltman, line 1: In general, we've put decorators into the rules folder and I think they should go into a decorators folder instead.

adamaltman:
```suggestion
import { Oas3Decorator, Oas2Decorator } from '../../visitors';
```
Also, I think these would be the correct types...

packages/core/src/rules/oas3/index.ts
```diff
@@ -106,4 +107,5 @@ export const decorators = {
   'operation-description-override': OperationDescriptionOverride as Oas3Decorator,
   'tag-description-override': TagDescriptionOverride as Oas3Decorator,
   'info-description-override': InfoDescriptionOverride as Oas3Decorator,
+  'hide-internals': HideInternals as Oas3Decorator,
```
Marshevskyy, line 110: would it work for OAS 3.1?

packages/core/src/walk.ts
```diff
@@ -212,16 +212,15 @@ export function walkDocument<T>(opts: {
...
                 resolvedNode,
                 context,
                 ruleId,
                 severity,
               );
-              if (ignoreNextVisitorsOnNode) {
-                break;
-              }
+              if (ignoreNextVisitorsOnNode) break;
+              if (isDeleted) return;
```
Marshevskyy, line 223: need unit test for this case

packages/core/src/walk.ts
```diff
@@ -235,7 +234,9 @@ export function walkDocument<T>(opts: {
           const itemsType = type.items;
           if (itemsType !== undefined) {
             for (let i = 0; i < resolvedNode.length; i++) {
+              const item = resolvedNode[i];
               walkNode(resolvedNode[i], itemsType, resolvedLocation.child([i]), resolvedNode, i);
+              if (item !== resolvedNode[i]) i--;
```
Marshevskyy, line 239: same, add unit test for this case

packages/core/src/rules/common/hide-internals.ts
```diff
@@ -0,0 +1,65 @@
...
+  'MediaType',
+  'Server',
+  'Callback',
+  'Parameter',
+];
+
+export const HideInternals: Oas3Decorator | Oas2Decorator = ({ hideTag }) => {
+  const hiddenTag = hideTag || DEFAULT_HIDDEN_TAG;
+  const isEmptyObject = (obj: any) => isPlainObject(obj) && Object.keys(obj).length === 0;
+  const isEmptyArray = (arr: any) => Array.isArray(arr) && arr.length === 0;
```
Marshevskyy, line 21: add better typing instead of `any`

packages/core/src/rules/common/hide-internals.ts
```diff
@@ -0,0 +1,65 @@
...
+  'Response',
+  'RequestBody',
+  'Example',
+  'MediaType',
+  'Server',
+  'Callback',
+  'Parameter',
+];
+
+export const HideInternals: Oas3Decorator | Oas2Decorator = ({ hideTag }) => {
```
Marshevskyy, line 18: there are a lot of conditions, please cover all of them with tests

Marshevskyy (commented):
good progress

packages/core/src/rules/oas3/index.ts
andriyl: @Marshevskyy - yes.

packages/core/src/rules/common/hide-internals.ts
andriyl: `tagToHide` option is covered with the test.

packages/core/src/rules/common/hide-internals.ts
andriyl: @Marshevskyy I've moved these functions to utils module.
And about typing - the parameter is type `any` because we expected any type there and that's why we check: is it object or array. Also, we've already such js utils in [CLI](https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/js-utils.ts).

adamaltman (changes requested):
This naming doesn't match the issue requesting it. `hide-internals` is different from `hide-x-internal`.

docs/commands/bundle.md
```diff
@@ -42,6 +42,7 @@ Option                 | Type      | Required?    | Default     | Description
 `--skip-preprocessor`  | `array`   | no           | -           | Ignore certain preprocessors. See the [Skip preprocessor, rule, or decorator section](#skip-preprocessor-rule-or-decorator) below
 `--skip-rule`          | `array`   | no           | -           | Ignore certain rules. See the [Skip preprocessor, rule, or decorator section](#skip-preprocessor-rule-or-decorator) below
 `--version`            | `boolean` | no           | -           | Show version number
+`--clean-components`   | `boolean` | no           | -           | Clean unused components
```
adamaltman, line 45: This should be `remove-unused-components`. Clean can mean different things to different people.

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
...
+The `tagToHide` option uses to define name of tag.
+
+```yaml
+lint:
+  decorators:
+    hide-internals:
+      tagToHide: 'hideit'
+```
+
+### clear-unused-components
```
adamaltman, line 94: Is this a decorator or a command line argument or both? Also, I would recommend to rename it.
```suggestion
### remove-unused-components
```

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
...
+
+Remove unused components. A good point to use with `hide-internals` decorator.
```
adamaltman, line 96:
```suggestion
Remove unused components. We recommend to use it with `hide-x-internal` decorator.
```

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### hide-internals
```
adamaltman, line 69: The opening issue calls for it to be `hide-x-internal` but I wonder if `remove-x-internal` would be more clear.

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
...
+  Response
+  RequestBody
+  Example
+  MediaType
+  Server
+  Callback
+  Parameter
+```
+
+The `tagToHide` option uses to define name of tag.
```
adamaltman, line 85: I don't understand this sentence or the example.

adamaltman: Oh, this is the tag that is evaluated -- default is `x-internal`?

adamaltman: I think given this the decorator name `hide-internals` could make sense (or `remove-internals`). However, this particular argument is unclear, because it doesn't just hide that tag, it hides the node/object it is included in such as an operation, property, etc...

docs/resources/built-in-decorators.md
adamaltman: I mention below I see this is more flexible by allowing another token instead of `x-internal` for evaluation of what to remove.

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
...
+  PathItem
+  Operation
+  Schema
+  Response
+  RequestBody
+  Example
+  MediaType
+  Server
+  Callback
+  Parameter
```
adamaltman, line 82: We are missing `Tag` too. That should be added. 

We are missing Schema property (this is different from hiding an entire schema -- just hiding a specific property).

docs/resources/built-in-decorators.md
adamaltman: I think this should not have the word `tag` at all because the word has semantic meaning in OpenAPI.

packages/cli/src/index.ts
```diff
@@ -209,6 +209,11 @@ yargs
           array: true,
           type: 'string',
         },
+        'clean-components': {
```
adamaltman, line 212: I would recommend this change to `remove-unused-components`. It used in a few different ways in this PR "clear components" and "clean components" and "clean unused components" and I think it should be standardized on "remove-unused-components".

docs/resources/built-in-decorators.md
andriyl: @adamaltman done.

docs/resources/built-in-decorators.md
andriyl: @adamaltman done.

docs/commands/bundle.md
andriyl: @adamaltman done.

docs/resources/built-in-decorators.md
andriyl: @adamaltman 
`remove-x-internal` sounds good, for now using `hide-x-internal`.
@RomanHotsiy what do you think?

packages/cli/src/index.ts
andriyl: @adamaltman done

docs/resources/built-in-decorators.md
andriyl: @adamaltman @RomanHotsiy, guys could you help me to define the name of this option please.

docs/resources/built-in-decorators.md
RomanHotsiy: `remove-x-internal` sounds better 👍

docs/resources/built-in-decorators.md
RomanHotsiy: I have a few suggestions:

- `propertyName`
- `byPropertyName` (it will have logical flow: remove-internals -> byPropertyName)
- `byProperty` (it will have logical flow: remove-internals -> byProperty)
- `internalPropertyName`

I don't have a strong opinion here. @adamaltman what do you think?

docs/resources/built-in-decorators.md
RomanHotsiy: Removing tags won't remove corresponding operations though (and I'm not sure this would be a good idea).

We should probably add a note to the docs because it may be confusing otherwise.

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,32 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### hide-internals
+
+Using `x-internal` tag by default to hide nodes:
```
RomanHotsiy, line 71:
```suggestion
Using a specific property (`x-internal` by default ) to remove nodes:
```

packages/core/src/rules/common/hide-internals.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+    if (isEmptyObject(node)) { delete parent[key]; ignoreRemoved(); }
+  }
+
+  const rootLeave = (root: any, { ignoreRemoved }: any) => {
+    const keysForRemove = getKeysForRemove(root);
+    if (keysForRemove.length) {
+      for (const rKey of keysForRemove) {
+        delete root[rKey];
+        ignoreRemoved();
```
RomanHotsiy: no need to call it here

packages/core/src/rules/common/hide-internals.ts
```diff
@@ -0,0 +1,63 @@
...
+      ignoreRemoved();
+    }
+  }
+
+  const leave = (node: any, { parent, key, ignoreRemoved }: any) => {
+    const keysForRemove = getKeysForRemove(node);
+    if (keysForRemove.length) {
+      for (const rKey of keysForRemove) {
+        delete parent[key][rKey];
+        ignoreRemoved();
```
RomanHotsiy: `removeSelf` instead `ignoreRemoved`

packages/core/src/walk.ts
```diff
@@ -358,16 +360,14 @@ export function walkDocument<T>(opts: {
           key,
           parentLocations: collectParentsLocations(context),
           oasVersion: ctx.oasVersion,
-          ignoreNextVisitorsOnNode: () => {
-            ignoreNextVisitorsOnNode = true;
-          },
+          ignoreNextVisitorsOnNode: () => { ignoreNextVisitorsOnNode = true; },
+          ignoreRemoved: () => { isDeleted = true; },
```
RomanHotsiy, line 364:
```suggestion
          removeSelf: () => {
            Array.isArray(parent)) ? parent.splice(key as number, 1) : delete parent[key];
            isDeleted = true;
          },
```

docs/resources/built-in-decorators.md
andriyl: done

packages/core/src/walk.ts
andriyl: @Marshevskyy `delete` logic removed, use another better approach.

docs/resources/built-in-decorators.md
andriyl: done

packages/core/src/walk.ts
andriyl: @Marshevskyy also removed this line.

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,295 @@
...
+          get:
+            parameters:
+              - $ref: '#/components/parameters/x'
+      components:
+        parameters:
+          x:
+            name: x
+		`);
+
+  it('should use `tagToHide` option to remove internal paths', async () => {
```
adamaltman, line 24: I think this should be changed?

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    const { bundle: res } = await bundleDocument({
+      document: testDocument,
+      externalRefResolver: new BaseResolver(),
+      config: makeConfig({}, { 'remove-x-internal': { 'tagToHide': 'hideit' } })
```
adamaltman, line 28: And this

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    `);
+  });
+
+  it('should remove unused components', async () => {
+    const { bundle: res } = await bundleDocument({
+      document: testDocument,
+      externalRefResolver: new BaseResolver(),
+      config: makeConfig({}, {
+        'remove-x-internal': { 'tagToHide': 'hideit' },
```
adamaltman, line 46: And this

adamaltman (changes requested):
A few minor changes. Looks much better.

docs/resources/built-in-decorators.md
adamaltman: I think @andriyl implemented `internalPropertyName` and it seems good.

docs/resources/built-in-decorators.md
adamaltman: No, we also need to remove tags though. Remember, we do render tags with no operations.

packages/core/src/rules/__tests__/hide-internals.test.ts
andriyl: done

packages/core/src/rules/__tests__/hide-internals.test.ts
andriyl: done

packages/core/src/rules/__tests__/hide-internals.test.ts
andriyl: done

docs/resources/built-in-decorators.md
andriyl: done

docs/resources/built-in-decorators.md
RomanHotsiy: What do you mean by "we also need to remove tags though"? Remove removed tag from all the operations that use it? Or remove operations? (I don't think we should remove operations)

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,20 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### remove-x-internal
+
+Using a specific property (`x-internal` by default ) to remove any node.
+
+The `internalPropertyName` option uses to define name of tag.
```
RomanHotsiy, line 73:
```suggestion
The `internalPropertyName` option can be used to define the name of the property that is used as flag for removal.
```

packages/cli/src/commands/bundle.ts
```diff
@@ -37,13 +37,20 @@ export async function handleBundle(
...
+    'remove-unused-components'?: boolean
   },
   version: string,
 ) {
   const config = await loadConfig(argv.config, argv.extends);
+  const removeUnusedComponentsArgv = 'remove-unused-components';
+  const removeComponents = argv[removeUnusedComponentsArgv];
+  if (removeComponents && !config.rawConfig.lint?.decorators?.hasOwnProperty(removeUnusedComponentsArgv)) {
+    config.lint.addDecorator(removeUnusedComponentsArgv);
+  }
```
RomanHotsiy, line 49:
```suggestion
  if (argv[removeUnusedComponentsArgv] && !config.rawConfig.lint?.decorators?.hasOwnProperty(removeUnusedComponentsArgv)) {
    config.lint.addDecorator(removeUnusedComponentsArgv);
  }
```

packages/core/src/rules/common/remove-x-internal.ts
```diff
@@ -0,0 +1,42 @@
+import { Oas3Decorator, Oas2Decorator } from '../../visitors';
+import { isEmptyArray, isEmptyObject, isPlainObject } from '../../utils';
+import { UserContext } from '../../walk';
+
+const DEFAULT_HIDDEN_TAG = 'x-internal';
```
RomanHotsiy, line 5: Rename to DEFAULT_INTERNAL_PROPERTY_NAME

packages/core/src/rules/oas2/remove-unused-components.ts
```diff
@@ -0,0 +1,55 @@
...
+    definitions.set(location.absolutePointer, {
+      used: definitions.get(location.absolutePointer)?.used || false,
+      name,
+    });
+  }
+
+  return {
+    ref(ref, { type, resolve, key }) {
+      if (
+        ['Schema', 'Header', 'Parameter', 'Response', 'Example', 'RequestBody'].includes(type.name)
```
RomanHotsiy, line 18: This is invalid

oas2 supports only 

- Schema
- Parameter
- Response
- SecurityScheme

packages/core/src/rules/oas2/remove-unused-components.ts
```diff
@@ -0,0 +1,55 @@
...
+    DefinitionRoot: {
+      leave(root) {
+        definitions.forEach(usageInfo => {
+          const { used, name } = usageInfo;
+          if (!used) { delete root.definitions![name]; }
+        });
+        if (isEmptyObject(root.definitions)) { delete root.definitions; }
+      },
+    },
+    NamedSchemas: {
```
RomanHotsiy, line 37: It's missing `NamedSecuritySchemas`

packages/core/src/rules/oas2/remove-unused-components.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (!resolvedRef.location) return;
+        definitions.set(resolvedRef.location.absolutePointer, {
+          used: true,
+          name: key.toString(),
+        });
+      }
+    },
+    DefinitionRoot: {
+      leave(root) {
+        definitions.forEach(usageInfo => {
```
RomanHotsiy, line 30: You have to do same also for
- `parameters`
- `responses`
- `securityDefinitions`

Not only `definitions`

packages/core/src/rules/oas3/remove-unused-components.ts
```diff
@@ -0,0 +1,103 @@
...
+  return {
+    ref(ref, { type, resolve, key }) {
+      if (
+        ['Schema', 'Header', 'Parameter', 'Response', 'Example', 'RequestBody'].includes(type.name)
+      ) {
+        const resolvedRef = resolve(ref);
+        if (!resolvedRef.location) return;
+        components.set(resolvedRef.location.absolutePointer, {
+          used: true,
+          parent: parsePointer(resolvedRef.location.pointer)[1] as keyof Oas3Components,
```
RomanHotsiy, line 26: no need to calculate parent here, is is not used anyways because `used: true`.

packages/core/src/rules/oas3/remove-unused-components.ts
```diff
@@ -0,0 +1,103 @@
+import { Oas3Rule } from '../../visitors';
+import { Location, parsePointer } from '../../ref-utils';
+import { Oas3Components } from '../../typings/openapi'
+import { isEmptyObject } from '../../utils';
+
+export const RemoveUnusedComponents: Oas3Rule = () => {
+  let components = new Map<string, { used: boolean; parent: keyof Oas3Components; name: string }>();
+
+  function registerComponent(location: Location, parent: keyof Oas3Components, name: string): void {
```
RomanHotsiy, line 9: rename `parent` to `componentType`

packages/core/src/rules/oas3/remove-unused-components.ts
```diff
@@ -0,0 +1,103 @@
...
+        });
+        if (isEmptyObject(root.components)) { delete root.components; }
+      },
+    },
+    NamedSchemas: {
+      Schema(schema, { location, key }) {
+        if (!schema.allOf) {
+          registerComponent(
+            location,
+            parsePointer(location.pointer)[1] as keyof Oas3Components,
```
RomanHotsiy, line 51: Do not parse pointer here, it is error prone, pointer can be to external file, instead just hardcode component type, much cleaner:

```suggestion
            'schemas',
```

RomanHotsiy (changes requested):

packages/core/src/rules/common/remove-x-internal.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const RemoveXInternal: Oas3Decorator | Oas2Decorator = ({ internalPropertyName }) => {
+  const hiddenTag = internalPropertyName || DEFAULT_HIDDEN_TAG;
```
adamaltman, line 8: This should also be renamed.

adamaltman (changes requested):
As I review this, I had minor changes to the documentation. Besides that, I wonder if `internalFlagProperty` would be better than `internalPropertyName`?

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,20 @@ lint:
...
+```yaml
+lint:
+  decorators:
+    remove-x-internal:
+      internalPropertyName: 'removeit'
+```
+
+### remove-unused-components
+
+Remove unused components. We recommend using it with `remove-x-internal` decorator.
```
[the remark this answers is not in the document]
andriyl: done

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,20 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### remove-x-internal
+
+Using a specific property (`x-internal` by default ) to remove any node.
```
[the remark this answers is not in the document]
andriyl: done

packages/cli/src/index.ts
```diff
@@ -209,6 +209,11 @@ yargs
           array: true,
           type: 'string',
         },
+        'remove-unused-components': {
+          description: 'Remove unused components',
```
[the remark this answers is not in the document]
andriyl: done

packages/core/src/rules/common/remove-x-internal.ts
andriyl: done

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,20 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### remove-x-internal
+
+Remove all nodes that have a specific `internalPropertyName` property (`x-internal` by default ).
```
adamaltman, line 71:
```suggestion
Remove all nodes that have a specific `internalPropertyName` property (`x-internal` by default).
```

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,20 @@ lint:
...
+```yaml
+lint:
+  decorators:
+    remove-x-internal:
+      internalPropertyName: 'removeit'
+```
+
+### remove-unused-components
+
+Removes unused components.
```
adamaltman, line 84:
```suggestion
Removes unused components.
Also can be used as a CLI argument `--remove-unused-components` to the `bundle` command.
```

adamaltman (changes requested):
I have left minor docs suggestions.

In addition, I have a few open questions:

1. Would `internalFlagProperty` be a better name than `internalPropertyName`? I think it may convey its meaning better.
2. I'm not clear on how we decide to make the `remove-unused-components` a CLI argument to the `bundle` command but not do the same for other built-in decorators.
3. I'm confused how someone will be able to use `remove-x-internal` in practice. If they enable the decorator in the `.redocly.yaml` file, then they need to set up an environment variable of `OPENAPI_CLI_BUNDLE_ARGS` with value `--skip-decorator=remove-x-internal` or something similar?

docs/resources/built-in-decorators.md
andriyl: done

docs/resources/built-in-decorators.md
andriyl: done

RomanHotsiy:
@adamaltman 

1. I agree that `internalFlagProperty` is a better name, @andriyl let's rename it
2. `remove-unused-components`, it has to go as the last decorator always so while it technically is implemented as decorator (similar to bundle decorator) I consider it more like a CLI argument. We should not document this as a decorator and we should not expose this decorator to be used externally.
3. Should be solved by https://github.com/Redocly/openapi-cli/pull/443

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,21 @@ lint:
...
+Also can be used as a CLI argument `--remove-unused-components` to the `bundle` command.
```
RomanHotsiy, line 85: I would not expose it as a decorator as it has to be run last which is a potential place for confusing/unexpected behaviors.

packages/core/src/rules/oas2/remove-unused-components.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+      },
+    },
+
+    NamedSchemas: {
+      Schema(schema, { location, key }) {
+        if (!schema.allOf) {
+          registerComponent(
+            location,
+            parsePointer(location.pointer)[0] as keyof Oas2Components,
```
RomanHotsiy, line 54: This is not correct in all the cases, I already mentioned it before but it was not resolved: https://github.com/Redocly/openapi-cli/pull/443#discussion_r776966814)

> Do not parse pointer here, it is error prone, pointer can be to external file, instead just hardcode component type here

RomanHotsiy (changes requested):
Left some change requests.

docs/resources/built-in-decorators.md
andriyl: done

packages/core/src/rules/oas2/index.ts
```diff
@@ -88,4 +90,6 @@ export const decorators = {
   'operation-description-override': OperationDescriptionOverride as Oas2Decorator,
   'tag-description-override': TagDescriptionOverride as Oas2Decorator,
   'info-description-override': InfoDescriptionOverride as Oas2Decorator,
+  'remove-x-internal': RemoveXInternal as Oas2Decorator,
+  'remove-unused-components': RemoveUnusedComponents as Oas2Decorator,
```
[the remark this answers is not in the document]
andriyl: done

docs/resources/built-in-decorators.md
```diff
@@ -65,3 +65,21 @@ lint:
       tagNames:
         pet: ./my-custom-description.md
 ```
+
+### remove-x-internal
+
+Remove all nodes that have a specific `internalPropertyName` property (`x-internal` by default).
+
+Remove additional remnants from components by also using the `remove-unused-components` decorator.
```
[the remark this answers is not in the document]
andriyl: done

packages/core/src/rules/common/test-decorator.ts
```diff
@@ -0,0 +1,14 @@
+import { UserContext } from '../../walk'
+
+export const TestDecorator: any = () => {
```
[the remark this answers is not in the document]
andriyl: I used this for a test, removed it.

packages/core/src/rules/oas2/remove-unused-components.ts
andriyl: done

__tests__/bundle/bundle-remove-unused-components/oas2/openapi.yaml
```diff
@@ -0,0 +1,18 @@
...
+  '/locations/{location-id}':
+    get:
+      description: Get information about a location.
+      responses:
+        '200':
+          description: Location information response.
+          x-internal: true
+          schema:
+            $ref: '#/definitions/MediaListResponse'
+definitions:
```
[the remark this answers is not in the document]
andriyl: done

__tests__/bundle/bundle-remove-unused-components/oas3/openapi.yaml
```diff
@@ -0,0 +1,13 @@
+openapi: 3.1.0
+
+paths:
+  /pet:
+    x-internal: true
+    get:
+      parameters:
+        - $ref: '#/components/parameters/x'
+
+components:
```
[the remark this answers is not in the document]
andriyl: done

adamaltman:
Additionally, Roman had an idea about changing it to `filter-by-permission`. 

Something like:

```
decorators:
  filter-by-permission:
    flagProperty: x-internal
    flagValue: true
```

This would allow for something more complex like:

```
decorators:
  filter-by-permission:
    flagProperty: x-permission
    flagValue: guest
```

And the property can be defined as a string or array of strings inside of any node in the OpenAPI definition like this:

```
sample:
  type: string
  x-permission:
    - guest
    - customer
    - admin
```

Finally, it would be more useful/powerful if the `flagProperty` and `flagValue` could be set by environment variable. (This env variable part could be a separate PR?)

---

Separately, I wonder if you should break `--remove-unused-components` into a separate PR so we can review/merge that faster?

RomanHotsiy:
I think we can merge both today. Env variables are not complicated so I would put them into this PR.

The only question is the names of those variables.

I was thinking about something like: `REDOCLY_ FILTER_BY_PERMISSION_PROPERTY` and `REDOCLY_ FILTER_BY_PERMISSION_VALUE`? What do you think?

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    });
+    expect(res.parsed).toMatchInlineSnapshot(
+    `
+    openapi: 3.0.0
+    components: {}
+
+    `);
+  });
+
+  it('should use `internalFlagProperty` and default value of `permissionValue`', async () => {
```
adamaltman, line 63:
```suggestion
  it('should use `permissionProperty` and default value of `permissionValue`', async () => {
```

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    openapi: 3.0.0
+    components:
+      parameters:
+        x:
+          name: x
+
+    `);
+  });
+
+  it('should use `internalFlagProperty` and `permissionValue`', async () => {
```
adamaltman, line 95:
```suggestion
  it('should use `permissionProperty` and `permissionValue`', async () => {
```

adamaltman (commented):
This seemed like a good idea. But I think it is a bit backwards now. I'll try to explain it.

Filter gives me the impression of inclusion... so if I filter by fruits, I expect to see apples, oranges, and bananas. 

Instead, this is more like filter out fruits... remove all the fruits, and you're left with the rest.

Filter in/out are both viable. One means to include everything that matches a condition, and the other is to exclude everything that does not match a condition. And I don't think we can have just one. 

In other words, imagine this scenario:

```
permissionProperty: x-internal
possible values:
  - true
  - false
```

Filter out (remove) seems easy and like the clear use case.

Now, imagine this scenario:

```
permissionProperty: x-permission
possible values:
  - guest
  - customer
  - admin
```

Filter out doesn't seem to make sense because we would need to filter out customer and admin. Also, customer and admin would have permissions of the guest too (in all practical cases).  So, the system breaks down in this case.

Instead, it needs a **filter-in** approach. And when you filter in, because you may not have tagged up every single node, you may need a default permission option too. This sets a default value for any node that doesn't have any permissions set explicitly.

```
permissionProperty: x-permission
permissionValue: customer
defaultPermissionValue: guest
```
Or the `defaultPermissionValue` is assumed to be the permission value itself.

So, I would imagine this actually more like two separate decorators:

- `filter-out-permission` (removes any node with the given permission)
- `filter-by-permission` (includes any node with the given permission)

packages/core/src/rules/__tests__/hide-internals.test.ts
andriyl: done

packages/core/src/rules/__tests__/hide-internals.test.ts
andriyl: done

packages/cli/src/commands/bundle.ts
```diff
@@ -156,6 +156,13 @@ export async function handleBundle(
           `📦 Created a bundle for ${blue(entrypoint)} at ${blue(outputFile)} ${green(elapsed)}.\n`,
         );
       }
+
+      const removedCount = meta.visitorsData['remove-unused-components']?.removedCount;
+      if (removedCount) {
+        process.stderr.write(
```
adamaltman, line 162: why is this stderr and not stdout?

RomanHotsiy (approved):
LGTM

Don't forget to add two records to the change log.

packages/cli/src/commands/bundle.ts
RomanHotsiy: Bundle goes into sdtout (if -o is not provided). All the logs go into stderr. This is common for CLI tools.

andriyl (approved):

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,234 @@
...
+            parameters:
+              - name: api_key
+            requestBody: {}
+      components: {}
+
+      `);
+  });
+});
+
+describe('oas2 hide-x-internal', () => {
```
adamaltman, line 190:
```suggestion
describe('oas2 remove-x-internal', () => {
```

packages/core/src/rules/__tests__/hide-internals.test.ts
```diff
@@ -0,0 +1,234 @@
+import { outdent } from 'outdent';
+import { bundleDocument } from '../../bundle'
+import { BaseResolver } from '../../resolve';
+import { parseYamlToDocument, yamlSerializer } from '../../../__tests__/utils';
+import { makeConfig } from './config';
+
+describe('oas3 hide-x-internal', () => {
```
adamaltman, line 7:
```suggestion
describe('oas3 remove-x-internal', () => {
```

adamaltman (approved):

PR #459, opened 2021-11-25
feat: support region for login and other commands

andriyl:
## What/Why/How?
Closes: #382

## Check yourself
- [x] Code is linted

## Security


packages/cli/src/commands/login.ts
```diff
@@ -0,0 +1,21 @@
+import { RedoclyClient } from '@redocly/openapi-core';
+import { blue, green } from 'colorette';
+import { promptUser } from '../utils';
+import { getRawConfigContent } from '../../../core/src/config/load';
```
Marshevskyy, line 4: use @redocly/openapi-core instead. do not couple cli with core this way

packages/cli/src/commands/login.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Region } from '@redocly/openapi-core';
```
Marshevskyy, line 5: join this with first import { RedoclyClient, Region } from '@redocly/openapi-core';

packages/core/src/config/load.ts
```diff
@@ -58,3 +55,15 @@ function findConfig() {
   }
   return undefined;
 }
+
+export async function getRawConfigContent(configPath = findConfig()) {
```
Marshevskyy, line 59: add unit test

packages/cli/src/commands/push.ts
```diff
@@ -14,32 +14,32 @@ import {
...
 import {
-  promptUser,
   exitWithError,
   printExecutionTime,
   getFallbackEntryPointsOrExit,
   pluralize,
   dumpBundle,
 } from '../utils';
+import { promptClientToken } from './login';
+import { getRawConfigContent } from '../../../core/src/config/load';
```
Marshevskyy, line 27: same

packages/cli/src/commands/push.ts
```diff
@@ -14,32 +14,32 @@ import {
...
   const isAuthorized = await client.isAuthorizedWithRedocly();
+
   if (!isAuthorized) {
-    const clientToken = await promptUser(
-      green(
-        `\n  🔑 Copy your API key from ${blue('https://app.redoc.ly/profile')} and paste it below`,
-      ),
-      true,
-    );
+    const clientToken = await promptClientToken(client.getDomain());
```
Marshevskyy, line 42: it looks similar to https://github.com/Redocly/openapi-cli/pull/459/files#diff-fa0a4d0e8d3c6b6651500902a739e43cfa0935d9d5c9ab939e62032ba1588e3dR16
why do not reuse it?

packages/core/src/redocly/index.ts
```diff
@@ -1,58 +1,78 @@
...
 import { resolve } from 'path';
 import { homedir } from 'os';
-import { yellow, red, green, gray } from 'colorette';
+import { red, green, gray } from 'colorette';
 import { RegistryApi } from './registry-api';
+import { DOMAINS, Region } from '../config/config';
 
 const TOKEN_FILENAME = '.redocly-config.json';
+const DEFAULT_REGION = 'us';
+const DEFAULT_DOMAIN = 'redoc.ly';
```
Marshevskyy, line 10: move this into constants since you use it in multiple places

packages/core/src/redocly/index.ts
```diff
@@ -1,58 +1,78 @@
...
     if (existsSync(credentialsPath)) {
-      const credentials = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
-      this.accessToken = credentials && credentials.token;
+      this.accessTokens = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
     }
   }
 
-  async isAuthorizedWithRedocly(): Promise<boolean> {
-    return this.hasToken() && !!(await this.getAuthorizationHeader());
+  async getValidTokens() {
```
Marshevskyy, line 57: add unit test

Marshevskyy (commented):
I know this is a draft version, I made a quick look and left a few comments, maybe most of them are. in your to-do list. I think it may be useful for you 😏

packages/core/src/config/load.ts
```diff
@@ -1,43 +1,40 @@
...
+
+  if (tokens.length) {
     if (!rawConfig.resolve) rawConfig.resolve = {};
     if (!rawConfig.resolve.http) rawConfig.resolve.http = {};
-    rawConfig.resolve.http.headers = [
-      {
-        matches: `https://api.${process.env.REDOCLY_DOMAIN || 'redoc.ly'}/registry/**`,
+    rawConfig.resolve.http.headers = [...(rawConfig.resolve.http.headers ?? [])];
+
+    for (const item of tokens) {
```
RomanHotsiy, line 27: This won't work with redocly.com domain (we discussed it we wan't to have it for future compatibility).

RomanHotsiy, line 27: This won't work with redocly.com domain (we discussed it we wan't to have it for future compatibility).

packages/core/src/redocly/index.ts
```diff
@@ -1,58 +1,78 @@
...
+  private region: Region;
   registryApi: RegistryApi;
 
-  constructor() {
-    this.loadToken();
-    this.registryApi = new RegistryApi(this.accessToken);
+  constructor(region?: Region) {
+    this.region = region || DEFAULT_REGION;
+    this.loadTokens();
+    this.setDomain(region);
```
RomanHotsiy, line 21: Let's get rid of this extra setter, it is not used anywhere else:

```suggestion
    this.domain = (region
      ? DOMAINS[region]
      : process.env.REDOCLY_DOMAIN || DOMAINS[DEFAULT_REGION]) || DEFAULT_DOMAIN;
```

We can also get rid of getter (if we add `|| DEFAULT_DOMAIN;` here

packages/core/src/config/load.ts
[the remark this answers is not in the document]
andriyl: @RomanHotsiy, yes - you are right, it's on my list.

packages/core/src/redocly/index.ts
```diff
@@ -1,58 +1,78 @@
...
+    return !!this.accessTokens;
+  }
+
+  loadTokens(): void {
     if (process.env.REDOCLY_AUTHORIZATION) {
-      this.accessToken = process.env.REDOCLY_AUTHORIZATION;
+      this.accessTokens = {};
+      for (const domain in DOMAINS) {
+        this.accessTokens[domain as Region] = process.env.REDOCLY_AUTHORIZATION;
+      }
```
RomanHotsiy, line 48: I thought a bit more about it. Let's use this token not for all domains but just for the current one: `this.domain`.

I believe this makes more sense.

```suggestion
      this.accessTokens[this.domain] = process.env.REDOCLY_AUTHORIZATION;
```

packages/core/src/redocly/index.ts
```diff
@@ -68,11 +88,11 @@ export class RedoclyClient {
       process.exit(1);
     }
 
-    this.accessToken = accessToken;
+    this.accessTokens = { [this.region]: accessToken };
     const credentials = {
-      token: accessToken,
+      ...(existsSync(credentialsPath) && JSON.parse(readFileSync(credentialsPath, 'utf-8'))),
```
RomanHotsiy, line 93: Move this into a separate function:

`existsSync(credentialsPath) && JSON.parse(readFileSync(credentialsPath, 'utf-8')`.

It can return `{}` by default (if file doesn't exist) which will simplify the code.

packages/core/src/redocly/index.ts
```diff
@@ -1,58 +1,78 @@
...
+        this.accessTokens[domain as Region] = process.env.REDOCLY_AUTHORIZATION;
+      }
       return;
     }
-
     const credentialsPath = resolve(homedir(), TOKEN_FILENAME);
     if (existsSync(credentialsPath)) {
-      const credentials = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
-      this.accessToken = credentials && credentials.token;
+      this.accessTokens = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
```
RomanHotsiy, line 53: This is backward-incompatible.

Most of the users have the following creds content on their machines:

```
{
  token: '....'
}
```

Those tokens will stop working. We need to have some logic to map old format to new.

I think if you move whole `JSON.parse(readFileSync(credentialsPath, 'utf-8'));` to separate function you can map it there.

I think it should do something like that:

```
const tokens = JSON.parse(readFileSync(credentialsPath, 'utf-8'));
if (tokens.token && !tokens[this.domain]) tokens[domain] = tokens.token;
```

RomanHotsiy (changes requested):
Left a few comments.

packages/core/src/config/load.ts
andriyl: this function was removed, now use the previous version of loadConfig

packages/cli/src/commands/push.ts
andriyl: thx, will do if it reasonable

packages/core/src/redocly/registry-api.ts
```diff
@@ -3,22 +3,25 @@ import { RegistryApiTypes } from './registry-api-types';
...
+  }
+
+  private async request(path = '', options: RequestInit = {}, accessToken?: string) {
+    if (!this.accessToken && !accessToken) {
       throw new Error('Unauthorized');
     }
 
     const headers = Object.assign({}, options.headers || {}, {
-      authorization: this.accessToken,
+      authorization: accessToken || this.accessToken,
```
Marshevskyy, line 20: what is the reason for adding this argument here?

packages/core/src/redocly/index.ts
```diff
@@ -1,78 +1,106 @@
...
-    }
+  getTokenByRegion() {
+    return this.accessTokens && this.accessTokens[this.region];
+  }
 
+  hasTokens(): boolean {
+    return !!this.accessTokens;
+  }
+
+  loadTokens(): void {
```
Marshevskyy, line 43: add unit test for this method

packages/core/src/config/load.ts
```diff
@@ -25,19 +24,31 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
...
-        value: (redoclyClient && (await redoclyClient.getAuthorizationHeader())) || '',
+        value: item.token,
       },
-      ...(rawConfig.resolve.http.headers ?? []),
-    ];
+      //support redocly.com domain for future compatibility
+      ...(item.region === 'us' ? [{
+        matches: `https://api.redocly.com/registry/**`,
+        name: 'Authorization',
+        envVariable: undefined,
```
Marshevskyy, line 46: why `envVariable: undefined`?

packages/core/src/redocly/registry-api.ts
andriyl: refactored

packages/core/src/config/load.ts
RomanHotsiy: Probably because there is check for undefined later in the code:

https://github.com/Redocly/openapi-cli/blob/master/packages/core/src/utils.ts#L48

docs/commands/login.md
```diff
@@ -14,6 +14,22 @@ Also, you will be able to access your members-only (private) API definitions in
 
 If you're having issues with the `login` command, use the `--verbose` option to display a detailed error trace (if any).
 
+## Region
```
RomanHotsiy, line 17: @skadinna I think we should reshape so it doesn't feel like users can choose where to login. It depends on the region their org is in. It can be only one region per org.

I would change it to something like:

if your redocly organization is in EU region then ...

packages/cli/src/__tests__/commands/login.test.ts
```diff
@@ -0,0 +1,19 @@
+import { RedoclyClient } from '@redocly/openapi-core';
+
+jest.unmock('@redocly/openapi-core');
+
+describe('login', () => {
+  it('should call login with setAccessTokens function', async () => {
```
Marshevskyy, line 6: move this test into core

packages/core/src/config/load.ts
```diff
@@ -25,19 +24,31 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
...
     if (!rawConfig.resolve.http) rawConfig.resolve.http = {};
-    rawConfig.resolve.http.headers = [
-      {
-        matches: `https://api.${process.env.REDOCLY_DOMAIN || 'redoc.ly'}/registry/**`,
+    rawConfig.resolve.http.headers = [...(rawConfig.resolve.http.headers ?? [])];
+
+    for (const item of tokens) {
+      const domain = DOMAINS[item.region as Region];
+      rawConfig.resolve.http.headers.push({
+        matches: `https://api.${domain}/registry/**`,
```
Marshevskyy, line 37: why not use getBaseUrl?

packages/core/src/config/load.ts
```diff
@@ -25,19 +24,31 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
...
         name: 'Authorization',
         envVariable: undefined,
-        value: (redoclyClient && (await redoclyClient.getAuthorizationHeader())) || '',
+        value: item.token,
       },
-      ...(rawConfig.resolve.http.headers ?? []),
-    ];
+      //support redocly.com domain for future compatibility
+      ...(item.region === 'us' ? [{
+        matches: `https://api.redocly.com/registry/**`,
```
Marshevskyy, line 44: same

docs/configuration/configuration-file.mdx
```diff
@@ -69,6 +72,42 @@ You may also import plugins and extends their configurations.
 
 Read more about the [lint configuration section](../guides/lint.md).
 
+
+### region
```
skadinna, line 76: I moved everything about the region setting to this config file, since it can be configured here as well. Then we can just link to this from the commands pages instead of repeating the same content.

However, I'm not sure the info here is 100% correct so it would be great if @adamaltman or @RomanHotsiy could take another look. Thanks! :)

andriyl: @skadinna tnx a lot

packages/core/src/redocly/index.ts
```diff
@@ -1,78 +1,114 @@
...
+      this.setAccessTokens({
+        ...this.accessTokens,
+        [this.region]: process.env.REDOCLY_AUTHORIZATION
+      })
     }
   }
 
-  async isAuthorizedWithRedocly(): Promise<boolean> {
-    return this.hasToken() && !!(await this.getAuthorizationHeader());
+  async getValidTokens() {
```
AlexVarchuk, line 63: If you want, please add a return type for the class methods.

packages/core/__tests__/login.test.ts
```diff
@@ -0,0 +1,17 @@
+import { RedoclyClient } from '../src/redocly';
+
+describe('loginnnnnn', () => {
```
Marshevskyy, line 3: too many nnnnnn )

Marshevskyy: I think this test should be moved into packages/core/src/redocly/__tests__ folder

packages/core/src/redocly/index.ts
Marshevskyy: @andriyl why did you resolve this with no comments? and no tests? 

add tests like this 
```
it('should loadTokens with no credentials ( NO .redocly-config.json)', async () => {
    const client = new RedoclyClient();
    const data = {};
    jest.spyOn(RedoclyClient.prototype, 'setAccessTokens');
    let spy = jest.spyOn(RedoclyClient.prototype, 'readCredentialsFile').mockImplementation(() => data);

    await client.loadTokens();
    expect(client.setAccessTokens).toHaveBeenCalledTimes(0);
    spy.mockRestore();
  });

  it('should loadTokens with old format credentials', async () => {
    const client = new RedoclyClient();
    const data = {"token": "test"};
    const expected = {"token": "test", "us": "test"};
    jest.spyOn(RedoclyClient.prototype, 'setAccessTokens');
    let spy = jest.spyOn(RedoclyClient.prototype, 'readCredentialsFile').mockImplementation(() => data);

    await client.loadTokens();
    expect(client.setAccessTokens).toHaveBeenLastCalledWith(expected);
    spy.mockRestore();
  });
  // it('should loadTokens with combination of old and new format', async () => {
    //   const data = {
    //     "token": "token old",
    //     "us": "token new"
    //   };
  // });
  // it('should loadTokens with "US"" region only', async () => {
  // });
  // it('should loadTokens with "EU" region only', async () => {
  // });
  // it('should loadTokens with process.env.REDOCLY_AUTHORIZATION + credentials', async () => {
  // });
  // it('should loadTokens with process.env.REDOCLY_AUTHORIZATION - credentials', async () => {
  // });
  ```
  
  maybe there are more use cases...

Marshevskyy (approved):

packages/core/src/redocly/index.ts
andriyl: done

packages/core/src/redocly/index.ts
andriyl: done

RomanHotsiy (approved):

PR #483, opened 2021-12-23
feat: option to skip bundling on push

SridarDhandapani:
A new cli option to skip bundling in push command. The default value of
the option (false) preserves the existing behaviour of the command.

## What/Why/How?
In the guide, [Hide your internal APIs](https://redoc.ly/docs/cli/guides/hide-apis/), it is suggest to add a special variable `OPENAPI_CLI_BUNDLE_ARGS` in registry environment variables setting to skip select decorators from running during build. However,  when pushing to the registry via `openapi push`, the bundle is invoked automatically  and the resulting API definition (after rules, preprocessor & decorators were applied) is pushed to the registry. This makes `OPENAPI_CLI_BUNDLE_ARGS` ineffective  and no way to restrict decorators when bundling on API registry.  

This PR adds a new cli option for `openapi push` to skip calling bundle before pushing to the API registry.  

## Check yourself

- [x] Code is linted

## Security


RomanHotsiy:
Hi @SridarDhandapani,

This PR is not complete. It doesn't account for any other dependent files. We're considering [adding an `enable-decorator` argument](https://github.com/Redocly/openapi-cli/issues/495) so you can make your decorators disabled by default and enable them using bundle args.

Will it solve your use case?

SridarDhandapani:
@RomanHotsiy, #495 looks like a good option. Happy to close this PR in favour of it.

adamaltman:
Let's keep this open for now as we're still reviewing the feasibility of #495. Sometimes decorators require configuration values. It's unclear (at the moment) if/how we'll take those from the command line.

However, I think we are scratching away at the right underlying problem -- of making decorators easier to use with multiple cases.

SridarDhandapani:
Fix for the issue addressed here is merged in https://github.com/Redocly/openapi-cli/pull/614.

SridarDhandapani:
Fix for the issue addressed here is merged in https://github.com/Redocly/openapi-cli/pull/614.

PR #489, opened 2021-12-28
feat: assertions rules

ohorbachevskyi:
closes https://github.com/Redocly/openapi-cli/issues/429
## What/Why/How?

This PR introduces new `openapi-cli` feature - `assertions` rules.
See detailed documentation here 
https://github.com/Redocly/openapi-cli/blob/e720a31e6d9cbb091468a325e745ca06fe0f2dd6/docs/resources/rules/assertions.md

## Testing
The changes are covered by unit and snapshot tests

## Screenshots (optional)
Screenshot 2022-03-30 at 11 18 47

## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/common/enforcements.ts
```diff
@@ -0,0 +1,87 @@
...
+        rulesMap[path][lastProp] = rulesToApply;
+      }
+    }
+  }
+
+  for (let path in rulesMap) {
+    let visitor = {};
+    const pathParts = path.split('.');
+    const lastNode = pathParts.pop() as string;
+    if (pathParts.length) {
```
RomanHotsiy, line 70: result of split is always non-empty array so this check seems to be redundant.

ohorbachevskyi: at line 69 we do `.pop()` (it extracts the last element of the array). So there we check if there are some elements after that.

RomanHotsiy: Oh, got it.

__tests__/lint/enforcements-casing-camel-case-error/.redocly.yaml
```diff
@@ -0,0 +1,11 @@
+apiDefinitions:
+  main: ./openapi.yaml
+
+lint:
+  rules:
+    enforcements:
+      - on:
+          - Operation.operationId
+        description: Operation id sould be camelCase
+        casing: PascalCase
```
slavikbez, line 10: from the file name looks like you're testing camel-case but here you specify PascalCase

__tests__/lint/enforcements-sort-order-missed-field-error/.redocly.yaml
```diff
@@ -0,0 +1,13 @@
+apiDefinitions:
+  main: ./openapi.yaml
+
+lint:
+  rules:
+    enforcements:
+      - on:
+          - Operation.tags
+        description: Operation tags should be ordered in DESC direction
```
slavikbez, line 9: I think message here should say something about missing fields in addition to incorrect sort order.

packages/core/src/config/minimal.ts
```diff
@@ -18,6 +18,7 @@ export default {
     'operation-description': 'off',
     'operation-2xx-response': 'warn',
     'operation-4xx-response': 'off',
+    'enforcements': 'warn',
```
slavikbez, line 21: should we turn off enforcements for `minimal` configuration ?

packages/core/src/rules/common/enforcements/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  let visitors = {};
+  let rulesMap: {[key: string]: any} = {};
+
+  // As 'enforcements' rule has an array of sub-rules,
+  // that array spreads into an 'opts' object on init rules phase,
+  // that is why we need to iterate through 'opts' values
+  const enforcements: any[] = Object.values(opts);
+
+  for (const enforcement of enforcements) {
+    if (enforcement.on) {
```
slavikbez, line 15: I think you can avoid nesting in code if you try to return early:
```js
if (!enforcement.on) {
  continue;
}
...
```

packages/core/src/rules/common/enforcements/index.ts
```diff
@@ -0,0 +1,58 @@
...
+      const onProps: string[] = Array.isArray(enforcement.on) ? enforcement.on : [enforcement.on];
+      const rulesToApply: Rule[] = Object.keys(genericRules).filter((rule: string) => enforcement[rule] !== undefined)
+        .map((rule: string) => ({
+          name: rule,
+          conditions: enforcement[rule],
+          description: enforcement.description,
+          severity: enforcement.severity || 'error'
+        }));
```
slavikbez, line 23:
```suggestion
      const rulesToApply: Rule[] =
        Object.keys(genericRules)
          .filter((rule: string) => enforcement[rule] !== undefined)
          .map((rule: string) => ({
            name: rule,
            conditions: enforcement[rule],
            description: enforcement.description,
            severity: enforcement.severity || 'error'
          }));
```

slavikbez: just a code formatting suggestion ^

__tests__/lint/enforcements-sort-order-missed-field-error/.redocly.yaml
ohorbachevskyi: The idea behind that is if we set non-existent field it means that the collection of objects is not ordered by that field, we return false. So it is not about missed field. Probably the folder name (which is actually describe's text) is confusing, I'll rename it to `enforcements-sort-order-error-when-missed-field`

packages/core/src/config/minimal.ts
ohorbachevskyi: good question, I think yes, we should do that. @RomanHotsiy @andriyl are you also agree?

slavikbez (approved):

packages/core/src/config/minimal.ts
adamaltman: what are enforcements?

adamaltman: Sorry, the PR description started with `enforcement` (but I think that's a typo in the PR description).

__tests__/lint/enforcements-all-rules-warn/.redocly.yaml
```diff
@@ -0,0 +1,34 @@
...
+        description: Operation summary example warn
+        severity: warn
+        pattern: '/test example/'
+        enum:
+          - summary test example
+        defined: false
+        nonEmpty: false
+        length: 16
+        minLength: 20
+        maxLength: 10
```
adamaltman, line 18: These aren't in the PR description of supported use cases.

adamaltman:
💪 This is going to be a powerful feature.

adamaltman:
I think we are missing some `casing` options such:

- flat
- macro
- cobol

https://stackoverflow.com/questions/17326185/what-are-the-different-kinds-of-cases

I have uses for both `flat` and `macro` so these are practical. 🤣

__tests__/lint/enforcements-all-rules-warn/.redocly.yaml
```diff
@@ -0,0 +1,34 @@
+apiDefinitions:
+  main: ./openapi.yaml
+
+lint:
+  rules:
+    enforcements:
+      - on:
+          - PathItem.Operation.summary
+        description: Operation summary example warn
```
adamaltman, line 9: Is the description a required property? What happens if it is not filled?

packages/core/src/config/minimal.ts
RomanHotsiy: @slavikbez @ohorbachevskyi I don't understand. What is it changing? If there are no enforcements defined in .redocly.yaml it's not going to report anything anyways. No?

@adamaltman what do you mean?

__tests__/lint/enforcements-all-rules-warn/.redocly.yaml
RomanHotsiy: Good question! @ohorbachevskyi make sure we have some good default description (ideally based on the failed enforcement)

__tests__/lint/enforcements-all-rules-warn/.redocly.yaml
ohorbachevskyi: Updated, thnx.

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
 
 OpenAPI-cli supports [custom rules](./custom-rules.md).
 However, if you have an idea for a built-in rule you believe will benefit the greater API community, please [open an issue](https://github.com/Redocly/openapi-cli/issues/new).
+
+## Generic built-in rules
+
+You can use special `enforcements` rule for cases which our other built-in rules do not cover.
+There are [a set of generic built-in rules](#list-of-generic-built-in-rules) which can be configured under `enforcements` to enforce different generic usecases.
```
RomanHotsiy, line 335: Won't it be confusing to have rules inside a rule? How about calling them `enforcements`? @skadinna @adamaltman

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### How it works?
+
+Let's say you want to validate Operation summary and Info description fields - they should be defined, not empty, 
+minimum 30 chars length and both have "_hello_" substring inside.
```
RomanHotsiy, line 340: I would use some more realistic example here. `_hello_` substring is not a realistic use case. How about "should end with a full stop"?

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+For these usecases you can create next enforcements:
```
RomanHotsiy, line 341: Isn't it a single usecase?

docs/resources/built-in-rules.md
RomanHotsiy: Then, nonEmpty and minimum 30 chars are redundant. You need only 30 chars.

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on
+        - Operation.summary
+        - Info.description
+        description: Summary and Description fields don't match enforcements
```
RomanHotsiy, line 350:
```suggestion
        description: Operation summary and description fields should be at least 30 characters and end with a full stop.
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+        severity: error
+        defined: true
+        nonEmpty: true
+        minLength: 30
+        regex: /hello/
+```
+
+In other words, it generates 8 unique custom rules `operation-summary-defined`, `operation-summary-non-empty`, `operation-summary-min-length`, `operation-summary-regex`,
+`info-description-defined`, `info-description-non-empty`, `info-description-min-length`, `info-description-regex`.
```
RomanHotsiy, line 359: This part is confusing. It doesn't generate 8 unique custom rules technically. Also what are those names? We don't have such rules. I would remove it.

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### Format of `enforcements` rule
```
RomanHotsiy, line 361:
```suggestion
### Format of enforcements
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+Basically `enforcements` rule is an array of enforcements, where each one has the next fields structure:
+- `on` - the path or array of the paths to the properties on which we should do the linting. The path is the string splitted by `.`.
+- `description` - the message which shows in case of fail. Not required.
```
RomanHotsiy, line 365:
```suggestion
- `description` - optional error message.
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+        regex: /hello/
+```
+
+In other words, it generates 8 unique custom rules `operation-summary-defined`, `operation-summary-non-empty`, `operation-summary-min-length`, `operation-summary-regex`,
+`info-description-defined`, `info-description-non-empty`, `info-description-min-length`, `info-description-regex`.
+
+### Format of `enforcements` rule
+
+Basically `enforcements` rule is an array of enforcements, where each one has the next fields structure:
+- `on` - the path or array of the paths to the properties on which we should do the linting. The path is the string splitted by `.`.
```
RomanHotsiy, line 364: This is not easy to understand unfortunately . We need to rewrite it.

It's not clear what should be put into on. We need to link to the list of node types.

@skadinna @adamaltman could you help?

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+- `description` - the message which shows in case of fail. Not required.
+- `severity` - the severity level of particular case (`off`, `warn`, `error`). Not required, by default `error`.
+- the [list of generic rules](#list-of-generic-built-in-rules) with their options.
+
+
+## List of generic built-in rules
```
RomanHotsiy, line 370: Again, shouldn't be called rules.

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### enum
+
+To enforce a value to be among set of predefined values.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 380:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### pattern
+
+To enforce a value to match a regex.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 396:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### casing
+
+To enforce specific casing style (camelCase, kebab-case, snake_case, PascalCase)
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.operationId
```
RomanHotsiy, line 410:
```suggestion
      - on: Operation.operationId
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+To mark some properties as mutually exclusive (`a` or `b` but not `a` and `b` together).
+This rule runs only on node. It means that the `on` path should end with the node, but not the property.
+In case if the last part of the path is the property, the rule will be ignored.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation
```
RomanHotsiy, line 426:
```suggestion
      - on: Operation
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+To mark some properties as mutually required (`a` and `b` but not only `a` or only `b`).
+This rule runs only on node. It means that the `on` path should end with the node, but not the property.
+In case if the last part of the path is the property, the rule will be ignored.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation
```
RomanHotsiy, line 444:
```suggestion
      - on: Operation
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### defined
+
+To check if some property is defined. 
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.operationId
```
RomanHotsiy, line 460:
```suggestion
      - on: Operation.operationId
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### undefined
+
+To check if some property is undefined.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 474:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### nonEmpty
+
+To check if some property is not empty.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 488:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### length
+
+To enforce length of specific array or string.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 502:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### minLength
+
+To enforce minimal length of specific array or string.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 516:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+### maxLength
+
+To enforce maximal length of specific array or string.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.summary
```
RomanHotsiy, line 530:
```suggestion
      - on: Operation.summary
```

docs/resources/built-in-rules.md
```diff
@@ -328,3 +328,225 @@ Here is the equivalent of the `recommended` configuration values:
...
+
+To enforce sort order in array of strings or array of objects (collection). 
+If the `property` fields is present, it means that we check the sort order direction on collection. 
+Also, in case of array of strings the direction could be also set like `sortOrder: asc`.
+
+```yaml
+lint:
+  rules:
+    enforcements:
+      - on Operation.tags
```
RomanHotsiy, line 546:
```suggestion
      - on: Operation.tags
```

RomanHotsiy (changes requested):
The docs are not yet ready to be shipped unfortunately.

I left a few suggestions but we need from help @skadinna or @adamaltman .

docs/resources/built-in-rules.md
RomanHotsiy: I think assertions is best

ohorbachevskyi:
> I think we are missing some `casing` options such:
> 
> * flat
> * macro
> * cobol
> 
> https://stackoverflow.com/questions/17326185/what-are-the-different-kinds-of-cases
> 
> I have uses for both `flat` and `macro` so these are practical. 🤣

@adamaltman just added next 3 `casing` options: `MACRO_CASE`, `COBOL-CASE`, `flatcase`

__tests__/lint/enforcements-all-rules-warn/.redocly.yaml
ohorbachevskyi: Added this one default description: "The assertion under <RULE_NAME> rule doesn't meet required conditions"

ohorbachevskyi: Added next one default description:
"The assertion under '${rule.name}' rule doesn't meet required conditions"

docs/resources/built-in-rules.md
adamaltman: If you don't mind, I'll make a commit adjusting these docs.

adamaltman:
- The `description` should be renamed to `message` to be consistent with what we name that.
- We are missing `suggest`. Suggestions are used in certain problem formats and also in the VS Code extension. We already support it in built-in rules and custom rules.
- I dislike the `sortOrder` because it is not a string or list of strings (so it is inconsistent with the other assertions) and would like to discuss it in more depth. Maybe there are no other solutions but I would like to hear what else was considered. If this is the least bad solution, then I'm all for it and I'll add it to the docs.
- Let's get rid of `length` and only have `minLength` and `maxLength`. If someone wants the `length` they can accomplish it by setting both min and max length to the same value. The benefit to us? Shorter and more clear docs (and I don't actually understand a practical use-case yet for exact length).

I've drafted changes to the docs here (it is incomplete... please read the PR description):
https://github.com/Redocly/openapi-cli/pull/506

adamaltman:
I've completed #506 . If there is any assert that does not support `$keys` please let me know. If so, that is not documented yet. Also, we should consider supporting it, because there is probably a use case.

In addition, I would like some way to assert that file names and folder names follow specific casing conventions. For example, at Rebilly they use singular-tense (not plural) PascalCase for schema names such as `Customer.yaml`. But I didn't find any "type" for file name or folder name.

adamaltman:
The biggest outstanding issue is with the `on` property. The dot notation is non-standard as far as being able to only use one single dot. It also mutes the potential of this rule.

__tests__/bundle/primitive-types/snapshot.js
```diff
@@ -30,7 +30,7 @@ paths:
                   decimal: 12.34
                   numberWithLeadingZero: 127
                   boolean: true
-                  date: 2020-01-01
+                  date: '2020-01-01'
```
Marshevskyy, line 33: why?

adamaltman:
Should we merge #506 into here?

__tests__/bundle/primitive-types/snapshot.js
ohorbachevskyi: because of that https://github.com/Redocly/openapi-cli/pull/523

ohorbachevskyi:
> Should we merge #506 into here?

I think yes, let me do that and all the rebase things, plan to do it asap, when we'll fix all the remaining issues that we have

ohorbachevskyi:
> The docs are not yet ready to be shipped unfortunately.
> 
> I left a few suggestions but we need from help @skadinna or @adamaltman .
Docs already updated. Please review

adamaltman:
Confirming we decided to omit `sort` assertions for now until we have an agreed upon design? I'm good with that.

adamaltman (approved):

adamaltman:
Note: My review is not on the code

__tests__/lint/assertions-casing-camel-case-error/.redocly.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - context:
+        - type: Operation
+          matchParentKeys: [get]
+        subject: Operation
+        property: operationId
+        message: Operation id should be canmelCase
+        casing: camelCase
+      - subject: Operation
+        property: operationId
+        message: Operation id should be camelCase
```
RomanHotsiy, line 17: Let's have different messages in this rule. Otherwise it's not clear in the test result.

I also think we should add some assertion id so it's easy to find it later from the error message.

What do you think?

__tests__/lint/assertions-casing-camel-case-error/snapshot.js
```diff
@@ -0,0 +1,54 @@
...
+Operation id should be camelCase
+
+29 |       description: example description
+30 | post:
+31 |   operationId: PostExampleId
+   |   ^^^^^^^^^^^
+32 |   summary: summary example
+33 |   responses:
+
+Error was generated by the assertions rule.
```
RomanHotsiy, line 45: Can we add assertion index (or also ideally assertion id, see above).

__tests__/lint/assertions-casing-camel-case-error/.redocly.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+lint:
+  rules:
+    assertions:
+      - context:
+        - type: Operation
+          matchParentKeys: [get]
+        subject: Operation
+        property: operationId
+        message: Operation id should be canmelCase
```
RomanHotsiy, line 13:
```suggestion
        message: Operation id for get operation should be canmelCase
```

__tests__/lint/assertions-enum-error/snapshot.js
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E lint assertions-enum-error 1`] = `
+
+validating /openapi.yaml...
+[1] openapi.yaml:20:7 at #/paths/~1pet~1findByStatus/get/summary
+
+Operation summary value should be among of predefined values
```
RomanHotsiy, line 8: can we show those values in the error message? We also have `suggest` in `report` args, we can use it here maybe.

__tests__/lint/assertions-error/snapshot.js
```diff
@@ -0,0 +1,328 @@
...
+Error was generated by the assertions rule.
+
+
+[8] openapi.yaml:25:7 at #/paths/~1pet~1findByStatus/put/summary
+
+Operation summary must have a maximum of 2 characters
+
+23 | put:
+24 |   operationId: exampleOne
+25 |   summary: My resource
```
RomanHotsiy, line 112: The error location is wrong, should be on value

__tests__/lint/assertions-error/snapshot.js
```diff
@@ -0,0 +1,328 @@
...
+
+[9] openapi.yaml:31:9 at #/paths/~1pet~1findByStatus/put/requestBody/content
+
+Only application/pdf can be used
+
+29 |   url: "https://example.com"
+30 | requestBody:
+31 |   content:
+   |   ^^^^^^^
+32 |     application/json:
```
RomanHotsiy, line 128: shouldn't the error location be on the wrong key instead of the on parent?

__tests__/lint/assertions-error/snapshot.js
```diff
@@ -0,0 +1,328 @@
...
+
+Summary must be one of the predefined values
+
+Did you mean:
+  - change to 'My resource'
+  - change to 'My collection'
+
+56 | post:
+57 |   operationId: EXAMPLE
+58 |   summary: ''
```
RomanHotsiy, line 172: Wrong location, should be on the value.

__tests__/lint/assertions-error/snapshot.js
```diff
@@ -0,0 +1,328 @@
...
+Error was generated by the assertions rule.
+
+
+[14] openapi.yaml:57:7 at #/paths/~1pet~1findByStatus/post/operationId
+
+NamedExamples key must be in PascalCase
+
+55 |                 description: hooray
+56 | post:
+57 |   operationId: EXAMPLE
```
RomanHotsiy, line 200: Should be on the value

__tests__/lint/assertions-min-length-string-error/snapshot.js
```diff
@@ -0,0 +1,26 @@
...
+exports[`E2E lint assertions-min-length-string-error 1`] = `
+
+validating /openapi.yaml...
+[1] openapi.yaml:9:3 at #/info/description
+
+Info description should be minimum 20 chars length
+
+ 7 |   url: http://example.xyz
+ 8 | title: Example OpenAPI 3 definition.
+ 9 | description: Example description
```
RomanHotsiy, line 12: Error location again (and everywhere)

packages/core/src/config/rules.ts
```diff
@@ -24,14 +24,25 @@ export function initRules<T extends Function, P extends RuleSet<T>>(
...
+            ruleId,
+            visitor: visitor,
+          }))
+        }
 
         return {
           severity: ruleSettings.severity,
           ruleId,
-          visitor,
+          visitor: visitors,
```
RomanHotsiy, line 40: Should we rename this prop to `visitors` too?

packages/core/src/rules/common/assertions/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (shouldRunOnKeys && assertion.property) {
+      throw new Error(`${shouldRunOnKeys.name} can't be used on a single property. Please use 'property'.`);
+    }
+
+    for (const subject of subjects) {
+      const lastVisitor = formLastVisitor(assertion.property, assertsToApply, assertion.context);
+      const visitorObject = assertion.context
+        ? buildVisitorObject(subject, assertion.context, lastVisitor)
+        : { [subject]: lastVisitor };
```
RomanHotsiy, line 49: We can incapsulate this ternary operator into `buildVisitorObject`. Would be easier to read.

ohorbachevskyi:
> Confirming we decided to omit `sort` assertions for now until we have an agreed upon design? I'm good with that.

Yes, we'll not introduce `sortOrder` now, it is removed from docs, but it is left in the code base

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    tmp = tmp[node.type];
+  }
+
+  tmp[subject] = lastVisitor;
+
+  return visitor;
+}
+
+export const formLastVisitor = (properties: string | string[], asserts: Assert[], _context?: Record<string, any>[]) =>
```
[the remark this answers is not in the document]
RomanHotsiy: I would rename it to `buildSubjectVisitor`

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+        }
+      } else {
+        doLint(Object.keys(node), assert, location.key(), report);
+      }
+  }
+}
+
+
+export const getCounts = (keys: string[], properties: string[]): number => {
+  let counter = 0;
```
RomanHotsiy, line 91:
```suggestion
  let count = 0;
```

RomanHotsiy (changes requested):
Great job. This is very close.

Left a bunch of code improvement comments as well as some suggestions for better functionality.

__tests__/lint/assertions-casing-camel-case-error/.redocly.yaml
ohorbachevskyi: Will be like this (where `Hello world` is predefined assertion id, if no  assertion id then assertion #number)

Screenshot 2022-03-30 at 16 54 29
Screenshot 2022-03-30 at 16 51 06

__tests__/lint/assertions-enum-error/snapshot.js
ohorbachevskyi: added `suggest`

packages/core/src/config/rules.ts
ohorbachevskyi: Actually no, because we always return only one visitor or array of visitors. 
Agree that it's little bit confusing here in line 40, because `visitors` variable is an object (only one single visitor). 
Left a comment in code which clarifies that.

packages/core/src/rules/common/assertions/index.ts
```diff
@@ -0,0 +1,56 @@
...
+export const Assertions: Oas3Rule | Oas2Rule = (opts: object) => {
+  let visitors: any[] = [];
+
+  // As 'Assertions' has an array of asserts,
+  // that array spreads into an 'opts' object on init rules phase,
+  // that is why we need to iterate through 'opts' values
+  const assertions: any[] = Object.values(opts);
+
+  for (const assertion of assertions) {
+    if (!assertion.subject) {
```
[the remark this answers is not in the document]
ohorbachevskyi: Will throw "assertion #number `subject` is required for assertion"

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    if (matchParentKeys || excludeParentKeys) {
+      tmp[node.type] = {
+        skip: (_value: any, key: string) => {
+          if (matchParentKeys) {
+            return !matchParentKeys.includes(key);
+          }
+          if (excludeParentKeys) {
+            return excludeParentKeys.includes(key);
+          }
```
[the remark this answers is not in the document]
ohorbachevskyi: Will throw next error
`Both 'matchParentKeys' and 'excludeParentKeys' should not be present under one context item`

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
...
+
+## OpenAPI node types
+
+Redocly defines a type tree based on the document type.
+OpenAPI 2.0 has one type tree.
+OpenAPI 3.0 and OpenAPI 3.1 share a type tree.
+
+### List of OpenAPI types
+
+Implementation of types for each specific OAS version:
```
skadinna, line 419:
```suggestion
For technical details on the implementation of types for each OAS version, consult the source files in the OpenAPI CLI repository:
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
...
+### `context` example
+
+The following example asserts that PUT responses with HTTP status 200 or 201 cannot return an `application/pdf`content type.
+Without the `context`, the assertion would evaluate every MediaTypeMap including:
+- Responses with all codes, including codes other than 200 or 201
+- Responses for all HTTP methods, including DELETE, GET, POST, and more.
+To restrict the evaluation, use the `context` feature to limit what will be evaluated.
+
+```yaml
+- context:
```
skadinna, line 164: issue: this example looks incomplete - it doesn't show where to place `context` in the `lint` section. It should be on the level of a single assertion, right?

__tests__/lint/assertions-error/snapshot.js
ohorbachevskyi: It is not possible if we will have more than 1 wrong key. It's just current example has only 1.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
+# `assertions`
+
+Configure assertions to enforce your API design standards without coding custom rules.
```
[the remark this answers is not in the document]
ohorbachevskyi: @skadinna Yes, I thinks we can start with your explanation, it's good enough! Thank you!

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
...
+
+Property | Type | Description
+-- | -- | --
+assertions | [[Assertion object]](#assertion-object) | A list of assertions to enforce your custom API design standards. Add or edit your assertions in the configuration file. The `assertions` rule is composed of a list of `assertion` objects. This rule may result in more than one problem message as you may define more than one assertion. More than one assertion may be defined, and an assertion may have more than one assert.
+
+## Assertion object
+
+Property | Type | Description
+-- | -- | --
+subject | `string` \| [`string`] | **REQUIRED.** The [OpenAPI node type](#openapi-node-types) that the the lint subjects to its evaluation. Use with `context` for more control.
```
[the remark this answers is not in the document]
ohorbachevskyi: I think both variants are ok, lets leave as it is if you don't mind.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
...
+property | `string` \| [`string`] \| null | Property name corresponding to the [OpenAPI node type](#openapi-node-types). If a list of properties is provided, assertions will evaluate against each property in the sequence. If not provided (or null), assertions will evaluate against the key names for the subject node type. See [property example](#property-example).
+context | [Context object](#context-object) | The context influences evaluation for assertions. When `matchParentKeys` or `excludeParentKeys` is used in the `context` object, it evaluates the specified subset of the subject type. The resolution of reference objects is done at the context level. If no context is provided, it evaluates the assertion for all instances of the given type. See [context example](#context-example). 
+message | `string` | Problem message displayed if the assertion is false.
```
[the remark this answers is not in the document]
ohorbachevskyi: So all that fields are optional. I'll add explanation about default values

docs/resources/rules/assertions.md
ohorbachevskyi: yes, it should be on the level of a single assertion, so I can add `assertions` at the top:
```suggestion
assertions:
  - context:
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,471 @@
...
+        severity: error
+        mutuallyExclusive:
+          - description
+          - externalDocs
+```
+
+### `mutuallyRequired` example
+
+The following example asserts that a response body schema must have both `amount` and `currency` properties (and not either one by itself).
+This assertion runs only on node's keys.
```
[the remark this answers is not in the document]
ohorbachevskyi: @skadinna Maybe something like this?
```suggestion
This assertion evaluates only node's properties keys, but not properties values.
```

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
ohorbachevskyi: Put your text after the first yaml example

__tests__/lint/assertions-casing-camel-case-error/.redocly.yaml
RomanHotsiy: Great!

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
RomanHotsiy: I would simplify. It sounds too complicated to me too.

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
RomanHotsiy: I like it more @ohorbachevskyi 👍 .

@skadinna how about you?

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
RomanHotsiy: It's good explanation @skadinna. I would fix it a bit:

> the linter shows you a warning message in the log

change to:

the linter shows you a warning or error message in the log."

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
RomanHotsiy: I suppose they will see some default message. right?

packages/core/src/rules/common/assertions/__tests__/asserts.test.ts
```diff
@@ -166,26 +166,26 @@ describe('oas3 assertions', () => {
...
-        expect(asserts.sortOrder(['test', 'foo', 'example'], 'desc')).toBeTruthy();
-        expect(asserts.sortOrder(['test', 'foo', 'example'], {direction: 'desc'})).toBeTruthy();
-        expect(asserts.sortOrder(['example'], 'desc')).toBeTruthy();
-        expect(asserts.sortOrder(['example', 'test', 'foo'], 'desc')).toBeFalsy();
-        expect(asserts.sortOrder(['test', 'foo', 'example'], 'asc')).toBeFalsy();
-        expect(asserts.sortOrder([{name: 'foo'}, {name: 'baz'}, {name: 'bar'}], {direction: 'desc', property: 'name'})).toBeTruthy();
-        expect(asserts.sortOrder([{name: 'foo'}, {name: 'baz'}, {name: 'bar'}], {direction: 'asc', property: 'name'})).toBeFalsy();
-      });
-    });
+    // describe('sortOrder', () => {
```
RomanHotsiy, line 169: Maybe replace it with `describe.skip` and keep it uncommented. Thx.

docs/resources/rules/assertions.md
```diff
@@ -24,6 +24,7 @@ context | [Context object](#context-object) | The context influences evaluation
 message | `string` | Problem message displayed if the assertion is false.
 suggest | [`string`] | List of suggestions to display if the problem occurs.
 severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
+assertionId | `string` | Id which appears in the lint result if the assertion is false. It helps to define the assertion.
```
RomanHotsiy, line 27: `if the assertion is false.` -> `if the assertion fails`

RomanHotsiy (approved):
Thanks for improvements! 

Looks great. I would like to further improve some error locations for some specific assertions but this is pretty good as is now. We can improve those later.

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
ohorbachevskyi: thnx, already added.

docs/resources/rules/assertions.md
ohorbachevskyi: already added assertions at the top

docs/resources/rules/assertions.md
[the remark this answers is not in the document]
ohorbachevskyi: yes, there is default `message`. All that things already added.

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const getIntersectionLength = (keys: string[], properties: string[]): number => {
+  const props = new Set(properties);
+  let count = 0;
+  for (const key of keys) {
+    if (props.has(key)) {
+      count++;
+    }
+  }
+  return count;
```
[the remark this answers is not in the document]
RomanHotsiy: I would keep the original due to perf reasons.
This function is on the hot-path so performance matters.

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+      currValue = value[i][property];
+      prevVal = value[i-1][property]
+    }
+
+    result = direction === 'asc' ? currValue >= prevVal : currValue <= prevVal;
+    if (!result) {
+      break;
+    }
+  }
+  return result;
```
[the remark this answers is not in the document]
RomanHotsiy: This function is pretty easy to read though. I'm not sure if simplification is needed.

I would do some tiny change, instead of:

```diff
  if (!result) {
-   break;
+   return;
  }
```

and then


```diff
return result;
return true;
```

It makes it a little bit simpler to follow (in my opinion).

packages/core/src/rules/common/assertions/utils.ts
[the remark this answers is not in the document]
tatomyr: The performance should be about the same, 6-7% difference according to https://jsbench.me. Is this really worth it?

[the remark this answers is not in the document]
RomanHotsiy: Yes. This is on hot path, this function may be executed for every value in the api definition (multiple times). So even a small difference matters here.

Especially when readability is still not too bad.

We can add extra benchmark for assertions later and see if it really matters (it may not in fact). But i would keep it as is for now.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,476 @@
...
+context | [Context object](#context-object) | The context influences evaluation for assertions. When `matchParentKeys` or `excludeParentKeys` is used in the `context` object, it evaluates the specified subset of the subject type. The resolution of reference objects is done at the context level. If no context is provided, it evaluates the assertion for all instances of the given type. See [context example](#context-example). 
+message | `string` | Optional problem message displayed if the assertion is false. By default displayed "The assertion doesn't meet required conditions".
+suggest | [`string`] | Optional list of suggestions to display if the problem occurs.
+severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Optional field, default value is `error`.
```
skadinna, line 29:
```suggestion
severity | `string` | Optional field to configure the severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,476 @@
...
+assertionId | `string` | Id which appears in the lint result if the assertion fails. It helps to define the assertion.
```
skadinna, line 30:
```suggestion
assertionId | `string` | ID displayed in the lint log if the assertion fails. Useful for troubleshooting and identifying the assertion.
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,476 @@
...
+        severity: error
+        mutuallyExclusive:
+          - description
+          - externalDocs
+```
+
+### `mutuallyRequired` example
+
+The following example asserts that a response body schema must have both `amount` and `currency` properties (and not either one by itself).
+This assertion evaluates only node's properties keys, but not properties values.
```
skadinna, line 270:
```suggestion
This assertion evaluates only property keys for the node, but not property values.
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,476 @@
...
+- kebab-case
+- snake_case
+- PascalCase
+- MACRO_CASE
+
+
+### `mutuallyExclusive` example
+
+The following example asserts the operation `description` and `externalDocs` must be mutually exclusive.
+This assertion evaluates only node's properties keys, but not properties values.
```
skadinna, line 253:
```suggestion
This assertion evaluates only property keys for the node, but not property values.
```

skadinna (approved):
Docs look great :+1: thanks for your help, everyone!

packages/core/src/rules/common/assertions/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (shouldRunOnKeys && assertion.property) {
+      throw new Error(`${shouldRunOnKeys.name} can't be used on a single property. Please use 'property'.`);
+    }
+
+    for (const subject of subjects) {
+      const subjectVisitor = buildSubjectVisitor(assertion.property, assertsToApply, assertion.context);
+      const visitorObject = buildVisitorObject(subject, assertion.context, subjectVisitor);
+      visitors.push(visitorObject);
+    }
+  }
```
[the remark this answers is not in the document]
ohorbachevskyi: What the benefit of avoiding mutation here?
Then it should be something like that:
```suggestion
    visitors = [
       ...visitors,
       ...subjects.map(subject => {
            const subjectVisitor = buildSubjectVisitor(assertion.property, assertsToApply, assertion.context);
            return buildVisitorObject(subject, assertion.context, subjectVisitor);
       })
    ]
```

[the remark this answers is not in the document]
tatomyr: 1. It always easier to keep track on your code if your variables don't change. 
2. In this case you don't need to spread `...visitors` since you there's no value assigned before unless I'm missing something. The only usage of `visitors` I see is here: https://github.com/Redocly/openapi-cli/pull/489/files?file-filters%5B%5D=.ts&show-viewed-files=true#diff-d9c7573397e9ca9187db6db4319df03e40bb5ec9a3eb067c973b0c225167978bR6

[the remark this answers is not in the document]
ohorbachevskyi: The `visitors` array mutates through the 2 cycles, and as result we want to see a flat array of visitors.
Just discussed with you on the call 😃

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+  return visitor;
+}
+
+export function buildSubjectVisitor(properties: string | string[], asserts: Assert[], context?: Record<string, any>[]) {
+  return function(node: any, { report, location, key, type }: UserContext) {
+
+    // check context for same node type parent includes/excludes:
+    if (context) {
+      const lastContextNode = context[context.length-1];
+      if (lastContextNode.type === type.name) {
```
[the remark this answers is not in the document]
ohorbachevskyi: yes, added

PR #496, opened 2022-01-03
feat: add checkstyle as new lint format

twz123:
## What/Why/How?

Add another format for linting: Checkstyle. Can be used to pipe linting results into Jenkins, for example. Implementation is adapted from the one [found in ESLint](https://github.com/eslint/eslint/blob/v8.6.0/lib/cli-engine/formatters/checkstyle.js).

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/docs/commands/lint.md
+++ b/docs/commands/lint.md
@@ -159,12 +159,10 @@ openapi lint --format=checkstyle
 </checkstyle>
 ```
 
-Writes linting results to standard output in the [Checkstyle] XML report format.
-Due to the limitations of this format, only file name, line, column, severity
+In this format, `lint` uses the [Checkstyle](https://checkstyle.org/) XML report format.
+Due to the limitations of this format, only file name, line, column, severity,
 and rule ID (in the `source` attribute) are included. All other information is
 omitted.
-
-[Checkstyle]: https://checkstyle.org/
 
 ### Max problems
 
--- a/docs/commands/lint.md
+++ b/docs/commands/lint.md
@@ -27,7 +27,7 @@ Option                   | Type      | Required    | Default     | Description
 `entrypoints`            | `array`   | no           | -           | Array of API definition filenames that need to be linted. See [the Entrypoints section](#entrypoints) for more options
 `--config`               | `string`  | no           | -           | Specify path to the [config file](#custom-configuration-file)
 `--extends`              | `array`   | no           | -           | [Extend a specific configuration](#extend-configuration) (defaults or config file settings)
-`--format`               | `string`  | no           | `codeframe` | Format for the output.<br />**Possible values:** `codeframe`, `stylish`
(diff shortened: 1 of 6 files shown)


docs/commands/lint.md
```diff
@@ -142,6 +142,30 @@ openapi/core.yaml:
...
+</file>
+</checkstyle>
+```
+
+Writes linting results to standard output in the [Checkstyle] XML report format.
+Due to the limitations of this format, only file name, line, column, severity
+and rule ID (in the `source` attribute) are included. All other information is
+omitted.
+
+[Checkstyle]: https://checkstyle.org/
```
RomanHotsiy, line 167: Same as above.

```suggestion
In this format, `lint` uses the [Checkstyle](https://checkstyle.org/) XML report format.
Due to the limitations of this format, only file name, line, column, severity,
and rule ID (in the `source` attribute) are included. All other information is
omitted.
```

packages/core/src/format/format.ts
```diff
@@ -126,6 +126,33 @@ export function formatProblems(
...
+        writeln(`<file name="${xmlEscape(path.relative(cwd, file))}">`);
+        for (let i = 0; i < fileProblems.length; i++) {
+          const problem = fileProblems[i];
+          const { line, col } = problem.location[0].start;
+          const severity = problem.severity == 'warn' ? 'warning' : 'error';
+          const message = xmlEscape(problem.message);
+          const source = xmlEscape(problem.ruleId);
+          writeln(
+            `<error line="${line}" column="${col}" severity="${severity}" message="${message}" source="${source}" />`,
+          );
```
RomanHotsiy, line 147: Could we move this into a separate function `formatCheckstyle` similar to `stylish` to make it consistent? Also, switch to `for of` instead of regular for loop.

RomanHotsiy (commented):
Wow! This looks pretty good!

Thanks a lot for the contribution! I've left a few suggestions/comments.

Would you be able to resolve them? We can finish this too.

packages/core/src/format/format.ts
twz123: I somehow felt that, without a function, it's easier to grasp the report file format, given that it's just a few lines.

twz123: Oh, then I also have to toss away the writeln arrow function.

RomanHotsiy: @twz123 I agree it's somehow easier but I from the other perspective consistency also is important.

You should extract writeln to the module scope. It doesn't have any closure deps.

RomanHotsiy:
Adding @skadinna to review docs.

RomanHotsiy (approved):

RomanHotsiy:
Thanks for updates @twz123!

I'll wait for somebody to confirm docs update and merge it after.

skadinna (approved):
Docs look good to me :+1: thanks for your contribution, @twz123! :clap:

adamaltman:
Thank you @twz123 !

PR #498, opened 2022-01-06
docs: simplify language in changelog and use past tense

adamaltman:
## What/Why/How?

I think we should be more simple in the changelog:

- Use "fixed" instead of "resolved".
- Use "added" instead of "implemented".
- Consistently use past-tense.

## Security


docs/changelog.md
```diff
@@ -144,17 +144,17 @@ lint:
...
 
 ## 1.0.0-beta.64 (2021-10-20)
 
 ### Fixes
 
-- Resolved an issue with the `--format` option not working with the `bundle` command.
+- Fixed an issue with the `--format` option not working with the `bundle` command.
 
-- Resolved a validation issue with the non-string `openapi` value in API definitions. The `lint` command now warns if the value is not string instead of crashing.
+- Fixed a validation issue with the non-string `openapi` value in API definitions. The `lint` command now warns if the value is not string instead of crashing.
```
adamaltman, line 157:
```suggestion
- Fixed a validation issue with the non-string `openapi` value in API definitions. The `lint` command warns if the value is not string instead of crashing.
```

skadinna (approved):
I'm OK with this change as long as we don't lose the details (e.g. if we start writing just "Added X" instead of "Added X to [benefit for the user], I think the changelog won't be as valuable anymore).

Next steps:
- updating these templates: https://github.com/Redocly/intranet/blob/main/engineering/documentation-process.md
- applying this retroactively to all other changelogs

PR #503, opened 2022-01-10
fix: cleanup version in e2e tests

andriyl:
## What/Why/How?
- remove version in bundle-lint-format test - json format
- add `cleanUpVersion` function to print in Serializer - all snapshots printed without package version

## Check yourself
- [x] Code is linted

## Security

### DIFF
--- a/__tests__/index.test.ts
+++ b/__tests__/index.test.ts
@@ -17,7 +17,7 @@ addSerializer({
 });
 
 function cleanUpVersion(str: string): string {
-  return str.replace(/"version":\s(\".*\"),\s*/g, '');
+  return str.replace(/"version":\s(\".*\")*/g, '"version": "<version>"');
 }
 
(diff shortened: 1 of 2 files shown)


__tests__/index.test.ts
```diff
@@ -13,9 +13,13 @@ expect.extend({
 
 addSerializer({
   test: (val: any) => typeof val === 'string',
-  print: (v: any) => v as string,
+  print: (v: any) => cleanUpVersion(v),
 });
 
+function cleanUpVersion(str: string): string {
+  return str.replace(/"version":\s(\".*\"),\s*/g, '');
```
RomanHotsiy, line 20: I would keep version to check if it is not removed in the future:

```suggestion
  return str.replace(/"version":\s(\".*\"),\s*/g, '"version": "<version>"');
```

RomanHotsiy (approved):

Issue #429, opened 2021-11-01
feature: assertions


RomanHotsiy:
## Request for comments
This feature request is not yet actionable. We need to collect some feedback first. 

---

**Is your feature request related to a problem? Please describe.**
We need a generic rules to be used in cases which our other built-in rules do not cover.

**Describe the solution you'd like**
I'd like to have a set of built-in rules which can be configured to enforce different generic usecases:

- `enum` to enforce a value to be among set of predefined values
- `pattern` - to enforce a value to match a regex
- `casing` - to enforce specific casing style (camelCase, kebab-case, snake_case, PascalCase)
- `mutuallyExclusive` - to mark some properties as mutually exclusive (`a` or `b` but not `a` and `b` together)
- `defined`/`undefined`/`nonEmpty`
- `length` - to enforce length of specific array or string
- `sortOrder` - to enforce sort order in array of strings or array of objects (collection)

Example usage:

```yaml
lint:
  rules:
    enforcements: # I would like a better suggestion here
      - on: SecuritySchema.type
        description: Only API key authorization is allowed
        severity: error
        enum:
          - apiKey
      - on: 'MediaTypeMap.$keys'
          enum:
          - apiKey
        - on: MediaTypeMap.$keys
          description: Only application/json can be used
          enum:
            - 'application/json'
    nonEmpty:
      description: Schema title and description should be non-empty
      locations:
        - on: Schema.description
        - on: Schema.title
```

**Additional context**
none

slavikbez:
In the near (or far) future will users be able to generate these rules through UI?

At a first glance configuring enforcements will help writing definitions but on the other side it gives complexity to the configuration file itself (we may need another linter for it).

RomanHotsiy:
> In the near (or far) future will users be able to generate these rules through UI?

This is a great question! Yes, we plan to work on this in hopefully near future. Do you see any issues with the suggested format that can complicate the UI in the future.

> At a first glance configuring enforcements will help writing definitions but on the other side it gives complexity to the configuration file itself (we may need another linter for it).

I agree 💯 . Do you have any suggestions? Provided we definitely need this kind of rules to be configurable.
Also, what do you think about enforcements approach vs alternative per-rule approach?
While it complicates these rules it still makes them a bit more accessible over custom JS rules.

jacobator:
I like per-key approach better (allowing "on" to be an array makes it even better)

slavikbez:
[...]

PR #506, opened 2022-01-11
docs: add assertions docs

adamaltman:
## What/Why/How?

Add docs for the assertions feature. A few notes:
- I did not try out the feature yet.
- There are "TODO"s inside of the docs. Please look at the PR and find/comment.
- I omitted the `sortOrder` for now and would like to discuss it a bit further.
- In code: rename `description` to `message` for consistency with the naming throughout OpenAPI cli.
- In code: I think we should add support for `suggest` too (like we have that support in our built-in rules as well as custom rules).

This moves the bulk of the assertions docs to a stand-alone page.

In general, each rule should have its own page, but that will take a good deal of time to set up. There is no way this content can go inside of the built-in rules page, so instead I added a link to it from there.

### TODO

- [x] Add `sortOrder` docs
- [x] Add `suggest` to docs
- [x] Add OpenAPI node types section
- [x] Remove `length` from docs
- [ ] Check examples

## Reference

https://github.com/Redocly/openapi-cli/pull/489

## Testing

TODO

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/resources/rules/assertions.md
+++ b/docs/resources/rules/assertions.md
@@ -1,0 +1,312 @@
+# `assertions`
+
+Configure assertions to enforce your API design standards without coding custom rules.
+
+Define the `assertions` key in the lint rules of the configuration file.
+
+```yaml
+lint:
+  rules:
+    assertions: ...
+```
+
+Property | Type | Description
+-- | -- | --
+assertions | [[Assertion object]](#assertion-object) | A list of assertions to enforce your custom API design standards. Add or edit your assertions in the configuration file. The `assertions` rule is composed of a list of `assertion` objects. This rule may result in more than one problem message as you may define more than one assertion. More than one assertion may be defined, and an assertion may have more than one assert.
+
+## Assertion object
+
+Property | Type | Description
+-- | -- | --
+on | `string` \| [`string`] | **REQUIRED.** The node type, or list of node types, to the lint targets. Nodes are traversed with a dot `.` between nodes, and can start from any [OpenAPI node type](#openapi-node-types). A special value `$keys` can be used to target the keys of any map (for example, `ResponseMap.$keys`).
+description | `string` | Problem message displayed if the assertion is false.
+severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
+enum | [`string`] | Asserts value is within a predefined list of values. See [enum example](#enum-example).
+pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#regex-pattern-example).
+casing | `string` | Asserts a casing style from this possible list: `camelCase`, `kebab-case`, `snake_case`, `PascalCase`, `MACRO_CASE`, `COBOL-CASE`, `flatcase`. See [casing example](#casing-example).
+mutuallyExclusive | [`string`] | Asserts list of properties (key names only) are mutually exclusive. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyExclusive example](#mutuallyexclusie-example).
+mutuallyRequired | [`string`] | Asserts list of properties (key names only) are mutually required. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyRequired example](#mutuallyrequired-example).
+defined | `boolean` | Asserts a property is defined. See [defined example](#defined-example).
+undefined | `boolean` | Asserts a property is undefined. See [undefined example](#undefined-example).
+nonEmpty | `boolean` | Asserts a property is not empty. See [nonEmpty example](#nonempty-example).
+length | `integer` | Asserts a length of a string or list (array). See [length example](#length-example).
+minLength | `integer` | Asserts a minimum length (inclusive) of a string or list (array). See [minLength example](#minlength-example).
+maxLength | `integer` | Asserts a maximum length (exclusive) of a string or list (array). See [maxLength example](#maxlength-example).
+
+
+## Examples
+
+The following example shows two assertions each multiple asserts (`defined`, `minLength`, `maxLength`, `pattern`).
+
+The `Operation`, `Tag`, and `Info` properties must:
+- be defined
+- have at least 30 characters
+- end with a "_full stop_".
+
+In addition, the `Operation` summary property must:
+- be defined
+- be between 20 and 60 characters
+- not end with a "_full stop_".
+
+The following shows how to write that configuration:
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on:
+          - Tag.description
+          - Operation.description
+          - Info.description
+        message: Description must be at least 30 characters and end with a full stop.
+        severity: error
+        defined: true
+        minLength: 30
+        pattern: /(full stop)$/
+      - on:
+          - Operation.summary
+        message: Operation summary must be at least 20 characters and not end with a full stop.
+        severity: error
+        defined: true
+        minLength: 20
+        maxLength: 60
+        pattern: /^(full stop)$/
+      
+```
+<!-- TODO: is the regex not /\\.$/? -->
+
+### `enum` example
+
+Assert that keys or values are in a list of values.
+Append `$keys` to the `on` node value to run on the node's keys.
+
+```yaml keys
+lint:
+  rules:
+    assertions:
+      - on: MediaTypeMap.$keys
+        message: Only application/json can be used
+        severity: error
+        enum:
+          - application/json
+```
+
+```yaml values
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Summary must be one of the predefined values
+        severity: error
+        enum:
+          - one
+          - two
+```
+
+### `pattern` example
+
+This example shows a regex pattern to assert that the operation summary contains "test".
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Summary should match a regex
+        severity: error
+        pattern: /test/
+```
+
+### `casing` example
+
+This example enforces `PascalCase` casing style for named Examples keys.
+
+<!-- TODO: Does casing work on keys? I would like to do it on media type examples PascalCase. Please add an example.  -->
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: NamedExamples.$keys
+        message: NamedExamples key must be in PascalCase
+        severity: error
+        casing: PascalCase
+```
+
+Casing supports these styles: 
+- camelCase
+- COBOL-CASE
+- flatcase
+- kebab-case
+- snake_case
+- PascalCase
+- MACRO_CASE
+
+
+### `mutuallyExclusive` example
+
+This example asserts the operation `description` and `externalDocs` must be mutually exclusive.
+This assert runs only on node's keys.
+Append `$keys` to the `on` node value to run on the node's keys.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.$keys
+        message: "Operation must not define both properties together: description and externalDocs"
+        severity: error
+        mutuallyExclusive:
+          - description
+          - externalDocs
+```
+
+### `mutuallyRequired` example
+
+This example asserts that a response body schema must have both `amount` and `currency` properties (and not either one by itself).
+This assertion runs only on node's keys.
+Append `$keys` to the `on` node value to run on the node's keys.
+
+```yaml Schema example
+lint:
+  rules:
+    assertions:
+      # - on: Operation.Response.$keys check this
+        message: The created_at and updated_at properties are mutually required
+        severity: error
+        mutuallyRequired:
+          - created_at
+          - updated_at
+```
+
+This example asserts that PUT requests have both 200 and 201 responses defined.
+
+```yaml Response example
+lint:
+  rules:
+    assertions:
+      - on: PathItem.put.responses.$keys
+        message: Must define 200 and 201 responses for PUT requests.
+        severity: error
+        mutuallyRequired:
+          - '200'
+          - '201'
+```
+
+### `defined` example
+
+This example asserts that `x-codeSamples` is defined.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.x-codeSamples
+        message: x-codeSamples must be defined
+        severity: error
+        defined: true
+```
+
+### `undefined` example
+
+This example asserts that `x-code-samples` is undefined.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.x-code-samples
+        message: use x-codeSamples instead of x-code-samples
+        severity: error
+        undefined: true
+```
+
+### `nonEmpty` example
+
+This example asserts that the operation summary is not empty.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Operation summary should not be empty
+        severity: error
+        nonEmpty: true
+```
+
+### `length` example
+
+The example asserts the exact length of the summary string is 20 characters.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Operation summary must have 20 characters
+        severity: error
+        length: 20
+```
+
+The example asserts the list of the status enums is 20 characters exactly.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Schema.status
+        message: The status property enum must have 20 items
+        severity: error
+        length: 20
+```
+<!-- Help check this example is correct -->
+
+### `minLength` example
+
+This example asserts that the minimum length of each operation summary is 20 characters.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Operation summary must have minimum of 20 chars length
+        severity: error
+        minLength: 20
+```
+
+### `maxLength` example
+
+This example asserts that the maximum length of each operation summary is 20 characters.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.summary
+        message: Operation summary must have maximum of 20 characters
+        severity: error
+        maxLength: 20
+```
+
+<!-- 
+TODO: discuss
+
+### `sortOrder` example
+
+To enforce sort order in array of strings or array of objects (collection). 
+If the `property` fields is present, it means that we check the sort order direction on collection. 
+Also, in case of array of strings the direction could be also set like `sortOrder: asc`.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Operation.tags
+        message: Tags should be ordered in ASC direction
+        severity: error
+        sortOrder:
+          direction: asc
+          property: name
+``` -->
--- a/docs/resources/rules/assertions.md
+++ b/docs/resources/rules/assertions.md
@@ -23,9 +23,9 @@ message | `string` | Problem message displayed if the assertion is false.
 suggest | [`string`] | List of suggestions to display if the problem occurs.
 severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
 enum | [`string`] | Asserts value is within a predefined list of values. See [enum example](#enum-example).
-pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#regex-pattern-example).
+pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#pattern-example).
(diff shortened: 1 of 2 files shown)


docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+## Assertion object
+
+Property | Type | Description
+-- | -- | --
+on | `string` \| [`string`] | **REQUIRED.** The node type, or list of node types, to the lint targets. Nodes are traversed with a dot `.` between nodes, and can start from any [OpenAPI node type](#openapi-node-types). A special value `$keys` can be used to target the keys of any map (for example, `ResponseMap.$keys`).
+message | `string` | Problem message displayed if the assertion is false.
+suggest | [`string`] | List of suggestions to display if the problem occurs.
+severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
+enum | [`string`] | Asserts value is within a predefined list of values. See [enum example](#enum-example).
+pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#regex-pattern-example).
```
ohorbachevskyi, line 26:
```suggestion
pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#pattern-example).
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+casing | `string` | Asserts a casing style from this possible list: `camelCase`, `kebab-case`, `snake_case`, `PascalCase`, `MACRO_CASE`, `COBOL-CASE`, `flatcase`. See [casing example](#casing-example).
+mutuallyExclusive | [`string`] | Asserts list of properties (key names only) are mutually exclusive. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyExclusive example](#mutuallyexclusie-example).
```
ohorbachevskyi, line 28:
```suggestion
mutuallyExclusive | [`string`] | Asserts list of properties (key names only) are mutually exclusive. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyExclusive example](#mutuallyexclusive-example).
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+        minLength: 30
+        pattern: /\.$/
+      - on:
+          - Operation.summary
+        message: Operation summary must be at least 20 characters and not end with a full stop.
+        severity: error
+        defined: true
+        minLength: 20
+        maxLength: 60
+        pattern: /^\.$/
```
ohorbachevskyi, line 80:
```suggestion
        pattern: /[^\.]$/
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+          - updated_at
+```
+
+The following example asserts that `PUT` requests have both `200` and `201` responses defined.
+
+```yaml Response example
+lint:
+  rules:
+    assertions:
+      - on: PathItem.put.responses.$keys
```
ohorbachevskyi, line 197: We do not support properties inside properties, only property inside NodeType.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+
+Sort order can be applied on lists (arrays) of strings or lists of objects (collections).
+
+The following example asserts that the status schema's enum list is in alphabetical order.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - on: Schema.status.enum
```
ohorbachevskyi, line 287: We do not support properties inside properties, only property inside NodeType.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,393 @@
...
+The following example shows ways to traverse to the equivalent destination.
+One starts from the definition root into the info description.
+The other starts from the info section.
+
+```yaml Traversal from DefinitionRoot object type
+lint:
+  rules:
+    assertions:
+      - on:
+        - DefinitionRoot.info.description
```
ohorbachevskyi, line 328:
```suggestion
        - DefinitionRoot.Info.description
```

adamaltman: I don't understand. Can you traverse objects but not regular properties?

RomanHotsiy: No, this is not about traversal at all. I'll explain you on a call.

adamaltman:
The docs are still not quite there. First, this is turning into a docs-driven development. I think we should work towards finalizing the docs, and then finalize #429 .

This proposes to eliminate the `on` property, and adds support for these:

- subject (required)
- property
- context (list of context objects)

In particular, I don't believe I explained the context correctly. I believe it can be used not only to limit evaluation, but also to expand evaluations. I need help with the docs there, and examples.

Finally, it also adds two additional asserts:
- required
- disallowed

The issue is an assert like enum is great, but leaves a gap to show that everything is required.

For example:
```
enum:
  - application/json
  - application/xml
```

An API will be compliant if it has any of those Media Types. However, what if someone wanted to make it requires that both are defined. It is cumbersome to use "defined" and set up separate asserts for each of those cases (and this was a simple example with only two cases).

And another is disallowed... this is equivalent to "undefined" but in a list format (where the user can supply multiple values).

An alternative idea to this may be to change `property` to be `properties` and be a list of properties. Thoughts?

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,487 @@
...
+
+```yaml
+- context:
+    - type: Operation
+      filterKeys: [put]
+    - type: ResponsesMap
+      filterKeys: [201, 200]
+  subject: MediaTypeMap
+  disallowed: ['application/pdf']
+```
```
ohorbachevskyi, line 124: I like the idea of `context`, especially how it expands the previous "dot notation" way to traverse Node types. But I see there couple things to improve:

1. It's not clear from the notation above what is the Node types hierarchy - I understand that it's Operation -> ResponsesMap, but would be nice to introduce it clearer, like `next` key (or more suitable word) 
    ```js
    - context:
        - type: Operation
          filterKeys: [put]
          next:
          - type: ResponsesMap
            filterKeys: [201, 200]
    ```
    It will also expand the `context` power - now `context` will spread in many directions, like in the next example 
    ```js
    - context:
        - type: Operation
          filterKeys: [put]
          next:
          - type: ResponsesMap
            filterKeys: [201, 200]
        - type: Type1
          filterKeys: [key1, key2]
          next:
          - type: Type2
            filterKeys: [key3]
            next:
            - type: Type3
              filterKeys: [key4]     
    ```
    It tells us that the first context "direction" is Operation -> ResponsesMap, and the second Type1 -> Type2 -> Type3

2. I would rather use `onKeys` not `filterKeys`, as it sounds like "filter out". For me it's clearer :) 

3. It would be nice to have ability to exclude keys, with `offKeys` (or more suitable word) :
    ``` js
    - context:
      - type: Operation
        onKeys: [put]
        next:
        - type: ResponsesMap
          offKeys: [201, 200]
    ```

RomanHotsiy: @ohorbachevskyi thanks for the great feedback!

1. I believe you did not get the context correctly. The context is not an array of alternative contexts, this is actually a hierarchy. I see why you got confused though. But I think your suggestion is over-complicated and won't be used in any real use-cases. I would keep it as is.

2. I believe we were going to rename it to `filterParentKeys`. But I like `onParentKeys` much more. @adamaltman what do you think?

3. I'm not sure if `off` is the correct usage here but I agree it would be nice to have some keyword like that.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,487 @@
...
+## Sort Order object
+
+Property | Type | Description
+-- | -- | --
+direction | `string` | **REQUIRED.** Asserts values are sorted in ascending or descending order using one of these possible values: `asc` or `desc`.
+property | `string` | **REQUIRED.** The name of the property in the collection of objects that will hold values to assess the sort order.
+
+## Examples
+
+The following example shows two assertions each multiple asserts (`defined`, `minLength`, `maxLength`, `pattern`).
```
ohorbachevskyi, line 59: Frankly I see four but not two assertions in example below, but I understand that technically there are only two :) I think you divided it because of message as it should be exact about the subject. We could combine it within `subject` key, like

```yaml
lint:
  rules:
    assertions:
      - subject: 
        - name: Tag
          message: Tag description must be at least 30 characters and end with a full stop.
        - name: Operation
          message: Operation description must be at least 30 characters and end with a full stop.
        - name: Info
          message: Info description must be at least 30 characters and end with a full stop.
        property: description
        severity: error
        defined: true
        minLength: 30
        pattern: /\.$/
```
If the message is about the property, we can put it under `property` key, for example
```yaml
lint:
  rules:
    assertions:
      - subject: Tag
        property: 
        - name: description
          message: Tag description must be at least 30 characters and end with a full stop.
        - name: name
          message: Tag name must be at least 30 characters and end with a full stop.
        severity: error
        defined: true
        minLength: 10
        pattern: /\.$/
```
If we want to use one message for all the asserts, so let's do it as previous at the assertion level.
So now we will have three ways to use `message` - under `subject`, under `property` and at the assertion level.

adamaltman: Actually, you are right. I adjusted it to four assertions but didn't show it here. We are trying to limit nesting as much as possible. While I am okay with the idea of make subject and property lists, I think it may result in overcomplexity and documenting it would be more difficult than making four assertions.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,487 @@
...
+- not end with a _full stop_.
+
+The following shows how to write that configuration:
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - subject: Tag
+        property: description
```
ohorbachevskyi, line 78: Agree that the key name should be `properties` and it should be an array, so we can evaluate on more that one property from `subject` Node type. But how about nested properties? Is it possible case when we have prop1.prop2 in some Node type?

docs/resources/rules/assertions.md
adamaltman: How about `matchKeys` and `excludeKeys`?

docs/resources/rules/assertions.md
RomanHotsiy: As Adam mentioned above better to keep it flat to simplify docs/usage.

I'm not sure about nested properties. We can technically support it but I'm afraid people may think it will traverse through $ref's... I would keep it simple first and then make it more complex if needed.

adamaltman: No nesting.  I'm okay to make it properties like: `string` | [`string`] type. I think it could be helpful.

docs/resources/rules/assertions.md
RomanHotsiy: @adamaltman I think it's important to add `parent` word into naming... so `matchParentKeys`, `excludeParentKeys`?

docs/resources/rules/assertions.md
adamaltman:
```suggestion
The following example shows four assertions each multiple asserts (`defined`, `minLength`, `maxLength`, `pattern`).
```

adamaltman: I think this is okay for properties:  ` string | [string] ` but not something like ` string | [NameMessage Object]`. Flatter with more assertions is easier to use.

docs/resources/rules/assertions.md
ohorbachevskyi: @adamaltman Yeah, `matchKeys` and `excludeKeys` much better! @RomanHotsiy why parent is so important here?

adamaltman: I think it depends on how well people read and understand the docs. The issue is I only understand it personally as the parent. Let's take `put` for example... I think of it as belonging to the `PathItem` but it won't work as a `matchKey` of the `PathItem`. Instead, it is viewed as the parent of the `Operation`.

adamaltman:
I'm going to adjust the docs based on the discussion above. I think there is one issue remaining with the `SortOrder` object, now that the subject can have properties.

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,487 @@
...
+    assertions:
+      - subject: Tag
+        property: name
+        message: Tags should be ordered by name in alphabetical order
+        severity: error
+        sortOrder:
+          direction: asc
+          property: name
+```
+<!-- TODO: Does the sortOrder need a property "property" now? -->
```
ohorbachevskyi: Yep, you're right, now a property "property" is redundant here; 
and we can use the simpler `sortOrder` notation like `sortOrder: asc`, because we can set the order's property right after the `subject`.

docs/resources/rules/assertions.md
ohorbachevskyi: I got it, thnx for clarifying.

docs/resources/rules/assertions.md
adamaltman: I noticed it was already defined like `string | [string]`. I went to rename it to properties, and then as I was updating the examples I become uncertain if `properties` is a better name than `property` and decided to stick with `property`. I have a small concern that `properties` is overloaded and may be confusing in the documentation between OpenAPI/JSON Schema and that should dominate compared to our "assertions".

docs/resources/rules/assertions.md
adamaltman: I've adjusted it (I think)

adamaltman:
Thanks @ohorbachevskyi your feedback was very helpful on the docs. I think we should proceed with the next round of development updates to the assertions so we can :shipit: .

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+- Responses with all codes including others than 200 or 201
+- Responses for all HTTP methods including delete, get, post, and more.
+To restrict the evaluation, use the context feature to limit where context is used.
+
+```yaml
+- context:
+    - type: Operation
+      matchParentKeys: [put]
+    - type: ResponsesMap
+      matchParentKeys: [201, 200]
```
anastasiia-developer, line 169:
```suggestion
      matchParentKeys: ['201', '200']
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+### `mutuallyRequired` example
+
+The following example asserts that a response body schema must have both `amount` and `currency` properties (and not either one by itself).
+This assertion runs only on node's keys.
+
+```yaml Schema example
+lint:
+  rules:
+    assertions:
+      - subject: Response
```
anastasiia-developer, line 272:
```suggestion
      - subject: SchemaProperties
         context:
           - type: Response
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+          - updated_at
+```
+
+The following example asserts that when `PUT` requests have either `200` or `201` defined that both `200` and `201` responses defined.
+
+```yaml Response example
+lint:
+  rules:
+    assertions:
+      - subject: ResponseMap
```
anastasiia-developer, line 286:
```suggestion
      - subject: ResponsesMap
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+### `disallowed` example
+
+The following example asserts that `x-code-samples` and `x-internal` are not defined.
+
+```yaml
+lint:
+  rules:
+    assertions:
+      - subject: Operation
+        message: x-codeSamples and x-internal must not be defined
```
anastasiia-developer, line 328:
```suggestion
        message: x-code-samples and x-internal must not be defined
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+message | `string` | Problem message displayed if the assertion is false.
+suggest | [`string`] | List of suggestions to display if the problem occurs.
+severity | `string` | The severity level of the problem if the assertion is false. It must be one of these values: `error`, `warn`, `off`. Default value is `error`.
+enum | [`string`] | Asserts value is within a predefined list of values. See [enum example](#enum-example).
+pattern | `string` | Asserts value matches a regex pattern. See [regex pattern example](#pattern-example).
+casing | `string` | Asserts a casing style from this possible list: `camelCase`, `kebab-case`, `snake_case`, `PascalCase`, `MACRO_CASE`, `COBOL-CASE`, `flatcase`. See [casing example](#casing-example).
+mutuallyExclusive | [`string`] | Asserts list of properties (key names only) are mutually exclusive. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyExclusive example](#mutuallyexclusive-example).
```
anastasiia-developer, line 30:
```suggestion
mutuallyExclusive | [`string`] | Asserts list of properties (key names only) are mutually exclusive. See [mutuallyExclusive example](#mutuallyexclusive-example).
```

docs/resources/rules/assertions.md
```diff
@@ -0,0 +1,532 @@
...
+mutuallyRequired | [`string`] | Asserts list of properties (key names only) are mutually required. Use `$keys` in the `on` when this is enabled (for example, `Operation.$keys`). See [mutuallyRequired example](#mutuallyrequired-example).
```
anastasiia-developer, line 31:
```suggestion
mutuallyRequired | [`string`] | Asserts list of properties (key names only) are mutually required. See [mutuallyRequired example](#mutuallyrequired-example).
```

docs/resources/rules/assertions.md
adamaltman: We need to look at other options. I'm pinging @RomanHotsiy (and discussed it with him in a meeting).

RomanHotsiy: I don't like the extra `property` in the `sortOrder`. This will be confusing. We should come up with some better idea.

I think we are mixing a few things here. Sorting by a property value and sorting the property value itself.

I don't think we should rely on the type of the property value to determine the behaviour because some properties can have mixed values, e.g. `type` can be string or array.

One possible solution is to bind the behaviour to our type tree. We clearly know whether some type is a collection or a map (usually these have `_List` or `_Map` suffix). So if the type is a collection we should use `property` as a sort key vs sort the `property` itself:

**Sort collection by property `name`:**
```yaml
lint:
  rules:
    assertions:
      - subject: Tag_List
        property: name
        sortOrder: asc
```

**Sort value of a property:**
```yaml
lint:
  rules:
    assertions:
      - subject: Schema
        property: enum
        sortOrder: asc
```

The main concern with this approach is we can't actually sort something that is untyped (e.g. vendor extensions). Imagine x-metadata vendor extension which may look like:

```yaml
x-metadata:
  - name: Description
    value: Lorem ipsum
  - name: Location
    value: Lviv
```

```yaml
lint:
  rules:
    assertions:
      - subject: Schema
        property: x-metadata # how to sort by `name` ?
        sortOrder: asc # what sort order even means here ?
```

adamaltman: Here is another idea. What about two different sort assertions:

- `sortCollection`
- `sortValues`

### Sorts the **collection** by the property `name`
```yaml
lint:
  rules:
    assertions:
      - subject: Tag
        property: name
        sortCollection: asc
```

### Sorts the **values** of the property `enum`
```yaml
lint:
  rules:
    assertions:
      - subject: Schema
        property: enum
        sortValues: asc
```

For something that may have mixed values, like `product` in this example:

```yaml
x-metadata:
  - name: Sample name as a string
     product: elf
  - name: What is this?
     product:
        - bar
        - foo
```

Someone can specify if they want to sort by the **values** or by the collection by choosing the appropriate assertion `sortValues` or `sortCollection`. 

The open question is what is the logic of `sortCollection` when the value is not a scalar. In the example above, will it use the array's first value for the comparison logic? Or what?

adamaltman: > One possible solution is to bind the behaviour to our type tree. We clearly know whether some type is a collection or a map (usually these have _List or _Map suffix). So if the type is a collection we should use property as a sort key vs sort the property itself:

One thing I don't like about this idea is that we already have types for some maps like `ResponsesMap`, so this feels like we missed types like `TagsMap`. I think we are missing content (tech docs) on the type tree. And this idea may also make things like typeahead less intuitive too.

RomanHotsiy: > What about two different sort assertions:

I don't like this idea because `sortValues` and `sortCollection` feel more like a decorator, not as an assertion.

> The open question is what is the logic of sortCollection when the value is not a scalar. 

Also `sortCollection` has other open questions. What if the subject is not part of any collection.

> so this feels like we missed types like TagsMap

What is `TagsMap`? Tags are not a map, it's an array (list). There is `Tag_List`.

ohorbachevskyi (approved):

PR #507, opened 2022-01-11
fix: remove-unused-components strips remotely referenced components

### LINKED ISSUE #504: --remove-unused-components strips remotely referenced components
I am currently testing --remove-unused-components with a spec that has $refs pointing to an external specification:
```
          properties:
            items:
              type: array
              items:
                $ref: 'https://my-other-api.my-service.io/openapi.yaml#/components/schemas/Order'
```

If I run openapi bundle, I can produce a dereferenced version:
```
openapi bundle -o openapi.combined.yaml --ext yaml openapi.yaml
```
This gives me a file with an Order schema included. However, if I add --remove-unused-components:
```
openapi bundle --remove-unused-components -o openapi.combined.yaml --ext yaml openapi.yaml
```
Then the Order schema gets immediately removed, because it's not detected as used.

andriyl:
## What/Why/How?
Closes #504  
## Check yourself
- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -1,5 +1,5 @@
 import isEqual = require('lodash.isequal');
-import { BaseResolver, resolveDocument, Document } from './resolve';
+import { BaseResolver, resolveDocument, Document, ResolvedRefMap } from './resolve';
 import { Oas3Rule, normalizeVisitors, Oas3Visitor, Oas2Visitor } from './visitors';
 import { Oas3Types } from './types/oas3';
 import { Oas2Types } from './types/oas2';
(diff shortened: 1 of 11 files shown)


packages/core/src/bundle.ts
```diff
@@ -267,6 +269,17 @@ function makeBundleVisitor(
     };
   }
 
+  function replaceInlineRef(node: OasRef, resolved: ResolveResult<any>, ctx: UserContext) {
```
RomanHotsiy, line 272: Rename it to `resolveBundledComponent` plz

packages/core/src/bundle.ts
```diff
@@ -267,6 +269,17 @@ function makeBundleVisitor(
...
+    const newRefId = ctx.location.source.absoluteRef + '::' + node.$ref;
```
RomanHotsiy, line 273: Make this a helper function in `resolve.ts` module:

`function makeRefId(absoluteRef, pointer)`

RomanHotsiy (changes requested):
Left a few suggestion.

packages/core/src/bundle.ts
andriyl: done

packages/core/src/bundle.ts
andriyl: done

RomanHotsiy (approved):

PR #508, opened 2022-01-11
chore: refactor decorators file structure

andriyl:
## Check yourself
- [x] Code is linted

## Security


adamaltman (commented):
I love this change 🎉

packages/core/src/config/config.ts
```diff
@@ -623,3 +624,12 @@ function assignExisting<T>(target: Record<string, T>, obj: Record<string, T>) {
     }
   }
 }
+
+export function makeConfig(rules: Record<string, RuleConfig>, decorators?:  Record<string, DecoratorConfig>) {
```
RomanHotsiy, line 628: Keep this function in __tests__ helpers. We don't need this function anywhere in non-test code so why have it here.

RomanHotsiy (changes requested):
Looks great, just move the test helper function `makeConfig` back to __tests__ folder

packages/core/src/config/config.ts
andriyl: @RomanHotsiy moved to `core/__test__/utils`

RomanHotsiy (approved):

Issue #509, opened 2022-01-12
oauth2-redirect.html is absent when served by preview-docs command


nalesnichok:
**Describe the bug**
The client is getting a 404 error after running `openapi preview-docs` command and utilizing the `Try it` console.

**To Reproduce**
1. Open the definition file locally(ping me for details) and run `openapi preview-docs -p 8081 openapi.yaml`
2. Go to any endpoint and click the `Try it` button.
3. Click `Authorize`. You don't need to change a predefined `Client ID` 
4. After that you'll need to authorize with credentials(ping me as well)
5. See the 404 errors message after entering the correct credentials.

@Oprysk tried to reproduce the issue directly in Reference docs but it's working fine there. Also, the issue is not reproducible in hosted docs(the same definition). Customer is using an oauth2 + implicit flow, and it seems like `oauth2-redirect.html` file is missing after performing the above steps, which brings the 404 error.

**`openapi-cli` Version(s)**
What version of `openapi-cli` are you using? 
1.0.0-beta.79

**`Node.js` Version(s)**
What version of `node.js` are you using?

v16.13.1

**Additional context**
Reported here:
https://app.crisp.chat/website/a4d2dd27-7665-4213-b5b0-864b6fe69e31/inbox/session_5009d514-bdbd-43d9-a29c-3f5807c2cd62/

nalesnichok:
Closing this issue as it is already resolved

PR #512, opened 2022-01-13
fix: oauth2-redirect.html is absent when served by preview-docs command #509

Oprysk:
## What/Why/How?
fix: oauth2-redirect.html is absent when served by preview-docs command #509
## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
+++ b/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
@@ -13,7 +13,6 @@ function getPageHTML(
   redocOptions: object = {},
   useRedocPro: boolean,
   wsPort: number,
-  oauth = false,
 ) {
   let templateSrc = readFileSync(htmlTemplate, 'utf-8');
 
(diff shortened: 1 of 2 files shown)


andriyl (approved):

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -36,6 +37,11 @@ function getPageHTML(
       ? 'https://cdn.jsdelivr.net/npm/@redocly/reference-docs@latest/dist/redocly-reference-docs.min.js'
       : 'https://cdn.jsdelivr.net/npm/redoc@latest/bundles/redoc.standalone.js'
   }"></script>
+  ${
```
RomanHotsiy, line 40: better to use a separate page template: https://cdn.jsdelivr.net/npm/@redocly/reference-docs@latest/dist/oauth2-redirect.html

Oprysk: I know, but in that case we need to host it somewhere. Is it ok ?

Oprysk: done

RomanHotsiy (approved):

PR #515, opened 2022-01-14
feat: enable decorators logic

### LINKED ISSUE #495: enable-rule, enable-decorator args

There is a way to skip rules and decorators but no way to enable them by args.

I think this causes a user experience problem. This means to enable any decorator or rule requires adding it explicitly to the `.redocly.yaml` file. This is problematic when someone doesn't even have a `.redocly.yaml` file -- they need to:

1. Create a `.redocly.yaml` file (which has its own challenges because commonly used operating systems hide dot files by default)
2. Learn the structure of the file and fill it out to enable a rule or decorator. Save the file.
3. Commit to source control, upload, or CLI push again.

I would like to be able to specify enabling some rule or decorator at runtime. For example, removing internal APIs.... 

`--enable-decorator=remove-x-internal,remove-unused-components`

Auto-generating `.redocly.yaml` file on the fly from a wizard.

**Additional context**

I am envisioning what the usage may be like from within the API registry in workflows, where someone may be able to select from specific rules and decorators.

andriyl:
## What/Why/How?
Closes #495

Next steps:
- add tests
- decide about subparameter format
- decide about enabling rules

## Check yourself
- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/commands/bundle.ts
+++ b/packages/cli/src/commands/bundle.ts
@@ -37,13 +37,14 @@ export async function handleBundle(
     format: OutputFormat;
     metafile?: string;
     extends?: string[];
-    'remove-unused-components'?: boolean
+    'remove-unused-components'?: boolean;
+    'enable-decorators'?: string[];
   },
   version: string,
 ) {
-  const config = await loadConfig(argv.config, argv.extends);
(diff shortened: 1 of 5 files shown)


tatomyr:
Closing this as the main issue is resolved.

PR #530, opened 2022-01-25
fix: no-invalid-media-type-examples doesn't work with example that uses

### LINKED ISSUE #518: no-invalid-media-type-examples doesn't work with example that uses $ref
**Describe the bug**

It reports:

Example value must conform to the schema: must NOT have additional properties `$ref`.

```
16 |         $ref: '../components/schemas/Schema.yaml'
17 |       example:
18 |         $ref: '../components/examples/example.yaml'
             ^^^^
```

**To Reproduce**

Use $ref to an external file in the example.

**Expected behavior**
Should work.

andriyl:
## What/Why/How?
Closes #518 

## Check yourself
- [x] Code is linted

## Security


packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -13,31 +13,27 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
...
+          for (const exampleName of Object.keys(mediaType.examples)) {
+            resolveAndValidateExample(mediaType.examples[exampleName], location.child(['examples', exampleName, 'value']), true);
+          }
+        }
+
+        function resolveAndValidateExample(example: Oas3Example | any, location: Location, isMultiple?: boolean) {
+          if (isRef(example)) {
+            const resolved = resolve<Oas3Example>(example);
+            if (!resolved.location) return;
+            location = resolved.location.child('value');
```
RomanHotsiy, line 27:
```suggestion
            location = isMultiple ? resolved.location.child('value') : resolved.location;
```

__tests__/lint/no-invalid-media-type-examples/openapi.yaml
```diff
@@ -0,0 +1,28 @@
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                property:
+                  type: string
+            example:
+              $ref: '#/components/examples/State'
```
RomanHotsiy, line 24: Add one more tests that tests bad example to verify it reports location correctly. Thanks!

RomanHotsiy (changes requested):

RomanHotsiy (approved):

PR #535, opened 2022-01-27
feat: response-contains-property built-in rule

### LINKED ISSUE #389: `response-contains-property` built-in rule

Yes. Often people want some consistent props in responses. This is about enforcing that.

https://redoc.ly/docs/cli/guides/response-contains-property/

Make this into a built-in rule. Someone would need to enable it manually and configure it.

Sometimes people want different response properties for collections compared to an individual resource. For example, 

```
GET /apis <-- returns all apis
GET /apis/123 <-- returns api 123.
POST /apis <-- returns single resource
PUT /apis/123 <-- returns single resource
```

I think it is possible to infer if an operation is a collection or individual resource, but it is challenging.

Some may use an array in the response, and that can be an indicator that it is a collection. However, others may use an object in the response where the actual collection is buried inside a property of that collection.

Stripe does that, see a collection of customers:

```
{
  "object": "list",
  "url": "/v1/customers",
  "has_more": false,
  "data": [
    {
      "id": "cus_8epDebVEl8Bs2V",
      "object": "customer",
      "address": null,
      "balance": 0,
      "created": 1466202923,
      "currency": "usd",
      "default_source": "card_18NVYR2eZvKYlo2CQ2ieV9S5",
      "delinquent": true,
      "description": "Mia Wilson",
      "discount": null,
      "email": "mia.wilson.99@example.com",
      "invoice_prefix": "D86E170",
      "invoice_settings": {
        "custom_fields": null,
        "default_payment_method": null,
        "footer": null
      },
      "livemode": false,
      "metadata": {
        "order_id": "6735"
      },
      "name": null,
      "next_invoice_sequence": 249348,
      "phone": null,
      "preferred_locales": [],
      "shipping": null,
      "tax_exempt": "none"
    },
    {...},
    {...}
  ]
}
```

Even in the case that a collection contains an object, the object is not likely to contain the properties that are desired to be enforced... samples of common props include: `id`, `created_at`, `updated_at`. 

This rule should be configurable from the `.redocly.yaml` file. 

None.

**Additional context**

Similarly, people want to enforce `response-contains-headers`. A real world example would be to return pagination headers for collections. This also needs to be able to be targeted to collections or individual resources.

andriyl:
## What/Why/How?
Closes #389 

## Check yourself
- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/oas2/response-contains-property.ts
+++ b/packages/core/src/rules/oas2/response-contains-property.ts
@@ -1,0 +1,23 @@
+import { Oas2Rule } from '../../visitors';
+
+export const ResponseContainsProperty: Oas2Rule = (options) => {
+  const mustExist = options.mustExist || [];
+  return {
+    Response: {
+      skip: (_response, key) => {
+        return !['200', '201', '202'].includes(key.toString());
+      },
+      Schema (schema, { report, location }) {
+        if (schema.type !== 'object') return;
+        for (let element of mustExist) {
+          if (!schema.properties?.[element]) {
+            report({
+              message: `Response object must have a top-level "${element}" property.`,
+              location: location.child('properties'),
+            });
+          }
+        }
+      }
+    }
+  }
+};
--- a/packages/core/src/rules/oas2/__tests__/response-contains-property.test.ts
+++ b/packages/core/src/rules/oas2/__tests__/response-contains-property.test.ts
@@ -1,0 +1,92 @@
+import { outdent } from 'outdent';
+import { lintDocument } from '../../../lint';
+import { parseYamlToDocument, makeConfig } from '../../../../__tests__/utils';
+import { BaseResolver } from '../../../resolve';
+
+describe('Oas2 response-contains-property', () => {
+  it('oas2-response-contains-property: should report on response object when not contains properties', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+      swagger: '2.0'
+      schemes:
+        - https
+      basePath: /v2
+      paths:
+        '/accounts/{accountId}':
+          get:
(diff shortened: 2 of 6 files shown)


packages/core/src/rules/oas2/response-contains-property.ts
```diff
@@ -0,0 +1,23 @@
+import { Oas2Rule } from '../../visitors';
+
+export const ResponseContainsProperty: Oas2Rule = (options) => {
+  const mustExist = options.mustExist || [];
+  return {
+    Response: {
+      skip: (_response, key) => {
+        return !['200', '201', '202'].includes(key.toString());
```
tatomyr, line 8: What do these numbers mean?

tatomyr (approved):

packages/core/src/rules/oas2/response-contains-property.ts
adamaltman: Well, they may have response content but may be errors, and that's why they aren't applicable either.

204 - no content
4XX - bad requests
5XX - server errors

It's common that things like 4XX have different properties.

RomanHotsiy: Same as https://github.com/Redocly/openapi-cli/pull/537#discussion_r795133078

But here also different mime types are in play, so I'm not sure what to suggest yet.

@adamaltman do you have some ideas?

adamaltman: Yes, I think the 2XX, 4XX configuration options is a good idea.

tatomyr:
Closing this since the changes are incorporated into https://github.com/Redocly/openapi-cli/pull/537

PR #536, opened 2022-01-28
Fix joining paths with metadata

TejasQ:
## What/Why/How?
The `join` command assumes all paths have _only_ operations that are expected to be objects, but paths can also have `string` fields like `summary` and `description`. This PR only treats objects as objects.

## Reference
https://swagger.io/docs/specification/paths-and-operations/#:~:text=In%20OpenAPI%20terms%2C%20paths%20are,as%20GET%2C%20POST%20or%20DELETE.

image

## Testing
There are no tests for this file, so no tests were added.

## Check yourself

- [x] Code is linted
- [] All new/updated code is covered with tests

## Security

### DIFF
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -248,28 +248,30 @@ packageVersion: string
       for (const path of Object.keys(paths)) {
         if (!joinedDef.paths.hasOwnProperty(path)) { joinedDef.paths[path] = {}; }
         if (!potentialConflicts.paths.hasOwnProperty(path)) { potentialConflicts.paths[path] = {}; }
-        for (const operation of Object.keys(paths[path])) {
+        for (const operationOrMetadata of Object.keys(paths[path])) {
           // @ts-ignore
-          const pathOperation = paths[path][operation];
-          joinedDef.paths[path][operation] = pathOperation;
-          potentialConflicts.paths[path][operation] = [...(potentialConflicts.paths[path][operation] || []), entrypoint];
-          const { operationId } = pathOperation;
+          const pathOperationOrMetadata = paths[path][operationOrMetadata];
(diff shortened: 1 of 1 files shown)


adamaltman:
Thanks @TejasQ  for raising the issue and providing a solution.

The path item object can also have a `parameters` object at the top level.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object

We should solve that as well.

alesbrelih:
> Thanks @TejasQ for raising the issue and providing a solution.
> 
> The path item object can also have a `parameters` object at the top level.
> 
> https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object
> 
> We should solve that as well.

Please do 🙏 Thanks for your effort 🙇

TejasQ:
So who does “we” in “we should solve that as well” exactly refer to? 😬

adamaltman:
image

(We're deep into some other projects, and I hope we can get to this within a few weeks. I'll see if anyone else on the team can hop in.)

urso:
Hi,

> The path item object can also have a parameters object at the top level.
>
> [OAI/OpenAPI-Specification@main/versions/3.0.3.md#path-item-object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object)
>
> We should solve that as well.

Is this really a blocker for this PR? The change looks rather minimal and it is solving an actual bug. Is there a chance to get it in and we create an issue for the path item object?

IgorKarpiuk:
Hi @TejasQ, thanks for the PR.

Based on this PR we created [this one](https://github.com/Redocly/redocly-cli/pull/778) to handle all possible fields in a PathItem object

PR #537, opened 2022-01-28
feat: `response-contains-header` & `response-contains-property` built-in rules

### LINKED ISSUE #390: `response-contains-header` built-in rule
See #389 except this is for response headers.

andriyl:
## What/Why/How?
Closes #390 & #389 respectively

## Check yourself
- [x] Code is linted

## Security

### DIFF
--- a/docs/resources/built-in-rules.md
+++ b/docs/resources/built-in-rules.md
@@ -335,6 +335,25 @@ Verifies that each tag has a description.
 
 Verifies that tags (names) are declared in alphabetical order.
 
+### response-contains-header
+
+Enforces using specified response headers.
+
+```yaml
+lint:
+  rules:
+    response-contains-header:
+      severity: error
+      names:
+        2xx: 
+          - x-request-id
+        400:
+          - Content-Length
+```
+
+### response-contains-property
+
+The same as `response-contains-header` except for a response properties.
 
 ## Recommended config
 
--- a/docs/resources/built-in-rules.md
+++ b/docs/resources/built-in-rules.md
@@ -357,12 +357,13 @@ Enforces definition of specific response properties based on HTTP status code or
 Priority is given to more precise status codes over the status code range.
 ```yaml
 response-contains-header:
-      severity: error
-      names:
-        2XX: 
-          - created_at
-        400:
-          - title
+  severity: error
+  names:
+    2XX: 
+      - created_at
+    400:
+      - title
+```
 
 ## Recommended config
 
--- a/docs/resources/built-in-rules.md
+++ b/docs/resources/built-in-rules.md
@@ -359,7 +359,7 @@ Priority is given to more precise status codes over the status code range.
 response-contains-header:
   severity: error
   names:
-    2XX: 
+    2XX:
       - created_at
     400:
       - title
--- a/docs/resources/built-in-rules.md
+++ b/docs/resources/built-in-rules.md
@@ -345,7 +345,7 @@ lint:
     response-contains-header:
       severity: error
       names:
-        2xx: 
+        2XX: 
           - x-request-id
         400:
           - Content-Length
@@ -353,7 +353,16 @@ lint:
 
 ### response-contains-property
 
-The same as `response-contains-header` except for a response properties.
+Enforces definition of specific response properties based on HTTP status code or HTTP status code range.
+Priority is given to more precise status codes over the status code range.
+```yaml
+response-contains-header:
+      severity: error
+      names:
+        2XX: 
+          - created_at
+        400:
+          - title
 
 ## Recommended config
 
--- a/packages/core/src/rules/common/response-contains-header.ts
+++ b/packages/core/src/rules/common/response-contains-header.ts
@@ -1,0 +1,25 @@
+import { Oas2Rule, Oas3Rule } from '../../visitors';
+import { UserContext } from '../../walk';
+import { Oas3Response } from '../../typings/openapi';
+import { Oas2Response } from '../../typings/swagger';
+
+export const ResponseContainsHeader: Oas3Rule | Oas2Rule = (options) => {
+  const mustExist = options.mustExist || [];
+  return {
+    Response: {
+      skip: (_response: Oas2Response | Oas3Response, key: any) => {
+        return !['200', '201', '202'].includes(key.toString());
+      },
+      enter: (response: Oas2Response | Oas3Response, { report, location }: UserContext) => {
+        for (let element of mustExist) {
+          if (!response.headers?.[element]) {
+            report({
+              message: `Response object must have a top-level "${element}" header.`,
+              location: location.child('headers'),
+            });
+          }
+        }
+      }
+    }
+  }
+};
--- a/packages/core/src/rules/common/response-contains-header.ts
+++ b/packages/core/src/rules/common/response-contains-header.ts
@@ -2,24 +2,25 @@ import { Oas2Rule, Oas3Rule } from '../../visitors';
 import { UserContext } from '../../walk';
 import { Oas3Response } from '../../typings/openapi';
 import { Oas2Response } from '../../typings/swagger';
+import { generalizeResponseStatusCode } from '../../utils';
 
 export const ResponseContainsHeader: Oas3Rule | Oas2Rule = (options) => {
-  const mustExist = options.mustExist || [];
+  const names: Record<string, string[]> = options.names || {};
   return {
-    Response: {
-      skip: (_response: Oas2Response | Oas3Response, key: any) => {
-        return !['200', '201', '202'].includes(key.toString());
-      },
-      enter: (response: Oas2Response | Oas3Response, { report, location }: UserContext) => {
-        for (let element of mustExist) {
-          if (!response.headers?.[element]) {
-            report({
-              message: `Response object must have a top-level "${element}" header.`,
-              location: location.child('headers'),
-            });
+    Operation: {
+      Response: {
+        enter: (response: Oas2Response | Oas3Response, { report, location, key }: UserContext) => {
+          const expectedHeaders = names[key] || names[generalizeResponseStatusCode(key)] || [];
+          for (const expectedHeader of expectedHeaders) {
+            if (!response.headers?.[expectedHeader]) {
+              report({
+                message: `Response object must have a "${expectedHeader}" header.`,
+                location: location.child('headers'),
+              });
+            }
           }
-        }
-      }
-    }
-  }
+        },
+      },
+    },
+  };
 };
--- a/packages/core/src/rules/common/response-contains-header.ts
+++ b/packages/core/src/rules/common/response-contains-header.ts
@@ -2,7 +2,7 @@ import { Oas2Rule, Oas3Rule } from '../../visitors';
 import { UserContext } from '../../walk';
 import { Oas3Response } from '../../typings/openapi';
 import { Oas2Response } from '../../typings/swagger';
-import { generalizeResponseStatusCode } from '../../utils';
+import { getMatchingStatusCodeRange } from '../../utils';
 
 export const ResponseContainsHeader: Oas3Rule | Oas2Rule = (options) => {
   const names: Record<string, string[]> = options.names || {};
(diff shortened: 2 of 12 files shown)


packages/core/src/rules/common/response-contains-header.ts
```diff
@@ -0,0 +1,25 @@
...
+  return {
+    Response: {
+      skip: (_response: Oas2Response | Oas3Response, key: any) => {
+        return !['200', '201', '202'].includes(key.toString());
+      },
+      enter: (response: Oas2Response | Oas3Response, { report, location }: UserContext) => {
+        for (let element of mustExist) {
+          if (!response.headers?.[element]) {
+            report({
+              message: `Response object must have a top-level "${element}" header.`,
```
adamaltman, line 17:
```suggestion
              message: `Response object must have a "${element}" header.`,
```

All headers are top-level, right? I think the top-level thing was carried over from body schema which can be nested.

RomanHotsiy, line 17:
```suggestion
        for (let headerName of mustExist) {
          if (!response.headers?.[headerName]) {
            report({
              message: `Response object must have a top-level "${headerName}" header.`,
```

packages/core/src/rules/common/response-contains-header.ts
```diff
@@ -0,0 +1,25 @@
+import { Oas2Rule, Oas3Rule } from '../../visitors';
+import { UserContext } from '../../walk';
+import { Oas3Response } from '../../typings/openapi';
+import { Oas2Response } from '../../typings/swagger';
+
+export const ResponseContainsHeader: Oas3Rule | Oas2Rule = (options) => {
+  const mustExist = options.mustExist || [];
```
RomanHotsiy, line 7: I don't quite like `mustExists` option name.

Does anybody have some better ideas? How about just `names`:

```
response-contains-header:
  severity: error
  names: [Content-Type]
```

packages/core/src/rules/common/response-contains-header.ts
```diff
@@ -0,0 +1,25 @@
...
+  return {
+    Response: {
+      skip: (_response: Oas2Response | Oas3Response, key: any) => {
+        return !['200', '201', '202'].includes(key.toString());
```
RomanHotsiy, line 11: So this rule checks only 200 201 and 202? What if I want to check other responses headers?

Maybe we should change option to look like this:

```yaml
response-contains-header:
  severity: error
  names:
    2xx: [ 'x-request-id' ]
    400: [ 'Content-Type' ]
```

RomanHotsiy (changes requested):
Left a bunch of suggestions.

@adamaltman could you review them too? Thx.

packages/core/src/rules/common/response-contains-header.ts
adamaltman: Remove "top-level"

packages/core/src/rules/common/response-contains-header.ts
adamaltman: Sure. `names` seems correct:

> An HTTP header consists of its case-insensitive name followed by a colon (:), then by its value. [Whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) before the value is ignored.

packages/core/src/rules/common/response-contains-header.ts
adamaltman: Yes, this would be great.

tatomyr:
I updated the MR. @adamaltman, @RomanHotsiy please take a look. If the changes work for you, I'll proceed to the `response-contains-property` rule (I'm going to incorporate it in this MR).

packages/core/src/utils.ts
```diff
@@ -171,3 +171,6 @@ export function isPathParameter(pathSegment: string) {
 export function isNotEmptyObject(obj: any) {
   return !!obj && Object.keys(obj).length > 0;
 }
+
+export const generalizeResponseStatusCode = (code: number | string): string =>
+  `${code}`.replace(/^(\d)\d\d$/, (_, firstDigit) => `${firstDigit}xx`);
```
adamaltman, line 176: Can you explain this in plain language to me?

tatomyr: It translates status codes to general template, e.g. `200` -> `2xx` (see examples in [tests](https://github.com/Redocly/openapi-cli/pull/537/files#diff-ef67a52004653b8e678abb028729aa47bac58239986d6f7afc7317850fe45427R75-R84)). Not sure I named the function correctly though.

adamaltman: The issue is there are cases where different headers are common for different response codes -- for example, 200, 201, 202 may have different headers. I think what this needs to do is generalize the other way around.

tatomyr: The current approach is: we take a code of the response, and if we find the corresponding header names in the config, we apply the rule. If we don't, we are trying to find if there's a common rule for the code's group (like 2xx, 3xx, &c.). 
I only use the generalization if there's no exact match for the current status code. 
For the example below status code `201` will pick up `[ 'Content-Type' ]`, and `202` will pick up `[ 'x-request-id' ]`: 

```yaml
response-contains-header:
  severity: error
  names:
    2xx: [ 'x-request-id' ]
    201: [ 'Content-Type' ]
```

Do you think it should be a different approach? If yes, could you elaborate?

adamaltman: Sorry, I think I misunderstood the direction of the generalization. OpenAPI calls this a range: 2XX is 200-299. I'm still not exactly clear what the method returns. Also, OpenAPI says the field must be a string like '200' and not a number like 200.

This is purely supposed to return the range match. For example, '200' becomes '2XX'. But my question is more like this... let's say there is an API with a responses map with '200', '201', '202', and '204' response codes.

And, let's say there are rules like:

```yaml
response-contains-header:
  severity: error
  names:
    '2xx': [ 'x-request-id' ]
    '201': [ 'Content-Type' ]
```

In this case, all of them must have the `x-request-id`, and in addition to that, the 201 must also have the `Content-Type` header?

What about a name like `convertToStatusCodeRange`?

RomanHotsiy: I would rename to `getMatchingStatusCodeRange`

tatomyr: > I'm still not exactly clear what the method returns.

It returns a string. If the original code matches the pattern (three digits) - it returns the appropriate code range. Otherwise it returns the input converted to string.

> In this case, all of them must have the x-request-id, and in addition to that, the 201 must also have the Content-Type header?

In my implementation the required headers for a specific code override the required headers for the range. So in your example `201` will only require the `Content-Type` header. I reckon, override is a more obvious option (especially when there is a large set of codes defined) but I might be wrong.

tatomyr:
@adamaltman, @RomanHotsiy 
Added changes for the `response-contains-property` rule. Please take a look.

docs/resources/built-in-rules.md
```diff
@@ -330,6 +330,25 @@ Verifies that each tag has a description.
...
+
+Enforces using specified response headers.
+
+```yaml
+lint:
+  rules:
+    response-contains-header:
+      severity: error
+      names:
+        2xx:
```
adamaltman, line 343:
```suggestion
        2XX: 
```
In OAS it is using capital `XX`.

docs/resources/built-in-rules.md
```diff
@@ -330,6 +330,25 @@ Verifies that each tag has a description.
...
+      names:
+        2xx: 
+          - x-request-id
+        400:
+          - Content-Length
+```
+
+### response-contains-property
+
+The same as `response-contains-header` except for a response properties.
```
adamaltman, line 351: Let's define it inline.
```suggestion
Enforces definition of specific response properties based on HTTP status code or HTTP status code range.
Priority is given to more precise status codes over the status code range.
```yaml
response-contains-header:
      severity: error
      names:
        2XX: 
          - created_at
        400:
          - title
```
```

File: docs/resources/built-in-rules.md [hunk 7]
```diff
@@ -335,6 +335,34 @@ Verifies that each tag has a description.
...
+          - Content-Length
+```
+
+### response-contains-property
+
+Enforces definition of specific response properties based on HTTP status code or HTTP status code range.
+Priority is given to more precise status codes over the status code range.
+```yaml
+response-contains-header:
+      severity: error
```
RomanHotsiy, line 360: identation

packages/core/src/rules/common/response-contains-header.ts
```diff
@@ -0,0 +1,30 @@
...
+          const expectedHeaders =
+            names[key] ||
+            names[getMatchingStatusCodeRange(key)] ||
+            names[getMatchingStatusCodeRange(key).toLowerCase()] ||
+            [];
+          for (const expectedHeader of expectedHeaders) {
+            if (!response.headers?.[expectedHeader]) {
+              report({
+                message: `Response object must have a "${expectedHeader}" header.`,
+                location: location.child('headers'),
```
RomanHotsiy, line 22:
```suggestion
                location: location.child('headers').key(),
```

packages/core/src/rules/common/response-contains-header.ts
```diff
@@ -0,0 +1,30 @@
...
+        enter: (response: Oas2Response | Oas3Response, { report, location, key }: UserContext) => {
+          const expectedHeaders =
+            names[key] ||
+            names[getMatchingStatusCodeRange(key)] ||
+            names[getMatchingStatusCodeRange(key).toLowerCase()] ||
+            [];
+          for (const expectedHeader of expectedHeaders) {
+            if (!response.headers?.[expectedHeader]) {
+              report({
+                message: `Response object must have a "${expectedHeader}" header.`,
```
RomanHotsiy, line 21:
```suggestion
                message: `Response object must contain a "${expectedHeader}" header.`,
```

packages/core/src/rules/oas2/response-contains-property.ts
```diff
@@ -0,0 +1,36 @@
...
+          if (schema.type !== 'object') return;
+          const expectedProperties =
+            names[key] ||
+            names[getMatchingStatusCodeRange(key)] ||
+            names[getMatchingStatusCodeRange(key).toLowerCase()] ||
+            [];
+          for (const expectedProperty of expectedProperties) {
+            if (!schema.properties?.[expectedProperty]) {
+              report({
+                message: `Response object must have a top-level "${expectedProperty}" property.`,
```
RomanHotsiy, line 27:
```suggestion
                message: `Response object must contain a top-level "${expectedProperty}" property.`,
```

packages/core/src/rules/oas2/response-contains-property.ts
```diff
@@ -0,0 +1,36 @@
...
+                location: location.child('properties'),
```
RomanHotsiy, line 28:
```suggestion
                location: location.child('properties').key(),
```

packages/core/src/rules/oas3/response-contains-property.ts
```diff
@@ -0,0 +1,38 @@
...
+            if (schema.type !== 'object') return;
+            const expectedProperties =
+              names[key] ||
+              names[getMatchingStatusCodeRange(key)] ||
+              names[getMatchingStatusCodeRange(key).toLowerCase()] ||
+              [];
+            for (const expectedProperty of expectedProperties) {
+              if (!schema.properties?.[expectedProperty]) {
+                report({
+                  message: `Response object must have a top-level "${expectedProperty}" property.`,
```
RomanHotsiy, line 28:
```suggestion
                  message: `Response object must contain a top-level "${expectedProperty}" property.`,
```

packages/core/src/rules/oas3/response-contains-property.ts
```diff
@@ -0,0 +1,38 @@
...
+                  location: location.child('properties'),
```
RomanHotsiy, line 29:
```suggestion
                  location: location.child('properties').key(),
```

RomanHotsiy (approved):
LGTM, I left a few cosmetic suggestions 💅

docs/resources/built-in-rules.md
```diff
@@ -335,6 +335,34 @@ Verifies that each tag has a description.
...
+      names:
+        2XX: 
+          - created_at
+        400:
+          - title
```
tatomyr, line 365:
```suggestion
```yaml
response-contains-header:
  severity: error
  names:
    2XX: 
      - created_at
    400:
      - title
```
```

Issue #539, opened 2022-01-31
Core package should not depend on colorette


akx:
**Describe the bug**

The core package depends on `colorette` to be able to color error messages using ANSI codes.

That package attempts to read `process.env` during import time, which isn't a thing in browsers (unless transpiled), causing trouble trying to use `@redocly/openapi-core` with e.g. https://vitejs.dev/.

(Another related issue is `node-fetch`, which doesn't work OOTB in a Vite environment either.)

**To Reproduce**
1. Try to import `@redocly/openapi-core` 1.0.0-beta.80 in a non-bundling browser environment.

**Expected behavior**

The library should not depend on a console coloring library.

AlexVarchuk:
It also relates to https://github.com/Redocly/redoc/issues/1820.

roman-sainchuk:
Hello @akx
In `1.0.0-beta.108` was added browser support. We're still depending on `colorette` but it is not used in browser env.
Please let us know if you still have issues with it

lornajane:
I think this has been fixed, so I'll close the issue but please let us know if this still causes a problem.

PR #541, opened 2022-01-31
Avoid process.* in core package

akx:
## What/Why/How?

Related to #539, this changes references to the Node `process` global to be less... fatal in e.g. browser environments.

In addition, it fixes that failing to create a `RedoclyClient` would just happily immediately shut down the Node process.

## Check yourself

- [x] Code is linted

## Security

  - has a net positive impact on security since you can't DoS things by tricking a Node process into misinitializing a RedoclyClient 


packages/core/src/config/config.ts
```diff
@@ -125,20 +125,26 @@ export type ResolveConfig = {
...
-  eu: 'eu.redocly.com',
-};
 
-// FIXME: temporary fix for our lab environments
-if (REDOCLY_DOMAIN?.endsWith('.redocly.host')) {
-  DOMAINS[REDOCLY_DOMAIN.split('.')[0] as Region] = REDOCLY_DOMAIN;
-}
-if (REDOCLY_DOMAIN === 'redoc.online') {
-  DOMAINS[REDOCLY_DOMAIN as Region] = REDOCLY_DOMAIN;
+export const env = process?.env || {};
```
RomanHotsiy, line 129: `process` can be not defined as well. I think you need 

```suggestion
export const env = typeof process !== 'undefined' ? process?.env || {} : {};
```

akx: Good catch, fixed.

RomanHotsiy (approved):
LGTM!

@andriyl @tatomyr I would appreciate if if somebody can test it with our other tools before merging.

adamaltman:
There are more conflicts now. We should revisit this one in a couple weeks @tatomyr (maybe during the cooldown?)

akx:
Hi @RomanHotsiy, @andriyl, @tatomyr, @adamaltman: I rebased this – could you re-review, please?

tatomyr:
Thank you @akx for your contribution! It was very kind of you to do the rebasing. 
The code looks quite valid, and the only thing remaining here is to test it with other products. I'll try to do this as soon as I can.

tatomyr (approved):

tatomyr:
Tested with redoc/reference-docs/workflows - seems to work correctly. 
@akx thank you again. The changes will be available with the next release.

PR #550, opened 2022-02-08
feat: support for webhooks to the split command

### LINKED ISSUE #534: Add support for webhooks to the split command
> I've tried to use openapi split command for a file with webhooks, but the webhooks stayed in openapi.yaml.

Add support for webhooks to the split command.

N/a

**Additional context**
None.

AntonKozachuk:
## What/Why/How?
Add support for webhooks and x-webhooks to the `split` command.

## Reference
Closes #534 

## Testing
Tested locally. Also, tested on my Windows machine.
Will add test cases for webhooks after we agree on the implementation.

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/split/index.ts
```diff
@@ -237,49 +242,73 @@ function gatherComponentsFiles(
...
+            method + langToExt(sample.lang),
           );
 
           fs.mkdirSync(path.dirname(sampleFileName), { recursive: true });
           fs.writeFileSync(sampleFileName, sample.source);
           // @ts-ignore
           sample.source = {
-            $ref: slash(path.relative(pathsDir, sampleFileName))
+            $ref:
+              rootDocObj === PATHS
```
andriyl, line 291: maybe a good point to move it to function

packages/cli/src/commands/split/index.ts
```diff
@@ -237,49 +242,73 @@ function gatherComponentsFiles(
...
+  rootDocObj: string,
+  customDir?: string,
 ) {
-  const { paths } = openapi;
-  if (paths) {
-    for (const oasPath of Object.keys(paths)) {
-      const pathFile = path.join(pathsDir, pathToFilename(oasPath)) + '.yaml';
-      const pathData: Oas3PathItem = paths[oasPath] as Oas3PathItem;
+  const docObj =
+    rootDocObj === PATHS
```
andriyl, line 253: A bit confusing ternarity).

packages/cli/src/commands/split/index.ts
```diff
@@ -237,49 +242,73 @@ function gatherComponentsFiles(
...
+      ? openapi.paths
+      : openapi[WEBHOOKS]
+      ? openapi[WEBHOOKS]
+      : openapi[xWEBHOOKS];
+
+  if (docObj) {
+    const checkDir = customDir ? customDir : path.join(openapiDir, rootDocObj);
```
andriyl, line 260: const checkDir = customDir || path.join(openapiDir, rootDocObj);

andriyl (commented):
LGTM, just leave some minor suggestions above.

packages/cli/src/commands/split/index.ts
AntonKozachuk: Done - https://github.com/Redocly/openapi-cli/blob/split-webhooks/packages/cli/src/commands/split/index.ts#L352

packages/cli/src/commands/split/index.ts
```diff
@@ -237,49 +242,62 @@ function gatherComponentsFiles(
   componentsFiles[componentType][componentName] = { inherits, filename };
 }
 
-function iteratePaths(
-  openapi: Oas3Definition,
-  pathsDir: string,
-  openapiDir: string
+function iteratePathItem(
+  openapi: any,
```
Marshevskyy, line 246: please specify a type

AntonKozachuk: It was renamed because ow iterates on webhooks also. The common thing is that both paths and webhooks has path item inside. 
customDir is used in tests right now - https://github.com/Redocly/openapi-cli/blob/split-webhooks/packages/cli/src/commands/split/__tests__/index.test.ts#L34, as I didn't come up with a better solution there.

AntonKozachuk: > please specify a type, There should be something better then `any` ;)

We only pass there an `openapi` parameter which can be of 2 types only:  `Oas3Definition | Oas3_1Definition`;
Duplicating those or having `object` type produces errors here - https://github.com/Redocly/openapi-cli/blob/split-webhooks/packages/cli/src/commands/split/index.ts#L252

RomanHotsiy: > The common thing is that both paths and webhooks has path item inside.

then rename it to `iteratePathItems`.

So how about changing the function accept path times object as a first arg and outDir as the second and call it like:

```
iteratePathItem(openapi.paths, path.join(openapiDir, 'paths'), componentsFiles);
iteratePathItem(openapi.webhooks || openapi['x-webhooks'] , path.join(openapiDir, 'webhooks'), componentsFiles);

Comment (AntonKozachuk) [reply packages/cli/src/commands/split/index.ts:246]: > Also two extra args seem to complex. What is customDir? looks like it's not used. How about changing the function accept paths object as a first arg and outDir as the second and call it like:
> 
> ```
> iteratePathItem(openapi.paths, path.join(openapiDir, 'paths'), componentsFiles);
> iteratePathItem(openapi, path.join(openapiDir, 'webhooks'), componentsFiles);
> ```

I can agree on the `customDir` part, but `rootDocObj` value is used in several places to define to logic between the webhooks and paths, mostly because of the different paths (e.g path starts with slash, webhooks don't). 
I believe the func itself is cleaner with aliases WEBHOOKS or PATHS, avoiding the actual strings.
Let me know if it makes sense.

File: packages/cli/src/commands/split/index.ts [hunk 5]
```diff
@@ -237,49 +242,62 @@ function gatherComponentsFiles(
...
-    for (const oasPath of Object.keys(paths)) {
-      const pathFile = path.join(pathsDir, pathToFilename(oasPath)) + '.yaml';
-      const pathData: Oas3PathItem = paths[oasPath] as Oas3PathItem;
+  const docObj = rootDocObj === PATHS ? openapi.paths : (openapi[WEBHOOKS] ? openapi[WEBHOOKS] : openapi[xWEBHOOKS]);
+
+  if (docObj) {
+    const checkDir = customDir || path.join(openapiDir, rootDocObj);
+    fs.mkdirSync(checkDir, { recursive: true });
+
+    for (const oasPath of Object.keys(docObj)) {
```
Marshevskyy, line 258: I know it's out of this issue, but I would like to refactor this function a bit, it includes 3 nested `for` (

Marshevskyy (commented):
good progress! please add more tests

packages/cli/src/commands/split/__tests__/index.test.ts
```diff
@@ -11,21 +15,23 @@ describe('#split', () => {
   it('should have correct paths for mac', () => {
     const pathsDir = 'test/paths';
     const openapiDir = 'test';
+    const componentsFiles: ComponentsFiles = {};
 
     jest.spyOn(openapiCore, 'slash').mockImplementation(() => 'paths/test.yaml');
     jest.spyOn(path, 'relative').mockImplementation(() => 'paths/test.yaml');
-    iteratePaths(require("./fixtures/spec.json"), pathsDir, openapiDir);
+    iteratePathItem(require("./fixtures/spec.json"), openapiDir, componentsFiles, PATHS, pathsDir);
```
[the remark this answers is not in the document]
AntonKozachuk: As I mentioned in the PR description, I want to add tests when we all be good with implementation.

packages/core/src/typings/openapi.ts
```diff
@@ -7,6 +7,7 @@ export interface Oas3Definition {
   security?: Oas3SecurityRequirement[];
   tags?: Oas3Tag[];
   externalDocs?: Oas3ExternalDocs;
+  'x-webhooks'?: Oas3_1Webhooks;
```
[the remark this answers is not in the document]
AntonKozachuk: Could you please describe your question in more detail?
x-webhooks is our extension to use Webhooks in 3.0.

packages/cli/src/commands/split/__tests__/index.test.ts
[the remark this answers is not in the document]
Marshevskyy: great

packages/core/src/typings/openapi.ts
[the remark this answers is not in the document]
Marshevskyy: I mean, I'm a bit confused with naming. sometimes you call it `xWebhooks ` )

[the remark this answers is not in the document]
AntonKozachuk: I just added alias to use in the logic: 
https://github.com/Redocly/openapi-cli/blob/split-webhooks/packages/cli/src/commands/split/types.ts#L25

packages/cli/src/commands/split/__tests__/index.test.ts
```diff
@@ -1,32 +1,50 @@
-import { iteratePaths } from '../index';
-import * as path from 'path';
-import * as openapiCore from '@redocly/openapi-core';
+// import { iteratePathItems } from '../index';
```
adamaltman, line 1: We should remove commented out code, right?

adamaltman: Or you are planning to fix the test, I guess...

AntonKozachuk: Yes, I will be working on those today. Thanks.

__tests__/split/oas3-no-errors/snapshot.js
```diff
@@ -0,0 +1,119 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E split openapi with no errors 1`] = `
+
+type: object
```
RomanHotsiy, line 5: Why it prints definition into stdout? I don't undeerstand. Must be some mistake.

__tests__/index.test.ts
```diff
@@ -72,6 +72,46 @@ describe('E2E', () => {
...
+    test('openapi with no errors', () => {
+      const folderPath = join(__dirname, `split/oas3-no-errors`);
+      const args = [
+        '--transpile-only',
+        '../../../packages/cli/src/index.ts',
+        'split',
+        '../../../__tests__/split/oas3-no-errors/openapi.yaml',
+        '--outDir=output'
+      ];
+      const result = getBundleResult(args, folderPath);
```
RomanHotsiy, line 110: rename it to getCommandOutput

RomanHotsiy (changes requested):
Looks great! Left a few clean up requests!

packages/cli/src/commands/split/index.ts
```diff
@@ -41,23 +42,16 @@ export async function handleSplit (argv: {
...
-function splitDefinition(openapi: Oas3Definition, openapiDir: string) {
+function splitDefinition(openapi: Oas3Definition | Oas3_1Definition, openapiDir: string) {
   fs.mkdirSync(openapiDir, { recursive: true });
-  const pathsDir = path.join(openapiDir, PATHS);
-  fs.mkdirSync(pathsDir, { recursive: true });
 
   const componentsFiles: ComponentsFiles = {};
-  iteratePaths(openapi, pathsDir, openapiDir);
   iterateComponents(openapi, openapiDir, componentsFiles);
+  iteratePathItems(openapi.paths, openapiDir, path.join(openapiDir, 'paths'), componentsFiles);
```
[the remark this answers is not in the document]
AntonKozachuk: I believe it was this way even in your PR - https://github.com/Redocly/openapi-cli/commit/000f45acb34ecaa05e80da50616761c4ba3de3e5

__tests__/split/oas3-no-errors/snapshot.js
AntonKozachuk: If I place it after, components aren't splitter. The files are created, but refs are missing in the paths or webhooks. 
@RomanHotsiy I can investigate this.

packages/cli/src/commands/split/index.ts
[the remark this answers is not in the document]
RomanHotsiy: hmmm, weird, it used to work so I'm just worried something went wrong....

__tests__/split/oas3-no-errors/snapshot.js
AntonKozachuk: @RomanHotsiy the reason for the print is this condition - https://github.com/Redocly/openapi-cli/blob/master/packages/cli/src/utils.ts#L141
And we call `writeYaml` func several times while running the `split` command.
Should I delete this section? It might have been added for testing purposes. 
@ohorbachevskyi maybe you can recall?

AntonKozachuk: It's actually printing the content instead of creating the actual files. We can change it if needed or just leave the snapshot as-is.

RomanHotsiy: Okay, let it be then.

packages/cli/src/commands/split/index.ts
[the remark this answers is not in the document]
AntonKozachuk: I believe this happens because `traverseDirectoryDeep` is now called inside `iteratePathItems` (and we call `replace$Refs` there).
Previuosly, `traverseDirectoryDeep` was called after `iterateComponents` and this way refs were resolved correctly.
We can change this again, but I believe it's easier to have iterateComponents above. I believe it doesn't break any logic.
@RomanHotsiy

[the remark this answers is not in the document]
RomanHotsiy: Just test it with some definitions and I'm good.

RomanHotsiy (approved):

PR #557, opened 2022-02-11
feat: introduce redocly.yaml

tatomyr:
## What/Why/How?

[Pitch: change .redocly.yaml config structure](https://intranet.redoc.ly/roadmap/tooling/2022-02/redocly-yaml/)
Introduces `redocly.yaml` file name.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/config/__tests__/load.test.ts
+++ b/packages/core/src/config/__tests__/load.test.ts
@@ -1,5 +1,7 @@
-import { loadConfig } from '../load';
+import { loadConfig, findConfig } from '../load';
 import { RedoclyClient } from '../../redocly';
+
+const fs = require('fs')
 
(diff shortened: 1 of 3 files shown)


adamaltman:
Looks good to me. :tada: 

I think we should also consider the same approach for consistency with the `.redocly.lint-ignore.yaml` file.

tatomyr:
@adamaltman  I'm not sure we can safely do the same for the `.redocly.lint-ignore.yaml` since it could be generated automatically. This should be thoroughly thought through so our users won't end up with several config files &c. 
Could we postpone this and make it separately? What do you think?

RomanHotsiy (approved):
Great 👍

PR #561, opened 2022-02-20
chore: support redocly registry at .com domain

RomanHotsiy:
## What/Why/How?

Some small updates required to migrate registry URLs from api.redoc.ly to api.redocly.com.

## Reference

Depends on https://github.com/Redocly/workflows/pull/1820

## Check yourself

- [x] Code is linted

## Security


slavikbez (approved):

packages/core/src/config/load.ts
```diff
@@ -41,7 +41,7 @@ export async function loadConfig(configPath?: string, customExtends?: string[]):
       },
       //support redocly.com domain for future compatibility
```
AyeOne, line 42: It seems that the comment need to be updated.
~Do I understand it correctly, that we don't need similar config for 'eu' region?~
(ok, I see, EU domain is already `.com`)

packages/core/src/redocly/index.ts
```diff
@@ -160,16 +162,15 @@ export class RedoclyClient {
 
 export function isRedoclyRegistryURL(link: string): boolean {
   const domain = process.env.REDOCLY_DOMAIN || DOMAINS[DEFAULT_REGION];
-  if (!link.startsWith(`https://api.${domain}/registry/`)) return false;
-  const registryPath = link.replace(`https://api.${domain}/registry/`, '');
 
-  const pathParts = registryPath.split('/');
+  const legacyDomain = domain === 'redocly.com' ? 'redoc.ly' : domain;
```
AyeOne, line 166: IMHO, it's a bit hard to understand logic here.

AyeOne (approved):

packages/core/src/redocly/index.ts
ozapotichnyi: I think it'll be more simple if you just hardcode 'redoc.ly' (you're comparing it with hardcoded values btw)

ozapotichnyi (approved):

packages/core/src/redocly/index.ts
RomanHotsiy: >  if you just hardcode 'redoc.ly' 

In such case someone on EU region can get `redoc.ly/*` links matching too but it should not.

Marshevskyy (approved):

tatomyr (approved):

PR #573, opened 2022-03-01
chore: add getAllTokens method to RedoclyClient

roman-sainchuk:
## What/Why/How?
Added `getAllTokens` method to `RedoclyClient` class to avoid code duplication of finding `.redocly-config.json` file, reading tokens from there and taking proper token field.
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/redocly/index.ts
+++ b/packages/core/src/redocly/index.ts
@@ -4,6 +4,7 @@ import { homedir } from 'os';
 import { red, green, gray, yellow } from 'colorette';
 import { RegistryApi } from './registry-api';
 import { AccessTokens, DEFAULT_REGION, DOMAINS, Region } from '../config/config';
+import { RegionalToken, RegionalTokenWithValidity } from './redocly-client-types';
 import { isNotEmptyObject } from '../utils';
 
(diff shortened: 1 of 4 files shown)


roman-sainchuk:
This change atm is needed for `openapi-vs-code` to avoid duplication of retrieving tokens from config file

packages/core/src/redocly/index.ts
```diff
@@ -91,16 +92,22 @@ export class RedoclyClient {
...
-  }[]> {
-    return (await Promise.all(
-      Object.entries(this.accessTokens).map(async ([key, value]) => {
-        return { region: key, token: value, valid: await this.verifyToken(value, key as Region) }
-      })
-    )).filter(item => Boolean(item.valid));
+  getAllTokens (): RegionalToken[] {
+    return Object.entries(this.accessTokens).map(([key, value]) => {
+      return { region: key as Region, token: value };
+    });
```
Marshevskyy, line 98:
```suggestion
    return Object.entries(this.accessTokens).map(([key, value]) =>
      ({ region: key as Region, token: value });
    );
```

Marshevskyy (approved):

PR #578, opened 2022-03-06
Update installation.md

JodyWinter:
## What/Why/How?
Substantive edit of this page.

## Testing
Fully tested npm global method, as discussed. Other methods will need testing.

## Check yourself

- [ ] Code is linted

## Security


docs/installation.md
```diff
@@ -1,13 +1,13 @@
 # Install OpenAPI CLI
 
-You can install and use OpenAPI CLI in several ways:
+There are two methods that you can use to install OpenAPI CLI: global and runtime.
```
skadinna, line 3:
```suggestion
There are two methods you can use to install OpenAPI CLI: global and runtime.
```

skadinna: Or even just "There are two ways to install" if we want to simplify it further.

docs/installation.md
```diff
@@ -188,4 +186,7 @@ The command output contains installation instructions. For example, to install t
 openapi completion >> ~/.bashrc
 ```
 
-The approach is similar for other shells. After the installation, restart your terminal for changes to take effect.
+The approach is similar for other shells. After the installation, restart your terminal for the changes to take effect.
+
+## What to do next
+Go back to the [Quickstart guide](./docs/quickstart.md) and finish it.
```
skadinna, line 192: issue: The quickstart guide should not require the user to switch back and forth between pages. In this section, we should recommend going to the configuration file and learning how to adjust it, and/or reading more about the supported commands and using them.

docs/installation.md
```diff
@@ -141,11 +141,9 @@ petstore.yaml: validated in 51ms
 Woohoo! Your OpenAPI definition is valid. 🎉
 ```
 
-## Post-installation
+## Set up tab completion
 
-### Set up tab completion
-
-To set up tab completion for the terminal, generate the completion script with the command:
+After a successful install, you should set up tab completion for the terminal. Generate the completion script using the following command:
```
skadinna, line 146: question: Do you think we could somehow indicate that this is not required? Tab completion is a convenient option that makes using OpenAPI CLI easier and faster, but it's not like the tool won't work without it. Our current docs are not great at indicating this...

skadinna (changes requested):

skadinna (approved):

swapnilogale (approved):
Looks great @JodyWinter. Simplifies the installation process substantially! Well done.

PR #583, opened 2022-03-08
fix: detect parameters without schema or content fields

### LINKED ISSUE #568: Parameters without schema or content fields
**Describe the bug**

OpenAPI CLI should detect `Parameters` without `schema` or `content` fields while the OpenAPI specification states that one of these fields MUST be specified. This should be part of the `spec` rule.

**To Reproduce**
TBD

**Expected behavior**
Show a problem if none of schema or content are defined or if both schema and content are defined.

**OpenAPI definition**
3.0 and 3.1

**`openapi-cli` Version(s)**
1.1.0-beta.83

**`Node.js` Version(s)**
14

**Additional context**

This came up in a slack discussion.

### LINKED ISSUE #570: Request for Schema validation
No - This is not related to any problems.

It seems like currently, we don't have any schema validation for request parameters. I think it would be great if you could add a schema evaluation like the online swagger editor.

If we do not have a scheme for the request parameters, I think we can raise an error as shown below. 

image

**OpenAPI Specification**

```yaml
openapi: 3.0.0
info:
  title: Petstore API
  version: 1.0.0
  description: description

paths:

  /v1/pets/{petId}:
    get:
      tags:
        - pets
      summary: Gets a specific pet
      parameters:
      - name: petId
        in: path
      - name: limit
        in: query
      responses:
        '200':
          description: Details of the requested pet.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'

components:

  schemas:
    Pet:
      type: object
      properties:
        resourceType:
          type: string
```

AntonKozachuk:
## What/Why/How?
OpenAPI CLI should detect Parameters without schema or content fields while the OpenAPI specification states that one of these fields MUST be specified.
image

## Reference
Fixes #568 
Fixes #570 

## Testing
Tested locally

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/common/spec.ts
+++ b/packages/core/src/rules/common/spec.ts
@@ -2,11 +2,11 @@ import type { Oas3Rule, Oas2Rule } from '../../visitors';
 import { isNamedType } from '../../types';
 import { oasTypeOf, matchesJsonSchemaType, getSuggest } from '../utils';
 import { isRef } from '../../ref-utils';
-import { isPlainObject, hasOneOfProperty } from '../../utils';
+import { isPlainObject } from '../../utils';
 
 export const OasSpec: Oas3Rule | Oas2Rule = () => {
   return {
-    any(node: any, { report, oasVersion, type, location, key, resolve, ignoreNextVisitorsOnNode } ) {
+    any(node: any, { report, type, location, key, resolve, ignoreNextVisitorsOnNode } ) {
       const nodeType = oasTypeOf(node);
 
       if (type.items) {
@@ -32,18 +32,6 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
         if (!(node as object).hasOwnProperty(propName)) {
           report({
             message: `The field \`${propName}\` must be present on this level.`,
-            location: [{ reportOnKey: true }],
-          });
-        }
-      }
-
-      if (oasVersion !== 'oas2') {
-        if (
-          type.name === 'Parameter' &&
-          !hasOneOfProperty(node as object, ['schema', 'content'])
-        ) {
-          report({
-            message: `A parameter must contain either a \`schema\` property, or a \`content\` property, but not both.`,
             location: [{ reportOnKey: true }],
           });
         }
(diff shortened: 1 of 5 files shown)


packages/core/src/rules/common/spec.ts
```diff
@@ -37,6 +37,18 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
         }
       }
 
+      if (oasVersion !== 'oas2') {
```
RomanHotsiy, line 40: Better to add something like `oneOfProperty` into the type tree here https://github.com/Redocly/openapi-cli/blob/master/packages%2Fcore%2Fsrc%2Ftypes%2Foas3.ts#L121

RomanHotsiy (approved):

packages/core/src/types/oas3.ts
```diff
@@ -136,6 +136,7 @@ const Parameter: NodeType = {
     content: 'MediaTypeMap',
   },
   required: ['name', 'in'],
+  requiredOneOf: ['schema', 'content']
```
Marshevskyy, line 139: would it work with OAS 3.1?

AntonKozachuk: Yes, because Oas3_1Types include Oas3Types (and Parameter type too):

adamaltman: Do we have a code style for a trailing comma on these? I notice, for example, a trailing comma on the line above.

tatomyr (approved):

packages/core/src/types/oas3.ts
AntonKozachuk: Not sure.

AntonKozachuk: @adamaltman can I merge this or hold it till we found a common ground on commas?

RomanHotsiy: We do have but looks like this project does not have prettier set up. @tatomyr would be great to do it on the cool-down.

RomanHotsiy, line 139:
```suggestion
  requiredOneOf: ['schema', 'content'],
```

RomanHotsiy: @AntonKozachuk I added the comma 😄

PR #585, opened 2022-03-10
fix: terser crash in client-site bundles due to process.env assignment

RomanHotsiy:
## What/Why/How?

The previous release crashes client-side builds (e.g. in our portal) because it contains `process.env` assignment which is converted to a const by DefinePlugin.

## Reference

https://github.com/Redocly/redocly-portal/runs/5487874186?check_suite_focus=true#step:7:100

## Check yourself

- [x] Code is linted

## Security


tatomyr (approved):

akx:
This is related to #541... If only that had gotten merged earlier, hey? :)

PR #591, opened 2022-03-11
fix: use correct domain when logging to labs

tatomyr:
## What/Why/How?

At the moment, when trying to log in to a lab environment, the request is always sent to the `api.redocly.com/registry`. It happens because there are no 'lab' regions in `ALLOWED_DOMAINS` so OpenapiCLI uses default 'us' region.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -130,7 +130,6 @@ export const DOMAINS: { [region in Region]: string } = {
   us: 'redocly.com',
   eu: 'eu.redocly.com',
 };
-export const AVAILABLE_REGIONS = Object.keys(DOMAINS) as Region[];
 
 // FIXME: temporary fix for our lab environments
 if (REDOCLY_DOMAIN?.endsWith('.redocly.host')) {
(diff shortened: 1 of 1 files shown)


roman-sainchuk (approved):

roman-sainchuk:
Cannot add as a suggestion due to unchanged `DOMAINS` but can we write it like this to avoid issues with the ordering?
```
export const DOMAINS: { [region in Region]: string } = {
  us: 'redoc.ly',
  eu: 'eu.redocly.com',
  ...(REDOCLY_DOMAIN?.endsWith('.redocly.host')
    ? { [REDOCLY_DOMAIN.split('.')[0] as Region]: REDOCLY_DOMAIN }
    : {}),
  ...(REDOCLY_DOMAIN === 'redoc.online' ? { [REDOCLY_DOMAIN as Region]: REDOCLY_DOMAIN } : {}),
};
export const AVAILABLE_REGIONS = Object.keys(DOMAINS) as Region[];
```

tatomyr:
@roman-sainchuk it looks nicer, indeed. However, since the comment defines these lines as a temporary fix, I doubt it's a good idea to incorporate them into the stable code.

RomanHotsiy (approved):

PR #593, opened 2022-03-14
Fix/improve error locations

roman-sainchuk:
## What/Why/How?
Improve error highlighting for cases:
- Servers field missing
- Root fields have empty (nullable) value

Currently for these cases entire file is highlighted instead of specific problematic place
## Check yourself

- [x] Code is linted

## Security


packages/core/src/format/codeframes.ts
```diff
@@ -183,8 +183,10 @@ export function getAstNodeByPointer(root: YAMLNode, pointer: string, reportOnKey
   for (const key of pointerSegments) {
     if (currentNode.kind === yamlAst.Kind.MAP) {
       const mapping = currentNode.mappings.find((m) => m.key.value === key);
-      if (!mapping?.value) break;
-      currentNode = mapping?.value as YAMLNode;
+      if (!mapping) break;
```
tatomyr, line 186: Could we somehow avoid using the `break`s? Would extracting these checks into a separate function be an option?

roman-sainchuk: I think we will have the same picture but with `return`s instead of `break`s

RomanHotsiy (approved):

tatomyr (approved):

PR #597, opened 2022-03-17
hotfix: do not crash on empty redocly.yaml

tatomyr:
## What/Why/How?

- Fixes crash when there's an empty [.]redocly.yaml file
- Fixed some types
- Applied Prettier

## Check yourself

- [ ] Code is linted

## Security


Marshevskyy (approved):
since this is hotfix, it's good to go, but please cover this use case with tests

PR #603, opened 2022-03-23
fix: improper handling of filenames with multiple dots

### LINKED ISSUE #560: bundle issue when files use multiple dots
We're having a similar issue, in that our object files are named `foo.bar.baz.v1.yaml` - we've verified that the bundler behaves sensibly if we rename the files/references to `foo-bar-baz-v1.yaml`, however it would be nice(r) for the bundler not to assume that the first period it encounters is the filename and everything after it is the extension.

As it is, `foo.bar.baz.v1.yaml`, `foo.bar.qux.v1.yaml` and `foo.quux.quuz.v1.yaml` get folded into the components collection as `foo-1`, `foo-2` and `foo-3` - along with a load of "duplicate name" errors.

_Originally posted by @jasonbstanding in https://github.com/Redocly/openapi-cli/issues/260#issuecomment-1041396491_

And678:
## What/Why/How?
Fixes the issue when `foo.bar.baz.v1.yaml`, `foo.bar.qux.v1.yaml` and `foo.quux.quuz.v1.yaml` would get folded into the components collection as foo-1, foo-2 and foo-3 - along with a load of "duplicate name" errors.

This happened because it treated everthing after the **first** dot as a file extension. Now it matches the **last** dot of the filename.

## Reference
closes #560 
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/__tests__/__snapshots__/bundle.test.ts.snap
+++ b/packages/core/__tests__/__snapshots__/bundle.test.ts.snap
@@ -144,10 +144,24 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/vendor'
+                $ref: '#/components/schemas/vendor.schema'
 components:
   schemas:
     vendor:
+      $ref: '#/components/schemas/vendor.schema'
+    myvendor:
+      $ref: '#/components/schemas/vendor.schema'
+    simple:
+      type: string
+    A:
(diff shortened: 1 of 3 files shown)


tatomyr:
Could you write a test for this?

jasonbstanding:
👍

PR #604, opened 2022-03-23
fix: update .redocly.yaml types

roman-sainchuk:
## What/Why/How?
Update types used to lint .redocly.yaml according to new config file structure
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/__tests__/lint.test.ts
+++ b/packages/core/src/__tests__/lint.test.ts
@@ -46,7 +46,7 @@ describe('lint', () => {
   it('lintConfig should work', async () => {
     const document = parseYamlToDocument(
       outdent`
-      apiDefinitions: error string
+      apis: error string
       lint:
         plugins:
           - './local-plugin.js'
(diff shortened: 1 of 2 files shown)


tatomyr (approved):

packages/core/src/__tests__/lint.test.ts
```diff
@@ -105,6 +105,40 @@ describe('lint', () => {
...
+      Array [
+        Object {
+          "location": Array [
+            Object {
+              "pointer": "#/apis/lint/plugins",
+              "reportOnKey": true,
+              "source": "",
+            },
+          ],
+          "message": "Property \`plugins\` is not expected here.",
```
adamaltman, line 133: PR looks good to me. Are we sure this is actually correct though? I heard:

- apis lint rules (`preprocessors`, `rules`, `decorators`) merge and take priority over the global configuration
- other things override the global configuration (this was explicitly `extends`, but I'm unclear is `plugins` explicitly not allowed)?

tatomyr: Yes. As far as I remember, we discussed `plugins` usage with @RomanHotsiy and decided it should be declared only on the top level.

adamaltman: We'll adjust the docs PR.

PR #616, opened 2022-04-04
feat: add support local files to extends

AlexVarchuk:
## What/Why/How?

Implements https://intranet.redoc.ly/roadmap/tooling/2022-04/cli-extends/

## Check yourself

- [x] Code is linted

## Security


packages/core/src/config/__tests__/config.test.ts
```diff
@@ -51,103 +50,103 @@ const testConfig: Config = {
...
-        "resolve": Object {
-          "http": Object {
-            "customFetch": undefined,
-            "headers": Array [],
-          },
-        },
-      }
-    `);
-  });
+//   it('should merge lint defined in "apis" section the root one', () => {
```
tatomyr, line 53: Fix this

__tests__/bundle/bundle-lint-format/checkstyle-format-snapshot.js
```diff
@@ -9,6 +9,7 @@ exports[`E2E bundle lint format bundle lint: should be formatted by format: chec
 </checkstyle>
 
 No configurations were defined in extends -- using built in recommended configuration by default.
+Warning! This default behavior is going to be deprecated soon..
```
RomanHotsiy, line 12: Let's have only one snapshot with this warning and keep other tests with happy-path.

packages/core/__tests__/login.test.ts
```diff
@@ -11,7 +11,7 @@ describe('login', () => {
       writable: true,
       configurable: true
     });
-    await client.login('token');
+    await client.login('token'); // TODO: bug with rewrite local config file
```
RomanHotsiy, line 14: Can this todo be fixed?

packages/core/src/config/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+export function parsePresetName(presetName: string): { pluginId: string; configName: string } {
+  if (presetName.indexOf('/') > -1) {
+    const [pluginId, configName] = presetName.split('/');
+    return { pluginId, configName };
+  } else {
+    return { pluginId: '', configName: presetName };
+  }
+}
+
+export function assignExisting<T>(target: Record<string, T>, obj: Record<string, T>) {
```
RomanHotsiy, line 24: Let's move it to generic utils, not config utils.

packages/core/src/config/utils.ts
```diff
@@ -0,0 +1,188 @@
...
+      customFetch: undefined,
+    },
+  };
+}
+
+export function getUniquePlugins(plugins: Plugin[]): Plugin[] {
+  return plugins.reduce<Plugin[]>(
+    (acc, item) => (acc.some(({ id }) => id === item.id) ? acc : [...acc, item]),
+    [],
+  );
```
RomanHotsiy, line 187: This is hard to read and also it has O(n^2) complexity (which doesn't matter here because it is not hot path and n is going to be small but still):

```suggestion
  const seen = new Set();
  const results = [];
  for (const p of plugins) {
    if (!seen.has(p.id)) {
      results.push(p);
    } else {
      // we need to show some warning here maybe?
    }
    seen.add(p.id);
  }
  return results;
```

If we don't need to show a warning then it can be as simple as:

```suggestion
  return [ ...new Map(plugins.map(p => [p.id, p])).values() ];
```

packages/core/src/config/__tests__/__snapshots__/config-resolvers.test.ts.snap
```diff
@@ -0,0 +1,2886 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`resolveApis should not merge recommended extends by default by every level 1`] = `
```
RomanHotsiy, line 3: These snapshots look not ideal. We should focus one some specific aspect. Why we have plugins and other duplicated thing here?

The concert is we will never review those snapshots because they are unmaintainable. 

We may have a single full snapshots but other tests should focus on the specific area. Not on whole config dump.

RomanHotsiy (changes requested):
Great job!

Left a few small requests.

anastasiia-developer (approved):

packages/core/__tests__/login.test.ts
tatomyr: I'd like to left it as a tech debt for now.

packages/core/src/config/__tests__/__snapshots__/config-resolvers.test.ts.snap
AlexVarchuk: we reduced the count of snapshots and left only two snapshots for the lint part of the config.

tatomyr (approved):

RomanHotsiy (approved):

PR #622, opened 2022-04-05
hotfix: fix getMergedConfig to keep config file location

RomanHotsiy:
## What/Why/How?

Push command skips any dependencies because the merged config loses the original config filename and there is a check here: https://github.com/Redocly/openapi-cli/blob/hotfix-get-merged-config/packages/cli/src/commands/push.ts#L224

## Reference

Reported by customer.

## Check yourself

- [x] Code is linted

## Security


Marshevskyy (approved):
after release we should cover this case with tests

tatomyr (approved):

PR #637, opened 2022-04-12
feat: merge plugins by unique paths

tatomyr:
## What/Why/How?

We have to merge plugins in a way that should avoid duplication (they should be unique for each path). For this, I'm leaving only the last entry of each unique plugin path. Entries of type `Plugin` remain unchanged.
> Additionally:
> + Updated tests settings.
> + Added one not related test (for `getConfig` function).

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/config/__tests__/load.test.ts
+++ b/packages/core/src/config/__tests__/load.test.ts
@@ -100,7 +100,7 @@ describe('resolveNestedPlugins', () => {
 
 describe('getConfig', () => {
   jest.spyOn(fs, 'hasOwnProperty').mockImplementation(() => false);
-  it('should return empty object if there is no configPath', () => {
+  it('should return empty object if there is no configPath and config file is not found', () => {
     expect(getConfig()).toEqual(Promise.resolve({}));
   });
 });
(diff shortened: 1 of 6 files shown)


packages/core/src/config/__tests__/load.test.ts
```diff
@@ -97,3 +97,10 @@ describe('resolveNestedPlugins', () => {
     expect(path).toStrictEqual('/api/test-plugin.js');
   });
 });
+
+describe('getConfig', () => {
+  jest.spyOn(fs, 'hasOwnProperty').mockImplementation(() => false);
+  it('should return empty object if there is no configPath', () => {
```
AlexVarchuk, line 103:
```suggestion
  it('should return empty object if there is no configPath and config file is not found', () => {
```

packages/core/src/config/load.ts
```diff
@@ -137,7 +137,7 @@ export function getLintRawConfigWithMergedContentByPriority(
 
   return {
     ...lintConfig,
-    plugins: mergeArrays(extendedContent?.plugins, lintConfig.plugins), // FIXME: plugins should be uniq
+    plugins: getUniquePlugins(mergeArrays(extendedContent?.plugins, lintConfig.plugins)),
     // TODO: think about unique default rules/plugins group
```
AlexVarchuk, line 141: do we need this `TODO` after the fix?

PR #646, opened 2022-04-18
fix: resolvable scalars not working with asserts

RomanHotsiy:
## What/Why/How?

Resolvable scalars were not working with assertions.

## Reference

Fixes https://github.com/Redocly/openapi-cli/issues/645

## Testing

TODO:

- [x] add e2e test

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/common/assertions/utils.ts
+++ b/packages/core/src/rules/common/assertions/utils.ts
@@ -1,3 +1,4 @@
+import { isRef } from '../../../ref-utils';
 import { Problem, ProblemSeverity, UserContext } from '../../../walk';
 import { asserts } from './asserts';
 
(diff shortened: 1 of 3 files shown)


ohorbachevskyi (approved):

AntonKozachuk (approved):

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -102,7 +103,9 @@ export function buildSubjectVisitor(
     for (const assert of asserts) {
       if (properties) {
         for (const property of properties) {
-          runAssertion(node[property], assert, location.child(property), report);
+          // we can have resolvable scalar so need to resolve value here.
+          const value = isRef(node[property]) ? resolve(node[property])?.node : node[property];
```
adamaltman, line 107: I have an open question if we can detect that it is a `$ref`? For example, enforce `$ref` usage for path items (or anything else that you may want to enforce it with). So in other words, I would like to be able to assert `isRef: true`.

adamaltman: Also, I would like to assert some things about the ref, like if it is a ref to a another file (as opposed to the components section of the same file).

RomanHotsiy: > I would like to be able to assert isRef: true.

We can do it I think. Needs some changes but it should be possible.

> I would like to assert some things about the ref, like if it is a ref to a another file (as opposed to the components section of the same file).

I think we need some syntax idea there then.

adamaltman: I made a separate issue for it: https://github.com/Redocly/openapi-cli/issues/648

adamaltman (approved):

PR #653, opened 2022-04-20
chore: update node version requirements in docs

tatomyr:
## What/Why/How?

Update of node version requirements in the Contribution guide.

## Check yourself

- [ ] Code is linted

## Security


adamaltman:
Question: Node 14 is LTS and supported for another 1 year. Do we not want to support that?

tatomyr:
@adamaltman this  is not about supporting the **Node** version but rather about **npm v7** required to run the monorepo with workspaces. And since the minimal supported **npm** version comes with **Node 15**, it seems logical to require at least this version. 
Albeit we may support **Node v14+** in the packages themselves (since we currently do this for **v12+**), I'd recommend against mixing different **Node** versions in the project unless we have to support the older versions.
@RomanHotsiy do we have such a requirement?

RomanHotsiy:
@tatomyr when you specify engine in package.json of packages it will impact end users. They will see the warning I believe.
But what's important, the packages work just fine with older versions of node and npm.

We need npm@7 only for the development purposes so I would not put node requirement into the package.lock of individual packages. Maybe let's put it on the top-level one.

>  I'd recommend against mixing different Node versions in the project unless we have to support the older versions.

We don't have to mix them. We should just use single node for development. But we should let users use any supported node version they can.

AntonKozachuk (approved):

PR #662, opened 2022-04-28
feat: adjust assertions syntax

volodymyr-rutskyi:
## What/Why/How?

Simplify the assertions syntax by getting rid of the nested arrays.

The code changes are only these few lines: https://github.com/Redocly/openapi-cli/pull/662/files#diff-47607adc7025266f655cda3df6abbd5cc9f1fbf4e9f493c7b0ba6ed6c3350614R716-R738

The rest is examples and docs changes.

This change basically collects all of the assertions under `rules` and puts them in an `assertions` rule. Thus only changing the definition in `.redocly.yaml` and not the internal handling. 

## Reference

https://github.com/Redocly/openapi-cli/issues/647

## Check yourself

- [ ] Code is linted

## Security


__tests__/lint/assertions-casing-camel-case-error/.redocly.yaml
```diff
@@ -4,16 +4,17 @@ apis:
...
-        property: operationId
-        message: Operation id should be camelCase
-        casing: camelCase
+    assert/operation-get:
+      context:
+      - type: Operation
+        matchParentKeys: [get]
+      subject: Operation
+      property: operationId
+      message: Operation id for get operation should be canmelCase
```
adamaltman, line 13:
```suggestion
      message: Operation id for get operation should be camelCase
```

I think this is an opportunity to fix this typo. The corresponding snapshot will need to be updated too.

__tests__/lint/assertions-casing-snake-case-error/.redocly.yaml
```diff
@@ -4,16 +4,17 @@ apis:
...
-        property: operationId
-        message: Operation id should be snake_case
-        casing: snake_case
+    assert/operation-id-pascal-case:
+      context:
+        - type: Operation
+          matchParentKeys: [get]
+      subject: Operation
+      property: operationId
+      message: Operation id should be PascalCase
```
adamaltman, line 13: This one seems like the message should be improved to help document the functionality better. Corresponding snapshots should be updated too.

```suggestion
      message: Operation id for get requests should be PascalCase
```

__tests__/lint/assertions-defined-error/.redocly.yaml
```diff
@@ -4,16 +4,17 @@ apis:
...
-        message: Response content should be defined
-        defined: true
-      - subject: Operation
-        property: security
-        message: Operation security should be defined
-        defined: true
+    assert/response-content-defined:
+      context:
+      - type: Response
+        excludeParentKeys: ['200']
```
adamaltman, line 10:
```suggestion
        excludeParentKeys: ['204']
```
A practical use case is to adjust this one for 204 (HTTP code means no content). I realize you didn't write these test cases, but this improves them for real-world use-cases. This will allow future developers to contribute more easily, and prospective users to review the tests which will have better semantic meaning.

adamaltman:
I had minor nitpicks and didn't consider the actual implementation. 😆 (I think it seems good in concept. 🎉 )

__tests__/lint/assertions-defined-error/.redocly.yaml
volodymyr-rutskyi: It makes perfect sense, thanks :)

packages/core/src/config/config.ts
```diff
@@ -687,6 +687,43 @@ function transformApiDefinitionsToApis(
...
+      
+      } else {
+        // If it's not an assertion, keep it as is
+        transformedRules[ruleKey] = rules[ruleKey];
+      }
+    }
+
+    if (assertions.length > 0) {
+      transformedRules.assertions = assertions
+    }
```
tatomyr, line 722:
```suggestion
  const assertions = [];
  for (const [ruleKey, rule] of Object.entries(rules)) {
    if (ruleKey.startsWith('assert/') && typeof rule === 'object' && rule !== null) {
      const assertion = rule;
      assertions.push({
        ...assertion,
        assertionId: ruleKey.replace('assert/', ''),
      });
    } else {
      // If it's not an assertion, keep it as is
      transformedRules[ruleKey] = rule;
    }
  }
  if (assertions.length > 0) {
    transformedRules.assertions = assertions;
  }
```

packages/core/src/config/config.ts
```diff
@@ -712,9 +749,14 @@ export function transformConfig(rawConfig: DeprecatedRawConfig | RawConfig): Raw
       )} instead. Read more about this change: https://redocly.com/docs/api-registry/guides/migration-guide-config-file/#changed-properties\n`,
     );
   }
+  
   return {
     'features.openapi': referenceDocs,
     apis: transformApiDefinitionsToApis(apiDefinitions),
+    lint: lint && {
```
tatomyr, line 756: I'm afraid it won't work correctly with nested configs we've just implemented (see [here](https://github.com/Redocly/openapi-cli/pull/616)). It looks like `transformConfig` will fire each time a nested config is loaded and thus convert all `assertions/...` entries into the single `assertions` rule. And then, it may be overwritten by the following `assertions` arrays from other configs if such exist. 
@adamaltman, is the actual intent to inherit assertion rules from the `extends` and then merge with the root rules, isn't it?

adamaltman: @tatomyr yes. Of course, if there is a conflict on the individual assert name then it would override. For example,

**redocly.yaml**
```yaml
lint:
  extends:
    ./another.yaml
  rules:
    assert/operation-get:
      context:
      - type: Operation
        matchParentKeys: [get]
      subject: Operation
      property: operationId
      message: Operation id for get operation should be camelCase
      casing: camelCase
```

**another.yaml**
```yaml
lint:
  extends:
    ./another.yaml
  rules:
    assert/operation-get:
      context:
      - type: Operation
        matchParentKeys: [get]
      subject: Operation
      property: operationId
      message: Operation id for get operation should be PascalCase
      casing: PascalCase
    assert/operation-id-camel-case:
      subject: Operation
      property: operationId
      message: Operation id should be camelCase
      casing: camelCase
```

I would expect the previous two files to actually merge this way:

```yaml
lint:
  rules:
    assert/operation-get:
      context:
      - type: Operation
        matchParentKeys: [get]
      subject: Operation
      property: operationId
      message: Operation id for get operation should be camelCase
      casing: camelCase
    assert/operation-id-camel-case:
      subject: Operation
      property: operationId
      message: Operation id should be camelCase
      casing: camelCase

```

Because priority is given to the `redocly.yaml` file and there is one conflict where the priority matters.

Thanks for spotting this issue @tatomyr

volodymyr-rutskyi: This is a good catch. I'll handle that once the [nested config PR](https://github.com/Redocly/openapi-cli/pull/616) is merged.

adamaltman: Thanks @volodymyr-rutskyi

volodymyr-rutskyi: @tatomyr The PR is now updated to handle nested configs and is ready to review again: https://github.com/Redocly/openapi-cli/pull/662/commits/69729babe4a71160ba67315a7971047d0935c3dd

adamaltman:
I will prepare the docs as a separate PR. Thank you. I'm really looking forward to this rule.

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٤:٠٢ م، كتب Adam Altman ***@***.***>:

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٤:٠٤ م، كتب minalabib223 ***@***.***

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٤:٣٩ م، كتب minalabib223 ***@***.***

PR #665, opened 2022-05-01
docs: update lint.md

JodyWinter:
## What/Why/How?
Substantive edit, update and testing.

## Testing
Done.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/commands/lint.md
+++ b/docs/commands/lint.md
@@ -1,16 +1,12 @@
 # `lint`
 
-## Introduction
+## Why you need it
 
-Redocly OpenAPI CLI can identify and report on problems found in OpenAPI definitions. This helps you avoid bugs and make API definitions more consistent.
+One of the most powerful features of Redocly OpenAPI CLI is its ability to identify problems found in OpenAPI definition files. This is achieved through the `lint` command, which helps your definitions conform to a consistent set of rules.
 
-The `lint` command reports on problems and executes preprocessors and rules. Unlike the `bundle` command, `lint` doesn't execute decorators.
+The `lint` command ensures that your definition file structure is valid (according to the OpenAPI Specification and your own custom rules, if any) and contains no errors. It also executes preprocessors and rules (unlike the `bundle` command, `lint` doesn't execute decorators). To learn more about preprocessors and rules, see the [custom rules](../resources/custom-rules.md) page.
 
-:::success Tip
-To learn more about preprocessors and rules, refer to the [custom rules](../resources/custom-rules.md) page.
-:::
-
-## Usage
+Always the over-achiever, `lint` has options which you can execute alone, or combine with others:
 
 ```bash
 openapi lint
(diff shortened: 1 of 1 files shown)


docs/commands/lint.md
```diff
@@ -1,16 +1,12 @@
 # `lint`
 
-## Introduction
+## Why you need it
```
skadinna, line 3: issue: Changing this heading will make this page inconsistent with all the other pages for commands. Currently, they all follow the same structure: Introduction - Usage - Options - Examples. I believe we want to keep this "predictability" in the way they're internally organized. Do you plan to make this same change on all other command pages?

docs/commands/lint.md
```diff
@@ -1,16 +1,12 @@
...
 
-Redocly OpenAPI CLI can identify and report on problems found in OpenAPI definitions. This helps you avoid bugs and make API definitions more consistent.
+One of the most powerful features of Redocly OpenAPI CLI is its ability to identify problems found in OpenAPI definition files. This is achieved through the `lint` command, which helps your definitions conform to a consistent set of rules.
```
skadinna, line 5:
```suggestion
One of the most powerful features of Redocly OpenAPI CLI is the ability to identify problems found in OpenAPI definition files. This is achieved with the `lint` command, which helps your definitions conform to a consistent set of rules.
```

docs/commands/lint.md
```diff
@@ -1,16 +1,12 @@
...
 
-The `lint` command reports on problems and executes preprocessors and rules. Unlike the `bundle` command, `lint` doesn't execute decorators.
+The `lint` command ensures that your definition file structure is valid (according to the OpenAPI Specification and your own custom rules, if any) and contains no errors. It also executes preprocessors and rules (unlike the `bundle` command, `lint` doesn't execute decorators). To learn more about preprocessors and rules, see the [custom rules](../resources/custom-rules.md) page.
```
skadinna, line 7: suggestion: I would keep the information about not executing decorators more visible - it's an important distinction between those two commands, and if we keep it in parentheses as a sidenote, it's easy to miss.

docs/commands/lint.md
```diff
@@ -1,16 +1,12 @@
...
 
-:::success Tip
-To learn more about preprocessors and rules, refer to the [custom rules](../resources/custom-rules.md) page.
-:::
-
-## Usage
+Always the over-achiever, `lint` has options which you can execute alone, or combine with others:
```
skadinna, line 9: issue: We list some of the options here, but not all of them. Why? Also, again - removing the Usage/Options headings makes the page structure inconsistent with other commands pages.

docs/commands/lint.md
```diff
@@ -20,221 +16,375 @@ openapi lint [--generate-ignore-file] [--help]
...
-`--format`               | `string`  | no           | `codeframe` | Format for the output.<br />**Possible values:** `codeframe`, `stylish`, `json`, `checkstyle`
-`--generate-ignore-file` | `boolean` | no           | -           | [Generate ignore file](#generate-ignore-file)
-`--help`                 | `boolean` | no           | -           | Show help
-`--max-problems`         | `number`  | no           | 100         | Truncate output to display the specified [maximum number of problems](#max-problems)
-`--skip-preprocessor`    | `array`   | no           | -           | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--skip-rule`            | `array`   | no           | -           | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--version`              | `boolean` | no           | -           | Show version number
+## How to run the `lint` command
+In your terminal, cd to the root of your project directory, then type the command along with any of the options described below.
```
skadinna, line 22: issue: This seems out of order, or just confusing - options are listed above this section? But they're described below...and they're not described all in one place? Also, this section doesn't deliver on its title. It doesn't show how to run the command...

skadinna: suggestion: I think we should reconsider this new page structure. I'm totally OK with more specific, actionable headings for usage examples. But I think we're making the page confusing if we remove the whole list/table of all supported options. It's important for users to be able to see all supported options in one place. The "old" table was doing a good job of briefly describing each option and then linking to a specific section on how to use the option - maybe we can build on that and make it better?

docs/commands/lint.md
skadinna: suggestion: Maybe we can drop this `Usage` section and these options altogether, but keep the `Options` section with all supported options listed in one place? That way, we won't have this redundant/incomplete situation where some options are mentioned twice, and some just once. And then we can rename the `Examples` section to `Usage examples`. That structure won't be too drastically different from all the other pages (it's easy to make that change across all pages), and it will still have the effect of cleaning up the page. What do you think?

skadinna (changes requested):

docs/commands/lint.md
adamaltman: I think the table is very important.

Same with having consistent page structures. (comment above about: Introduction - Usage - Options - Examples)

docs/commands/lint.md
JodyWinter: I'll reinstate the current format. Due to my part-time nature, it might take months to sort all of the command pages out so I agree to keep the status quo.

docs/commands/lint.md
JodyWinter: To answer your question, I don't think the Usage section on the current page is meant to show an example of all commands, just a few, including combos. I will list all of them in the one place in the new page and show single commands as well as combos.

Re your suggestion - will do.

docs/commands/lint.md
JodyWinter: I will reverse-engineer as discussed. Hopefully the second offering will be better.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+| --format | string | [Applies formatting to the output](#how-to-format-linting-results). Possible values: `codeframe`, `stylish`, `json`, `checkstyle`. Default: `codeframe`. |
+| --generate-ignore-file | boolean | [Generates an ignore file](#how-to-ignore-specific-errors-and-warnings) which tells Redocly CLI to ignore certain warnings/errors when validating the definition. |
+| --help | boolean | [Displays a short help menu](#how-to-open-the-help), showing lint options. |
+| --max-problems | integer | [Truncates the output](#how-to-limit-issues-shown-in-the-output) to show only a specified number of errors and warnings. Default: `100`. |
+| --skip-preprocessor | array | [Ignores specific preprocessors](#how-to-skip-a-rule-or-preprocessor-during-linting). |
+| --skip_rule | array | [Ignores specific custom or built-in rules](#how-to-skip-a-rule-or-preprocessor-during-linting). |
+| --version | boolean | Adding `--version` to any command (not just `lint`) will show your current Redocly CLI version. |
```
adamaltman, line 46: I prefer the original documentation here because this is a page about the lint command and "Show version number" is more to the point.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-`--skip-rule`            | `array`   | no           | -           | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--version`              | `boolean` | no           | -           | Show version number
+```bash Combined
+redocly lint [--max-problems=<n>] [--config=<path>] [--format=<value>]
+```
 
-## Examples
+| Option | Type | Description |
+|---|---|---|
+| entrypoints | array | Lets you [pass the entrypoints](#how-to-use-entrypoints-to-check-all-or-selected-definitions) (locations) of specific definition files in the command so `lint` only checks those files. |
```
adamaltman, line 37: issue with: "only check those files".  Lint will also check any files pointed to by reference object.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-#### Pass entrypoints directly
+##### Method 1: Pass entrypoints directly
+In this case, `lint` will validate the definitions passed in the command and ignore the `apis` section in `redocly.yaml` or any other customized config file you may be using.
 
 ```bash
-openapi lint openapi/openapi.yaml
+redocly lint openapi/openapi.yaml
 ```
 
-In this case, `lint` will validate the definition(s) passed to the command. The configuration file is ignored.
```
adamaltman, line 50: I didn't check if the configuration file is ignored. If it is, this is very important documentation.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+  sandbox:
+    root: ./sandbox.yaml
+    labels:
+      - sandbox
+    lint: []
+    features.openapi: []
 ```
 
-In this case, after resolving the path behind the `main` name (see the `Configuration file` tab), `lint` will validate the `definition.json` file. The presence of the Redocly configuration file is mandatory.
+ In this example, we only want to validate files under `main`:
```
adamaltman, line 91: It's not about "files" (I think this is the wrong word).

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+The `entrypoints` argument can also use any glob format supported by your file system. For example, `redocly lint ./root-documents/*.yaml`.
 
-The `entrypoints` argument can also use any glob format supported by your file system. For example, `openapi lint ./root-documents/*.yaml`.
+##### Method 2: Pass entrypoints via the config file
 
-#### Pass entrypoints via configuration file
+The `apis` section in the Redocly configuration file (`redocly.yaml`) lets you configure one or more APIs. Each file path is specified (using the `root` property), but files are identified by their name and version in the format `name@version`.
 
-Instead of full paths, you can use names listed in the `apis` section of your Redocly configuration file as entrypoints.
+`@version` is optional, and when not provided, Redocly apps interpret it as `latest` by default. Every `name@version` combination listed in the `apis` section must be unique.
```
adamaltman, line 65: I don't think this is appropriate for documentation because the version should be supplied. There is a lot of things people can do in an incomplete manner, but we don't call out explicitly that they can do it in an incomplete manner and it will still work (partially).

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-```bash Command
-openapi lint main
-```
+You can pass `name` or `name@version` in commands like `lint` and `stats` to act only on those APIs. Below is an example `apis` section showing three files identified using `main`, `production` and `sandbox`:
 
 ```yaml Configuration file
```
adamaltman, line 69: This file goes beyond a minimal viable example. It is also incorrect everywhere:

Wrong
```
lint: []
```

Correct
```
lint: {}
```

One is an array, and one is an object. The same issue with `features.openapi`.

Those shouldn't be fixed though. They should be removed here because it is a configuration of nothing. The minimum way to do the same thing is to remove the lines, making the file shorter and easier to digest.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-:::
+Redocly CLI has three rule sets, each specifying different levels of severity, that can be applied to API definition file validation by adding the `--extends` option to the `lint` command:
+
+* For you rebels out there, the `minimal` rule set will apply the least severe config for each built-in rule, and deactivate others entirely. We like to call this our death-metal rule set. View the GitHub Source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/minimal.ts).
+* Late 80s easy-listening more your jam? Then our `recommended` rule set is for you. It will apply the recommended config for each built-in rule. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/recommended.ts).
+* Got Gregorian chants playing on loop? The `all` rule set will satisfy your need for strictness. All rules are on, and `lint` will return an error if your API definition is anything less than perfect. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/e6f99e57da339d89bba7f4a1ba897282d7cebbd2/packages/core/src/config/all.ts).
```
adamaltman, line 127: This shouldn't be used.

JodyWinter: Do you mean you don't like the jovial language/tone I've used here, or that the entire list should be deleted? If the latter, the explanation of the `--extends` option will lack an overview of its allowed values (rule sets). Can you let me know what action you'd like me to take?

docs/commands/lint.md
JodyWinter: How about 'APIs' or 'API definitions' instead of 'files'?

docs/commands/lint.md
JodyWinter: OK I will reinstate it.

docs/commands/lint.md
JodyWinter: But what if there are no versions? Are we saying we expect users to always express `name@version` regardless? How would that go for my one version...?

myapi@v1

docs/commands/lint.md
JodyWinter: Would it be better if we just say...

Lets you pass the entrypoints (locations) of specific definition files in the command.

skadinna: That suggestion sounds good to me :heavy_plus_sign:

docs/commands/lint.md
skadinna: APIs is enough, I think

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
 
 ## Introduction
 
-Redocly OpenAPI CLI can identify and report on problems found in OpenAPI definitions. This helps you avoid bugs and make API definitions more consistent.
+One of the most powerful features of Redocly CLI is its ability to identify problems found in OpenAPI definition files. This is achieved with the `lint` command, which helps your definitions conform to a consistent set of rules.
```
skadinna, line 5:
```suggestion
One of the most powerful features of Redocly CLI is its ability to identify problems found in OpenAPI definitions. This is achieved with the `lint` command, which helps your definitions conform to a consistent set of rules.
```

skadinna: I think we can keep it simple; it's generally understood that those are files.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-```bash
-openapi lint
-openapi lint <entrypoints>...
-openapi lint [--max-problems=<n>] [--config=<path>] [--format=<value>]
-openapi lint [--generate-ignore-file] [--help]
-openapi lint --version
+Always the over-achiever, `lint` has many options which you can execute alone, or combine with others in a single command:
+
+```bash Alone
```
skadinna, line 17:
```suggestion
```bash Standalone
```

skadinna: suggestion: this might be a more common word in the context of CLI commands

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+| --format | string | [Applies formatting to the output](#how-to-format-linting-results). Possible values: `codeframe`, `stylish`, `json`, `checkstyle`. Default: `codeframe`. |
+| --generate-ignore-file | boolean | [Generates an ignore file](#how-to-ignore-specific-errors-and-warnings) which tells Redocly CLI to ignore certain warnings/errors when validating the definition. |
+| --help | boolean | [Displays a short help menu](#how-to-open-the-help), showing lint options. |
+| --max-problems | integer | [Truncates the output](#how-to-limit-issues-shown-in-the-output) to show only a specified number of errors and warnings. Default: `100`. |
+| --skip-preprocessor | array | [Ignores specific preprocessors](#how-to-skip-a-rule-or-preprocessor-during-linting). |
+| --skip_rule | array | [Ignores specific custom or built-in rules](#how-to-skip-a-rule-or-preprocessor-during-linting). |
+| --version | boolean | Adding `--version` to any command (not just `lint`) will show your current Redocly CLI version. |
 
-### Entrypoints
+## Usage examples
```
skadinna, line 48: I think this was originally my own suggestion :see_no_evil: but looking at it now, it's simpler to just say "Examples", which is what we do in all the other command docs.

skadinna, line 48:
```suggestion
## Examples
```

docs/commands/lint.md
skadinna: I think we don't want to encourage users to omit `version` because it can cause compatibility problems with the API registry. I don't know if you've tried this already, but when you're adding an API to the registry, it doesn't let you complete the action unless you specify the `version`. So in the light of that, it makes sense to sort of "steer" them in this happy-path direction where they just always specify the version and don't have to think about what would happen if they didn't have it.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+```bash Command
+redocly lint main
+```
 
-You can omit entrypoints completely when executing the `lint` command.
+##### Method 3: Pass no entrypoints
 
-```bash Command
-openapi lint
+Running `lint` without specifying a particular API definition file checks all definition files in your project directory.
```
skadinna, line 99: question: doesn't `lint` in this case validate all APIs listed in the conf file? I don't think it cares about the project directory because you could technically have an API definition located entirely outside of the project directory, but still have it listed in the `apis` section of the conf file. Maybe this needs to be tested?

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-    root: ./openapi/production.yaml
-  sandbox:
-    root: ./openapi/sandbox.yaml
+:::warning Important
+If you run `redocly lint` without entrypoints, and your project has no configuration files, you'll get an error.
+:::
+
+### How to reference a custom config file
+
+By default, Redocly CLI looks for the [Redocly configuration file](/docs/cli/configuration/configuration-file.mdx) in the current working or project directory. Using the optional `--config` argument provides an alternative path to a different configuration file if one exists.
```
skadinna, line 111:
```suggestion
By default, Redocly CLI looks for the [Redocly configuration file](/docs/cli/configuration/configuration-file.mdx) in the current working or project directory. With the optional `--config` argument, you can provide the path to a different configuration file.
```

skadinna: Suggesting this just to simplify the sentence

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+
+```bash
+redocly lint --config=./another/directory/anotherconfig.yaml
 ```
 
-In this case, if no API definitions are specified, `lint` validates all entrypoints listed under `apis` in your Redocly configuration file. The presence of the configuration file is mandatory.
+### How to activate rule sets
```
skadinna, line 117:
```suggestion
### How to activate rule configurations
```

skadinna: This is my mistake. I think I used this term, _rule sets_, to explain this functionality to you at some point. I also used it in other places in the docs. Later, Adam warned me not to use that term anymore, so I removed it but forgot to tell you. Sorry about that. I think we must use "rule configurations" instead, or just "configuration" when it's clear from the context we're talking about rules.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-:::warning Important
+Redocly CLI comes with [built-in rules](./resources/built-in-rules.md) that are used to validate API definitions. If you want to override a rule or reduce its severity, you can do so by modifying `redocly.yaml` (globally or per API definition).
 
-If you try to execute the `lint` command without entrypoints when your project doesn't have any configuration files, the `lint` command will display an error.
+An alternative approach is to use **rule sets** to quickly invoke a lot of rules at the same time.
```
skadinna, line 121:
```suggestion
An alternative approach is to use **rule configurations** to quickly invoke a lot of rules at the same time.
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-:::
+Redocly CLI has three rule sets, each specifying different levels of severity, that can be applied to API definition file validation by adding the `--extends` option to the `lint` command:
```
skadinna, line 123:
```suggestion
Redocly CLI has three built-in rule configurations, each specifying different levels of severity, that can be used to validate API definitions by adding the `--extends` option to the `lint` command:
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+
+* For you rebels out there, the `minimal` rule set will apply the least severe config for each built-in rule, and deactivate others entirely. We like to call this our death-metal rule set. View the GitHub Source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/minimal.ts).
```
skadinna, line 125:
```suggestion
* For you rebels out there, the `minimal` configuration will apply the least severe config for each built-in rule, and deactivate others entirely. We like to call this our death-metal rule set. View the GitHub Source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/minimal.ts).
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+* Late 80s easy-listening more your jam? Then our `recommended` rule set is for you. It will apply the recommended config for each built-in rule. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/recommended.ts).
```
skadinna, line 126:
```suggestion
* Late 80s easy-listening more your jam? Then our `recommended` configuration is for you. It will apply the recommended config for each built-in rule. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/ed997e2586e7adb7e32d8107cac79c452891f765/packages/core/src/config/recommended.ts).
```

docs/commands/lint.md
skadinna, line 127:
```suggestion
* Got Gregorian chants playing on loop? The `all` configuration will satisfy your need for strictness. All rules are on, and `lint` will return an error if your API definition is anything less than perfect. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/e6f99e57da339d89bba7f4a1ba897282d7cebbd2/packages/core/src/config/all.ts).
```

skadinna: I've no idea what Adam meant exactly, but I'm not a fan of these random musical references at all. They appear out of nowhere, without any precedent on this page. If you had used some kind of a music-related "theme" throughout the text, then it would work just fine. But like this, they're just distracting/fluff.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+* Got Gregorian chants playing on loop? The `all` rule set will satisfy your need for strictness. All rules are on, and `lint` will return an error if your API definition is anything less than perfect. View the GitHub source [here](https://github.com/Redocly/openapi-cli/blob/e6f99e57da339d89bba7f4a1ba897282d7cebbd2/packages/core/src/config/all.ts).
 
-### Custom configuration file
+When you compare the source files, you will see that some rules are switched on, some are set to different severities, and some are switched off. Here's how to activate each rule set when linting your API definitions:
```
skadinna, line 129:
```suggestion
When you compare the source files, you will see that some rules are switched on, some are set to different severities, and some are switched off. Here's how to activate each rule configuration when linting your API definitions:
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+
+### How to skip a rule or preprocessor during linting
+
+You can use `--skip-rule` to ignore any [custom rules](./resources/custom-rules.md) or [built-in rules](./resources/built-in-rules.md). In this example, we will skip the `spec` and `operation-4xx-response` built-in rules:
+
+```bash
+redocly lint openapi.yaml --skip-rule=spec --skip-rule=operation-4xx-response
+```
+
+Want to know how to skip a preprocessor? Because we don't recommend using preprocessors, you'll understand why we're choosing to skip that advice.
```
skadinna, line 379: suggestion: This is clever :smile: but I think we'll be better off not mentioning preprocessors at all. The only information we're giving the user here is that we don't recommend preprocessors. We don't say why, or what will happen if they use (or skip) them, or anything else of value. I'd suggest removing "preprocessor" from the heading and just make this section about skipping rules. Then you don't need this outro anymore.

skadinna (changes requested):
I've made some suggestions and minor changes, but otherwise this looks pretty good :+1: Excited to merge it soon :clap:

docs/commands/lint.md
adamaltman:
```suggestion
One of the most powerful features of Redocly CLI is its ability to identify problems found in OpenAPI definitions. This is achieved with the `lint` command, which keeps your API design consistent.
```

docs/commands/lint.md
adamaltman: I meant about the `all` word. It doesn't make sense to be used. It's intended for testing purposes.

docs/commands/lint.md
adamaltman: You are correct @skadinna.
```suggestion
Running `lint` without specifying a particular API definition checks all APIs in your configuration file.
```

docs/commands/lint.md
JodyWinter: Sorry Adam. I don't know what you mean or what action I should take here.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-### Entrypoints
+## Usage examples
 
-The `lint` command behaves differently depending on how you pass entrypoints to it and whether the [configuration file](#custom-configuration-file) exists.
+### How to use entrypoints to check all or selected definitions
 
-#### Pass entrypoints directly
+##### Method 1: Pass entrypoints directly
+In this case, `lint` will validate the definitions passed in the command and ignore the `apis` section in `redocly.yaml` or any other customized config file you may be using.
```
adamaltman, line 53:
```suggestion
In this case, `lint` will validate the definitions passed in the command and ignore the `apis` object in the Redocly configuration file.
```

docs/commands/lint.md
JodyWinter: Just my attempt at some humour... they weren't meant to link to anything else :) Might be best if I keep the tone more even from now on.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 ```
 
-In this case, `lint` will validate the definition(s) passed to the command. The configuration file is ignored.
+The `entrypoints` argument can also use any glob format supported by your file system. For example, `redocly lint ./root-documents/*.yaml`.
 
-The `entrypoints` argument can also use any glob format supported by your file system. For example, `openapi lint ./root-documents/*.yaml`.
+##### Method 2: Pass entrypoints via the config file
 
-#### Pass entrypoints via configuration file
+The `apis` section in the Redocly configuration file (`redocly.yaml`) lets you configure one or more APIs. Each file path is specified (using the `root` property), but files are identified by their name and version in the format `name@version`.
```
adamaltman, line 63:
```suggestion
The `apis` object in the Redocly configuration file (`redocly.yaml`) lets you configure one or more APIs. Each file path is specified (using the `root` property), but files are identified by their name and version in the format `name@version`.
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+`@version` is optional, and when not provided, Redocly apps interpret it as `latest` by default. Every `name@version` combination listed in the `apis` section must be unique.
 
-```bash Command
-openapi lint main
-```
+You can pass `name` or `name@version` in commands like `lint` and `stats` to act only on those APIs. Below is an example `apis` section showing three files identified using `main`, `production` and `sandbox`:
 
 ```yaml Configuration file
 apis:
   main:
```
adamaltman, line 71:
```suggestion
  core@latest:
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-    root: ./openapi/definition.json
+    root: ./openapi.yaml
+    labels:
+      - main
```
adamaltman, line 74:
```suggestion
      - core
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+    labels:
+      - production
+    lint: []
+    features.openapi: []
+  sandbox:
+    root: ./sandbox.yaml
+    labels:
+      - sandbox
+    lint: []
+    features.openapi: []
```
adamaltman, line 88:
```suggestion
  production@v1:
    root: ./production.yaml
  sandbox@v1:
    root: ./sandbox.yaml
```

docs/commands/lint.md
JodyWinter: Sorry Adam. I don't know what you mean or what action I should take here.

docs/commands/lint.md
adamaltman, line 91:
```suggestion
 In this example, we validate the `core@latest` API:
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 ```
 
-In this case, after resolving the path behind the `main` name (see the `Configuration file` tab), `lint` will validate the `definition.json` file. The presence of the Redocly configuration file is mandatory.
+ In this example, we only want to validate files under `main`:
 
-#### Empty entrypoints
+```bash Command
+redocly lint main
```
adamaltman, line 94:
```suggestion
redocly lint core@latest
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-apis:
-  main:
-    root: ./openapi/definition.json
-  production:
-    root: ./openapi/production.yaml
-  sandbox:
-    root: ./openapi/sandbox.yaml
+:::warning Important
+If you run `redocly lint` without entrypoints, and your project has no configuration files, you'll get an error.
+:::
```
adamaltman, line 107: This does not seem important. They will see the error which will tell them:

>  error: missing required argument `entrypoints`.

Do we need to document this?

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 ```
 
-In this case, if no API definitions are specified, `lint` validates all entrypoints listed under `apis` in your Redocly configuration file. The presence of the configuration file is mandatory.
+### How to activate rule sets
 
-:::warning Important
+Redocly CLI comes with [built-in rules](./resources/built-in-rules.md) that are used to validate API definitions. If you want to override a rule or reduce its severity, you can do so by modifying `redocly.yaml` (globally or per API definition).
 
-If you try to execute the `lint` command without entrypoints when your project doesn't have any configuration files, the `lint` command will display an error.
+An alternative approach is to use **rule configurations** to quickly invoke a lot of rules at the same time.
```
adamaltman, line 121:
```suggestion
An alternative approach is to use **rule configurations** to quickly configure a lot of rules at the same time.
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+
+```bash
+redocly lint --config=./another/directory/anotherconfig.yaml
 ```
 
-In this case, if no API definitions are specified, `lint` validates all entrypoints listed under `apis` in your Redocly configuration file. The presence of the configuration file is mandatory.
+### How to activate rule sets
 
-:::warning Important
+Redocly CLI comes with [built-in rules](./resources/built-in-rules.md) that are used to validate API definitions. If you want to override a rule or reduce its severity, you can do so by modifying `redocly.yaml` (globally or per API definition).
```
adamaltman, line 119:
```suggestion
Redocly CLI comes with [built-in rules](./resources/built-in-rules.md) that are used to validate API definitions.
```
That sentence feels misplaced. The built-in rules aren't enabled automatically. The user can enable or disable them in their configuration file.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+redocly lint --config=<path>
+redocly lint --extends=<value>
+redocly lint --format=<value>
+redocly lint --generate-ignore-file
+redocly lint --help
+redocly lint --max-problems=<n>
+redocly lint --skip-preprocessor=<preprocessor>
+redocly lint --skip-rule=<rule>,<rule>
+redocly lint --version
 ```
```
adamaltman, line 29: Is this part necessary? There is a table with options below. And then examples further down below.

docs/commands/lint.md
adamaltman:
```suggestion
| entrypoints | array | [Sets the entrypoints](#how-to-use-entrypoints-to-check-all-or-selected-definitions) (locations) of specific definition files to lint. If no entrypoints are provided, all the APIs defined in the Redocly configuration file are linted. |
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+| --extends | array | [Extends the existing config](#how-to-activate-rule-sets) by invoking rule sets that apply lots of rules at once. Possible values: `minimal`, `recommended`, `all`. Default: `recommended`. |
+| --format | string | [Applies formatting to the output](#how-to-format-linting-results). Possible values: `codeframe`, `stylish`, `json`, `checkstyle`. Default: `codeframe`. |
+| --generate-ignore-file | boolean | [Generates an ignore file](#how-to-ignore-specific-errors-and-warnings) which tells Redocly CLI to ignore certain warnings/errors when validating the definition. |
+| --help | boolean | [Displays a short help menu](#how-to-open-the-help), showing lint options. |
+| --max-problems | integer | [Truncates the output](#how-to-limit-issues-shown-in-the-output) to show only a specified number of errors and warnings. Default: `100`. |
+| --skip-preprocessor | array | [Ignores specific preprocessors](#how-to-skip-a-rule-or-preprocessor-during-linting). |
+| --skip_rule | array | [Ignores specific custom or built-in rules](#how-to-skip-a-rule-or-preprocessor-during-linting). |
```
adamaltman, line 45:
```suggestion
| --skip-rule | array | [Ignores specific custom or built-in rules](#how-to-skip-a-rule-or-preprocessor-during-linting). |
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+| --version | boolean | Adding `--version` to any command (not just `lint`) will show your current Redocly CLI version. |
 
 ## Examples
 
-### Entrypoints
-
-The `lint` command behaves differently depending on how you pass entrypoints to it and whether the [configuration file](#custom-configuration-file) exists.
+### How to use entrypoints to check all or selected definitions
```
adamaltman, line 50:
```suggestion
### How to use entrypoints to lint all or selected definitions
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-#### Pass entrypoints via configuration file
+The `apis` object in the Redocly configuration file (`redocly.yaml`) lets you configure one or more APIs. Each file path is specified (using the `root` property), but files are identified by their name and version in the format `name@version`.
 
-Instead of full paths, you can use names listed in the `apis` section of your Redocly configuration file as entrypoints.
+`@version` is optional, and when not provided, Redocly apps interpret it as `latest` by default. Every `name@version` combination listed in the `apis` section must be unique.
 
-```bash Command
-openapi lint main
-```
+You can pass `name` or `name@version` in commands like `lint` and `stats` to act only on those APIs. Below is an example `apis` section showing three files identified using `main`, `production` and `sandbox`:
```
adamaltman, line 67:
```suggestion
You can pass `name@version` in commands like `lint` and `stats` to act only on those APIs. Below is an example `apis` object showing three files identified using `core@latest`, `production@v1` and `sandbox@v1`:
```

docs/commands/lint.md
adamaltman:
```suggestion
 In this example, we only want to validate the `core@latest` API:
```

docs/commands/lint.md
adamaltman: We need to check if this sentence is correct. If it is, it needs to be in the docs.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-`--generate-ignore-file` | `boolean` | no           | -           | [Generate ignore file](#generate-ignore-file)
-`--help`                 | `boolean` | no           | -           | Show help
-`--max-problems`         | `number`  | no           | 100         | Truncate output to display the specified [maximum number of problems](#max-problems)
-`--skip-preprocessor`    | `array`   | no           | -           | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--skip-rule`            | `array`   | no           | -           | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--version`              | `boolean` | no           | -           | Show version number
+| Option | Type | Description |
+|---|---|---|
+| entrypoints | array | Lets you [pass the entrypoints](#how-to-use-entrypoints-to-check-all-or-selected-definitions) (locations) of specific definition files in the command so `lint` only checks those files. |
+| --config | string | Lets you [specify a path to a custom config file](#how-to-reference-a-custom-config-file). |
```
adamaltman, line 38: I'm not sure "Lets you" makes sense in the context of CLI options.

I'm used to a more declarative style:

https://awscli.amazonaws.com/v2/documentation/api/latest/reference/index.html
https://cli.github.com/manual/gh_auth_login
https://curl.se/docs/manpage.html
https://git-scm.com/docs/git-switch

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
+| --extends | array | [Extends the existing config](#how-to-activate-rule-sets) by invoking rule sets that apply lots of rules at once. Possible values: `minimal`, `recommended`, `all`. Default: `recommended`. |
```
adamaltman, line 39:
```suggestion
| --extends | array | [Extends the existing configuration](#how-to-activate-rule-sets). |
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
-`--extends`              | `array`   | no           | -           | [Extend a specific configuration](#extend-configuration) (defaults or config file settings)
-`--format`               | `string`  | no           | `codeframe` | Format for the output.<br />**Possible values:** `codeframe`, `stylish`, `json`, `checkstyle`
-`--generate-ignore-file` | `boolean` | no           | -           | [Generate ignore file](#generate-ignore-file)
-`--help`                 | `boolean` | no           | -           | Show help
-`--max-problems`         | `number`  | no           | 100         | Truncate output to display the specified [maximum number of problems](#max-problems)
-`--skip-preprocessor`    | `array`   | no           | -           | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--skip-rule`            | `array`   | no           | -           | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below
-`--version`              | `boolean` | no           | -           | Show version number
+| Option | Type | Description |
```
adamaltman, line 35: This matched the actual tool's help before. Now it doesn't.  This is also outdated in some places. I would recommend parity between the command's help and this content. A ticket can be opened to adjust the built in help, but I believe it should be in the standard declarative style all CLI tools I use adhere to.

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
 
 ## Introduction
 
-Redocly OpenAPI CLI can identify and report on problems found in OpenAPI definitions. This helps you avoid bugs and make API definitions more consistent.
+One of the most powerful features of Redocly CLI is its ability to identify problems found in OpenAPI definitions. This is achieved with the `lint` command, which helps your definitions conform to a consistent set of rules.
```
adamaltman, line 5:
```suggestion
One of the most powerful features of Redocly CLI is its ability to lint OpenAPI definitions.
This helps your definitions conform to a consistent API styleguide.
```

docs/commands/lint.md
```diff
@@ -2,239 +2,378 @@
...
 
-The `lint` command reports on problems and executes preprocessors and rules. Unlike the `bundle` command, `lint` doesn't execute decorators.
+The `lint` command ensures that your definition file structure is valid (according to the OpenAPI Specification and your own custom rules, if any) and contains no errors. It also executes preprocessors and rules.
```
adamaltman, line 7: This isn't really correct. 
```suggestion
The `lint` command ensures that your definition file structure is valid (according to the OpenAPI Specification) and adheres to your API styleguide.
```

docs/commands/lint.md
JodyWinter: It can't hurt to leave it in.

docs/commands/lint.md
JodyWinter: I used the word 'invoke' here because in this context, isn't the user using `--extends` to trigger pre-configured rule sets?

docs/commands/lint.md
JodyWinter: I can get rid of it.

docs/commands/lint.md
JodyWinter: OK, I'll stick with the original table columns.

docs/commands/lint.md
JodyWinter: I have several api definition files in my project directory and all are listed in my `redocly.yaml` file under the `apis` object. When I run this command, only the specified definition file is linted, all others are ignored. So I have concluded that this statement is correct.

adamaltman: I read this in a different way.

If I have a lint configuration, and I run the lint command on an API using the filename instead of the API name then the lint configurations are ignored and it will use the default lint configuration. I haven't tried to verify what the behavior is though.

JodyWinter: I ran the command first using the API name then the filename, and both validated:

```
C:\Users\jodyw\Filing Cabinet\Git\redocly-openapi-starter\openapi>redocly lint production@v1
No configurations were defined in extends -- using built in recommended configuration by default.

validating ./production.yaml...
./production.yaml: validated in 40ms

```

```
C:\Users\jodyw\Filing Cabinet\Git\redocly-openapi-starter\openapi>redocly lint production.yaml
No configurations were defined in extends -- using built in recommended configuration by default.

validating production.yaml...
production.yaml: validated in 38ms

```

Should our statement in fact read:

In this case, `lint` will only validate the API definition(s) passed in the command (you can use either the API name or filename). For example:

`redocly lint core@v1`
` redocly lint openapi.yaml`

lornajane:
This looks really good but quite outdated, I'll review as I update the docs.

Issue #648, opened 2022-04-18
`ref` assertions


adamaltman:
**Is your feature request related to a problem? Please describe.**

I want to be able to detect if a particular type of thing is defined using a `$ref`, and in particular a `$ref` to a separate file or url.

**Describe the solution you'd like**

Option A is flatter and has two data types (which is a bit unorthodox).

Option B requires a nested object, which is a different type of complexity.

Which is the less of the two evils?

## Option A

Add to the `Assertion Object`:
| Property | Type | Description |
|---|---|---|
|`ref`| string \| boolean | Evaluates the subject's property. A boolean value of `true` means the property has a `$ref` defined. A boolean value of `false` means the property has not defined a `$ref` (it has an in-place value). A string value means that the `$ref` is defined and the unresolved value must match the pattern (for example, `'/paths\/.*\.yaml$/'`).    | 

Example with pattern

```yaml
subject: PathItem
ref:  '/paths\/.*\.yaml$/'
```
Example with defined
```yaml
subject: PathItem
ref: true
```

## Option B

Add to the `Assertion Object`:
| Property | Type | Description |
|---|---|---|
|`ref`|Ref assertion object|Asserts that the subject property meets the corresponding ref assertion.| 

### Ref assertion object

| Property | Type | Description |
|---|---|---|
|undefined|boolean|Set to true means there is not a `$ref` defined for the subject property.|
|pattern|string|Pattern checked against the unresolved `$ref` value. For example, `'/paths\/.*\.yaml$/'`.|

Example with pattern

```yaml
subject: PathItem
ref:
  pattern: '/paths\/.*\.yaml$/'
```
Example with defined
```yaml
subject: PathItem
ref:
  undefined: false
```

**Describe alternatives you've considered**

Multiple boolean properties like `isRef`, `isFileRef`, `isUrlRef`.

**Additional context**

To enforce that PathItems are made using `$ref` to a separate file. To enforce that request and response schemas are made using a `$ref` to a separate "any" `$ref` schema.

RomanHotsiy:
I like A) more in terms of UX. The B) is more flexible though if we decide to add other asserts to the ref string.
But I lean towards A) though.

PR #675, opened 2022-05-06
feat: ref assertion

And678:
## What/Why/How?
Added a new assertation that checks if ref is present.

Add to the `Assertion Object`:
| Property | Type | Description |
|---|---|---|
|`ref`| string \| boolean | Evaluates the subject's property. A boolean value of `true` means the property has a `$ref` defined. A boolean value of `false` means the property has not defined a `$ref` (it has an in-place value). A string value means that the `$ref` is defined and the unresolved value must match the pattern (for example, `'/paths\/.*\.yaml$/'`).    | 

Example with pattern

```yaml
subject: PathItem
ref:  '/paths\/.*\.yaml$/'
```
Example with defined
```yaml
subject: PathItem
ref: true
```

**Also**, added a small related fix to _pattern_ assert, because it failed on complicated regexes (e.g `/^(\.\/)?components\/.*\.yaml$/`)
## Reference
#648 
## Testing
Added unit tests
## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/common/assertions/asserts.ts
```diff
@@ -24,19 +25,16 @@ export const runOnValuesSet = new Set([
...
     const values = typeof value === 'string' ? [value] : value;
-    const regexOptions = condition.match(/(\b\/\b)(.+)/g) || ['/'];
-    condition = condition.slice(1).replace(regexOptions[0], '');
-    const regx = new RegExp(condition, regexOptions[0].slice(1));
+    const regx = regexFromString(condition);
     for (let _val of values) {
-      if (!_val.match(regx)) {
-        return false;
-      }
+      return regx?.test(_val) || false;
```
tatomyr, line 37: This will always return with the first item. Shouldn't we preserve the conditional returning, should we?

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٣:١٠ م، كتب Adam Altman ***@***.***>:

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -165,3 +166,8 @@ function runAssertion(
     });
   }
 }
+
+export function regexFromString(input: string): RegExp | null {
+  const matches = input.match(/^\/(.*)\/(.*)|(.*)/);
+  return matches && new RegExp(matches[1] || matches[3], matches[2]);
+}
```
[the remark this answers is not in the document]
And678: My bad, added it

packages/core/src/rules/common/assertions/asserts.ts
And678: Fixed it, thanks

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٤:١٥ م، كتب Andriy Kaminskyy <
***@***.***>:

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٤:٣٩ م، كتب minalabib223 ***@***.***

docs/resources/rules/assertions.md
```diff
@@ -49,6 +49,8 @@ undefined | `boolean` | Asserts a property is undefined. See [undefined example]
 nonEmpty | `boolean` | Asserts a property is not empty. See [nonEmpty example](#nonempty-example).
 minLength | `integer` | Asserts a minimum length (inclusive) of a string or list (array). See [minLength example](#minlength-example).
 maxLength | `integer` | Asserts a maximum length (exclusive) of a string or list (array). See [maxLength example](#maxlength-example).
+ref | `boolean | string` | Asserts a reference object presence in object's property. A boolean value of `true` means the property has a `$ref` defined. A boolean value of `false` means the property has not defined a `$ref` (it has an in-place value). A string value means that the `$ref` is defined and the unresolved value must match the pattern (for example, `'/paths\/.*\.yaml$/'`). See [ref example](#ref-exmaple).|
```
adamaltman, line 52:
```suggestion
ref | `boolean | string` | Asserts a reference object presence in object's property. A boolean value of `true` means the property has a `$ref` defined. A boolean value of `false` means the property has not defined a `$ref` (it has an in-place value). A string value means that the `$ref` is defined and the unresolved value must match the pattern (for example, `'/paths\/.*\.yaml$/'`). See [ref example](#ref-example).| 
```

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -165,3 +166,8 @@ function runAssertion(
     });
   }
 }
+
+export function regexFromString(input: string): RegExp | null {
+  const matches = input.match(/^\/(.*)\/(.*)|(.*)/);
```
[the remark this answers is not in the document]
adamaltman: Thank you

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٥:٤٤ م، كتب Adam Altman ***@***.***>:

minalabib223:
في جمعة، ٦ مايو، ٢٠٢٢ في ٥:٥٣ م، كتب minalabib223 ***@***.***

packages/core/src/rules/common/assertions/asserts.ts
```diff
@@ -134,4 +134,13 @@ export const asserts: Asserts = {
       ? getIntersectionLength(value, condition) === condition.length
       : true;
   },
+  ref: (_value: any, condition: string | boolean, rawValue: any): boolean => {
```
RomanHotsiy, line 137: I think this is not going to work correctly in terms it won't use `context` or `subeject` settings.

Try to add an end-to-end test when you have multiple references (PathItem and Parameter for example) but you want only some specific type of reference to be asserted.

RomanHotsiy (changes requested):
It seems to be not working appropriately from what I can see in the code.

I left a comment. Please test it out plz. I can help you to update the code so ping me next week plz.

packages/core/src/rules/common/assertions/asserts.ts
adamaltman: @And678 were you able to test it? Also #662 is likely to merge in the morning, and I don't know if it will impact your PR.

And678: @adamaltman Yes, managed to test it, fixing final issue, probably will release on monday morning

adamaltman: Thanks for the update! 🎉

RomanHotsiy: @And678 I just double-checked the code. Sorry, I thought you were using `ref` visitor but this is a ref assertion 🤦 

Disregard my comment above.

packages/core/src/rules/common/assertions/utils.ts
```diff
@@ -105,10 +105,10 @@ export function buildSubjectVisitor(
         for (const property of properties) {
           // we can have resolvable scalar so need to resolve value here.
           const value = isRef(node[property]) ? resolve(node[property])?.node : node[property];
-          runAssertion(value, assert, location.child(property), report);
+          runAssertion(value, node[property], assert, location.child(property), report);
         }
       } else {
-        runAssertion(Object.keys(node), assert, location.key(), report);
+        runAssertion(Object.keys(node), Object.keys(node), assert, location.key(), report);
```
RomanHotsiy, line 111: should it use Object.values?

packages/core/src/rules/common/assertions/asserts.ts
RomanHotsiy: But, please, add e2e test. I can't see any. Thanks.

RomanHotsiy: I just tested it. It still doesn't work correctly even with the example from the PR description.

It always reports $ref to be missing (even if it's there). And if you try `ref: false` it never reports at all.

Please, add positive and negative e2e tests.

RomanHotsiy (changes requested):
Still doesn't work correctly.

docs/resources/rules/assertions.md
And678: ✅

packages/core/src/rules/common/assertions/utils.ts
And678: I believe it shouldn't, because we use run assertions on keys when property is not specified, no assertions expects an array of values

And678:
@RomanHotsiy Added e2e tests (for true, false and regex asserts), could you check them, the functionality seems to work there.
Can you send your .redocly.yaml/openapi.yaml that doesn't work with this assertion?
Thanks

RomanHotsiy:
@And678 I tried case from this PR description and it doesn't work. Add tests for the case when there is no `property`:

image

__tests__/lint/assertions-ref-forbidden-error/snapshot.js
```diff
@@ -0,0 +1,41 @@
...
+49 |   $ref: components/schemas/Owners.yaml
+
+Error was generated by the ref-forbidden assertion rule.
+
+
+[2] components/schemas/Owners.yaml:1:1 at #/
+
+PathItems should NOT should have a ref
+
+ 1 | get:
```
RomanHotsiy, line 24: The report location is wrong. Should highlight $ref itself.

RomanHotsiy: Maybe we need to add unresolved location additionally into the context.

__tests__/lint/assertions-ref-pattern-error/.redocly.yaml
```diff
@@ -0,0 +1,18 @@
...
+      context:
+        - type: Response
+      subject: MediaType
+      property: schema
+      message: Response MediaType schema should contain ref to components folder
+      ref: ^(\.\/)?components\/schema\/.*\.yaml$
+    assert/ref-pattern-no-properties:
+      subject: PathItem
+      message: PathItem should contain ref to components folder
+      ref: ^(\.\/)?components\/schema\/.*\.yaml$
```
RomanHotsiy, line 17: I would change it to components/paths (so the test makes sense)

__tests__/lint/assertions-ref-pattern-error/components/schemas/Owners.yaml
```diff
@@ -0,0 +1,27 @@
+get:
```
RomanHotsiy, line 1: I would move those files out of `components/schemas` folder into `components/pathItems`.

__tests__/lint/assertions-ref-pattern-error/snapshot.js
```diff
@@ -0,0 +1,116 @@
...
+
+
+[2] openapi.yaml:45:17 at #/paths/~1pets/get/responses/default/content/application~1json/schema
+
+Response MediaType schema should contain ref to components folder
+
+43 |           application/json:
+44 |             schema:
+45 |               $ref: ./components/schemas/Error.yaml
+   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
RomanHotsiy, line 27: This looks like schema does contain ref to the components folder. Why error here?

__tests__/lint/assertions-ref-pattern-error/snapshot.js
```diff
@@ -0,0 +1,116 @@
...
+Error was generated by the ref-pattern-no-properties assertion rule.
+
+
+[5] components/schemas/Owners.yaml:25:13 at #/get/responses/200/content/application~1json/schema
+
+Response MediaType schema should contain ref to components folder
+
+23 | application/json:
+24 |   schema:
+25 |     properties:
```
RomanHotsiy, line 69: Let's highlight `schema` word, not whole object.

RomanHotsiy (changes requested):

adamaltman:
@And678 how's it going? When do you think this will be ready to review again?

And678:
@adamaltman Doing it today, will be ready for review soon, didn't do it yesterday because i had to be ooo

tatomyr:
I've configured reporting on key/values from inside the `asserts` object methods depending on the context. 
Please look at the updated snapshots and examine whether the highlightings are proper.

__tests__/lint/assertions-casing-camel-case-error/snapshot.js
```diff
@@ -3,7 +3,7 @@
 exports[`E2E lint assertions-casing-camel-case-error 1`] = `
 
 validating /openapi.yaml...
-[1] openapi.yaml:19:20 at #/paths/~1pet~1findByStatus/get/operationId
+[1] openapi.yaml:19:20 at #/paths/~1pet~1findByStatus/get/operationId/GetExampleId
```
RomanHotsiy, line 6: Something wrong her... This pointer seems invalid.

RomanHotsiy: There are many places like this and all of them seem to have extra segment at the end.

RomanHotsiy (changes requested):
Great approach! 👏 

I left a few small comments and there is one more small issue with location.

__tests__/lint/assertions-casing-camel-case-error/snapshot.js
tatomyr: I see. Not sure where does it come from though. Will take a look.

And678 (commented):
💪

__tests__/lint/assertions-casing-camel-case-error/snapshot.js
tatomyr: Updated the snapshots. Please take a look.

ohorbachevskyi (approved):
overall lgtm 🚢

docs/resources/rules/assertions.md
```diff
@@ -435,7 +436,35 @@ lint:
...
+
+### `ref` example
+
+The following example asserts that schema in MediaType contains a Reference object ($ref).
+
+```yaml
+lint:
+  rules:
+    assert/mediatype-schema-has-ref:
+      - subject: MediaType
```
adamaltman, line 448: The `-` is wrong. It's not an array here.

PR #688, opened 2022-05-16
add bundle option to keep url references

pvonr:
## What/Why/How?
By default bundling resolves all references into a single file. With the new `keep-url-references` flag we can keep external url references intact.

This is useful to keep 'global' references intact that are coming from e.g. an internal registry. With this option we can basically bundle multiple files inside a repo into a single file without loosing trace of external dependencies.

## Testing
New test is added to test the functionality.

## Check yourself

- [ ] Code is linted

## Security


pvonr:
Note: I didn't run `npm run prettier` as the command was changing many untouched files. Happy to do this if you think these changes should go into this PR as well. I would prefer a separate PR to bring the repo to compliance with the prettier rules.

RomanHotsiy (approved):

packages/core/src/bundle.ts
```diff
@@ -224,6 +235,10 @@ function makeBundleVisitor(
           return;
         }
 
+        if (keepUrlRefs && isUrlRef(node.$ref)) {
```
tatomyr, line 238: Could you use `isAbsoluteUrl` from **packages/core/src/ref-utils.ts**? You may probably adjust it to be case insensitive.

pvonr: Thanks for pointing me to this function. Question: it seems you assume lower case `http(s)://` prefixes for absolute urls in your codebase (also in the `isMappingRef` in **packages/core/src/ref-utils.ts**). I'm fine keeping the `isAbsoluteRef` as is and dropping the case-insensitive check for consistency. Wdyt?

tatomyr: I think it should be case insensitive all over the codebase, but I'd rather do this a separate merge request.

tatomyr (approved):

packages/core/src/bundle.ts
adamaltman: If we open an issue, please be aware: Only the scheme and host are case insensitive. The remaining part (paths, parameters, etc) of the URL are case sensitive.

https://www.rfc-editor.org/rfc/rfc3986

> [6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.2.1).  Case Normalization
> 
>    For all URIs, the hexadecimal digits within a percent-encoding
>    triplet (e.g., "%3a" versus "%3A") are case-insensitive and therefore
>    should be normalized to use uppercase letters for the digits A-F.
> 
>    When a URI uses components of the generic syntax, the component
>    syntax equivalence rules always apply; namely, that the scheme and
>    host are case-insensitive and therefore should be normalized to
>    lowercase.  For example, the URI <HTTP://www.EXAMPLE.com/> is
>    equivalent to <http://www.example.com/>.  The other generic syntax
>    components are assumed to be case-sensitive unless specifically
>    defined otherwise by the scheme (see [Section 6.2.3](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.3)).
>

packages/cli/src/commands/bundle.ts
```diff
@@ -98,6 +99,7 @@ export async function handleBundle(
         ref: path,
         dereference: argv.dereferenced,
         removeUnusedComponents,
+        keepUrlRefs: argv['keep-url-references'],
```
adamaltman, line 102: Two questions I have: should it be `keep-url-references` or `keep-url-refs`? 

What if someone should want to keep other refs such as file refs? Will it require another option? Or should the option not be boolean?

tatomyr: What the bundle will it be if we keep file $refs? :) 
Regarding the name, I'm ok with the current one, it is clear enough for me.

adamaltman: Okay, that makes sense. I wonder though about references vs refs in the option name.

adamaltman:
https://twitter.com/adamaltman/status/1527261391446581250

PR #692, opened 2022-05-18
fix: operation-security-defined warn -> error

Oprysk:
## What/Why/How?
fix: Empty dropdown is shown in case security scheme is not defined [#24](https://github.com/Redocly/reference-docs/issues/24)
## Check yourself

- [ ] Code is linted

## Security


tatomyr (approved):

adamaltman (changes requested):
I don't think this should be an `error` in the `minimal` config. Some people don't have security on some APIs, and it seems that is okay.

packages/core/src/config/minimal.ts
```diff
@@ -24,7 +24,7 @@ export default {
     'operation-operationId-unique': 'warn',
     'operation-parameters-unique': 'warn',
     'operation-tag-defined': 'off',
-    'operation-security-defined': 'warn',
+    'operation-security-defined': 'error',
```
RomanHotsiy, line 27: This is a breaking change. We can't merge it so easily.

PR #694, opened 2022-05-19
chore: add missing array types for 3.1

AntonKozachuk:
## What/Why/How?
Add missing Array props to type tree: https://json-schema.org/understanding-json-schema/reference/array.html

## Testing
N/a

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security


packages/core/src/types/oas3_1.ts
```diff
@@ -127,6 +127,8 @@ const Schema: NodeType = {
     dependentSchemas: listOf('Schema'),
     prefixItems: listOf('Schema'),
     contains: 'Schema',
+    minContains: { type: 'number' },
```
adamaltman, line 130: Is there an integer type?

AntonKozachuk: @adamaltman looks like we used it as type in 2.0, but don't have it starting from 3.0. Maybe @RomanHotsiy has more context regarding this.

RomanHotsiy: We should use `integer` and fix 3.0 schema usage too.

AntonKozachuk: @RomanHotsiy I rechecked and looks like the integer type in 2.0 was used according to the spec. And in 3.0 and older there seem no fields with type integer, besides the `type` itself.

Example:
image

RomanHotsiy: OAS3 and later Schema type is covered by JsonSchema. See here: https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.3.1

RomanHotsiy (approved):
LGTM 🚢

tatomyr (approved):

PR #703, opened 2022-05-24
fix: escape symbols in the samples lang name

### LINKED ISSUE #685: # symbol is not properly encoded while forming the $ref path
***Describe the issue**

 The # symbol in C# code samples is not encoded properly if we run `openapi bundle` command.

**To Reproduce**

1. Add C# `x-codeSamples` to your definition file - `- lang: C#`
2. Run `openapi split` command on your definition
3. Run `openapi bundle` command after that
4. Notice the error messages:
Screenshot 2022-05-13 at 18 17 35

**Expected behavior**
(#) symbol should be properly encoded while forming the $ref path. As a workaround, we can replace `-lang: C#` to `-lang: C-sharp` for example, and it will solve the problem

**OpenAPI definition**

openapi: 3.0.0

What version of Redocly CLI are you using?

1.0.0-beta.94

**`Node.js` Version(s)**

v16.13.1

AntonKozachuk:
## What/Why/How?
Resolve an issue with the # symbol in C# code samples is not encoded properly if we run `openapi split` command.

## Reference
Fix #685

## Testing
Checked with several lang names, including C#.

## Check yourself

- [x] Code is linted

## Security


tatomyr:
Thanks. Could you add a few test cases?

packages/cli/src/commands/split/__tests__/index.test.ts
```diff
@@ -19,6 +19,7 @@ jest.mock('@redocly/openapi-core', () => ({
 describe('#split', () => {
   const openapiDir = 'test';
   const componentsFiles: ComponentsFiles = {};
+  const utils = require('../../../utils');
```
tatomyr, line 22: Why import it here? Could it be imported on the top?

tatomyr (approved):

PR #717, opened 2022-05-27
fix: allow scalar property with falsy values

AntonKozachuk:
## What/Why/How?
`scalar-property-missing-example` rule was reporting an error for falsy values like '', 0, false in the example.

## Reference
N/a

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/common/scalar-property-missing-example.ts
+++ b/packages/core/src/rules/common/scalar-property-missing-example.ts
@@ -36,7 +36,7 @@ export const ScalarPropertyMissingExample: Oas3Rule | Oas2Rule = () => {
 };
 
 function isNotDefined(value: unknown): boolean {
-  return value === null || value === undefined;
+  return value === undefined;
 }
 
 function isScalarSchema(schema: Oas2Schema | Oas3Schema | Oas3_1Schema) {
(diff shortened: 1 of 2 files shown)


AntonKozachuk:
@RomanHotsiy one concern is that an example can sometimes have a `null` value for nullable properties. What do you think about this?

tatomyr (approved):

packages/core/src/rules/common/scalar-property-missing-example.ts
```diff
@@ -32,6 +35,10 @@ export const ScalarPropertyMissingExample: Oas3Rule | Oas2Rule = () => {
   };
 };
 
+function isNotDefined(value: unknown): boolean {
+  return value === null || value === undefined;
```
RomanHotsiy, line 39: null is a valid example though.

RomanHotsiy (changes requested):

packages/core/src/rules/common/scalar-property-missing-example.ts
```diff
@@ -32,6 +35,10 @@ export const ScalarPropertyMissingExample: Oas3Rule | Oas2Rule = () => {
   };
 };
 
+function isNotDefined(value: unknown): boolean {
```
RomanHotsiy, line 38: can you move it to some utils? We have somewhere validateDefinedAndNotEmpty

tatomyr: Maybe drop this function and simply compare values inline? It doesn't bring much sense to me.

AntonKozachuk: validateDefinedAndNotEmpty won't work here. 
I just got rid of the func at all.

RomanHotsiy (approved):

Issue #617, opened 2022-04-05
Check errors in the configuration file


AlexVarchuk:
**Is your feature request related to a problem? Please describe.**
For improving stable work with the configuration file we need to add supporting linting(check errors) for configuration files(.redocly.yaml)

**Describe the solution you'd like**
when running the lint command with the configuration file, check the file of configuration and display the result(errors, warning, etc).

**Describe alternatives you've considered**
Add new command for linting conf file. But seems it is a not good idea.

adamaltman:
I thought we already did this 🤣

tatomyr:
Rising priority as it is a dependency for a [hight priority issue](https://github.com/Redocly/redocly-cli/issues/701).

PR #720, opened 2022-05-30
feat: check errors in the configuration file

roman-sainchuk:
## What/Why/How?
Closes https://github.com/Redocly/redocly-cli/issues/617
## Check yourself

- [ ] Code is linted

## Security


packages/cli/src/utils.ts
```diff
@@ -230,6 +234,33 @@ export function printLintTotals(totals: Totals, definitionsCount: number) {
...
+      green(`Your lint config is valid. 🎉\n`),
+    );
+    process.stdout.write(
+      yellow(`You have ${totals.warnings} ${pluralize('warning', totals.warnings)}.\n`),
+    );
+  } else {
+    process.stdout.write(
+      green(`Your lint config is valid. 🎉\n`),
+    );
+  }
```
tatomyr, line 259: Can we avoid code duplication here?

packages/cli/src/index.ts
```diff
@@ -149,6 +149,10 @@ yargs
           array: true,
           type: 'string',
         },
+        'skip-config-lint': {
```
tatomyr, line 152: I wouldn't introduce this flag. I think it's better to run lint config with `warn` severity in order not to break anything for customers.

RomanHotsiy:
Can you post a screenshot of the output?

AntonKozachuk:
> Can you post a screenshot of the tool output?

@RomanHotsiy looks like this at the moment: 
image

Config file:
https://github.com/Redocly/redocly-cli/blob/master/resources/.redocly.yaml

packages/cli/src/index.ts
roman-sainchuk: This flag just skips lint of the config and in case you dont like your CLI output to be polluted with config lint result - you can skip it

roman-sainchuk: Not sure I understand how it could break anything, because config lint doesnt interrupt definition lint in case of errors

packages/cli/src/commands/lint.ts
```diff
@@ -3,36 +3,54 @@ import {
...
+    await validateLintConfig({
+      configContent,
+      configPath,
+      outputFormat: argv.format,
+      maxProblems: argv['max-problems'],
+      version,
+    });
+  }
+
+  const config: Config = await getConfigFromString(configContent, configPath, argv.extends);
```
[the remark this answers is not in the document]
roman-sainchuk: To me it looks not good, because `loadConfig` should do the only thing it is supposed to do. And most probably there are libraries that import `loadConfig` which expect to get config and not to have it linted (and probably thrown lint results into console)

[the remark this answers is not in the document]
tatomyr: We may want to use config linting in other places where we load config. To prevent unwanted logs, we may pass an optional third parameter into `loadConfig` to indicate whether to lint config while loading.

packages/cli/src/index.ts
tatomyr: Ah, I thought you introduced this option because config linting may throw unwanted errors. For this cases, it may sense to use this flag.

packages/cli/src/utils.ts
tatomyr: Maybe nested `if`s would work?

```ts
  } else {
    process.stdout.write(
      green(`Your lint config is valid. 🎉\n`)
    );
    if (totals.warnings > 0) process.stdout.write(
      yellow(`You have ${totals.warnings} ${pluralize('warning', totals.warnings)}.\n`)
    );
  }
```

roman-sainchuk: indeed, thanks!

packages/cli/src/utils.ts
```diff
@@ -230,6 +234,33 @@ export function printLintTotals(totals: Totals, definitionsCount: number) {
   process.stderr.write('\n');
 }
 
+export function printConfigLintTotals(totals: Totals): void {
+  if (totals.errors > 0) {
+    process.stderr.write(
+      red(
+        `❌ Your lint config has ${totals.errors} ${pluralize('error', totals.errors)}${
```
adamaltman, line 241:
```suggestion
        `❌ Your config has ${totals.errors} ${pluralize('error', totals.errors)}${
```

packages/cli/src/utils.ts
```diff
@@ -230,6 +234,33 @@ export function printLintTotals(totals: Totals, definitionsCount: number) {
...
+          totals.warnings > 0
+            ? ` and ${totals.warnings} ${pluralize('warning', totals.warnings)}`
+            : ''
+        }.\n`,
+      ),
+    );
+  } else if (totals.warnings > 0) {
+    process.stdout.write(
+      green(`Your lint config is valid. 🎉\n`),
```
adamaltman, line 250:
```suggestion
      green(`Your config is valid. 🎉\n`),
```

tatomyr: We probably don't need to greet a user if the config is valid at all :) The main point is to have meaningful warnings when the config is invalid, am I right?
BTW @RomanHotsiy, should it be `stdout` where we put the messages? I remember once this broke something in **Workflows**.

RomanHotsiy: I would not show a "valid" message too. Only warnings or errors.

We should use `stderr` for all the user logs (even not errors). Stdout is used for the actual tool output (bundle or lint results JSON).

__tests__/lint-config/invalid-config--no-option/snapshot.js
```diff
@@ -0,0 +1,30 @@
...
+
+Warning was generated by the spec rule.
+
+
+You have 1 warning.
+No configurations were defined in extends -- using built in recommended configuration by default.
+Warning! This default behavior is going to be deprecated soon.
+
+validating /Users/yevhensmoliy/redocly/projects/openapi-cli/__tests__/lint-config/__fixtures__/valid-openapi.yaml...
+/Users/yevhensmoliy/redocly/projects/openapi-cli/__tests__/lint-config/__fixtures__/valid-openapi.yaml: validated in <test>ms
```
AntonKozachuk, line 23: All the snapshots with this type of path will fail for the user locally. 
image
Example:
image

__tests__/index.test.ts
```diff
@@ -49,76 +60,90 @@ describe('E2E', () => {
...
       });
     }
   });
 
+  describe('lint-config', () => {
+    const lintOptions: { dirName: string; option: string | null }[] = [
+      { dirName: 'invalid-conifg--lint-config-off', option: 'off' },
+      { dirName: 'invalid-conifg--lint-config-warn', option: 'warn' },
+      { dirName: 'invalid-conifg--lint-config-error', option: 'error' },
+      { dirName: 'invlid-lint-config-saverity', option: 'somethig' },
```
AntonKozachuk, line 77: Not that important, but you probably meant 'something'.

__tests__/lint-config/__fixtures__/valid-openapi.yaml
```diff
@@ -0,0 +1,21 @@
...
+  /pet/findByStatus:
+    get:
+     summary: needExist
+     operationId: findByStatus
+     responses:
+      '200':
+        description: example description
+      '400':
+        description: some desc
+
```
AntonKozachuk, line 21: Probably need to add a new line at the end of the file.

__tests__/lint-config/invalid-conifg--lint-config-warn/.redocly.yaml
```diff
@@ -0,0 +1,8 @@
+apis:
+  main:
+    root: ./openapi.yaml
+
+lint:
+  rules:
+    context:
+
```
AntonKozachuk, line 8: The new line is missing here too.

__tests__/lint-config/invalid-config--no-option/snapshot.js
SmoliyY: Great catch, thx, will fix!

AntonKozachuk (approved):

roman-sainchuk:
Looks good to me

packages/cli/src/commands/lint.ts
```diff
@@ -108,3 +123,53 @@ export async function handleLint(
...
+    await validateLintConfig({
+      problems,
+      outputFormat: argv.format,
+      maxProblems: argv['max-problems'],
+      version,
+    });
+  };
+}
+
+async function validateLintConfig({
```
RomanHotsiy, line 149: This function does not validate. It just prints the outputs. Please, rename.

packages/cli/src/commands/lint.ts
```diff
@@ -108,3 +123,53 @@ export async function handleLint(
...
+  formatProblems(problems, {
+    format: outputFormat,
+    maxProblems,
+    totals: fileTotals,
+    version,
+  });
+  printConfigLintTotals(fileTotals);
+}
+
+function printValidatingFile(filePath: string, validationDuration: number | string): void {
```
RomanHotsiy, line 171:
```suggestion
function printFileValidationTime(filePath: string, validationDuration: number | string): void {
```

packages/cli/src/commands/lint.ts
```diff
@@ -108,3 +123,53 @@ export async function handleLint(
...
+  process.stderr.write(
+    gray(`${relativeFilePath(filePath)}: validated in ${validationDuration}\n\n`),
```
RomanHotsiy, line 173: is `validationDuration` string or number? if it's a number do we need to add `ms` unit?

```suggestion
    gray(`${relativeFilePath(filePath)}: validated in ${validationDuration}ms\n\n`),
```

packages/cli/src/index.ts
```diff
@@ -153,6 +153,15 @@ yargs
           array: true,
           type: 'string',
         },
+        'lint-config': {
+          description: 'Apply severity for linting config file.',
```
RomanHotsiy, line 157:
```suggestion
          description: 'Apply severity for linting the config file.',
```

packages/cli/src/commands/lint.ts
RomanHotsiy: Do we need a function for this? maybe we can just inline it?

packages/cli/src/utils.ts
```diff
@@ -305,3 +323,7 @@ export function isSubdir(parent: string, dir: string): boolean {
   const relative = path.relative(parent, dir);
   return !!relative && !/^..($|\/)/.test(relative) && !path.isAbsolute(relative);
 }
+
+export function relativeFilePath(filePath: string): string {
+  return filePath.replace(process.cwd(), '');
```
RomanHotsiy, line 328:
```suggestion
  return path.relative(process.cwd(), filePath);
```

But maybe we don't need a separate function for it.

RomanHotsiy (commented):
Left some comments.

packages/cli/src/utils.ts
SmoliyY: yeah, thx, we can remove it I guess.

packages/cli/src/commands/lint.ts
SmoliyY: As argument `validationDuration = getExecutionTime(startedAt)` and already return it with `ms` but there is no need for number type in such case. Due to removing `relativeFilePath` there is no more need in this fn.

packages/cli/src/commands/lint.ts
SmoliyY: Comment below resolve this

packages/cli/src/commands/lint.ts
SmoliyY: Agree, this make sense for me!

__tests__/index.test.ts
```diff
@@ -49,79 +64,88 @@ describe('E2E', () => {
...
-        const result = `${out}\n${err}`;
+        const result = getCommandOutput(args, testPath);
         (<any>expect(result)).toMatchSpecificSnapshot(join(testPath, 'snapshot.js'));
       });
     }
   });
 
+  describe('lint-config', () => {
+    const lintOptions: { dirName: string; option: string | null }[] = [
+      { dirName: 'invalid-conifg--lint-config-off', option: 'off' },
```
RomanHotsiy, line 78: rename this plz

__tests__/lint-config/invalid-config-assertation-name/snapshot.js
```diff
@@ -0,0 +1,28 @@
...
+
+
+❌ Your config has 1 error.
+validating ../__fixtures__/valid-openapi.yaml...
+../__fixtures__/valid-openapi.yaml: validated in <test>ms
+
+Woohoo! Your OpenAPI definition is valid. 🎉
+
+[WARNING] Unused rules found in .redocly.yaml: asset/path-item-mutually-required.
+Check the spelling and verify you added plugin prefix.
```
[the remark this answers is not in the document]
RomanHotsiy: We can add it, yes, I would do it in a separate PR though.

__tests__/lint-config/invalid-conifg--lint-config-error/snapshot.js
```diff
@@ -0,0 +1,30 @@
...
+
+Property \`context\` is not expected here.
+
+4 | lint:
+5 |   rules:
+6 |     context: null
+  |     ^^^^^^^
+7 |
+
+Error was generated by the spec rule.
```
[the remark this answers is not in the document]
RomanHotsiy: We run it internally using `spec` rule. Maybe we can rewrite the name of the rule in the output. @SmoliyY what do you think?

__tests__/lint-config/invalid-config-assertation-config-type/snapshot.js
```diff
@@ -0,0 +1,26 @@
...
+\`type\` can be one of the following only: "DefinitionRoot", "Tag", "ExternalDocs", "Server", "ServerVariable", "SecurityRequirement", "Info", "Contact", "License", "PathMap", "PathItem", "Parameter", "Operation", "Callback", "RequestBody", "MediaTypeMap", "MediaType", "Example", "Encoding", "Header", "ResponsesMap", "Response", "Link", "Schema", "Xml", "SchemaProperties", "DiscriminatorMapping", "Discriminator", "Components", "NamedSchemas", "NamedResponses", "NamedParameters", "NamedExamples", "NamedRequestBodies", "NamedHeaders", "NamedSecuritySchemes", "NamedLinks", "NamedCallbacks", "ImplicitFlow", "PasswordFlow", "ClientCredentials", "AuthorizationCode", "SecuritySchemeFlows", "SecurityScheme", "XCodeSample", "WebhooksMap".
+
+ 6 | assert/path-item-mutually-required:
+ 7 |   context:
+ 8 |     - type: Invalid-type
+   |             ^^^^^^^^^^^^
+ 9 |   subject: Operation
+10 |   message: Operation should have summary and security fields simultaneously
+
+Error was generated by the spec rule.
```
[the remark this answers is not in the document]
RomanHotsiy: It has:

image

__tests__/lint-config/invalid-conifg--lint-config-error/snapshot.js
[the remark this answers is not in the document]
SmoliyY: Yes we can extend it with `configuration spec rule`

__tests__/lint-config/invalid-config-assertation-name/snapshot.js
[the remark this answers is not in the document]
SmoliyY: I agree. Sorry, I did not deep dive into code for this and will do it in separate PR.

SmoliyY:
@adamaltman @RomanHotsiy Fixed last comments above.

__tests__/lint-config/invalid-conifg--lint-config-error/snapshot.js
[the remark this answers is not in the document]
adamaltman: Can we make it say the configuration spec rule?

adamaltman (approved):
Thank you 🎉

PR #722, opened 2022-06-02
feat: join multiple files with same tags

### LINKED ISSUE #668: For join, the same tag in multiple input files should not be seen as a conflict
**Problem**
I have two (or more) OAS files that use the same tag to tag some operations. Running `openapi join` results in an error about "Conflict on tags."

**Improved Solution**
I don't see why it would be considered a conflict if multiple schemas that you're trying to join together use the same tag names. Maybe display a warning, but it's not an irreconcilable crisis that should prevent the join from succeeding. Use the first (or the last) description of the tag--authors can figure that out. 

This is what is already done with components that are defined in multiple input files but are identical. It's not considered an error.

Perhaps have a command line option to let the user choose what to do if the same tag name is used in multiple files:
1. Allow it.
2. Error out.
3. Prefix with info property
4. Prefix with filename

I tried using prefix-tags-with-info-prop but that gave me different tags with goofy names, which was not at all useful.

SmoliyY:
## What/Why/How?
Currently `join` command does not tolerate the same tags from multiple files because of preventing the conflicts in `x-tagGroups`. Was added one more option `--skip-tags-check` which allows having the same tags in a resolved file but without `x-tagGroups` . `description` for the same `tag` with this option will be taken from the first match
 
## Reference
Closes #668 

## Screenshots (optional)
After `join` command with tags conflicts
Was : 
Screenshot 2022-06-08 at 12 56 28

Now: 
Screenshot 2022-06-08 at 12 52 46

with `--without-x-tag-groups` option: 

Screenshot 2022-06-08 at 12 54 55

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/join.ts
```diff
@@ -383,21 +448,33 @@ function iteratePotentialConflicts(potentialConflicts: any) {
             }
           }
         } else {
-          showConflicts(green(group) +' => '+ key, conflicts);
-          potentialConflictsTotal += conflicts.length;
+          showConflicts(green(group) + ' => ' + key, conflicts);
+          if (key !== 'description') {
```
tatomyr, line 452: It's not obvious why we need this here.

packages/cli/src/commands/join.ts
```diff
@@ -383,21 +448,33 @@ function iteratePotentialConflicts(potentialConflicts: any) {
...
-        prefixTagSuggestion(group, conflicts.length);
+        prefixTagSuggestion(group, conflicts.length, key);
       }
     }
   }
 }
 
-function prefixTagSuggestion(group: string, conflictsLength: number) {
+function prefixTagSuggestion(group: string, conflictsLength: number, key?: string) {
+  if (group === 'tags' && key === 'description') {
```
tatomyr, line 463: Again, it's not obvious.

__tests__/index.test.ts
```diff
@@ -113,41 +113,50 @@ describe('E2E', () => {
...
+    });
 
-      const args = [
-        '--transpile-only',
-        '../../../packages/cli/src/index.ts',
-        'join',
-        'foo.yaml',
-        'bar.yaml'
+    describe('join with options', () => {
+      const options = [
```
tatomyr, line 143: could this be of the below type?
```ts
{ option: string; value: any }[]
```

__tests__/index.test.ts
```diff
@@ -113,41 +113,50 @@ describe('E2E', () => {
...
-          },
-        });
-
-        const out = r.stdout.toString('utf-8');
-        const err = r.stderr.toString('utf-8');
-        const result = `${out}\n${err}`;
+      test.each(options)('test with option: %s', (option) => {
+        const testPath = join(__dirname, `join/${option}`);
+        const isPropOption =
+          option === 'prefix-components-with-info-prop' || option === 'prefix-tags-with-info-prop';
```
tatomyr, line 153: See the above comment and avoid using specific options in the test.

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
...
+        ];
       }
 
-      const doesEntrypointExist = !potentialConflicts.tags.all[entrypointTagName] || (
-        potentialConflicts.tags.all[entrypointTagName] &&
-        !potentialConflicts.tags.all[entrypointTagName].includes(entrypoint)
-      )
+      if (!skipTagsCheck) createXTagGroups(entrypointFilename);
+
+      if (!skipTagsCheck) populateXTagGroups(entrypointTagName, getIndexGroup(entrypointFilename));
```
tatomyr, line 216: Why not joining these `if`s?

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
...
         joinedDef.tags.push(tag);
+
+        if (skipTagsCheck) {
+          potentialConflicts.tags.description[entrypointTagName] = [entrypoint];
+        }
       }
-      if (!joinedDef[xTagGroups][indexGroup][Tags].find((t: any) => t === entrypointTagName)) {
-        joinedDef[xTagGroups][indexGroup][Tags].push(entrypointTagName);
+      // If tag already exist do not overwvite tag proprs,
+      // instead check if description are different for potential conflicts warning
```
tatomyr, line 203: Split these comments

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
         tag['x-displayName'] = tag['x-displayName'] || tag.name;
```
tatomyr, line 194: It looks like you can reuse `.find...` in the above line https://github.com/Redocly/redocly-cli/pull/722/files#diff-3d0894febbd7923f278d468139041923d786a33f480741eabb8d465b9040eb4cR206
And also you can try avoiding using negative assertion.

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
...
+      // If tag already exist do not overwvite tag proprs,
+      // instead check if description are different for potential conflicts warning
+      else if (skipTagsCheck && tag.hasOwnProperty('description')) {
+        const isTagDescriptionNotEqual = joinedDef.tags.find(
+          (item: any) => item.description !== tag.description,
+        );
+        potentialConflicts.tags.description[entrypointTagName] = [
+          ...(potentialConflicts.tags.description[entrypointTagName] || []),
+          ...(isTagDescriptionNotEqual ? [entrypoint] : []),
+        ];
```
tatomyr, line 211: Could it be like this?
```suggestion
        potentialConflicts.tags.description[entrypointTagName].push(...(isTagDescriptionNotEqual ? [entrypoint] : []);
```

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
...
       potentialConflicts.tags.all[entrypointTagName] = [
-        ...(potentialConflicts.tags.all[entrypointTagName] || []), ...(doesEntrypointExist ? [entrypoint] : [])
+        ...(potentialConflicts.tags.all[entrypointTagName] || []),
+        ...(!skipTagsCheck && doesEntrypointExist ? [entrypoint] : []),
       ];
     }
   }
 
+  function getIndexGroup(entrypointFilename: string): number {
+    return joinedDef[xTagGroups].findIndex((item: any) => item.name === entrypointFilename);
```
tatomyr, line 230: Do we have a type for `xTagGroups` defined? Can we use it here?

packages/cli/src/commands/join.ts
```diff
@@ -174,21 +194,64 @@ packageVersion: string
...
+  }
+
+  function createXTagGroups(entrypointFilename: string) {
+    if (!joinedDef.hasOwnProperty(xTagGroups)) {
+      joinedDef[xTagGroups] = [];
+    }
+
+    if (!joinedDef[xTagGroups].some((g: any) => g.name === entrypointFilename)) {
```
tatomyr, line 238: The same as above

packages/cli/src/commands/join.ts
```diff
@@ -383,21 +448,33 @@ function iteratePotentialConflicts(potentialConflicts: any) {
...
         } else {
-          showConflicts(green(group) +' => '+ key, conflicts);
-          potentialConflictsTotal += conflicts.length;
+          showConflicts(green(group) + ' => ' + key, conflicts);
+          if (!skipTagsCheck) {
+            potentialConflictsTotal += conflicts.length;
+          }
         }
-        prefixTagSuggestion(group, conflicts.length);
+        prefixTagSuggestion(group, conflicts.length, skipTagsCheck);
```
tatomyr, line 456: Can we check if `group === tags` here?

tatomyr (changes requested):
Left a few comments, please resolve

packages/cli/src/commands/join.ts
SmoliyY: This is good point, unfortunately  we do not have such type yet

packages/cli/src/commands/join.ts
```diff
@@ -153,42 +174,87 @@ packageVersion: string
...
     for (const tag of tags) {
       const entrypointTagName = addPrefix(tag.name, tagsPrefix);
       if (tag.description) {
         tag.description = addComponentsPrefix(tag.description, componentsPrefix!);
       }
-      if (!joinedDef.tags.find((t: any) => t.name === entrypointTagName)) {
+
+      const tagDuplicate = joinedDef.tags.find((t: Oas3Tag) => t.name === entrypointTagName);
+
+      if (tagDuplicate && skipTagsCheck) {
```
tatomyr, line 196: Do we cover case when `tagsDuplicate && !skipTagsCheck`?

packages/cli/src/commands/join.ts
```diff
@@ -383,24 +451,36 @@ function iteratePotentialConflicts(potentialConflicts: any) {
...
-function prefixTagSuggestion(group: string, conflictsLength: number) {
-  if (group === 'tags') {
-    process.stderr.write(green(`
-    ${conflictsLength} conflict(s) on tags.
-    Suggestion: please use ${blue('prefix-tags-with-filename')} or ${blue('prefix-tags-with-info-prop')} to prevent naming conflicts. \n\n`
-    ));
+function prefixTagSuggestion(conflictsLength: number, skipTagsCheck?: boolean) {
+  if (!skipTagsCheck) {
+    process.stderr.write(
+      green(`
```
tatomyr, line 470: Do we use something like `outdent` or `dedent`?

packages/cli/src/commands/join.ts
SmoliyY: Currently there is no sense for this, because we want add `tag` into  `tags` that not yet exist in `joinDef`.

packages/cli/src/commands/join.ts
SmoliyY: Thx, let's do it, this is make sense for me!

tatomyr (approved):

RomanHotsiy (commented):
`--skip-tags-check` is not a good name.

It is more about `--no-x-tag-groups` and as a consequence, it supports tags with the same names.

I didn't check code yet.

SmoliyY:
@RomanHotsiy Looks like yargs do not tolerate the options that begin with `--no-` https://github.com/yargs/yargs/blob/main/lib/validation.ts#L341. But it will works ok if pass also value of option (e.g `--no-x-tag-groups=true`)

AntonKozachuk (approved):

tatomyr:
> Looks like yargs do not tolerate the options that begin with --no-
Maybe it could be `--skip-x-tag-groups` then?

RomanHotsiy:
I see, then maybe `--without-x-tag-groups`. Skip for me sounds more like skipping some existing tag groups and we are not generating them.

SmoliyY:
Added some screenshots after last changes @RomanHotsiy @adamaltman @tatomyr. Pls let me know if there is no other objection, thx)

tatomyr:
LGTM

RomanHotsiy (approved):
LGTM

PR #724, opened 2022-06-06
fix: improve types for config

AntonKozachuk:
## What/Why/How?
Improve config types for VS Code and CLI validation.

## Reference
Connected to #720

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/types/redocly-yaml.ts
+++ b/packages/core/src/types/redocly-yaml.ts
@@ -1,6 +1,5 @@
 import { NodeType, listOf } from '.';
-import { omitObjectProps, pickObjectProps } from '../utils';
-const customRulesRegexp = /^[a-zA-Z0-9]+\//;
+import { omitObjectProps, pickObjectProps, isCustomRuleId } from '../utils';
 const builtInRulesList = [
   'spec',
   'info-description',
@@ -194,7 +193,7 @@ const Rules: NodeType = {
   additionalProperties: (value: unknown, key: string) => {
     if (key.startsWith('assert/')) {
       return 'Assert';
-    } else if (customRulesRegexp.test(key) || builtInRulesList.indexOf(key) !== -1) {
+    } else if (isCustomRuleId(key) || builtInRulesList.indexOf(key) !== -1) {
       if (typeof value === 'string') {
         return { enum: ['error', 'warn', 'off'] };
       } else {
@@ -203,7 +202,7 @@ const Rules: NodeType = {
     }
     // Otherwise is considered as invalid
     return;
-  }
+  },
 };
 
 const ObjectRule: NodeType = {
--- a/packages/core/src/types/redocly-yaml.ts
+++ b/packages/core/src/types/redocly-yaml.ts
@@ -5,9 +5,21 @@ const ConfigRoot: NodeType = {
   properties: {
     organization: { type: 'string' },
     apis: 'ConfigApis',
+    apiDefinitions: {
+      type: 'object',
+      properties: {},
+      additionalProperties: { properties: { type: 'string' } }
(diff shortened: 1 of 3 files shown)


packages/core/src/types/redocly-yaml.ts
```diff
@@ -527,58 +565,103 @@ const ConfigReferenceDocs: NodeType = {
...
-    sideNavStyle: { type: 'string' },
+    sideNavStyle: { enum: ['summary-only', 'path-first', 'id-only'] },
     simpleOneOfTypeLabel: { type: 'boolean' },
-    sortEnumValuesAlphabetically: { type: 'boolean' },
-    sortOperationsAlphabetically: { type: 'boolean' },
+    sortEnumValuesAlphabetically: (value: unknown) => {
+      if (typeof value === 'boolean') {
+        return { type: 'boolean' };
+      } else {
+        return { type: 'string' };
```
tatomyr, line 647: What the case is for type `string`?

tatomyr, line 647: What the case is for type `string`?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -590,18 +673,42 @@ const ConfigMockServer: NodeType = {
   },
 };
 
+const ConfigHooks: NodeType = {
```
tatomyr, line 676: What are these config hooks? I don't recall anything similar in Redocly config.

tatomyr, line 676: What are these config hooks? I don't recall anything similar in Redocly config.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -1,13 +1,125 @@
 import { NodeType, listOf } from '.';
 import { omitObjectProps, pickObjectProps } from '../utils';
+const customRulesRegexp = /^[a-zA-Z0-9]+\//;
```
tatomyr, line 3:
```suggestion
const customRulesRegexp = /^[a-zA-Z0-9\-]+\/[a-zA-Z0-9\-]+$/;
```

tatomyr: I don't think it's going to be much faster (unless we expect a looot of looong rules). And it looks cumbersome to me.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -74,11 +189,87 @@ const RootConfigLint: NodeType = {
   },
 };
 
+const Rules: NodeType = {
+  properties: {},
+  additionalProperties: (value: unknown, key: string) => {
+    if (key.startsWith('assert/')) {
+      return 'Assert';
+    } else if (customRulesRegexp.test(key) || builtInRulesList.indexOf(key) !== -1) {
```
tatomyr, line 197:
```suggestion
    } else if (builtInRulesList.includes(key) || customRulesRegexp.test(key)) {
```

packages/core/src/types/redocly-yaml.ts
RomanHotsiy: We can move it into a function. Regexp is usually much harder to read than code (easy to write though).

But well, I don't think we need full parse here, it was just example. It can be as simple as:

```ts
function isCustomRuleId(id: string) {
  return id.includes('/');
}
```

and then instead of `customRulesRegexp.test(key)` we can use `isCustomRuleId(key)`

tatomyr: I like this. Later we can extend the function if we need.

AntonKozachuk: Added.

PR #748, opened 2022-06-29
chore: fix typing and update docs on `ctx.resolve`

tatomyr:
## What/Why/How?

Fixes wrong types for **ctx.resolve**. 

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/docs/resources/custom-rules.md
+++ b/docs/resources/custom-rules.md
@@ -207,7 +207,7 @@ The context object contains additional functionality that is helpful for rules t
 Additionally, the context object has the following methods:
 
 - `report(descriptor)` - reports a problem in the definition (see the dedicated section).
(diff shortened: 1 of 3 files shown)


SmoliyY (approved):

docs/resources/custom-rules.md
```diff
@@ -207,7 +207,7 @@ The context object contains additional functionality that is helpful for rules t
 Additionally, the context object has the following methods:
 
 - `report(descriptor)` - reports a problem in the definition (see the dedicated section).
-- `resolve(node)` - synchronously dereferences $ref node to its value. Works only with $refs from the original document.
+- `resolve(node)` - synchronously dereferences $ref node to its value. Works only with $refs from the original document. If you need to resolve a reference from another source, you can use optional second parameter: `resolve(node, from: string)`.
```
adamaltman, line 210:
```suggestion
- `resolve(node)` - synchronously dereferences $ref node to its value. Works only with $refs from the original document. If you need to resolve a reference from another source, you can use the optional second parameter: `resolve(node, from: string)`.
```

Issue #753, opened 2022-07-06
Filter-out does not filter out when value is set to false (bool)


deBarrosS:
**Description**
When bundling my openapi description I'm filtering **out** nodes with a property that is set to false.
But it does not really filter these nodes out. In fact, it'd seem that it comes from the fact that it is a boolean.
I tried setting a node to 'false' (string) and it did indeed filter this node out.

**To Reproduce**
1. Given this `.redocly.yaml` file
```
lint:
  decorators:
    filter-out:
      property: x-v1
      value: false
      matchStrategy: any
```
2. And this OpenAPI file(s)
```
openapi: 3.1.0
info:
  title: ''
  version: ''
paths:
  /mypath:
    get:
      deprecated: false
      responses:
        '204':
          description: No Content
      description: desc
      x-internal: true
      x-v3: true
      x-v4: true
      x-v1: false
      x-v2: false
  /myotherpath:
    get:
      description: other desc
      responses:
        '204':
          description: No Content
      x-internal: true
      x-v2: true
      x-v1: true
      x-v3: false
      x-v4: false
  /yetanotherpath:
    get:
      description: other desc
      responses:
        '204':
          description: No Content
      x-internal: true
      x-v2: true
      x-v1: 'false'
      x-v3: false
      x-v4: false
components: {}
```
3. Run
`redocly bundle path/to/openapi.yaml -o path/to/result.yaml --config=path/to/redocly.yaml `

**Expected behavior**
- The expected result was to have (at least) the _get_ node from "/mypath" removed on the result.yaml  (at most, the whole path once it was the only operation on this path.

**Buggy behavior**
- Nevertheless this hasn't been the case for me as the result.yaml (output of this command) is exactly the same as the openapi.yaml (input).

**But**
- When we change the redocly.yaml to :
```
lint:
  decorators:
    filter-out:
      property: x-v1
      value: 'false'
      matchStrategy: any
```
And re-run the same command, we get "/yetanotherpath" removed **as expected**.

- Also, if we try to "filter-in" with "value : false" (a boolean), the result is only going to have "/mypath" **as expected**

**Redocly Version(s)**
103

**`Node.js` Version(s)**
v16.15.0

deBarrosS:
[...]

PR #754, opened 2022-07-06
Fix : Filter-out does not filter out when value is set to false (bool) #753

deBarrosS:
### What ?

- Fix use of if guards on filtering in and out some nodes from an openapi file. 

### Why ?

- The use of **! operator*** for checking whether a value was undefined also checked out false (boolean) and null values. What made it impossible to use those values for filtering nodes out of the file.
- As a border effect, this problem didn't occur while filtering **in** properties that had these values (false/null) but not for the "proper reason".

### How ?

- Explicitly use **"=== undefined"**  to verify if the structure hasn't been given as a parameter to the function call instead of using the **! operator** 
- This allows the use of both false and null values to filter nodes in or out an openapi file 

## Reference

[Filter-out does not filter out when value is set to false (bool) #753](https://github.com/Redocly/redocly-cli/issues/753)

## Testing

- Test if the nodes containing the given properties with false or null values are properly removed from the input file.

## Check yourself

- [ ✅  ] Code is linted
- [ ❌  ] Tested with redoc/reference-docs/workflows
- [ ✅  ] All new/updated code is covered with tests

## Security

- [ ✅ ] Security impact of change has been considered
- [ ❔  ] Code follows company security practices and guidelines

### DIFF
--- a/packages/core/src/decorators/__tests__/filter-out.test.ts
+++ b/packages/core/src/decorators/__tests__/filter-out.test.ts
@@ -27,9 +27,7 @@ describe('oas3 filter-out', () => {
               callbacks:
                 x-access: protected
                 orderInProgress:
-                  x-internal: true
-             `
-  );
+                  x-internal: true`);
 
   it('should remove /pet path and y parameter', async () => {
     const testDocument = parseYamlToDocument(
@@ -62,7 +60,7 @@ describe('oas3 filter-out', () => {
               x:
                 name: x
 
-        `);
+                `);
   });
 
   it('should remove only /order path', async () => {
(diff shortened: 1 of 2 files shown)


deBarrosS:
@RomanHotsiy a bit later than I thought but there it is 😬

RomanHotsiy (approved):

RomanHotsiy:
Amazing, you even added tests 👏 !

@tatomyr plz merge it and release after you review it.

tatomyr:
LGTM.

@deBarrosS thanks for your contribution. I'm going to merge and release the new version early next week.

tatomyr (approved):

deBarrosS:
My pleasure, @tatomyr . Thank you for the reply (and obviously for this great tool) :+1: :smile:

PR #778, opened 2022-07-27
fix: join paths properly

roman-sainchuk:
## What/Why/How?
Currently `summary`, `description`, `servers`, `parameters` fields are ignored/incorrectly handled by `join` command.

This PR is rework for `join` command based on [this one](https://github.com/Redocly/redocly-cli/pull/536). The reason is that changes in the [latest](https://github.com/Redocly/redocly-cli/pull/536) is based just on checking if the value of `PathItem` field is of `object` type. I suppose that current logic performed on `object` values is  making sense only for `Operation Object` type. `summary`, `description`, `servers` and `parameters` should be joined in a different way.

Things required to finish this PR:
- [x] Merge only operations (`post`, `get`, etc.) using existing implementation
- [ ]  ~~Implement join of `Operation Object`s which are $refs~~
- [x]  Implement join of `summary` and `description`
- [x]  Implement join of `servers`
- [x]  Implement join of `parameters` 
- [x]  Add meaningful tests for join command

## Check yourself

- [x] Code is linted

## Security


roman-sainchuk:
For `description` and `summary` - take it from first document
For `servers` - create unique array by `url` from both documents

@adamaltman @RomanHotsiy please advice how to join `parameters` ? Should we create unique array by `name` and `in` fields ?

packages/cli/src/commands/join.ts
```diff
@@ -311,37 +337,94 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
...
+          joinedDef.paths[path] = {};
+        }
+        if (!potentialConflicts.paths.hasOwnProperty(path)) {
+          potentialConflicts.paths[path] = {};
+        }
+
+        const pathItem = paths[path];
+
+        if ('$ref' in pathItem) {
+          continue; // TODO: what do we do with refs?
```
RomanHotsiy, line 361: we do nothing I think

RomanHotsiy (commented):
Can you exclude prettier changes from this PR? It's hard to review :(

roman-sainchuk:
> Can you exclude prettier changes from this PR? It's hard to review :(

removed

RomanHotsiy:
> For description and summary - take it from first document

Yes, I would also show a warning if there are different values and they mismatch.

> For servers - create unique array by url from both documents

Yes. I would also show a warning or even crash if there are different values and they mismatch (e.g. variables)

> how to join parameters ? Should we create unique array by name and in fields ?

Yes. Unique by name/in seems valid. Make sure to crash if there are different schemas. For different descriptions we can show warnings only.

packages/cli/src/commands/join.ts
```diff
@@ -332,73 +334,179 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
...
+
+      if (!joinedDef.paths[path].hasOwnProperty('servers')) {
+        joinedDef.paths[path].servers = [];
+      }
+
+      for (const server of pathItem.servers) {
+        let isFoundServer = false;
+        for (const pathServer of joinedDef.paths[path].servers) {
+          if (pathServer.url === server.url) {
+            isFoundServer = true;
```
roman-sainchuk, line 399: do we need this variable if we anyway terminate the program in case of duplications ?

packages/cli/src/commands/join.ts
```diff
@@ -332,73 +334,179 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
...
+      }
+      if (!joinedDef.paths[path].hasOwnProperty('parameters')) {
+        joinedDef.paths[path].parameters = [];
+      }
+
+      for (const parameter of pathItem.parameters as Oas3Parameter[]) {
+        let isFoundParameter = false;
+        for (const pathParameter of joinedDef.paths[path].parameters) {
+          if (pathParameter.name === parameter.name && pathParameter.in === parameter.in) {
+            isFoundParameter = true;
```
roman-sainchuk, line 424: same here

packages/cli/src/commands/join.ts
```diff
@@ -332,73 +334,179 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
...
+        );
+      }
+    }
+  }
+
+  function isServersEqual(serverOne: Oas3Server, serverTwo: Oas3Server) {
+    if (serverOne.description !== serverTwo.description) {
+      return false;
     }
+    return isEqual(serverOne.variables, serverTwo.variables);
```
roman-sainchuk, line 509:
```suggestion
    if (serverOne.description === serverTwo.description) {
      return isEqual(serverOne.variables, serverTwo.variables);
    }
    
    return false;
```

packages/cli/src/commands/join.ts
IgorKarpiuk: No, we will terminate the program only if two conditions are met. We need this variable to check if we can add the current object to the result

roman-sainchuk:
LGTM 👍

packages/cli/src/commands/join.ts
```diff
@@ -332,73 +334,180 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
...
+            collectPathParameters(pathItem, path);
+          }
+          if (typeof pathItem[field] === 'string') {
+            collectPathStringFields(pathItem, path, field);
+          }
+        }
+      }
+    }
+
+    function collectPathStringFields(
```
tatomyr, line 370: Can we move this function out of the scope and avoid mutating of `joinedDef`?

packages/core/src/typings/openapi.ts
```diff
@@ -41,17 +41,24 @@ export interface OasRef {
 
 export type Referenced<T> = OasRef | T;
 
-export interface Oas3PathItem {
+export enum Oas3Operations {
```
tatomyr, line 44: Please reuse `OPENAPI3_METHOD` enum.

tatomyr (approved):

PR #782, opened 2022-07-29
chore: add user-agent field

RomanHotsiy:
## What/Why/How?

Add a user-agent field for Redocly requests.

## Check yourself

- [x] Code is linted

## Security


packages/core/src/redocly/registry-api.ts
```diff
@@ -4,6 +4,8 @@ import { DEFAULT_REGION, DOMAINS } from '../config/config';
 import { isNotEmptyObject } from '../utils';
 const version = require('../../package.json').version;
 
+export const currentCommand = typeof process !== 'undefined' ? process.env?.REDOCLY_CLI_COMMAND || '' : '';
```
tatomyr, line 7: Would this work?

```suggestion
export const currentCommand = process?.env?.REDOCLY_CLI_COMMAND || '';
```

RomanHotsiy: no, it will crash in the browser:

> ReferenceError: process is not defined

You can try it in your browser console.

adamaltman (commented):
What about `bundle`?

RomanHotsiy:
@adamaltman good catch. I think I've missed it 😄

packages/core/src/redocly/registry-api.ts
```diff
@@ -23,7 +25,10 @@ export class RegistryApi {
   }
 
   private async request(path = '', options: RequestInit = {}, region?: Region) {
-    const headers = Object.assign({}, options.headers || {}, { 'x-redocly-cli-version': version });
+    const headers = Object.assign({}, options.headers || {}, {
+      'x-redocly-cli-version': version,
```
tatomyr, line 29: Should we also put the same [here](https://github.com/Redocly/redocly-cli/blob/master/packages/cli/src/commands/push.ts#L371)?

packages/core/src/redocly/registry-api.ts
tatomyr: Ah, indeed.

packages/core/src/redocly/registry-api.ts
RomanHotsiy: We may but it's not essential. This goes directly to s3.

tatomyr: Got it

tatomyr (approved):

packages/core/src/rules/ajv.ts
```diff
@@ -1,4 +1,4 @@
-import Ajv, { ValidateFunction, ErrorObject } from '@redocly/ajv';
+import Ajv, { ValidateFunction, ErrorObject } from '@redocly/ajv/dist/2019';
```
RomanHotsiy, line 1:
```suggestion
import Ajv, { ValidateFunction, ErrorObject } from '@redocly/ajv';
```

resources/pets.yaml
```diff
@@ -545,6 +545,11 @@ paths:
               requestBody:
                 content:
                   application/json:
+                    example:
+                      {
+                        "orderId": 3,
+                        "test": 3
+                      }
```
RomanHotsiy, line 552:
```suggestion
```

PR #784, opened 2022-07-31
fix: fix enum type mismatch message

adamaltman:
## What/Why/How?

The current message is wrong, because in 3.1 you can define multiple types. The behavior works fine though except for `type: any` and an issue is opened for that (#786).

## Testing

Tests were adjusted as well.

## Check yourself

- [x] Code is linted

## Security


RomanHotsiy (approved):

adamaltman:
@tatomyr do I need to prepare a changelog? If so, we should also update this: https://github.com/Redocly/redocly-cli/blob/master/.github/CONTRIBUTING.md#pull-request-guidelines

tatomyr:
I don't think you need. We will do this when creating the next release.

PR #786, opened 2022-07-31
fix: match type any

### LINKED ISSUE #785: matchesJsonSchemaType incorrect results for type `any`
**Describe the bug**

This function `matchesJsonSchemaType` returns incorrect results for type `any`.

**To Reproduce**
1. Given a schema with type `any`.
2. Add some enum values, and check the `no-enum-type-mismatched` rule for example.
3. See error

**Expected behavior**

Any type of value should be acceptable for type any.

**Logs**

n/a

**OpenAPI definition**

```yaml
schema:
  type: object
  properties:
    demo:
      type: any
      enum: ['string', 123]
```

**Redocly Version(s)**

1.0.0-beta.105

**`Node.js` Version(s)**

16

**Additional context**

adamaltman:
## What/Why/How?

Fixes missing match for `type: any`.

## Reference

Fixes #785 

## Testing

Added tests for the method.

## Check yourself

I think there is an issue with the prettier config. I did not include the prettier run here as it adjusted so many files and seems also out of date with the current code style in some places.

- [ ] Code is linted

## Security

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -15,7 +15,7 @@ Hi! We're really excited that you are interested in contributing to Redocly CLI.
 
 - The best way to get your bug fixed is to provide a (reduced) test case. This means listing and explaining the steps we should take to try and hit the same problem you're having. It helps us understand in which conditions the issue appears, and gives us a better idea of what may be causing it.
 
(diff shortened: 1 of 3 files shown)


adamaltman:
This one is invalid, but adding tests was a good idea. I did that in #789 so this one can be closed.

PR #787, opened 2022-07-31
chore: run prettier with trailing comma all

adamaltman:
## What/Why/How?

I ran `npm run prettier` before I opened my pull request #786, but I realized I could not commit the prettier changes to that PR because there are so many changes to so many files. I also noticed that we use a lot of trailing commas which were adjusted by prettier. 

I adjusted the prettier config to have `all` for trailing commas. I'm not sure if that is correct or not, but either way, shouldn't we:

- run prettier and commit, such as I've done in this PR
- add a GitHub action to run prettier with the check flag: https://prettier.io/docs/en/cli.html#--check to prevent deviations so we don't end up with such a big difference?

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
   "printWidth": 100,
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "all"
 }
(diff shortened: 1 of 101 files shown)


adamaltman:
> Jest: "packages/cli/" coverage threshold for lines (39%) not met: 38.75%

This is purely due to formatting where there are additional lines added from prettier.

tatomyr:
I like the idea. I'll check if there's no substantial conflicts with other active branches and will prepare this for merging.

AntonKozachuk (approved):

tatomyr:
I'd like [this](https://github.com/Redocly/redocly-cli/pull/790/files) to have been finished firstly, since it contains a lot of changes. I'll try to do it as soon as possible.

adamaltman:
@tatomyr also it would be good if someone wrote the github action to run the `prettier --check` before this would be merged. This already has conflicts and is likely to be easier to produce a new PR at the time -- I only changed the config file (1 line), ran prettier, and git commit. We can redo that after the lint is renamed to styleguide.

Issue #696, opened 2022-05-20
Rename lint object to styleguide object in configuration


adamaltman:
**Is your feature request related to a problem? Please describe.**

The `lint` object in the configuration file contains information used by the `lint`, `bundle`, and (indirectly) `push` commands.

When criticizing the product, this is an area of confusion. In fact, in our Workflows app, we built a feature which leverages this configuration" and we called it **API styleguide**.

**Describe the solution you'd like**

I would like us to rename the `lint` object to the `styleguide` object.

It should show a deprecation notice to people who use the `lint` object.

**Describe alternatives you've considered**

- Doing nothing

**Additional context**

Docs need to be updated accordingly.

Issue #732, opened 2022-06-15
Validate 'redocly/cli` package before publishing to npm and workflows


SmoliyY:
### Description 

`redocly/cli` [release](https://github.com/Redocly/redocly-cli/pull/730) `1.0.0-beta101` was failed due to publishing code that use package that do not exist in `packages/cli/package.json`, but instead this package exist in root `package.json`. Also we need to add test to check `workflows` functionality

### Solution

- Add check (e.g eslint rule) that would check if package exist in `packages/cli/package.json`
- Add test that would build same packages that published to `npm`
- Add more tests for `workflows` functionality

tatomyr:
Turns out `npm pack` from [here](https://github.com/Redocly/redocly-cli/pull/773/files#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191R65) doesn't include the corresponding `openapi-core` what leads to inconsistency in this test. I'm reopening this issue as I  had to comment out this job and this should be fixed.

PR #790, opened 2022-08-01
chore: rename 'lint' to 'styleguide'

roman-sainchuk:
## What/Why/How?

Renaming `lint` field into `styleguide` in Redocly config. 
We still support the legacy format, but added deprecation notices if the `lint` field is found inside an api or on the root level. 

Closes https://github.com/Redocly/redocly-cli/issues/696

> Note: I had to comment out `cli-package-test` job for the time being as it incorrectly packs `cli` with `core` and I haven't found any quick solution for that. 

## Check yourself

- [x] Code is linted

## Security


.github/workflows/tests.yaml
```diff
@@ -40,35 +40,36 @@ jobs:
...
-          CI: true
-      - name: Prepare CLI package
-        run: npm run pack:prepare
-      - name: Install CLI Package
-        run: npm install -g redocly-cli.tgz
-      - name: Redocly version
-        run: redocly --version
-      - name: Definition test
-        run: redocly lint packages/core/src/benchmark/benches/rebilly.yaml --format=stylish
+  # FIXME: fix this so the `cli` package includes the corresponding `core`
```
adamaltman, line 43: Should this be addressed when?

packages/core/src/config/load.ts
```diff
@@ -6,7 +6,7 @@ import { Config, DOMAINS } from './config';
 import { transformConfig } from './utils';
 import { resolveConfig } from './config-resolvers';
 
-import type { RawConfig, Region } from './types';
+import type { DeprecatedInRawConfig, RawConfig, Region } from './types';
```
adamaltman, line 9: `DeprecatedInRawConfig` or `DeprecatedRawConfig`?

.github/workflows/tests.yaml
tatomyr: I reopened the [original issue](https://github.com/Redocly/redocly-cli/issues/732).

packages/core/src/config/load.ts
tatomyr: I reckon it should be `In` as it's only about fields being deprecated inside the raw config and not the entire deprecated raw config.

roman-sainchuk:
Looks good to me 😃

RomanHotsiy (approved):
LGTM 👍

AntonKozachuk (approved):

PR #797, opened 2022-08-05
chore: Setup and run Prettier 

tatomyr:
## What/Why/How?

Ran Prettier on all files.
Also I set up Github action to prevent styling deviation.

## Reference

Closes https://github.com/Redocly/redocly-cli/pull/787
Closes https://github.com/Redocly/redocly-cli/pull/788

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -90,3 +90,17 @@ jobs:
           skip-step: none
           annotations: none
           test-script: npm run jest
+
+  prettier-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.npm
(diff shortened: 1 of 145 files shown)


RomanHotsiy (approved):

adamaltman (commented):
What about adjusting the `.prettierrc.json` file to add  `"trailingComma": "all"` support?

tatomyr:
@adamaltman we discussed this internally and decided it doesn't make too much sense to have it enabled. Trailing commas are helpful for extensible structures like arrays and objects, so one can add/remove an item, and the diff will be only the item itself. But for argument lists (which essentially are tuples and meant to have a rigid length) it's better to have uglier diffs whenever they have been appended or reduced to attract more attention - they are better noticed this way. 

I'm going to merge this as it is, but if we decide to change the approach, next time it will be a lot easier to migrate as now we have the 'prettier check' enabled 🙂.

PR #798, opened 2022-08-08
fix: incorrect return codes when exit with error

### LINKED ISSUE #547: Incorrect return code when bundling an invalid JSON file
**Describe the bug**
When attempting to bundle a file that isn't valid JSON, the bundle command exits with an error:

```
Failed to parse entrypoint definition at invalid.json:

  - unexpected end of the stream within a flow collection in "[...]invalid.json" (2:1).
```

However the return code is zero, suggesting success. This makes it difficult to integrate this with other scripts/tools.

**To Reproduce**
1. Create an invalid.json file. To oversimplify, a file containing just the following demonstrates the issue:
```
{
```
2. Attempt to bundle this: `npx @redocly/openapi-cli bundle invalid.json --output out.json --ext json`. This will fail (as expected):
```
npx @redocly/openapi-cli bundle invalid.json --output out.json --ext json
bundling invalid.json...
Failed to parse entrypoint definition at invalid.json:

  - unexpected end of the stream within a flow collection in "[...]invalid.json" (2:1).

```
3. Check the return code. This will be zero even though it failed:
```
echo $?
0
```

**Expected behavior**
I expect a non-zero return code as the command didn't finish successfully. Other failure cases (such as attempting to bundle valid JSON that isn't an OpenAPI document) return a non-zero code as expected.

**Logs**
n/a

**OpenAPI definition**
n/a

**`openapi-cli` Version(s)**
1.0.0-beta.80

**`Node.js` Version(s)**
v16.12.0

**Additional context**
n/a

AntonKozachuk:
## What/Why/How?
Resolve an issue where we always return a 0 exit code, even when the result is not successful. 
Therefore, we decided to return code 1 in the handleError function to cover move cases then mentioned in the initial issue.

## Reference
Fixed #547

## Testing
Tested locally.

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/__tests__/commands/bundle.test.ts
+++ b/packages/cli/src/__tests__/commands/bundle.test.ts
@@ -63,6 +63,12 @@ describe('bundle', () => {
   it('bundles definitions w/ linting', async () => {
     const entrypoints = ['foo.yaml', 'bar.yaml', 'foobar.yaml'];
 
+    (getTotals as jest.Mock).mockReturnValue({
+      errors: 0,
+      warnings: 0,
+      ignored: 0,
+    });
+
     await handleBundle(
       {
         entrypoints,
@@ -74,7 +80,7 @@ describe('bundle', () => {
     );
 
     expect(lint).toBeCalledTimes(entrypoints.length);
-    expect(bundle).toBeCalledTimes(0);
+    expect(bundle).toBeCalledTimes(entrypoints.length);
   });
 
   it('exits with code 0 when bundles definitions w/linting w/o errors', async () => {
--- a/packages/cli/src/__tests__/commands/bundle.test.ts
+++ b/packages/cli/src/__tests__/commands/bundle.test.ts
@@ -139,7 +139,7 @@ describe('bundle', () => {
     expect(handleError).toHaveBeenCalledWith(new Error('Invalid definition'), 'invalid.json');
   });
 
-  it('handleError isn\'t called when bundles a valid definition', async () => {
+  it("handleError isn't called when bundles a valid definition", async () => {
     const entrypoints = ['foo.yaml'];
 
     (getTotals as jest.Mock).mockReturnValue({
--- a/packages/cli/src/__tests__/commands/bundle.test.ts
+++ b/packages/cli/src/__tests__/commands/bundle.test.ts
@@ -1,6 +1,7 @@
 import { lint, bundle, getTotals, getMergedConfig } from '@redocly/openapi-core';
 
 import { handleBundle } from '../../commands/bundle';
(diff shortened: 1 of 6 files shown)


packages/cli/src/__tests__/commands/bundle.test.ts
```diff
@@ -73,7 +74,7 @@ describe('bundle', () => {
     );
 
     expect(lint).toBeCalledTimes(entrypoints.length);
-    expect(bundle).toBeCalledTimes(entrypoints.length);
+    expect(bundle).toBeCalledTimes(0);
```
tatomyr, line 77: Why do we expect 0 times called bundle here?

packages/cli/src/__tests__/commands/bundle.test.ts
```diff
@@ -24,7 +25,7 @@ describe('bundle', () => {
   afterEach(() => {
     (lint as jest.Mock).mockClear();
     (bundle as jest.Mock).mockClear();
-    (getTotals as jest.Mock).mockClear();
+    (getTotals as jest.Mock).mockReset();
```
tatomyr, line 28: Why do we reset instead of clear? Do we maybe want to do the reset on the previous two as well?

AntonKozachuk: With `mockReset` here we pass the exit code from one test to the next one.

AntonKozachuk: `mockReset` does everything that [mockFn.mockClear()](https://jestjs.io/docs/mock-function-api#mockfnmockclear) does, and also removes any mocked return values or implementations.

packages/cli/src/__tests__/commands/bundle.test.ts
AntonKozachuk: Because the func fails before we even get to the bundle and therefore bundle can't be called.

tatomyr: But the point of the test is to ensure it will bundle & lint the files. You should either fix the test so it doesn't  fail or change the description.

packages/cli/src/__tests__/commands/bundle.test.ts
tatomyr: > With mockReset here we pass the exit code from one test to the next one.

Looks like you want to do the opposite: to NOT pass the exit code through tests 🙂 Am I not correct?

AntonKozachuk: Yes, you're right.

tatomyr (approved):

Issue #793, opened 2022-08-04
Change rules configuration options to default false


adamaltman:
**Is your feature request related to a problem? Please describe.**

Consistency is important in design -- that's the main use case of the `lint` command for API design. In this case, our options are designed inconsistently in 3 rules where a property has a default `true` value instead of a default `false` value like all other.

**Describe the solution you'd like**

Change rules config from `disallowAdditionalProperties` default `true` to `allowAdditionalProperties` default `false`.

The old config should be accepted, but a deprecation warning should show in the logs to encourage someone to change to `allowAdditionalProperties`. If both properties are defined, you can error.

**Describe alternatives you've considered**

Doing nothing. I think the time to do something is now before the GA release.

**Additional context**

- https://redocly.com/docs/cli/rules/no-invalid-media-type-examples/
- https://redocly.com/docs/cli/rules/no-invalid-parameter-examples/
- https://redocly.com/docs/cli/rules/no-invalid-schema-examples/

PR #800, opened 2022-08-09
feat: add allowAdditionalProperties option to rules

IgorKarpiuk:
## What/Why/How?

- Add allowAdditionalProperties to built in rules
- Mark disallowAdditionalProperties as deprecated

## Reference

Close: https://github.com/Redocly/redocly-cli/issues/793

## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/common/no-invalid-parameter-examples.ts
```diff
@@ -1,9 +1,9 @@
 import { UserContext } from '../../walk';
 import { Oas3Parameter } from '../../typings/openapi';
-import { validateExample } from '../utils';
+import { getAdditionalPropertiesOption, validateExample } from '../utils';
 
 export const NoInvalidParameterExamples: any = (opts: any) => {
-  const disallowAdditionalProperties = opts.disallowAdditionalProperties ?? true;
+  let allowAdditionalProperties = getAdditionalPropertiesOption(opts) ?? false;
```
tatomyr, line 6: Why do you need `let` here?

packages/core/src/config/minimal.ts
```diff
@@ -36,7 +36,7 @@ export default {
   oas3_0Rules: {
```
SmoliyY, line 36: Please take a look, also on `oas3_1Rules` set, maybe we want to add this possibility there too?

packages/core/src/rules/common/no-invalid-parameter-examples.ts
IgorKarpiuk: Fixed, thanks

packages/core/src/config/minimal.ts
IgorKarpiuk: @RomanHotsiy @adamaltman, oas 3.0 has a "no-invalid-media-type-examples" rule by default, but 3.1 doesn't have this rule by default. Do we need to add it to minimal and recommended for 3.1?, it's only for 3.0 now.

RomanHotsiy: Yes, let's add it.

packages/core/src/config/minimal.ts
```diff
@@ -47,6 +47,10 @@ export default {
     'no-servers-empty-enum': 'error',
   },
   oas3_1Rules: {
+    'no-invalid-media-type-examples': {
+      severity: 'warn',
+      allowAdditionalProperties: false,
+    },
```
adamaltman, line 53:
```suggestion
    'no-invalid-media-type-examples': 'warn',
```
Won't `allowAdditionalProperties` be `false` by default?

tatomyr (approved):

packages/core/src/config/minimal.ts
IgorKarpiuk: Yes, it is false by default, but do we need to turn it off in mininal? @adamaltman

Issue #702, opened 2022-05-24
Assertions severity off is considered invalid


adamaltman:
**Describe the bug**

The value `off` isn't accepted but it is described as an acceptable severity level.

  - Invalid severity level; accepted values: error or warn.

It also results in a hard (not elegant) crash.

**To Reproduce**
1. Given this `.redocly.yaml` file with

    ```yaml
        assert/properties-include-examples:
          subject: SchemaProperties
          property: example
          defined: true
          nonEmpty: true
          severity: off
    ```
2. See error

**Expected behavior**

It should work with off. However, if there is a value someone didn't expect, then don't do a hard crash.

**Logs**

```
  - Invalid severity level; accepted values: error or warn.

/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/node_modules/@redocly/openapi-core/lib/format/format.js:148
            throw new Error(ERROR_MESSAGE.INVALID_SEVERITY_LEVEL);
                  ^

Error: Invalid severity level; accepted values: error or warn
    at getBgColor (/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/node_modules/@redocly/openapi-core/lib/format/format.js:148:19)
    at formatCodeframe (/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/node_modules/@redocly/openapi-core/lib/format/format.js:153:25)
    at Object.formatProblems (/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/node_modules/@redocly/openapi-core/lib/format/format.js:71:41)
    at Object.<anonymous> (/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/lib/commands/lint.js:52:36)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/adam/.nvm/versions/node/v16.10.0/lib/node_modules/@redocly/cli/lib/commands/lint.js:5:58)
```

**OpenAPI definition**
If applicable, add an OpenAPI definition and `.redocly.yaml` configuration file that helps reproduce the problem.
At a minimum, please state the specification version(s) you're using (e.g. 2.0, 3.0, 3.1).

**Redocly Version(s)**
What version of Redocly CLI are you using?

**`Node.js` Version(s)**
What version of `node.js` are you using?

**Additional context**

adamaltman:
Raising priority so that docs can be simplified.

PR #806, opened 2022-08-15
fix: add assertions severity 'off'

tatomyr:
## What/Why/How?

Closes https://github.com/Redocly/redocly-cli/issues/702

Also adjusted tsconfig as it was extremely annoying to follow it while developing.

## Check yourself

- [x] Code is linted

## Security


tsconfig.json
```diff
@@ -8,8 +8,6 @@
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "resolveJsonModule": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
```
SmoliyY, line 12: I would leave this checks, or there are some specific reason for removing it?

tatomyr: Having unused variables does no harm, and in my experience it is very inconvenient to fix the code according to those checks every time I want to make some experimental  changes (e. g. commenting out some existing code). Moreover, TS still highlights unused variables (shows warnings). So it's just a matter of convenience for me. 
If it's only me who has this perception, I'll put them back 🙂.

tatomyr: Will move this out of the scope.

SmoliyY (approved):

PR #809, opened 2022-08-16
feat: create config dynamically

roman-sainchuk:
## What/Why/How?

In case we are running `openapi-core` in environments not having access to file system we have to be able to create config from string or object.

## Check yourself

- [ ] Code is linted

## Security


packages/core/src/config/load.ts
```diff
@@ -102,3 +108,21 @@ export async function getConfig(configPath: string | undefined = findConfig()):
...
+export async function createConfig (rawConfig: any, options: CreateConfigOptions = {}): Promise<Config> {
+  if (typeof rawConfig === 'string') {
+    rawConfig = parseYaml(rawConfig);
+  }
+
+  return addConfigMetadata({
+    rawConfig: transformConfig(rawConfig),
+    ...options
+  });
+}
```
tatomyr, line 128: I'd avoid overloading.

Maybe union types would work better?

```ts
export async function createConfig (rawConfig: rawConfig | string, options: CreateConfigOptions = {}): Promise<Config> {
 ...
}


```

It's shorter and it's JavaScript.

roman-sainchuk: agree

packages/core/src/config/load.ts
```diff
@@ -111,18 +111,19 @@ export async function getConfig(configPath: string | undefined = findConfig()):
...
-export async function createConfig (rawConfig: RawConfig, options?: CreateConfigOptions): Promise<Config>;
-export async function createConfig (rawConfig: any, options: CreateConfigOptions = {}): Promise<Config> {
-  if (typeof rawConfig === 'string') {
-    rawConfig = parseYaml(rawConfig);
+export async function createConfig(
+  config: string | RawConfig,
+  options?: CreateConfigOptions
+): Promise<Config> {
+  if (typeof config === 'string') {
+    config = parseYaml(config) as RawConfig;
```
tatomyr, line 122: Can you avoid mutating the `config` argument?

roman-sainchuk: Changed

packages/core/src/config/load.ts
roman-sainchuk: Changed

packages/core/src/config/load.ts
```diff
@@ -102,3 +108,20 @@ export async function getConfig(configPath: string | undefined = findConfig()):
     throw new Error(`Error parsing config file at '${configPath}': ${e.message}`);
   }
 }
+
+interface CreateConfigOptions {
```
tatomyr, line 112: We usually use `type` notation as more general.

```suggestion
type CreateConfigOptions = {
```

tatomyr (approved):

PR #811, opened 2022-08-16
feat: openapi-core in browser support

roman-sainchuk:
## What/Why/How?
Make `openapi-core` browser compatible.

## Check yourself

- [x] Code is linted

## Security


packages/core/src/colorize.ts
```diff
@@ -0,0 +1,17 @@
+import * as colorette from 'colorette';
+
+import { isBrowser } from './utils';
+
+const identity = (string: string): string => string;
```
tatomyr, line 5:
```suggestion
const identity = <T>(value: T): T => value;
```

packages/core/src/colorize.ts
```diff
@@ -0,0 +1,17 @@
+import * as colorette from 'colorette';
```
tatomyr, line 1: Can we safely import `colorette` in browser?

roman-sainchuk: Unfortunately no. It requires `Node >= 10` and has no support in browser.

packages/core/src/format/format.ts
```diff
@@ -177,7 +168,7 @@ export function formatProblems(
         fingerprint: `${p.ruleId}${p.location.length > 0 ? '-' + p.location[0].pointer : ''}`,
       };
     });
-    process.stdout.write(JSON.stringify(issues, null, 2));
+    logger.info(JSON.stringify(issues, null, 2));
```
tatomyr, line 171: I believe this doesn't make sense in the browser environment. We do both `info` and `error` into `process.stderr`. As far as I remember we only use `stdout` to convey some data in Workflows.

roman-sainchuk: but `error` and `warn` write to `stderr`

packages/core/src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+  process.stdout.write(...args);
+}
+class Logger {
+  info(...args: StdArgs) {
+    return isBrowser() ? console.log(...args) : stdout(...args);
+  }
+
+  warn(...args: StdArgs) {
+    return isBrowser() ? console.warn(...args) : stderr(...args);
+  }
```
tatomyr, line 34: Do we use `warn` level somewhere? If not, I'd refrain of introducing it.

packages/core/src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  error(...args: StdArgs) {
+    return isBrowser() ? console.error(...args) : stderr(...args);
+  }
+}
+
+export const logger = new Logger();
```
tatomyr, line 41: Do we really need the `class` here? A simple object literal wold work just as well I believe.

packages/core/src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  warn(...args: StdArgs) {
+    return isBrowser() ? console.warn(...args) : stderr(...args);
+  }
+
+  error(...args: StdArgs) {
+    return isBrowser() ? console.error(...args) : stderr(...args);
+  }
+}
```
tatomyr, line 39: Maybe we want to apply a default colorize here to enforce consistency?

packages/core/src/format/format.ts
tatomyr: I mean, we keep `stdout` only for some technical purposes, and put any user messages to `stderr`.

packages/core/src/logger.ts
roman-sainchuk: I decided to use `class` for extending purposes only (to be able to override some specific severity level method)

packages/core/src/format/format.ts
roman-sainchuk: should I create helper like `output` with one method `write` ? Which would write to logger in browser env ?

tatomyr: That makes sense.

packages/core/src/logger.ts
roman-sainchuk: We have 2 places in `openapi-core` where warnings are shown. Would it be ok if I change the style of these messages from black or white (depending on the system preferences of your terminal) to yellow (in case of warnings)?
Was:

image

Now: 

image

packages/core/src/colorize.ts
tatomyr: Then maybe we want to do a conditional import here? Something like this:
```ts 
const colorette = await isBrowser ? ... : import('colorette')
```

roman-sainchuk: we cannot do it on top level because dynamic import returns `Promise` and we cannot `await` it in `commonjs` module system

packages/core/src/config/config-resolvers.ts
```diff
@@ -71,6 +71,11 @@ export function resolvePlugins(
 ): Plugin[] {
   if (!plugins) return [];
 
+  if (isBrowser()) {
+    // TODO: show warning that plugins are not supported in browser
```
tatomyr, line 75: Can we at least put `console.error` here?

roman-sainchuk: This is not an issue to put `logger.warn` but im not sure about the condition (we almost always have at least one default plugin, would it be correct to check for other plugins where `plugin.id !== ''` or just for presence of `plugin.id`) ?

packages/core/src/logger.ts
tatomyr: If so, you can check if it's a browser or not in `constructor` or even pass it as an argument like so: 
```ts
new Logger(isBrowser())
```

packages/core/src/utils.ts
```diff
@@ -223,3 +229,17 @@ export type Falsy = undefined | null | false | '' | 0;
...
+
+export type StdArgs = [string | Uint8Array];
+
+export function stderr(...args: StdArgs) {
+  process.stderr.write(...args);
+}
+
+export function stdout(...args: StdArgs) {
+  process.stdout.write(...args);
+}
```
tatomyr, line 245: I believe we don't really need these façades. There're used each in one place only and shouldn't be further reused.

packages/core/src/colorize.ts
roman-sainchuk: > Can we safely import colorette in browser?

Yes, we can
Bu we cannot safely use it

tatomyr (approved):
Looks great!

PR #813, opened 2022-08-17
fix: rename rule no-empty-enum-servers to no-server-variables-empty-enum

### LINKED ISSUE #812: Mismatch between rules ids in docs and in-code
**Describe the bug**

https://redocly.com/docs/cli/rules/no-server-example-com/
https://redocly.com/docs/cli/rules/no-empty-enum-servers/

In these two doc pages the rule id documented is different to what is actually used in code. 
When using  `no-server-example-com`, in-code the id is `no-server-example.com`
When using `no-empty-enum-servers` in-code the id is `no-servers-empty-enum`

Because of this, when using the documented rule ids, and running `redocly lint`, the output will throw the `<Property> not expected here` warning.
Also, these rules are not validated given that the rule id was not found.

**To Reproduce**
1. Given this `redocly.yaml` file:
```---
styleguide:
  extends:
    - recommended
  rules:
    no-empty-enum-servers: error
    no-server-example-com: error
```
3. And any OpenAPI file
4. Run this command with these arguments... `redocly lint`
5. See error

**Expected behavior**
Either documentation gets updated to use the correct in-code rule ids, or in-code rule ids get updated to match documentation

**Logs**
```
[1] redocly.yaml:8:5 at #/styleguide/rules/no-empty-enum-servers

Property `no-empty-enum-servers` is not expected here.

 6 |   - recommended
 7 | rules:
 8 |   no-empty-enum-servers: error
 9 |   no-server-example-com: error
10 |

Warning was generated by the configuration spec rule.


[2] redocly.yaml:9:5 at #/styleguide/rules/no-server-example-com

Property `no-server-example-com` is not expected here.

 7 | rules:
 8 |   no-empty-enum-servers: error
 9 |   no-server-example-com: error
10 |

Warning was generated by the configuration spec rule.


You have 2 warnings.
validating /test.json...
/test.json: validated in 19ms

```

**OpenAPI definition**

OpenAPI definition: any
redocly.yaml file:
```
---
apis:
  test:
    root: ./test.json
styleguide:
  extends:
    - recommended
  rules:
    no-empty-enum-servers: error
    no-server-example-com: error
```

**Redocly Version(s)**
1.0.0-beta.106

**`Node.js` Version(s)**
v18.2.0

**Additional context**
N/A

adamaltman:
## What/Why/How?

- rename `no-empty-enum-servers` to `no-server-variables-empty-enum` and adjust corresponding code and docs. The current live docs are completely incorrect and wouldn't function, so it is safer to rename (the actual working one would have been `no-servers-empty-enum` but it wasn't documented that way due to the strange naming in the source code: https://redocly.com/docs/cli/rules/no-empty-enum-servers/ -- the changelog should note the rename.
- docs: fix `no-server-example.com` docs

## Reference

- Fixes #812 

## Testing

Passed.

## Check yourself

- [x] Code is linted
- [ ] **N/A** Tested with redoc/reference-docs/workflows

## Security


RomanHotsiy (approved):
This looks good.

The only concert is it is it technically a breaking change. Someone could have those rules ignored.

But I think this is okay.

tatomyr (approved):

adamaltman:
Are we sure we don't want to add a commit to the CHANGELOG in an unreleased section at the top? Anyway, here is what I propose for simplicity during the release.

Changes
- Renamed `no-servers-empty-enum` to `no-server-variables-empty-enum` and fixed incorrect docs of `no-empty-enum-servers`.

Fixes
- Fixed docs for no-server-example.com

Issue #750, opened 2022-07-05
Incorrect schema description dereferenced 


lucianojs:
**Describe the bug**
Incorrect schema description dereferenced 

**To Reproduce**

I use de sample specification:

```
openapi: 3.1.0
info:
  version: 1.0.0
  title: Example.com
  termsOfService: https://example.com/terms/
  contact:
    email: contact@example.com
    url: http://example.com/contact
  license:
    name: Apache 2.0
    url: http://www.apache.org/licenses/LICENSE-2.0.html
  description: |
    This is an **example** API to demonstrate features of the OpenAPI
    specification.
servers:
  - url: https://example.com/api/v1
paths:
  /users:
    post:
      summary: user
      description: User.
      operationId: createUser
      responses:
        '200':
          description: OK
      requestBody:
        description: Updated user object
        content:
          application/json:
            schema:
              description: Names and Numbers (specific)
              $ref: '#/components/schemas/NamesAndNumbers'
        required: true
components:
  schemas:
    Name:
      type: string
      description: Generic Name.
    Number:
      type: integer
    Names:
      type: object
      description: names description
      properties:
        oneName:
          $ref: "#/components/schemas/Name"
          description: One name (specific).
        otherName:
          $ref: "#/components/schemas/Name"
          description: Other name (specific).
    Numbers:
      type: object
      description: numbers description
      properties: 
        oneNumber:
          $ref: "#/components/schemas/Number"
          description: One number (specific)
        otherNumber:
          $ref: "#/components/schemas/Number"
          description: Other number (specific)
    NamesAndNumbers:
      type: object
      description: names and numbers description
      properties: 
        names: 
          $ref: "#/components/schemas/Names"
        numbers:
          $ref: "#/components/schemas/Numbers"
```

When I run this command:

`redocly bundle --dereferenced --remove-unused-components test.yaml -o test2.yaml`

The result is:

```
openapi: 3.1.0
info:
  version: 1.0.0
  title: Example.com
  termsOfService: https://example.com/terms/
  contact:
    email: contact@example.com
    url: http://example.com/contact
  license:
    name: Apache 2.0
    url: http://www.apache.org/licenses/LICENSE-2.0.html
  description: |
    This is an **example** API to demonstrate features of the OpenAPI
    specification.
servers:
  - url: https://example.com/api/v1
paths:
  /users:
    post:
      summary: user
      description: User.
      operationId: createUser
      responses:
        '200':
          description: OK
      requestBody:
[...]

PR #814, opened 2022-08-17
fix: description dereferenced

iosypov:
## What/Why/How?

Fix: #750. Make fix in object assign. Rewrite all info from parent, additional from $ref.

## Screenshots (optional)

image

## Check yourself

- [x] Code is linted

## Security


packages/core/src/bundle.ts
```diff
@@ -304,7 +304,8 @@ function makeBundleVisitor(
     } else {
       // @ts-ignore
       delete ref.$ref;
-      Object.assign(ref, resolved.node);
+      const obj = Object.assign({}, resolved.node, ref);
+      Object.assign(ref, obj);
```
RomanHotsiy, line 308:
```suggestion
      Object.assign(ref, resolved.node, ref); // assign ref itself again so ref fields take precedence
```

RomanHotsiy (approved):
We need to add tests for this case too.

tatomyr (approved):

packages/core/src/bundle.ts
iosypov: It's wrong, because we should get first val info from resolved.node.

RomanHotsiy: It's the same. I don't see how it's different. Can you show me some counter-example?

RomanHotsiy (approved):

Issue #757, opened 2022-07-07
Examples field is not validated against openAPI 3.0.0 spec


azimbelm:
**Describe the bug**
No validation rule applies on the "Examples" field name defined in
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#exampleObject
--> All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.

**To Reproduce**
I created an openapi.yaml containing the following lines

```
parameters:
    - name: identifier
    - required: true
          in: path
          allowEmptyValue: false
          description: Some description
          examples:
            valid identifier:
              description: 'Some description'
              value: '49039557299384'
```

Running 'redocly lint openapi.yaml' returns that the file is valid.

The field name '**valid identifier**' within the examples is invalid but I could not find any rule detecting this.
An custom built assertion rule did not work because it only works for the value of a given property.

**Expected behavior**
An error should be shown, that the 'valid identifier' is invalid

**OpenAPI definition**
 .redocly.yaml
```
lint:
  extends:
    - all
```

Specification version used: 3.0.0

**Redocly Version(s)**
**v1.0.0-beta.103**

**`Node.js` Version(s)**
**v16.12.0**

tatomyr:
Thanks @azimbelm for pointing this out. 
This would have been possible to validate via an assertion like this:
```yaml
assert/example-name-format:
      subject: ExampleMap
      pattern: ^[a-zA-Z0-9.-_]+$
```

However, at the moment `ExampleMap` and a few others aren't available. I think this should be fixed. Will create an issue for that.

azimbelm:
Workaround for anyone who needs this now:

Create a Plugin/Custom Rule (use this guide: https://redocly.com/docs/cli/guides/response-contains-property/#response-contains-property.js)  and use this code:

```
function ExampleKeyMatchesOASSpec () {
  return {
    Example(example, { key, report, location }) {
      if(new RegExp('^[a-zA-Z0-9\\.\\-_]+$').test(key) == false){
      //if (example.value || example.externalValue) {
        report({
          message: 'The key of the example "'+key+'" does not match the regular expression "^[a-zA-Z0-9\\.\\-_]+$"',
          location: location.key(),
        });
      }
  }
}}
```

adamaltman:
Shouldn't we add this to the `spec` rule (cc @tatomyr)?

RomanHotsiy:
[...]

PR #815, opened 2022-08-19
feat: add rule for validation field names inside components

IgorKarpiuk:
## What/Why/How?

Add rule for validation field names inside components

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/757

## Screenshots (optional)
Знімок екрана 2022-08-31 о 12 48 52

## Check yourself

- [x] Code is linted

## Security


IgorKarpiuk:
@RomanHotsiy @adamaltman Please correct me with the name of the rule if needed

adamaltman (commented):
Three things:
1. Rename to `spec-examples-invalid-map-name`
2. Improve the message template
3. Add to minimal, recommended, and all configurations
4. Add documentation


I think this should be renamed to `spec-examples-invalid-map-name` because the `Example Object` does have field names and this rule isn't about it at all.

> examples  "Map [string]"

I think this one:
> "message": "The key of the example \\"invalid identifier\\" does not match the regular expression \\"^[a-zA-Z0-9.\\\\-_]+$\\"",
should change to this:
> "message": "The map key in examples \\"invalid identifier\\" does not match the regular expression \\"^[a-zA-Z0-9.\\\\-_]+$\\"",


This rule should be added to the minimal and recommended configs with a `warn` level.

Also, when we add a rule, we need to add the docs. There are two things involved:

- adding a file to the `docs/rules` directory (it should follow the structure of the other files in that directory with the exception of assertions.md -- maybe check spec.md since that would be the closest)
- updating `rules.md` list with this rule in the appropriate spot

packages/core/src/rules/oas3/spec-components-invalid-map-name.ts
```diff
@@ -0,0 +1,53 @@
...
+      },
+      SecurityScheme(_node, { key, report, location }: UserContext) {
+        validateKey(key, report, location, 'securitySchemas');
+      },
+      Link(_node, { key, report, location }: UserContext) {
+        validateKey(key, report, location, 'links');
+      },
+      Callback(_node, { key, report, location }: UserContext) {
+        validateKey(key, report, location, 'callbacks');
+      },
```
tatomyr, line 50: It looks like it's better to use `any()` visitor and manually check whether we're inside of the `Components`.
Maybe we can go even more strict and require ALL keys to conform to this pattern? @adamaltman  do you think it makes sense?

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -0,0 +1,285 @@
+import { makeConfig, parseYamlToDocument } from '../../../../__tests__/utils';
+import { outdent } from 'outdent';
+import { lintDocument } from '../../../lint';
+import { BaseResolver } from '../../../resolve';
+
+describe('Oas3 spec-components-invalid-map-name', () => {
+  it('should report a key of the parameters does not match the regular expression', async () => {
```
tatomyr, line 7: Most of the tests are checking the same thing. You can left only unique ones.

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -0,0 +1,285 @@
...
+  });
+
+  it('should report a key of the examples does not match the regular expression', async () => {
+    const document = parseYamlToDocument(outdent`
+      openapi: 3.0.0
+      info:
+        version: 3.0.0
+      components:
+        examples:
+          valid identifier:
```
tatomyr, line 71: It doesn't look like a valid identifier :)

packages/core/src/rules/oas3/spec-components-invalid-map-name.ts
adamaltman: Well, look at this:

```yaml
    patternProperties:
      '^(schemas|responses|parameters|examples|requestBodies|headers|securitySchemes|links|callbacks|pathItems)$':
        $comment: Enumerating all of the property names in the regex above is necessary for unevaluatedProperties to work as expected
        propertyNames:
          pattern: '^[a-zA-Z0-9._-]+$'
```

I think we should check those (schemas, responses, parameters, examples, etc...). I'm not sure if that's `any()`, but it is a lot. Also, the examples map names don't only appear inside of `Components` but can be directly inside of operations too (probably the same is true with some other things).

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      info:
+        version: 3.0.0
+      components:
+        parameters:
+          my Param:
+            name: param
+            description: param
+            in: path
+        examples:
+          valid identifier:
```
adamaltman, line 93: isn't this an invalid identifier?

packages/core/src/rules/oas3/spec-components-invalid-map-name.ts
adamaltman: Well, yes, so it looks like this code misses this scenario where the map is not used in `Components`:

```yaml
openapi: 3.1.0
info: 
  title: Path examples bad map
  version: error-case
paths:
  /foo:
    get: 
      responses: 
        '200': 
          description: OK
          content: 
            application/json: 
              schema: 
                type: object
              examples:
                My invalid example:
                  value: 
                    foo: bar
                A_valid_example:
                  value: 
                    baz: quz
  
```

adamaltman: Wow, this is not specified: 

https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml

Seems like a bug in the schema. I opened up this issue: https://github.com/OAI/OpenAPI-Specification/issues/3013

IgorKarpiuk: @tatomyr It is related only to specific fields according specification https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#componentsObject

tatomyr: Thanks for pointing to the spec @IgorKarpiuk. Then your code makes sense. 

The only open question that remains is whether this restriction should be imposed on `Components` only or not. 

Meanwhile, I think it's okay to go with the current approach.

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
IgorKarpiuk: Fixed, thanks

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -0,0 +1,217 @@
...
+            description: bad request   
+        schemas: 
+          first schema:
+            type: integer
+            format: int64     ",
+                "mimeType": undefined,
+              },
+            },
+          ],
+          "message": "The map key in parameters \\"my Param\\" does not match the regular expression \\"^[a-zA-Z0-9.\\\\-_]+$\\"",
```
tatomyr, line 71: I believe it should refer a slightly different regex according to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#componentsObject): `^[a-zA-Z0-9\.\-_]+$`.

tatomyr (approved):
Left one more minor comment but overall it looks good to me.

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
IgorKarpiuk: ✅
Знімок екрана 2022-08-31 о 12 48 52

packages/core/src/rules/oas3/spec-components-invalid-map-name.ts
adamaltman: I agree. We can see or adjust it based on any responses here: https://github.com/OAI/OpenAPI-Specification/issues/3013

packages/core/src/config/recommended.ts
```diff
@@ -45,6 +45,7 @@ export default {
     'no-unused-components': 'warn',
     'no-undefined-server-variable': 'error',
     'no-server-variables-empty-enum': 'error',
+    'spec-components-invalid-map-name': 'warn',
```
adamaltman, line 48:
```suggestion
    'spec-components-invalid-map-name': 'error',
```

packages/core/src/config/recommended.ts
```diff
@@ -55,5 +56,6 @@ export default {
...
```
adamaltman, line 59:
```suggestion
    'spec-components-invalid-map-name': 'error',
```

adamaltman (changes requested):
This violates the spec and should be an error in the `recommended` ruleset.

adamaltman (approved):

Issue #392, opened 2021-10-07
`problem-details-rfc7807` built-in rule


adamaltman:
**Is your feature request related to a problem? Please describe.**

This is a common format. It would be good to enforce this with a single line in the `.redocly.yaml` file.

**Describe the solution you'd like**

`problem-details-rfc7807: error`

Enforcing a specific shape of the response for 4xx errors, such as following the `application/problem+json` https://datatracker.ietf.org/doc/html/rfc7807 format.

**Describe alternatives you've considered**
None

**Additional context**
None

iosypov:
@adamaltman i have on question. Should we only validate when response type `application/problem+json` or we should show error when in 4xx response type is not `application/problem+json` and then validate details by spec rfc7807?

RomanHotsiy:
I would do both.

adamaltman:
@RomanHotsiy do you mean by some configuration variable?

I think it would validate that all 4xx response types are content type `application/problem+json`.

RomanHotsiy:
But should we also validate the response schema? Like check if it has message, etc.?

PR #816, opened 2022-08-19
Feat: add rfc7807 rule

iosypov:
## What/Why/How?

Fix: #392 According the issue added new rules for check  rfc7807. https://datatracker.ietf.org/doc/html/rfc7807 

## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/oas3/operation-4xx-problem-details-rfc7807.ts
```diff
@@ -0,0 +1,42 @@
...
+export const Operation4xxProblemDetailsRfc7807: Oas3Rule = () => {
+  return {
+    ResponsesMap(responses: Record<string, object>, { report, location }: UserContext) {
+      const codes = Object.keys(responses);
+      for (const code of codes) {
+        if (!/4[Xx0-9]{2}/.test(code)) continue;
+        const response = responses[code] as Oas3Response;
+        if (!response.content || !response.content['application/problem+json']) {
+          report({
+            message: 'Response `4xx` must have content type `application/problem+json`.',
```
adamaltman, line 18:
```suggestion
            message: 'Response `4xx` must have content-type `application/problem+json`.',
```

adamaltman (commented):
Good start!

tatomyr:
Please add this to all rules set: `core/src/config/all.ts`.

iosypov:
> Please add this to all rules set: `core/src/config/all.ts`.

@tatomyr , done!

packages/core/src/rules/oas2/__tests__/operation-4xx-problem-details-rfc7807.test.ts
```diff
@@ -0,0 +1,111 @@
+import { outdent } from 'outdent';
+import { lintDocument } from '../../../lint';
+import { parseYamlToDocument, replaceSourceWithRef, makeConfig } from '../../../../__tests__/utils';
+import { BaseResolver } from '../../../resolve';
+
+describe('Oas2 operation-4xx-problem-details-rfc7807', () => {
```
tatomyr, line 6: I'm not completely sure we have to introduce this check for **OAS 2** at all. The rule itself is about the `application/problem+json` content type's shape which was introduced in **OAS 3** I believe. 

@adamaltman ,  @RomanHotsiy , what's your opinion on this?

RomanHotsiy: OAS2 had `produces`/`consumes` fields. If this is hard to apply, let's implement it for oas3 only.

adamaltman:
We also need:

- add file `docs/rules/operation-4xx-problem-details-rfc7807.md` following similar page template of `operation-4xx`
- add to docs/rules.md index file with link to details

I can help with that part if you need.

docs/rules/operation-4xx-problem-details-rfc7807.md
```diff
@@ -0,0 +1,84 @@
...
+readable details of errors in a HTTP response to avoid the need to
+define new error response formats for HTTP APIs.
+
+Every operation with (400-499) HTTP response should have content-type `application/problem+json` and fields `title` and `type`. According to [specification](https://datatracker.ietf.org/doc/html/rfc7807).
+
+## Configuration
+
+| Option   | Type   | Description                                                                               |
+| -------- | ------ | ----------------------------------------------------------------------------------------- |
+| severity | string | Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
```
adamaltman, line 23: Hmmm, I don't think we should warn on this one in the recommended. I think it should be `off` in the recommended config.

iosypov: I will change this. Please, review all document

packages/core/src/rules/oas3/operation-4xx-problem-details-rfc7807.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Validation according rfc7807 - https://datatracker.ietf.org/doc/html/rfc7807
+ */
+export const Operation4xxProblemDetailsRfc7807: Oas3Rule = () => {
+  return {
+    ResponsesMap(responses: Record<string, object>, { report, location }: UserContext) {
+      const codes = Object.keys(responses);
+      for (const code of codes) {
+        if (!/4[Xx0-9]{2}/.test(code)) continue;
+        const response = responses[code] as Oas3Response;
+        if (!response.content || !response.content['application/problem+json']) {
```
RomanHotsiy, line 16: Response here can be a `$ref`. It's better to register on `Response` object and `skip` all non-4xx responses.

packages/core/src/rules/oas3/operation-4xx-problem-details-rfc7807.ts
```diff
@@ -0,0 +1,42 @@
...
+          report({
+            message: 'Response `4xx` must have content-type `application/problem+json`.',
+            location: location.child(code).key(),
+          });
+        }
+      }
+    },
+    Response: {
+      skip(response: Oas3Response) {
+        if (!response.content || !response.content['application/problem+json']) {
```
RomanHotsiy, line 26: I think it's better to register on `MediaType` here.

RomanHotsiy (changes requested):

__tests__/lint-config/invalid-config--no-option/snapshot.js
```diff
@@ -20,9 +20,24 @@ No configurations were defined in extends -- using built in recommended configur
...
+20 |   description: some desc
+21 |
+
+Error was generated by the operation-4xx-problem-details-rfc7807 rule.
+
+
 ../__fixtures__/valid-openapi.yaml: validated in <test>ms
 
-Woohoo! Your OpenAPI definition is valid. 🎉
+❌ Validation failed with 1 error.
```
RomanHotsiy, line 39: why it breaks those tests?

packages/core/src/config/minimal.ts
```diff
@@ -55,5 +56,6 @@ export default {
     'no-unused-components': 'warn',
     'no-undefined-server-variable': 'warn',
     'no-server-variables-empty-enum': 'error',
+    'operation-4xx-problem-details-rfc7807': 'off',
```
RomanHotsiy, line 59: seems like we don't need to add rules with 'off' here.
Looks like it doesn't work correctly and enables those rules in fact. Check the test snapshots updated above. They should not.

RomanHotsiy (changes requested):

docs/rules/operation-4xx-problem-details-rfc7807.md
```diff
@@ -0,0 +1,84 @@
...
+                type: string
+              title:
+                type: string
+```
+
+Example of **correct** operation response:
+
+```yaml
+post:
+  rresponses:
```
adamaltman, line 64:
```suggestion
  responses:
```

RomanHotsiy (approved):
Code looks good so I'm approving it.

It would be great to add at least one e2e test before merging.

docs/rules/operation-4xx-problem-details-rfc7807.md
```diff
@@ -0,0 +1,84 @@
...
+
+| OAS | Compatibility |
+| --- | ------------- |
+| 2.0 | ❌            |
+| 3.0 | ✅            |
+| 3.1 | ✅            |
+
+## API design principles
+
+Problem Details for HTTP APIs as a way to carry machine-
```
adamaltman, line 13:
```suggestion
Problem Details for HTTP APIs are a way to carry machine-
```

docs/rules/operation-4xx-problem-details-rfc7807.md
```diff
@@ -0,0 +1,84 @@
...
+readable details of errors in a HTTP response to avoid the need to
+define new error response formats for HTTP APIs.
+
+Every operation with (400-499) HTTP response should have content-type `application/problem+json` and fields `title` and `type`. According to [specification](https://datatracker.ietf.org/doc/html/rfc7807).
```
adamaltman, line 17:
```suggestion
Every operation with (400-499) HTTP response should have content-type `application/problem+json` and fields `title` and `type` according to the [specification](https://datatracker.ietf.org/doc/html/rfc7807).
```

adamaltman (approved):
I made very minor docs changes and now approving.

Issue #481, opened 2021-12-22
`bundle` command squash referenced markdown file to one string


Boiarshinov:
**Describe the bug**

`bundle` command squash referenced markdown file to one string.
I'm not sure is that a bug or a feature absence 😅 

**To Reproduce**
Given this OpenAPI file:
```yaml
openapi: 3.0.1
info:
  title: Demo API
  version: '1.0'
  description:
    $ref: './description.md'
paths: {}
```

Which has reference to this markdown description:
```markdown
# Heading
List:
- first item
- second item

Line.
Same Line
```

Run this command
```sh
openapi bundle openapi.yaml --output bundled.yaml
```

And look at the squashed description in the bundled file:
```yaml
openapi: 3.0.1
info:
  title: Demo API
  version: '1.0'
  description: "# Heading\r\nList:\r\n- first item\r\n- second item\r\n\r\nLine.\r\nSame Line"
paths: {}
components: {}
```

FYI: swagger-cli have the same behavior

**Expected behavior**
In my opinion it's better not to squash markdown file to single line but to use multiline strings in [literal mode](https://oai.github.io/Documentation/specification-docs.html#providing-long-descriptions-in-yaml).
So the expected bundled file should be:
```yaml
openapi: 3.0.1
info:
  title: Demo API
  version: '1.0'
  description: |
    # Heading
    List:
    - first item
    - second item

    Line.
    Same Line
paths: {}
components: {}
```

**`openapi-cli` Version(s)**
1.0.0-beta.74

**`Node.js` Version(s)**
v16.13.0

iosypov:
I can not reproduce this. 

I tried file:
```yaml
openapi: 3.0.1
info:
  title: Demo API
  version: '1.0'
  description: "# Heading\r\nList:\r\n- first item\r\n- second item\r\n\r\nLine.\r\nSame Line"
paths: {}
components: {}
```

And get result:
```yaml
openapi: 3.0.1
info:
  title: Demo API
  version: '1.0'
  description: |
    # Heading

    List:

    - first item
    - second item

    Line.
    Same Line
paths: {}
components: {}
```

Boiarshinov:
I've opened a PR with instructions how to reproduce an issue
https://github.com/Redocly/redocly-cli/pull/823

PR #823, opened 2022-08-22
issue-481: Add cases to reproduce the bug

Boiarshinov:
Try to run in Linux terminal to reproduce the [issue](https://github.com/Redocly/redocly-cli/issues/481)
```sh
./reproduce.sh
```

### DIFF
--- a/reproduce/referal_text.md
+++ b/reproduce/referal_text.md
@@ -1,0 +1,7 @@
+# Heading
+List:
+- first item
+- second item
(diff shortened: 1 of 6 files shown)


adamaltman:
The issue was fixed, so I think we can close this demo of problem PR now. Thank you @Boiarshinov!

Boiarshinov:
Thanks for the fix!

Issue #462, opened 2021-12-01
$anchors in openapi documents are not properly parsed


karenetheridge:
**Describe the bug**

The presence of an `$anchor` in an openapi specification file generates errors.  Is OpenAPI 3.1 support not yet complete?

**To Reproduce**

With this openapi file:

```
---
openapi: 3.1.0
info:
  title: our application API
  version: 1.0
paths:
  /test-api:
    get:
      responses:
        '200':
          description: success
          content:
            application/json:
              schema:
                $defs:
                  main_data:
                    $anchor: main_data
                    type: object
                type: object
                oneOf:
                - properties:
                    wrapper:
                      $ref: '#main_data'
                - $ref: '#main_data'
```

...either attempt to load the rendered webpage, or run at the command line: `openapi bundle redoc-repro.yaml`

**Expected behavior**

No errors.

**Observed behaviour**

From the above command, I see the errors:

```
Can't resolve $ref: ENOENT: no such file or directory, open '/Users/ether/#main_data'

21 | - properties:
22 |     wrapper:
23 |       $ref: '#main_data'
24 | - $ref: '#main_data'
25 |

Error was generated by the bundler rule.


[2] docs/redoc-repro.yaml:24:19 at #/paths/~1test-api/get/responses/200/content/application~1json/schema/oneOf/1

Can't resolve $ref: ENOENT: no such file or directory, open '/Users/ether/#main_data'

22 |     wrapper:
23 |       $ref: '#main_data'
24 | - $ref: '#main_data'
25 |

Error was generated by the bundler rule.


❌ Errors encountered while bundling redoc-repro.yaml: bundle not created (use --force to ignore errors).
```

I have no redoc configuration file locally.

```
$ openapi --version
1.0.0-beta.69
$ npm --version
7.21.0
$ sw_vers
ProductName:	macOS
ProductVersion:	11.6
BuildVersion:	20G165
```

RomanHotsiy:
@karenetheridge yes, this is an issue in our 3.1 implementation. Thanks for reporting it.

PR #836, opened 2022-08-30
fix: $anchors in openapi documents are not properly parsed

tatomyr:
## What/Why/How?

Closes https://github.com/Redocly/redocly-cli/issues/462

In order to not fail when bundling, `$anchor`s need to be resolved correctly. 
I decided to collect all `$anchors` into `anchorRefsMap` while walking a document.

## Reference

[JSON Schema documentation on $anchor](https://json-schema.org/draft/2019-09/json-schema-core.html#anchor)
[More on $anchor](https://json-schema.org/understanding-json-schema/structuring.html#anchor)

## Check yourself

- [x] Code is linted

## Security


packages/core/src/resolve.ts
```diff
@@ -313,6 +319,22 @@ export async function resolveDocument(opts: {
       if (hasRef(refStack.prev, ref)) {
         throw new Error('Self-referencing circular pointer');
       }
+
+      if (isAnchor(ref.$ref)) {
+        // Wait for all anchors in the document to be collected firstly.
+        await delay();
```
RomanHotsiy, line 325: This is a bit dangerous and funky 😄 

What you are doing here is waiting for the next event loop iteration, I would rename this delay function appropriately. 

There are edge-cases I am worried about but let me think about it.

tatomyr: Yeah,  I agree. However, I've tried to isolate this behaviour for anchors only.  
Please let me know if you can think of a better option.

tatomyr: Renamed into `nextTick`.

RomanHotsiy (approved):

IgorKarpiuk (approved):

Issue #840, opened 2022-08-31
operation-4xx-response rule indicates to use "4xx" (downcase), but OAS standard defines "4XX" (upcase)


ahx:
**Describe the bug**
If I understand The OpenAPI Specification correctly, it [defines a "4XX"](https://spec.openapis.org/oas/latest.html#patterned-fields-0) (or 2XX …) response with uppercase "X". This is also reflected in the [OAS JSON Schema](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L559). 

The error message from redocly lint is:

> Operation must have at least one `4xx` response.

**Expected behavior**
The error message should be:

> Operation must have at least one `4XX` response.

**OpenAPI definition**

**Redocly Version(s)**
1.0.0-beta.106

**`Node.js` Version(s)**
v18.5.0

adamaltman:
Thanks for the PR @ahx (I hope we can release it soon).

PR #841, opened 2022-08-31
Upcase "4XX" in error message of operation-4xx-response

ahx:
## What/Why/How?

This changes the error message of `operation-4xx-response` to say "4XX" (upcase) instead of "4xx" (downcase). This is good, because it helps people to write their API description as defined in the standard. 

See also https://github.com/Redocly/redocly-cli/issues/840.

This also does the same for operation-2xx-response

## Reference
Solves https://github.com/Redocly/redocly-cli/issues/840

## Testing
✅ 

## Check yourself

- [x] Code is linted

## Security

This change is not security relevant


RomanHotsiy (approved):

adamaltman:
Thank you @ahx! 🎉 

@tatomyr can we include this in the next release?

tatomyr:
Sure, will do. 

@ahx thanks again for your contribution!

ahx:
Thanks folks ♥️

Issue #699, opened 2022-05-23
`push` command: ability to upload other files 


RomanHotsiy:
**Is your feature request related to a problem? Please describe.**
Our workflows app allows users to upload other files next to the openapi definition.

![image](https://user-images.githubusercontent.com/3975738/169823078-53c0c3d5-98c5-4447-bb9e-1bc0c6a7c73c.png)

Our `push` command doesn't have the same capabilities.

**Describe the solution you'd like**

Add a new argument to the `push` command: `--files`.

```sh
redocly push openapi/petstore.yaml @openapi-org/petstore-api@v1 --files ./files
```

Also, we may need to add `files` config to the redocly.yaml.

**Describe alternatives you've considered**
None.

**Additional context**
Should work similar to manual files upload.

adamaltman:
> Also, we may need to add `files` config to the `redocly.yaml`.

What possible configurations are there? Defining a list of files so that the argument can be avoided?

adamaltman:
I opened a can of worms. 🪱 🪱 

It could also be a list of folders.

```yaml
files:
  - ./xray
  - ./zebra
```

Also, are they with or without trailing slashes?

RomanHotsiy:
> It could also be a list of folders.

It's a little confusing why it's called `files` then. Or I overcomplicate?

> Also, are they with or without trailing slashes?

Doesn't matter.

PR #842, opened 2022-08-31
feat: ability to add other files and folders to push

iosypov:
## What/Why/How?

Fix: #699
Added ability to upload other files and folders. User can provide section `files` in `.redocly.yaml` or as argument `--files`.

## Screenshots (optional)

image

## Check yourself

- [x] Code is linted

## Security


tatomyr:
Should we also include `files` from other configs when extending? If so, we should add code here: https://github.com/Redocly/redocly-cli/blob/master/packages/core/src/config/utils.ts#L160

docs/commands/push.md
```diff
@@ -276,6 +277,25 @@ For more information on how to configure access to your APIs, check the [registr
...
+
+```bash
+redocly push openapi/petstore.yaml @openapi-org/petstore-api@v1 --files ./path/to/folder
+```
+
+You can also add files and folders providing them in `.redocly.yaml`:
+
+```yaml
+apiDefinitions:
+  main: ./openapi.yam;
```
tatomyr, line 292:
```suggestion
apis:
  main: 
     root: ./openapi.yam;
```

packages/core/src/config/load.ts
```diff
@@ -7,7 +7,7 @@ import { Config, DOMAINS } from './config';
 import { transformConfig } from './utils';
 import { resolveConfig } from './config-resolvers';
 
-import type { DeprecatedInRawConfig, RawConfig, Region } from './types';
+import { ArgsType, ConfigArgsOverrideFields, CONFIG_ARGS_OVERRIDE_FIELDS, DeprecatedInRawConfig, RawConfig, Region } from './types';
```
tatomyr, line 10: You can import 2 times: types and variables  separately.

packages/core/src/config/load.ts
```diff
@@ -62,11 +62,15 @@ async function addConfigMetadata({
...
+  options: {
+    configPath?: string;
+    customExtends?: string[];
+    processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>;
+    argv?: ArgsType;
+  } = {}
 ): Promise<Config> {
-  const rawConfig = await getConfig(configPath);
+  const { configPath = findConfig(), customExtends, processRawConfig, argv } = options;
+  const rawConfig = { ...(await getConfig(configPath)), ...getConfigArgv(argv), argv };
```
tatomyr, line 73: I'd prefer to be more explicit:

```suggestion
  const config = await getConfig(configPath);
  const rawConfig = { 
       ...config, 
       region: config.region ?? argv.region,
       files: config.files ?? argv.files, 
       argv 
   };
```

Then we don't need `getConfigArgv` at all.

packages/core/src/config/load.ts
```diff
@@ -62,11 +62,15 @@ async function addConfigMetadata({
...
 
 export async function loadConfig(
-  configPath: string | undefined = findConfig(),
-  customExtends?: string[],
-  processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>
+  options: {
+    configPath?: string;
+    customExtends?: string[];
+    processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>;
+    argv?: ArgsType;
```
tatomyr, line 69: What do you think of this: 
```suggestion
    files?: string[]
    region?: string
```
?

packages/cli/src/commands/push.ts
```diff
@@ -37,11 +37,12 @@ type PushArgs = {
   region?: Region;
   'skip-decorator'?: string[];
   public?: boolean;
+  files?: string[];
 };
 
 export async function handlePush(argv: PushArgs): Promise<void> {
-  const config = await loadConfig();
-  const region = argv.region || config.region;
+  const config = await loadConfig({ argv });
```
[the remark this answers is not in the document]
iosypov: We discussed that will be good to save all argv to rawConfig.

packages/core/src/config/load.ts
iosypov: For me, it's more cleaner. If we have much more argv to config, it will be long list of this.

tatomyr: Indeed. However, we have to support this list anyways [here in types](https://github.com/Redocly/redocly-cli/pull/842/files#diff-623912bbe54cdd04399ad03ecf4c603f03682cf4d64116197426943f96811cf9R181-R189), and it will be even longer :)

packages/core/src/config/types.ts
```diff
@@ -176,3 +177,15 @@ export type RulesFields =
...
+  files = 'files',
+  region = 'region',
+}
+
+export const CONFIG_ARGS_OVERRIDE_FIELDS = [
+  ConfigArgsOverrideFields.files,
+  ConfigArgsOverrideFields.region,
+];
+
+export type ArgsType = { [key: string]: any };
```
tatomyr, line 191: Can we write sth like this: 
```suggestion
export type ArgsType = A1 | A2 | A3...
```
?

tatomyr (approved):

docs/commands/push.md
```diff
@@ -276,6 +277,28 @@ For more information on how to configure access to your APIs, check the [registr
...
+```bash
+redocly push openapi/petstore.yaml @openapi-org/petstore-api@v1 --files ./path/to/folder
+```
+
+You can also add files and folders providing them in `.redocly.yaml`:
+
+```yaml
+apis:
+  main: 
+     root: ./openapi.yam;
```
RomanHotsiy, line 293: do we support files in `apis` section?

packages/core/README.md
```diff
@@ -10,7 +10,7 @@ See https://github.com/Redocly/redocly-cli
 import { formatProblems, lint, loadConfig } from '@redocly/openapi-core';
 
 const pathToApi = 'openapi.yaml';
-const config = loadConfig('optional/path/to/.redocly.yaml');
+const config = loadConfig({ configPath: 'optional/path/to/.redocly.yaml' });
```
RomanHotsiy, line 13: This is a breaking change. This function can be used in other repositories/projects. We need to double-check it.

packages/core/src/config/load.ts
```diff
@@ -62,11 +62,17 @@ async function addConfigMetadata({
...
 
 export async function loadConfig(
-  configPath: string | undefined = findConfig(),
-  customExtends?: string[],
-  processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>
+  options: {
+    configPath?: string;
+    customExtends?: string[];
+    processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>;
+    files?: string[];
```
RomanHotsiy, line 69: Do we need it here? I mean, the function name is `loadConfig` and seems to add external config values. How about making a new function for that? 

I don't know if this can be done easily. This is just an idea. Non-blocking.

packages/core/README.md
tatomyr: I believe TS should help in other projects while updating `openapi-core`. We did such changes before.

RomanHotsiy: Hopefully no one else except us uses it yet 🤞 
Good.

docs/commands/push.md
adamaltman: I think we should. Files might be specific to a specific API?

packages/core/src/config/utils.ts
```diff
@@ -157,6 +157,9 @@ export function getMergedConfig(config: Config, apiName?: string): Config {
             ...config['features.mockServer'],
             ...config.apis[apiName]?.['features.mockServer'],
           },
+          files: config.apis[apiName]
+            ? config.files.concat(config.apis[apiName].files ?? [])
+            : config.files,
```
tatomyr, line 162: I believe we don't need this ternary: 
```suggestion
          files: [...config.files, ...config.apis?.[apiName]?.files ?? []],
```

tatomyr, line 162: I believe we don't need this ternary: 
```suggestion
          files: [...config.files, ...config.apis?.[apiName]?.files ?? []],
```

tatomyr (changes requested):
Please

packages/core/src/config/utils.ts
```diff
@@ -157,6 +157,7 @@ export function getMergedConfig(config: Config, apiName?: string): Config {
             ...config['features.mockServer'],
             ...config.apis[apiName]?.['features.mockServer'],
           },
+          files: [...new Set([...config.files, ...(config.apis?.[apiName]?.files ?? [])])],
```
tatomyr, line 160: I'd place the uniqueness check  somewhere here:  https://github.com/Redocly/redocly-cli/pull/842/files#diff-51105770da26478a3f6a7c2ee8107952064a25f515f456f57b7a626ad5f9587dR269 to check all the uploaded files.

tatomyr (approved):

iosypov:
@adamaltman  @RomanHotsiy  need your reviews, please.

docs/commands/push.md
RomanHotsiy: Should we show an example of this in the docs?

RomanHotsiy (approved):
LGTM

I've left one small suggestion.

Issue #833, opened 2022-08-26
Flatten configuration file by removing styleguide object


adamaltman:
**Is your feature request related to a problem? Please describe.**

The configuration file has one extra level of nesting which isn't required.
This makes docs and copy/paste more difficult than it needs to be.

**Describe the solution you'd like**

- Move `doNotResolveExamples` to the `resolve` object.
- Move `extends`, `plugins`, `rules`, `decorators`, `preprocessors` to the root.
- Support `rules`, `decorators`, and `preprocessors` in the `apis` object too.

- I adjusted all the documentation here: https://github.com/Redocly/marketing-site-portal/pull/986

**Describe alternatives you've considered**

Do nothing. But the time to do it is now, before we have a general release.

**Additional context**

- Notice the number of lines we can remove from the docs. 
- This makes copy/paste of configs easier too.

tatomyr:
@adamaltman do we want to continue support `extends` and `doNotResolveExamples` in `apis` also (I see you haven't mentioned it in the changes)?
Please note that if we do want this for `doNotResolveExamples`, we would have to support the `resolve` object in `apis` as well.

adamaltman:
Let's support the `extends` in `apis` too. Let's not support the `doNotResolveExamples` in `apis`.

PR #847, opened 2022-09-02
chore: flatten configuration file by removing styleguide object

tatomyr:
## What/Why/How?

Closes https://github.com/Redocly/redocly-cli/issues/833

## Reference

Related to https://github.com/Redocly/marketing-site-portal/pull/986

## Check yourself

- [x] Code is linted

## Security


adamaltman (approved):

AntonKozachuk (approved):

packages/core/src/config/utils.ts
```diff
@@ -39,13 +42,76 @@ export function transformApiDefinitionsToApis(
...
+
+    decorators,
+    oas2Decorators,
+    oas3_0Decorators,
+    oas3_1Decorators,
+
+    doNotResolveExamples: (rawConfigRest as FlatRawConfig).resolve?.doNotResolveExamples,
+  };
+
+  return {
```
RomanHotsiy, line 95: should we add a check here and crash if there is both any of `rules` etc and `lint`?

RomanHotsiy (approved):
It looks good but I would add error message for mixed config.

Also, I'm not sure if we have a test for legacy config. If not, let's add one.

packages/core/src/config/utils.ts
tatomyr: Thanks, will add the error.

tatomyr: Added: https://github.com/Redocly/redocly-cli/pull/847/commits/bc76589e0b64402d8e9f73cfd8d2dbbd225094f8?file-filters%5B%5D=.ts&show-viewed-files=true

@adamaltman could you verify the error message?

Issue #839, opened 2022-08-31
Rename type: DefinitionRoot to Root 


adamaltman:
**Is your feature request related to a problem? Please describe.**

Trying to make our type names more clear. I think `Root` is better than `DefinitionRoot`. It will also look better in diagrams.

**Describe the solution you'd like**

Rename `DefinitionRoot` to `Root`.

**Describe alternatives you've considered**

Doing nothing.

**Additional context**

I think the time to do it is now, prior to any GA release.

tatomyr:
Do we want to update also other types to be more consistent? Like single VS plural forms in maps?

adamaltman:
I think that would be a good idea. But we'll need to keep the old ones for backwards compatibility for now too.

PR #852, opened 2022-09-06
chore: rename DefinitionRoot -> Root

tatomyr:
## What/Why/How?

Closes https://github.com/Redocly/redocly-cli/issues/839

Renamed also other types:
```
  Root: 'DefinitionRoot',
  ServerVariablesMap: 'ServerVariableMap',
  PathsMap: 'PathMap',
  CallbacksMap: 'CallbackMap',
  MediaTypesMap: 'MediaTypeMap',
  ExamplesMap: 'ExampleMap',
  EncodingsMap: 'EncodingMap',
  HeadersMap: 'HeaderMap',
  LinksMap: 'LinkMap',
```

## Check yourself

- [x] Code is linted

## Security


adamaltman (commented):
I like it but I think we should consider adding that `DefinitionRoot` is defined as type `Root` for some period of backwards compatibility.

adamaltman (approved):
LGTM

packages/core/src/visitors.ts
```diff
@@ -380,8 +393,11 @@ export function normalizeVisitors<T extends BaseVisitor>(
     }
 
     for (const typeName of visitorKeys as Array<keyof T>) {
-      const typeVisitor = visitor[typeName] as any as NestedVisitObject<any, T>;
-      const normalizedTypeVisitor = normalizedVisitors[typeName]!;
+      const typeVisitor = (visitor[typeName] ||
```
RomanHotsiy, line 396: do we have a test for a legacy names somewhere? if not we should add temporary.

RomanHotsiy (approved):
LGTM

packages/core/src/visitors.ts
tatomyr: Good point. Will add a them.

Issue #855, opened 2022-09-07
Include an assert function name into assertion messages


tatomyr:
**Is your feature request related to a problem? Please describe.**

If more than one assert functions are defined, it is not clear which one triggered the assertion failure. If both of the functions fail, the output contains two identical messages.

**Describe the solution you'd like**

Including an assert function name  into the output (like `patern`, `casing` or a custom name) would be handy to identify the cause of the failure.

**Describe alternatives you've considered**

Collapsing all the assertion messages into one. But it would require more efforts and also would be less clear to a user.

**Additional context**

[Internal discussion](https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1662061195536239)

adamaltman:
@RomanHotsiy and I discussed the required placeholders for assertion messages last night. @RomanHotsiy did you want to add it to this ticket?

RomanHotsiy:
Details here: https://github.com/Redocly/redocly-cli/pull/857/files#diff-278dfad5d57e4b60a03f414c74a1d4c5d5273dd755ffae1b201c8338e36fd824R37

adamaltman:
I copied it here in case that linked PR changes: 

> Problem message displayed if the assertion is false. If omitted, the default message is: "{{assertionName}} failed because the {{subject}} {{property}} didn't meet the assertions: {{problems}}" is displayed. The available placeholders are displayed in that message. In the case there are multiple properties, the `{{property}}` placeholder produces a comma and space separate list of properties. In case there are multiple problems, the `{{problems}}` placeholder produces a bullet-list with a new line between each problem.

Issue #781, opened 2022-07-28
Change context to if and change the object definition


RomanHotsiy:
**Is your feature request related to a problem? Please describe.**

The context object allows for targeting exact keys. However, sometimes you need more flexibility and want to target based on some specific parent property value 

**Describe the solution you'd like**

## Context object (current)

Property | Type | Description
-- | -- | --
type | `string` | **REQUIRED.** One of the [OpenAPI node types](#openapi-node-types).
matchParentKeys | [`string`] | The list of parent object key names to evaluate with respect to the subject.
excludeParentKeys | [`string`] | The list of parent object key names to not evaluate with respect to the subject.

## Context object (desired)

Property | Type | Description
-- | -- | --
type | `string` | **REQUIRED.** One of the [OpenAPI node types](#openapi-node-types).
includeParentKeys | [`string`] | The list of the exact parent object key names to evaluate with respect to the subject.
excludeParentKeys | [`string`] | The list of the exact parent object key names to not evaluate with respect to the subject.
includeByPropValue | [PropMatcher] | A list of prop matchers. They are combined using boolean AND. Subject is evaluated if all the matchers apply.
excludeByPropValue | [PropMatcher] | A list of prop matchers. They are combined using boolean AND. Subject is not evaluated if all the matchers apply.

## PropMatcher object (desired)
Property | Type | Description
-- | -- | --
key | `string` | **REQUIRED.** The property key
value | `any` | [`any`] | The property value to match. Supports only scalar values. If `value` is an array then the prop value should match at least one item.

Example usage:

```yaml
assert/only-string-in-params:
  context:
  - type: Operation
    matchParentKeys: [put]
  - type: Parameter
    includeByPropValue:
      - key: in # we can validate the key value!
        value: ["query", "path"]
  subject: Schema
  property: type
  enum: ["string"]
```

**Describe alternatives you've considered**

Make it a bit more simple without support for multiple matchers.

**Additional context**

Related to https://github.com/Redocly/redocly-cli/issues/714
Some more context in: https://github.com/Redocly/intranet/issues/370#issuecomment-1196703294

iosypov:
I think it will great to have ability to include parent key not only exact matching key and by regex. It can be make more flexibility.

adamaltman:
This issue should be replaced by this:

1. Add support for 3 additional locator properties: `filterInParentKeys`, `filterOutParentKeys`, `matchParentKeys` (regex). 

2. Change `context` to `if`.

[...]

Issue #780, opened 2022-07-28
Support for custom assertions


RomanHotsiy:
**Is your feature request related to a problem? Please describe.**
Using assertions context and subject syntax seems to be convenient. We need a simple way to add custom assertions as an alternative for writing custom rules.

Custom assertions will be easier to write and custom rules will be needed for complex cases only.

**Describe the solution you'd like**
I would implement it as an extension to plugins. To keep it simpler I would structure all the examples in our docs so it's using a single plugin file and does not split across multiple files.

The simplest plugin may look like:

**plugin.js**
```js
module.exports = {
  id: 'my-plugin',
  assertions: {
    customFn: (value, opts, location) => {
      if (...) return { isValid: false, location };
      return { isValid: true };
    }
  }
};
```

We need to implement a simple way to include plugin (using cli args without the need for redocly.yaml): `--include-plugin ./plugin.js`.

To use the assertion we may use one of the following approaches:

**A):**
```yaml
lint:
  rules:
    assert/operation-summary-max-length:
      subject: Operation
      property: summary
      message: Some custom assertion
      severity: error
      my-plugin/customFn: <options>
```

**B):**
```yaml
lint:
  rules:
    assert/operation-summary-max-length:
      subject: Operation
      property: summary
      message: Some custom assertion
      severity: error
      custom:
        name: my-plugin/customFn
        options: <options>
```

I would vote for the A).

**Describe alternatives you've considered**
None.

tatomyr:
I also like **A** better.

> We need to implement a simple way to include plugin (using cli args without the need for redocly.yaml)

It looks like we do need `redocly.yaml` anyway since assertions could only be used inside the  redocly config, right?

Issue #714, opened 2022-05-26
Add parent keys pattern to assertions context


adamaltman:
**Is your feature request related to a problem? Please describe.**

The context object allows for targeting exact keys. However, sometimes you need more flexibility and can't determine the exact preset parent key names. An example would be a path with a template parameter (it has `{` and `}` characters in it).  

**Describe the solution you'd like**

## Context object (current)

Property | Type | Description
-- | -- | --
type | `string` | **REQUIRED.** One of the [OpenAPI node types](#openapi-node-types).
matchParentKeys | [`string`] | The list of parent object key names to evaluate with respect to the subject.
excludeParentKeys | [`string`] | The list of parent object key names to not evaluate with respect to the subject.

## Context object (desired)

Property | Type | Description
-- | -- | --
type | `string` | **REQUIRED.** One of the [OpenAPI node types](#openapi-node-types).
includeParentKeys | [`string`] | The list of the exact parent object key names to evaluate with respect to the subject.
excludeParentKeys | [`string`] | The list of the exact parent object key names to not evaluate with respect to the subject.
matchParentKeys | `string` | A pattern that will evaluate against the parent keys.

Notice that we rename `matchParentKeys` to `includeParentKeys` and we use `matchParentKeys` keys for the new assertion context condition.

There is some discussion if the `includeParentKeys` should be mutually exclusive with `matchParentKeys` and `excludeParentKeys`. (I don't care either way.)

**Describe alternatives you've considered**

One alternative considered was removing the exact parent keys and have a pattern only. This makes it more difficult to read and write the exact key use cases.

**Additional context**

Attempting to limit context to paths with path template variables.

RomanHotsiy:
matchParentKeys
parentKeysPattern

parentKeys:
matchP

tatomyr:
BTW, why do we call them **parent** keys? As far as I understand, it is more about **current** ones, isn't it?
In this case, the assertion will apply to the key of the `Operation`, not to a key of its parent `PathItem`:
```yaml
         context:
            - type: Operation
              matchParentKeys:
                - get

```

adamaltman:
It's not a property of Operation from what I see but a property that contains the Operation as its value. Operation has these properties like `tags`, `summary`:

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-8

It's a property of the Path Item object: 

[...]

PR #857, opened 2022-09-07
docs: update to assertions

adamaltman:
## What/Why/How?

Docs for a desired update to assertions.

1. Change `context` to `if`.
2. Change `if` list to conform to assertion object (see the docs -- `if`, `message`, suggest`, `severity` are ignored when used inside of the `if` list. The assertion object in the `if` list must contain the subject and at least one assertion (such as `defined: true`).
3. Add support for 3 additional locator properties: `filterInParentKeys`, `filterOutParentKeys`, `matchParentKeys` (regex). See the docs for more details.
4. Remove `undefined` because it is the same as `defined: false`.
5. Add support for message placeholders as described in the documentation: `{{assertionName}}, {{subject}}, {{property}}, {{problems}}`.

## Reference

This would replace various other tickets:
- #855 
- #781 
- #714 

This would influence #780 because a custom function could provide a list of problems (which are added to the problems list used for the placeholders) so the return signature would need to be able to include that.

## Testing

This is the docs only.


docs/rules/assertions.md
```diff
@@ -373,18 +373,20 @@ The following example asserts that `PUT` requests have both `200` and `201` resp
...
+      if:
+        - type: Operation
+          filterInParentKeys:
+            - put
+          defined: true
+      message: Must define 200 and 201 responses for PUT requests.
+      severity: error
+      required:
+        - '200'
+        - '201'
```
adamaltman, line 389:
```suggestion
rules:
  assert/put-200-and-201:
    subject: ResponsesMap
    if:
      - type: Operation
        filterInParentKeys:
          - put
        defined: true
    message: Must define 200 and 201 responses for PUT requests.
    severity: error
    required:
      - '200'
      - '201'
```

docs/rules/assertions.md
```diff
@@ -162,23 +154,29 @@ rules:
...
+  - subject: Operation
+    filterByParentKeys: 
+      - put
+    defined: true
+  - subject: Response
+    filterByParentKeys: 
+      - '201'
+      - '200'
+    defined: true
+  subject: MediaTypeMap
```
SmoliyY, line 177: It is look a little bit confusing to have 3 subject.  Maybe it is better to rename subject that we specifically validation to something uniq - like `evaluateSubject` ?

docs/rules/assertions.md
```diff
@@ -162,23 +154,29 @@ rules:
...
 ```yaml
 assert/no-pdf-in-ok-response:
-  context:
-  - type: Operation
-    matchParentKeys: [put]
-  - type: Response
-    matchParentKeys: ['201', '200']
-  subject: MediaTypesMap
-  disallowed: ['application/pdf']
+  if:
```
SmoliyY, line 167: For me it looks like if we have `if` statement we also can have `else`. Maybe `context` suits here more the `if`.

docs/rules/assertions.md
sobanieca-redocly: Agree, I was thinking - why can't we define simple "if any item with given filters exist" rule, which when met, generates proper error. For example, yesterday's meeting ([link](https://us02web.zoom.us/rec/play/Ov8F9Wt1zIrtNXJfu2fQ3kvWedY0XxY2Jm1_80XnyW-A8Y5qoA-lXkV7N0OYgF_w_99vbxLCNAugOYZs.LZHS2eQRSFkXVA-8?_x_zm_rhtaid=820&_x_zm_rtaid=FzJlBy6FQ5WUDELmSdfpiQ.1663920708235.aa1d24cc93317830c10d3cc84ae2ea9d&autoplay=true&continueMode=true&startTime=1663857448000)) __"media type not pdf"__ rule:

```
assert:
	name: media-type-map-not-pdf
	when:
		- subject: Operation
		  filterOutParentKeys:
			- put
		- subject: Response
		  filterInParentKeys: ['201', '200']
		- subject: MediaTypesMap
		  filterInChildKeys: ['application/pdf']
	then:
		error: "Media type should not be pdf"
```

Inside `when` we can define conditions and if they are met (to be precise - item meeting such conditions is met), we can proceed with `then` actions.

docs/rules/assertions.md
RomanHotsiy: maybe `given`

docs/rules/assertions.md
RomanHotsiy: What else can be in `then`?

RomanHotsiy: Can we have else?

RomanHotsiy: also, it's essential to understand that the `when` statements are not separate conditions, it defines the hierarchy:

```
Operation(put) -> Response(201, 200) -> MediaTypesMap
```

sobanieca-redocly: For `then` I don't know, if nothing else, probably we could end up with

```
...
   then: "Media type should not be pdf"
```

or replace `then` with `error` (actually makes more sense to me).

I didn't think about any use-case for `else`, but if needed, we can add it later. What is more, in future we can add `or` and `and` operators:

```
assert:
	name: media-type-map-not-pdf
	when:
		-and:
			- subject: Operation
			  filterOutParentKeys:
				- put
			- subject: Response
			  filterInParentKeys: ['201', '200']
			- or:
				- subject: MediaTypesMap
				  filterInChildKeys: ['application/pdf']
				- subject: MediaTypesMap
				  filterInChildKeys: ['application/json']
	then:
		error: "Media type should not be pdf or json"
```


When it comes to hierarchy, maybe we should introduce `children` property? So for instance:

```
assert:
	name: media-type-map-not-pdf
	when:
		- subject: Operation
		  filterOutParentKeys:
			- put
		  children:
			- subject: Response
			  filterInParentKeys: ['201', '200']
			  children:
				- subject: MediaTypesMap
				  filterInChildKeys: ['application/pdf']
	then:
		error: "Media type should not be pdf"
```

sobanieca-redocly: We can also think about introducing instead of `when` two operators: `whenAny` and `whenNone`. One would be for searching if there exists any item that matches given filter, another one would be to ensure that there is none item that matches condition.

adamaltman: What about `whenAll`?

sobanieca-redocly: Actually, `whenNone` is more or less `whenAll`, but naming is not the best, I believe it should be more like `whenAnyNot` - when there is any item that doesn't satisfy condition. For instance, if we want to ensure that all operations contain media type "application/json" we can write:

```
assert:
	name: 
	whenAnyNot:
		- subject: Operation
		  filterOutParentKeys:
			- put
		  children:
			- subject: Response
			  filterInParentKeys: ['201', '200']
			  children:
				- subject: MediaTypesMap
				  filterInChildKeys: ['application/json']
	error: "All operations should contain application/json media type"
```

Basically, all such operators can be added gradually, for MVP I believe `whenAny` would be sufficient.

docs/rules/assertions.md
```diff
@@ -162,23 +154,29 @@ rules:
       - description
 ```
 
-### `context` example
+### `if` example
```
adamaltman, line 157:
```suggestion
### `where` example
```

docs/rules/assertions.md
tatomyr: Thanks all for bringing your ideas 🙂 

Putting aside the exact naming, what do you think about the assertion structure? I mean `subject` + `condition` + `optional container for narrowing down which consists of subject + condition again`? Do some maybe have an idea of a better approach?

docs/rules/assertions.md
SmoliyY: I think `given` suits more than `if`.  Same for [where](https://github.com/Redocly/redocly-cli/pull/857#discussion_r979583093). But as for me `where` is better than `given` )))

docs/rules/assertions.md
```diff
@@ -27,9 +27,12 @@ assert/my-assertion-name | [Assertion object](#assertion-object) | An assertion
...
+matchParentKeys | `string` | Applies a regex pattern to the subject's parent keys to determine where assertions run. An example value given the subject `Operation` could be `matchParentKeys: /^p/` means that `POST`, `PUT`, and `PATCH` operations are evaluated for the assertions.
+if | [`Assertion` object] | The `if` list influences evaluation for assertions by only evaluating assertions when the `if` arguments evaluate to true. The `if` argument must be structured like a valid assertion with the exception of no nested `if` values. When using the `if` list, it evaluates the locator specified with any assertions, and then continues to evaluate based on that subset of the values. The resolution of reference objects is done at the `if` level. The `if` list is evaluated from top to bottom in that order. If no `if` list is provided, it evaluates the assertion for all instances of the given type. See [if example](#if-example). The `if` evaluation does not result in any problems (it is used for filtering where to run assertions); therefore, the `message`, `suggest`, and `severity` properties are ignored.
```
tatomyr, line 35: We should state somewhere that the order of the array matters: it should go from the `Root` down to the subject (including).

roman-sainchuk:
In case we have assertion like this (I know it doesn't make much sense but Im interested in implementation behaviour):

```yaml
assert/put-200-and-201:
      subject: 
        - ResponsesMap
      if:
        - subject:
            - PathsMap
            - Operation
          filterInParentKeys:
            - put
            - /path
          defined: true
      message: Must mutually define 200 and 201 responses for PUT requests.
      severity: error
      mutuallyRequired:
        - '200'
        - '201'
 ```
 Will it apply assertion for `ResponsesMap` inside `PathsMap` AND `ResponsesMap` inside `Operation` ? Or `ResponsesMap` which is inside of `Operation` which is inside of `PathsMap`?

**Comment by RomanHotsiy** [comment]
@roman-sainchuk the first one (like 2 separate assertions).

I'm not sure if we want to support array for `subject` at all.

**Comment by roman-sainchuk** [comment]
I'd say to support array only for top level `subject` field. As for me its not clear how it supposed to work if we have several items in `where` list where each of them has multiple `subject`s

**Comment by RomanHotsiy** [comment]
I would just remove this to simplify things.

**Comment by roman-sainchuk** [comment]
> Add support for 3 additional locator properties: `filterInParentKeys`, `filterOutParentKeys`, `matchParentKeys` (regex). See the docs for more details.

From my understanding these properties are mutually excluding along with `property` locator. Is that correct?

**Comment by RomanHotsiy** [comment]
> From my understanding these properties are mutually exclusive along with property locator. Is that correct?

No, I don't think so. Let's sync tomorrow.

File: docs/rules/assertions.md (hunk 2 above)
Comment (adamaltman) [reply docs/rules/assertions.md:177]: I don't have a better idea.

File: docs/rules/assertions.md (hunk 3 above)
Comment (adamaltman) [reply docs/rules/assertions.md:167]: I prefer `where` too.

**Comment by adamaltman** [comment]
I agree with simplifying if possible. Flattening the subject may be fine.

Those properties (`filterInParentKeys`, `filterOutParentKeys`, `matchParentKeys`) shouldn't be mutually exclusive, and people should be aware they could create nonsense combinations (as is with most things in life):

```
filterInParentKeys:
  - put
filterOutParentKeys:
  - put
```

**Comment by roman-sainchuk** [comment]
But how should `filterInParentKeys` and  `matchParentKeys ` work together ? Should we collect matched fields by these 2 locators and remove from it matched fields by `filterOutParentKeys`?

**Comment by roman-sainchuk** [comment]
We had a small discussion with @ohorbachevskyi about new syntax and end up with idea to put top level `subject`, `property`, new locators and assertion inside `do` block. So this way users have better visual separation of preconditions and assertion's action itself.

Example:
```yaml
assert/put-200-and-201:
      where:
        - subject: Operation
           filterInParentKeys:
            - put
           defined: true
      do:
          subject:  ResponsesMap
          mutuallyRequired:
            - '200'
            - '201'
      message: Must mutually define 200 and 201 responses for PUT requests.
      severity: error
```

tatomyr:
I wonder if this:

```yaml


assert/media-type-map-not-pdf:
  
    subject: MediaTypesMap    
    disallowed: ['application/pdf']

    where:
      - subject: Operation
        filterOutParentKeys:
          - put
```

could be expressed with the following:

```yaml
assert/media-type-map-not-pdf:
  
    subject: MediaTypesMap    
    disallowed: ['application/pdf']

    where:
      - subject: PathItem
        disallowed:
          - put
```

If yes, having `filterIn/Out/matchParentKeys` sounds redundant. 
@adamaltman, @RomanHotsiy what do you think? Am I missing something?

tatomyr:
Ah, indeed, thanks. I'm constantly forgetting about this 🙂.

adamaltman:
I'm not sure about the `do` word. I feel it is used in combination with `while` and introduces the same confusion as `if` without `else`.

adamaltman:
There is a language Gherkin (used for describing behavior) which used the keywords `Given`, `When`, `Then`. I'm not saying this because those keywords make sense in our system (they don't IMO), but I think there are actually 3 parts here:

- subject evaluated (TBD)
- additional conditions (`where`)
- assertions (maybe we call them `assertions`? it's the most logical word)



The before:

```
    assert/operation-id-delete:
      context:
        - type: Operation
          matchParentKeys:
            - delete
      subject: Operation
      property: operationId
      pattern: /Delete/
      casing: PascalCase
```

The result:

```
    assert/operation-id-delete:
      where:
        - type: Operation
          filterInParentKeys:
            - delete
      subject: 
        type: Operation
        property: operationId
      assertions:      
        pattern: /Delete/
        casing: PascalCase
```

adamaltman:
I reworked it more and decided to open it as a separate PR: https://github.com/Redocly/redocly-cli/pull/917

PR #860, opened 2022-09-08
feat: new format option 'summary'

RomanHotsiy:
## What/Why/How?

New format option: `summary`:

image

Also, fixed json output.

## Testing

Covered with tests.

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/lint.ts
```diff
@@ -127,6 +127,11 @@ function lintConfigCallback(argv: LintOptions, version: string) {
     return;
   }
 
+  if (argv.format === 'json') {
```
tatomyr, line 130: Is this part covered with a test? If not, it's better to add one.

packages/core/src/format/format.ts
```diff
@@ -257,6 +266,26 @@ export function formatProblems(
...
+function formatSummary(problems: NormalizedProblem[]): void {
+  const counts: Record<string, { count: number; severity: ProblemSeverity }> = {};
+  for (const problem of problems) {
+    counts[problem.ruleId] = counts[problem.ruleId] || { count: 0, severity: problem.severity };
+    counts[problem.ruleId].count++;
+  }
+  const sorted = Object.entries(counts).sort((a, b) => {
+    const severityDiff = severityToNumber(a[1].severity) - severityToNumber(b[1].severity);
+    return severityDiff || b[1].count - a[1].count;
+  });
```
tatomyr, line 278:
```suggestion
  const sorted = Object.entries(counts).sort(([, a], [, b]) => {
    const severityDiff = severityToNumber(a.severity) - severityToNumber(b.severity);
    return severityDiff || b.count - a.count;
  });
```

tatomyr (approved):
LGTM. 
Left a couple of optional suggestions.

packages/cli/src/commands/lint.ts
RomanHotsiy: Added an e2e test.

packages/core/src/format/format.ts
RomanHotsiy: Thanks! 👍  Applied.

jeremyfiel:
please update `lint` command docs with examples for each type of `format`

https://redocly.com/docs/cli/commands/lint/#format

PR #862, opened 2022-09-09
fix: security-defined rule

### LINKED ISSUE #846: fix operation-security-defined and rename to security-defined
Fix operation-security-defined rule because it is incomplete. It should comply with the [documentation (which is hidden currently)](https://github.com/Redocly/redocly-cli/blob/master/docs/rules/operation-security-defined.md).

> Every API should have security defined. Even if the API is truly public without any credential required, define the empty security section to let people know you didn't forget.

This means we need to check if security is defined in the root, or if it is defined in all operations, then it passes the rule. If it is defined only in some operations and not in the root, it would not pass the rule.

An empty definition is okay, because it means they have explicitly decided that the security is disabled (which is different from accidentally forgetting to add security):

`- security: []` ✅ 

In addition, the rule should continue to enforce that the security requirements used are defined by the security schemes in the components (this is already part of the rule and seems to function properly).

Also, this rule should be renamed to `security-defined` (it doesn't need the `operation-` prefix because it applies to both the root and all operations).

AntonKozachuk:
## What/Why/How?
Fix and rename `operation-security-defined` rule because it is incomplete.

## Reference
Fixes #846.

## Testing
Locally.

## Screenshots (optional)
N/a

## Check yourself

- [ ] Code is linted

## Security


packages/core/src/config/recommended.ts
```diff
@@ -25,7 +25,7 @@ export default {
     'operation-operationId-url-safe': 'error',
     'operation-parameters-unique': 'error',
     'operation-tag-defined': 'off',
-    'operation-security-defined': 'error',
+    'security-defined': 'error',
```
RomanHotsiy, line 28: I'm not sure if this has to be `error`. Public APIs don't have to have a `security` defined in theory. I would change it back to `warn`.

@adamaltman what do you think?

RomanHotsiy (approved):
LGTM 👍

packages/core/src/config/recommended.ts
AntonKozachuk: It was an error before. Not sure what you mean by a turn back.

RomanHotsiy: Yes. I see. I mean how about making it `warn` at all?

packages/core/src/rules/common/security-defined.ts
```diff
@@ -25,6 +27,14 @@ export const OperationSecurityDefined: Oas3Rule | Oas2Rule = () => {
             });
           }
         }
+
+        if (root.security || eachOperationHasSecurity) {
+          return;
+        } else {
+          report({
+            message: `Every API should have security defined on the root level or for each operation.`,
```
AntonKozachuk, line 35: @adamaltman would use your advise on the message here. Thanks.

adamaltman: Sounds good to me.

packages/core/src/config/recommended.ts
RomanHotsiy: Let's keep it an error.

adamaltman (approved):

PR #863, opened 2022-09-12
feat: add build-docs command

AlexVarchuk:
## What/Why/How?
Add new command `build-docs` to cli. It builds definition and Redoc into a zero-dependency HTML file.
## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/build-docs/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+import { compile } from 'handlebars';
+import { join } from 'path';
+import { existsSync, lstatSync, readFileSync } from 'fs';
+
+import type { BuildDocsOptions } from './types';
+
+export function getObjectOrJSON(
+  options: string | Record<string, unknown>
+): JSON | Record<string, unknown> | Config {
+  const configFile = findConfig();
```
SmoliyY, line 15: I would call find `findConfig()` only in `default` case because this is only case where we use it

packages/cli/src/commands/build-docs/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    const start = Date.now();
+    const spec = await loadAndBundleSpec(isURL(pathToSpec) ? pathToSpec : resolve(pathToSpec));
+    const pageHTML = await getPageHTML(spec, pathToSpec, { ...config, redocCurrentVersion });
+
+    mkdirSync(dirname(config.output), { recursive: true });
+    writeFileSync(config.output, pageHTML);
+    const sizeInKiB = Math.ceil(Buffer.byteLength(pageHTML) / 1024);
+    const time = Date.now() - start;
+    console.log(
```
SmoliyY, line 31: Can we use instead `console.log` - `process.stderr.write` ?  And for all other case below to?

packages/cli/src/commands/build-docs/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  try {
+    const start = Date.now();
+    const spec = await loadAndBundleSpec(isURL(pathToSpec) ? pathToSpec : resolve(pathToSpec));
+    const pageHTML = await getPageHTML(spec, pathToSpec, { ...config, redocCurrentVersion });
+
+    mkdirSync(dirname(config.output), { recursive: true });
+    writeFileSync(config.output, pageHTML);
+    const sizeInKiB = Math.ceil(Buffer.byteLength(pageHTML) / 1024);
+    const time = Date.now() - start;
```
SmoliyY, line 30: Instead of this I would recommend to use `getExecutionTime` from `packages/cli/src/utils.ts`

packages/cli/src/commands/build-docs/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+export function escapeClosingScriptTag(str: string): string {
+  return str.replace(/<\/script>/g, '<\\/script>');
+}
+
+// see http://www.thespanner.co.uk/2011/07/25/the-json-specification-is-now-wrong/
+export function escapeUnicode(str: string): string {
+  return str.replace(/\u2028|\u2029/g, (m) => '\\u202' + (m === '\u2028' ? '8' : '9'));
+}
+
+export function handleError(error: Error) {
```
SmoliyY, line 113: Maybe we can reuse `exitWithError` function from `packages/cli/src/utils.ts` ?

SmoliyY:
Also it would be great to have at least one e2e test for this new command!

packages/cli/src/commands/build-docs/utils.ts
AlexVarchuk: I prefer it too. But eslint config has the rule `no-case-declarations`. I disabled it for this case.
But I see in the code we have other cases with ignoring this rule.
What do you think, should we disable this rule in the future? Does that make sense?

packages/cli/src/commands/build-docs/index.ts
AlexVarchuk: Goo point. Done

packages/cli/package.json
```diff
@@ -42,11 +42,20 @@
     "glob": "^7.1.6",
     "glob-promise": "^3.4.0",
     "handlebars": "^4.7.6",
+    "mobx": "^6.3.2",
     "portfinder": "^1.0.26",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "redoc": "^2.0.0",
     "simple-websocket": "^9.0.0",
+    "styled-components": "^5.3.0",
```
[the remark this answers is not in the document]
AlexVarchuk: It is Pure Dependencies for Redoc.
https://github.com/Redocly/redoc/blob/5fb4daa618bb6911ff566530001f1320d818134e/package.json#L136

packages/cli/src/index.ts
```diff
@@ -352,6 +354,60 @@ yargs
...
+      });
+
+      yargs.options('options', {
+        describe: 'ReDoc options, use dot notation, e.g. options.nativeScrollbars',
+      });
+
+      yargs.demandOption('spec');
+      return yargs;
+    },
+    async (argv) => {
```
tatomyr, line 407:
```suggestion
    async (argv) => {
      process.env.REDOCLY_CLI_COMMAND = 'build-docs';
```

packages/cli/src/__tests__/commands/build-docs.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const result = await getPageHTML({}, '../some-path/openapi.yaml', {
+      ...config,
+      redocCurrentVersion: '2.0.0',
+    });
+    expect(renderToString).toBeCalledTimes(1);
+    expect(createStore).toBeCalledTimes(1);
+    expect(result).toBe('<html></html>');
+  });
+
+  it('should correct work handlerBuildCommand', async () => {
```
tatomyr, line 43:
```suggestion
  it('should work correctly when calling handlerBuildCommand', async () => {
```

packages/cli/src/custom.d.ts
```diff
@@ -0,0 +1 @@
+type GenericObject = Record<string, any>;
```
tatomyr, line 1: Where does it come from?

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "test": "npm run typecheck && npm run unit",
     "jest": "jest ./packages",
-    "unit": "npm run jest -- --coverage --coverageReporters lcov text-summary",
+    "unit": "npm run jest -- --coverage --silent=false --coverageReporters lcov text-summary",
```
tatomyr, line 13: Do we need `--silent=false`?

packages/cli/package.json
```diff
@@ -42,11 +42,20 @@
     "glob": "^7.1.6",
     "glob-promise": "^3.4.0",
     "handlebars": "^4.7.6",
+    "mobx": "^6.3.2",
     "portfinder": "^1.0.26",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
+    "redoc": "^2.0.0",
```
tatomyr, line 49: Not `~`?

RomanHotsiy: I would use ~ for not at least.

packages/cli/src/commands/build-docs/utils.ts
SmoliyY: understand it

packages/cli/package.json
AlexVarchuk: thanks. Done

Oprysk (approved):

tatomyr (approved):
LGTM. 
Please test in Workflows before merging.

SmoliyY (approved):

webpack.config.ts
```diff
@@ -12,9 +12,6 @@ module.exports = {
           {
             loader: 'ts-loader',
           },
-          {
```
tatomyr, line 15: Could you also remove this https://github.com/Redocly/redocly-cli/pull/863/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L68 ?

tatomyr (changes requested):

webpack.config.ts
AlexVarchuk: Thanks, I missed that. removed.

tatomyr (approved):

PR #869, opened 2022-09-14
chore: remove recommended fallback if there is a config

tatomyr:
## What/Why/How?

Currently we provide the default fallback with the `recommended` ruleset if there's no `extends` explicitly set up in a Redocly config file or there's no config file at all. 

We put a deprecation message some time ago, and now it's the highest time to finally change the approach: to extend with the default ruleset ONLY when there's no config provided. 

Also it would be great to remove the deprecated suggestions (like `apiDefinitions` &c.).

- [x] change default fallback behaviour
- [x] check configs for braking changes
- [ ] ~remove deprecated suggestions~

## Check yourself

- [ ] Code is linted

## Security


tatomyr:
Closing this in favor of https://github.com/Redocly/redocly-cli/pull/923

PR #881, opened 2022-09-21
fix: improve build-docs command

AlexVarchuk:
## What/Why/How?

- fix `template.hbs` problem (we do not include `template.hbs` in publish flow for `build-docs` command. This should fix it.)
- add `apis` to build-docs
- rename `options` to `features.openapi`
## Screenshots (optional)
**BEFORE**
Screenshot 2022-09-21 at 17 09 10

**AFTER**
Screenshot 2022-09-21 at 17 07 42

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/index.ts
```diff
@@ -362,53 +362,53 @@ yargs
...
+          t: {
+            alias: 'template',
+            describe: 'Path to handlebars page template, see https://git.io/vh8fP for the example',
+            type: 'string',
+          },
+          templateOptions: {
+            describe:
+              'Additional options that you want pass to template. Use dot notation, e.g. templateOptions.metaDescription',
+          },
+          features: {
```
adamaltman, line 398: Isn't it features.openapi here?

packages/cli/src/commands/build-docs/types.ts
```diff
@@ -19,5 +19,8 @@ export type BuildDocsArgv = {
   disableGoogleFont?: boolean;
   template?: string;
   templateOptions: Record<string, any>;
-  options: string | Record<string, unknown>;
+  features: {
+    openapi: string | Record<string, unknown>;
+  };
```
adamaltman, line 24: Isn't `features.openapi` the single string? Not an object in an object?

packages/cli/src/index.ts
AlexVarchuk: according to the existing [issue](https://github.com/yargs/yargs/issues/1230) and this [comment](https://github.com/yargs/yargs/issues/995#issuecomment-352326378) `yargs` do not support straight forward solution for this `dot-notation`. It is a workaround solution for that.

packages/cli/src/commands/build-docs/types.ts
AlexVarchuk: No, it workaround flow for support `dot-notation`.  I tried update to the latest version of yargs, but they still have this issue.
```bash
redocly build-docs -- --config=./resources/.redocly.yaml --features.openapi.theme.colors.primary.main=red
```
Screenshot 2022-09-26 at 14 48 55

tatomyr (approved):

PR #886, opened 2022-09-26
fix: path parameter not defined

### LINKED ISSUE #211: Path parameter isn't defined
**Describe the bug**
If a path has a named placeholder but doesn't have a path parameter, then we should show the problem.

> Each template expression in the path MUST correspond to a path parameter that is included in the Path Item itself and/or in each of the Path Item's Operations.

**To Reproduce**
1. Given this `.redocly.yaml` file (just the `recommended` rules)
2. Contact Adam or see a sample below.
3. Run this command with these arguments... `openapi lint`
4. See no error

**Expected behavior**
I expect to see this error: https://redoc.ly/docs/cli/built-in-rules/#path-parameters-defined

**Logs**

```
validating /home/adam/Downloads/2020-09-23-13-03-openapi-od_3p8p9x.yaml...
/home/adam/Downloads/2020-09-23-13-03-openapi-od_3p8p9x.yaml: validated in in 86ms

```

**OpenAPI definition**

This is an excerpt.

The one with the path template was defined first:
```
paths:
  /something/{id}:
   parameters:
     - name: foo
       in: header
       required: true
       schema:
         type: string
  /something:
# ...
```

**`openapi-cli` Version(s)**
1.0.0-beta.14

**`Node.js` Version(s)**
v12.6.0

**Additional context**

Contact Adam if you need a full OpenAPI definition.

SmoliyY:
## What/Why/How?
If parameters ware not defined in the other operation but exist in first lint did not catch error.
For example 
```yaml

paths:
  /case/{a}:
    get:
      parameters: 
        - name: a
          in: path
          schema: 
            type: number 
      summary: Case GET 
      description: Should not fail cause parameters exist
      responses:
        '200':
          description: Successful operation

    post: 
      summary: Case POST
      description: Should fail cause no parameters
      responses:
        '200':
          description: Successful operation
```
Screenshot 2022-09-26 at 16 04 04

## Reference
Closes #211 

## Screenshots (optional)
Now it throw appropriate error
Screenshot 2022-09-26 at 16 04 46

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/common/path-params-defined.ts
+++ b/packages/core/src/rules/common/path-params-defined.ts
@@ -9,6 +9,7 @@ export const PathParamsDefined: Oas3Rule | Oas2Rule = () => {
   let pathTemplateParams: Set<string>;
   let definedPathParams: Set<string>;
   let currentPath: string;
+  let definedOperationParams: Set<string>;
 
   return {
     PathItem: {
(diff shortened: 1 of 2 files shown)


packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -31,19 +32,24 @@ export const PathParamsDefined: Oas3Rule | Oas2Rule = () => {
...
+        },
         leave(_op: object, { report, location }: UserContext) {
           for (const templateParam of Array.from(pathTemplateParams.keys())) {
-            if (!definedPathParams.has(templateParam)) {
-              report({
-                message: `The operation does not define the path parameter \`{${templateParam}}\` expected by path \`${currentPath}\`.`,
-                location: location.child(['parameters']).key(), // report on operation
-              });
+            if (!definedOperationParams.has(templateParam)) {
+              if (!definedPathParams.has(templateParam)) {
```
IgorKarpiuk, line 41: Can it be combined into one condition?

SmoliyY: Yes, thx)

IgorKarpiuk (approved):

RomanHotsiy (approved):
👍

Issue #850, opened 2022-09-05
lint command isn't working with URLs from API Registry used inline


AntonKozachuk:
**Describe the bug**
When you try to use the `lint` command and define API inline as a URL from API Registry, which has a branch query string, the command seems to break.
At the same time, it works when used inside the `apis` section of the configuration file. 

Example:
`redocly lint "https://api.redocly.com/registry/bundle/example/v1/latest/openapi.yaml?branch=release"
`
The `?branch=release` part seems to break the command.

**To Reproduce**
1. Add a definition to the API Registry or open an existing one.
2. Copy the link to the definition's [snapshot](https://redocly.com/docs/api-registry/guides/snapshots/#how-to-find-your-snapshot-links).
3. Try to use it with the lint command.
4. See error.

**Expected behavior**
The lint command works as expected when used with URLs from the API Registry.

**Logs**
N/a

**OpenAPI definition**
N/a

**Redocly Version(s)**
latest

**`Node.js` Version(s)**
15

**Additional context**
Some context can be found here: [redoc-ly.slack.com/archives/C02UHE7EL3E/p1661807377826659](https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1661807377826659)

lornajane:
I'm not sure when this old issue was fixed, but retesting it today I cannot replicate the issue (and I'm sure I've used this before without a problem)

PR #887, opened 2022-09-27
feat: handle error if api does not exist

### LINKED ISSUE #849: lint command not working when no API is added to options
**Describe the bug**
When you have added a configuration file (`.redocly.yaml`) and defined your `apis` there, you're still unable to run `redocly lint` without other arguments. 

Example:

Working:
`redocly lint someApi@v1`
Not working:
`redocly lint`  

**To Reproduce**
1. Create `redocly.yaml` file and add a definition to the `apis` section.
2. Run this command - `redocly lint`  
4. See error:
`error: missing required argument apis.
`

**Expected behavior**
The `lint` command picks up the definitions defined in the `apis` section of the configuration file and works as expected.

**Logs**
N/a

**OpenAPI definition**
N/a

**Redocly Version(s)**
latest

**`Node.js` Version(s)**
15

**Additional context**
Some context can be found here: https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1661807377826659

SmoliyY:
## What/Why/How?
Add error handle in cases if `definition` does not exist. Previously `cli` throw native error 
Screenshot 2022-09-26 at 17 19 41

Also there was and issue if `root` of apis provided as link. See #849 and #850

## Reference
Closes #849, #850
## Screenshots (optional)
Example with run `redocly lint resources/pets.yaml not-exist.yaml`
Screenshot 2022-09-28 at 18 21 50

## Check yourself

- [ ] Code is linted

## Security


packages/cli/src/utils.ts
```diff
@@ -45,6 +45,34 @@ function isNotEmptyArray<T>(args?: T[]): boolean {
   return Array.isArray(args) && !!args.length;
 }
 
+function isApiPathValid(apiPath: string): string | void {
+  if (!apiPath.trim()) {
+    exitWithError('Path can not be empty.');
```
adamaltman, line 50:
```suggestion
    exitWithError('Path cannot be empty.');
```

packages/cli/src/utils.ts
```diff
@@ -45,6 +45,34 @@ function isNotEmptyArray<T>(args?: T[]): boolean {
...
+    return;
+  }
+  if (fs.existsSync(apiPath) || isURL(apiPath)) {
+    return apiPath;
+  } else {
+    exitWithError(
+      yellow(`\n ${apiPath} ${red(`doe's not exist or invalid. Please provide valid path.`)}`)
```
adamaltman, line 57:
```suggestion
      yellow(`\n ${apiPath} ${red(`does not exist or is invalid. Please provide a valid path.`)}`)
```

redocly.yaml
```diff
@@ -0,0 +1,5 @@
+apis: 
+  main:
+    root: ./../../../do-not-exist.yaml
```
adamaltman, line 3:
```suggestion
    root: ./../../../does-not-exist.yaml
```

adamaltman (changes requested):
Some grammar changes to the messages. This is missing tests too.

packages/cli/src/utils.ts
```diff
@@ -323,3 +351,7 @@ export function isSubdir(parent: string, dir: string): boolean {
   const relative = path.relative(parent, dir);
   return !!relative && !/^..($|\/)/.test(relative) && !path.isAbsolute(relative);
 }
+
+export function isURL(str: string): boolean {
+  return /^(https?:)\/\//m.test(str);
```
adamaltman, line 356: What if it's an http url?

packages/cli/src/utils.ts
```diff
@@ -25,13 +25,20 @@ export async function getFallbackApisOrExit(
...
-  if (!isNotEmptyArray(res)) {
-    process.stderr.write('error: missing required argument `apis`.\n');
+
+  const filteredInvalidEntrypoints = res.filter(({ path }) => !isApiPathValid(path));
+  if (isNotEmptyArray(filteredInvalidEntrypoints)) {
+    for (const { path } of filteredInvalidEntrypoints) {
+      process.stderr.write(
+        yellow(
+          `\n ${relative(process.cwd(), path)} ${red(
+            `doe's not exist or invalid. Please provide valid path. \n\n`
```
adamaltman, line 37:
```suggestion
            `does not exist or is invalid. Please provide a valid path. \n\n`
```

redocly.yaml
SmoliyY: This file got here accidentally. Thx for pointing out

packages/cli/src/utils.ts
SmoliyY: It will pass, but there is another check that already exist and can be reused.

adamaltman (approved):

tatomyr:
Please check if the logic covers the `extends` functionality: https://redocly.com/docs/cli/configuration/#extends-list

RomanHotsiy (approved):

adamaltman:
And check it covers the `plugins` paths too.

SmoliyY:
> And check it covers the `plugins` paths too.
> Please check if the logic covers the `extends` functionality: https://redocly.com/docs/cli/configuration/#extends-list

This check do not handle both of those cases. It not hard so to add such check, but it depends if we want to proceed in cases if `plugin` or `config` does not exist. I think we should proceed with same behaviour as it is work now ( Throw error and stop executing but with more efficient error)

cc @adamaltman @tatomyr

tatomyr:
I believe we should also cover `plugins` and `extends`. To me there's sense to proceed with a config which is trying to extend from non-existing config or a plugin. I believe we should add errors for those cases as well.

adamaltman:
Yes, with a friendly error message.

packages/cli/src/utils.ts
```diff
@@ -327,6 +356,24 @@ export function exitWithError(message: string) {
...
+export async function loadConfigAndHandleErrors(
+  options: {
+    configPath?: string;
+    customExtends?: string[];
+    processRawConfig?: (rawConfig: RawConfig) => void | Promise<void>;
+    files?: string[];
+    region?: Region;
+  } = {}
+): Promise<Config> {
+  let config = new Config({ apis: {}, styleguide: {} });
```
tatomyr, line 372: I'd suggest using `const` instead of `let`.

packages/core/src/config/config-resolvers.ts
```diff
@@ -342,7 +346,7 @@ async function loadExtendStyleguideConfig(
 
     return rawConfig.styleguide;
   } catch (error) {
-    throw new Error(`Failed to load "${filePath}": ${error.message}`);
+    throw new Error(`Failed to load "${filePath}": Please provide a valid path`);
```
tatomyr, line 349: Why don't we want to keep the generic `error.message` here? In future there might be different errors inside the function, not necessarily related to  a wrong path.

tatomyr (approved):

Issue #856, opened 2022-09-07
type array list not validated in OAS 3.1


adamaltman:
**Describe the bug**

When a schema has type `null` defined in YAML, it is being treated as a null value and not "null" the string that defines the types.

https://json-schema.org/understanding-json-schema/reference/null.html

https://github.com/Redocly/openapi-starter/blob/main/openapi/components/schemas/Schema.yaml#L41

I believe that Schema needs to be adjusted to the following to define the type properly:

```yaml
  nullableOrStringProperty:
    description: Property name's description (type is string or null)
    type:
      - string
      - "null"
```

However, I then suspected that nothing in the type list is being validated:

https://github.com/Redocly/openapi-starter/pull/74

This bug does not exist when type is defined as a string (for example the following would produce a problem):
```
type: foo
```

**To Reproduce**
1. See https://github.com/Redocly/openapi-starter/pull/74
2. See error (that there is no problem)

**Expected behavior**

problem message: 

> `type` can be one of the following only: "object", "array", "string", "number", "integer", "boolean", "null".

**Logs**

There is no problem detected (which is the problem because there is a problem).

**OpenAPI definition**

OAS 3.1
https://github.com/Redocly/openapi-starter/blob/main/openapi/components/schemas/Schema.yaml#L41

**Redocly Version(s)**

1.0.0-beta.108

**`Node.js` Version(s)**

16.10

**Additional context**

- This was revealed by some crashing of our mock server.

razvanphp:
Hmm, interesting, so my schema is invalid? `null` needs to always be defined as a string? Why?

```
allOf:
  - type: object
    properties:
      car:
        $ref: ./Car.yaml
      user:
        type:
          - object
          - null
        $ref: ./User.yaml
      user_id:
        type:
          - integer
          - null
      car_id:
        type:
          - integer
          - null
```
This was not triggering an error on `1.0.0-beta.125`, but it does now.

PR #891, opened 2022-09-30
fix: validate types in spec rule

IgorKarpiuk:
## What/Why/How?

Validate types list in spec rule

## Reference
closes https://github.com/Redocly/redocly-cli/issues/856
## Screenshots (optional)
Знімок екрана 2022-09-30 о 11 46 27

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/common/spec.ts
+++ b/packages/core/src/rules/common/spec.ts
@@ -119,21 +119,22 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
             propValue.forEach((item) => {
               if (!propSchema.items?.enum?.includes(item !== null ? item : JSON.stringify(item))) {
                 report({
-                  message: 'type can be one of the following only: "object", "array", "string", "number", "integer", "boolean", "null".',
+                  location: propLocation,
+                  message: `\`${propName}\` can be one of the following only: ${propSchema.items
+                    ?.enum!.map((i) => `"${i}"`)
+                    .join(', ')}.`,
                   from: refLocation,
+                  suggest: getSuggest(item, propSchema.items?.enum as string[]),
                 });
               }
-            });
-          } else if (!propSchema.items.enum.includes(propValue !== null ? propValue : JSON.stringify(propValue))) {
-            report({
-              message: 'type can be one of the following only: "object", "array", "string", "number", "integer", "boolean", "null".',
-              from: refLocation,
             });
           }
         }
 
         if (propSchema.enum) {
-          if (!propSchema.enum.includes(propValue)) {
+          if (
+            !propSchema.enum.includes(propValue !== null ? propValue : JSON.stringify(propValue))
+          ) {
             report({
               location: propLocation,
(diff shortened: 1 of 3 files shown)


packages/core/src/rules/common/spec.ts
```diff
@@ -114,8 +114,27 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
...
+                });
+              }
+            });
+          }
+        }
+
         if (propSchema.enum) {
-          if (!propSchema.enum.includes(propValue)) {
+          if (
+            !propSchema.enum.includes(propValue !== null ? propValue : JSON.stringify(propValue))
```
tatomyr, line 136:
```suggestion
            !propSchema.enum.includes(propValue === null ? 'null' : propValue)
```

packages/core/src/rules/common/spec.ts
```diff
@@ -114,8 +114,27 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
...
+          ) {
```
tatomyr, line 137: I believe we have a duplication here. Can we avoid it?

IgorKarpiuk: I moved similar code to separate function

RomanHotsiy (approved):

packages/core/src/rules/common/spec.ts
```diff
@@ -114,17 +115,20 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
...
-                .map((i) => `"${i}"`)
-                .join(', ')}.`,
-              from: refLocation,
-              suggest: getSuggest(propValue, propSchema.enum),
-            });
+        if (propSchema.items && propSchema.items?.enum && Array.isArray(propValue)) {
+          for (const item of propValue) {
+            validateSchemaEnumType(propSchema.items?.enum, item, propName, refLocation, {
+              report,
+              location,
```
RomanHotsiy, line 122: Can we change the loop to regular `for` and use the index for location here:

```suggestion
              location: location.child([propName, i])
```

This way it will highlight a specific array item that is invalid vs highlighting the whole array

RomanHotsiy (changes requested):

packages/core/src/rules/common/spec.ts
IgorKarpiuk: Yes, Thank you

packages/core/src/rules/utils.ts
```diff
@@ -138,3 +138,25 @@ export function getAdditionalPropertiesOption(opts: Record<string, any>): boolea
...
+  propName: string,
+  refLocation: Location | undefined,
+  { report, location }: UserContext
+) {
+  if (!schemaEnum) {
+    return;
+  }
+  if (!schemaEnum.includes(propertyValue === null ? 'null' : propertyValue)) {
+    report({
+      location: location,
```
RomanHotsiy, line 154:
```suggestion
      location,
```

RomanHotsiy (approved):
Left one small suggestion

Issue #440, opened 2021-11-04
`media-type-examples-override` decorator


adamaltman:
**Is your feature request related to a problem? Please describe.**

To solve the same issue we solved with other decorators. 

Most developers are too lazy to define media type examples.

https://github.com/redocly-demo/decorators-demo/pull/1

In the event the OpenAPI file is auto-generated, then there is no reliable way to add examples (without them getting overwritten on the next auto-generation).

**Describe the solution you'd like**

I'd like to see:

```
decorators:
  media-type-examples-override:
    operationIds:
      PostPets: ./pet-examples.yaml
```

**Describe alternatives you've considered**
None - I am open to ideas.

IgorKarpiuk:
@adamaltman What should we do if we have few media type examples in one operation?

For example: 
```yaml
responses:
        200:
          description: json
          content:
            application/json:
              examples:
                 test1:
                   value:
                     a: test1
                     b: 35
        201:
          description: xml
          content:
            application/xml:
              examples:
                test2:
                  value:
                   a: test2
```

tatomyr:
@adamaltman , @RomanHotsiy any thoughts?

RomanHotsiy:
There is also request body example.

I'm not sure. @adamaltman what was your idea about the format of `./pet-examples.yaml` file? Was it supposed to have multiple examples somehow?

adamaltman:
Yes, it can have multiple examples. However, that's a good question. It needs to target the response code or the request media type.

So maybe it should be like this:
```
decorators:
  media-type-examples-override:
    operationIds:
      '200':
         PostPets: ./pet-examples.yaml
      '400':
         PostPets: ./pet-errors-examples.yaml

```

IgorKarpiuk:
@adamaltman  How we should handle $ref in responses?

```yaml
responses:
        200:
          $ref: '#/components/responses/okay'
        400:
         $ref: '#/components/responses/badRequest'
```
1. We can override examples directly inside components section
    But may be a problem with links to one response in several places. For example:
```yaml
responses:
        200:
          $ref: '#/components/responses/okay200'
        400:
         $ref: '#/components/responses/okay200'
```
2. We can resolve the refs and override the examples already in the right places. But the references needed in the decorator will be resolved and all others will not

IgorKarpiuk:
I implemented the second option because it looks more secure

PR #897, opened 2022-10-06
feat: add media type examples override decorator

IgorKarpiuk:
## What/Why/How?

Add media type examples override decorator

## Reference
closes: https://github.com/Redocly/redocly-cli/issues/440
## Check yourself

- [x] Code is linted

## Security


packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,41 @@
+import { Oas3Decorator } from '../../visitors';
+import { Oas3Operation, Oas3RequestBody } from '../../typings/openapi';
+import { readAndParseFileSync } from '../../utils';
+
+export const MediaTypeExamplesOverride: Oas3Decorator = ({ operationIds }) => {
+  let operationId: string | undefined;
```
tatomyr, line 6: Do you think it should be here? Let's try to remove it.

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,41 @@
...
+  return {
+    Operation: {
+      leave(operation: Oas3Operation) {
+        operationId = operation.operationId;
+        const properties = operationIds[operationId!];
```
tatomyr, line 11: I'd put some kind of guard instead of '!'.

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        if (properties.responses && operation.responses) {
```
tatomyr, line 13: Could you use a uniform approach both here and [here](https://github.com/Redocly/redocly-cli/pull/897/files#diff-150f23c15f2f75bc7e9556b690b68cd5c77fa61ce8801d6fd346ea86775c26feR20)?

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,41 @@
...
+          for (const responseCode of Object.keys(properties.responses)) {
+            if (!operation.responses[responseCode] || !operation.responses[responseCode]?.content) {
+              continue;
+            }
+
+            Object.values(operation.responses[responseCode].content!).forEach((mimeType) => {
+              if (mimeType.examples) {
+                mimeType.examples = readAndParseFileSync(properties.responses[responseCode]);
```
tatomyr, line 21: I'd suggest figuring out what type of example field is used, something like this:

```ts
const exampleField = mimeType.examples ? 'examples' : 'example';
mimeType[exampleField] = readAndParseFileSync(properties.responses[responseCode]);
```

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,41 @@
...
+        operationId = operation.operationId;
+        const properties = operationIds[operationId!];
+
+        if (properties.responses && operation.responses) {
+          for (const responseCode of Object.keys(properties.responses)) {
+            if (!operation.responses[responseCode] || !operation.responses[responseCode]?.content) {
+              continue;
+            }
+
+            Object.values(operation.responses[responseCode].content!).forEach((mimeType) => {
```
tatomyr, line 19: Could you use either `forEach` or `for ... of` here and [here](https://github.com/Redocly/redocly-cli/pull/897/files#diff-150f23c15f2f75bc7e9556b690b68cd5c77fa61ce8801d6fd346ea86775c26feR33)?

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,39 @@
...
+              const exampleField = mimeType.examples ? 'examples' : 'example';
+              mimeType[exampleField] = readAndParseFileSync(properties.responses[responseCode]);
+            });
+          }
+        }
+
+        if (properties.request && operation.requestBody) {
+          Object.keys(properties.request).forEach((mimeType) => {
+            (operation.requestBody as Oas3RequestBody).content[mimeType].example =
```
tatomyr, line 32: Would it work when we're referencing say, `components`?

packages/core/src/decorators/__tests__/medi-type-examples-override.test.ts
```diff
@@ -0,0 +1,432 @@
+import { makeConfig, parseYamlToDocument, yamlSerializer } from '../../../__tests__/utils';
```
adamaltman, line 1: The file name has a typo in "media".

packages/core/src/decorators/__tests__/medi-type-examples-override.test.ts
```diff
@@ -0,0 +1,432 @@
...
+      paths:
+        /pet:
+          get:
+            operationId: getUserById
+            responses:
+              '200':
+                description: json
+                content:
+                  application/json:
+                    example:
```
adamaltman, line 58: This seems like it is a mistake. I think this means to be `examples`?

adamaltman (changes requested):
This should always place target examples into the `examples` map and never into `example`. The examples map accepts a value in a specific structure (Example object)[https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#media-type-object] whereas the `example` field expects the raw example itself.

I didn't see the most common use case tested, which is no examples defined and then the override adds an example to the examples map.

packages/core/src/decorators/common/media-type-examples-override.ts
IgorKarpiuk: Yes

packages/core/src/decorators/__tests__/medi-type-examples-override.test.ts
IgorKarpiuk: Yes, thanks

IgorKarpiuk:
> This should always place target examples into the `examples` map and never into `example`. The examples map accepts a value in a specific structure (Example object)[https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#media-type-object] whereas the `example` field expects the raw example itself.
> 
> I didn't see the most common use case tested, which is no examples defined and then the override adds an example to the examples map.

@adamaltman I changed `example` to `examples` now it will be always place target examples into the `examples` field and i added new tests

packages/core/src/decorators/__tests__/media-type-examples-override.test.ts
```diff
@@ -0,0 +1,555 @@
...
+      components:
+        examples:
+          testExample:
+            value:
+              test: 1
+
+    `);
+  });
+
+  it('should not override the examples', async () => {
```
adamaltman, line 387: This is the wrong behavior. It should create the examples.

adamaltman (changes requested):
The behavior should be different. When you want to add an example and there isn't an example there yet, it should allow you to do that. The test I highlighted seems to do the opposite behavior.

packages/core/src/decorators/__tests__/media-type-examples-override.test.ts
adamaltman: How is the config currently?

IgorKarpiuk: @adamaltman 

Now it looks like this: 
```yaml
decorators:
  media-type-examples-override:
    operationIds:
       PostPets:
          request: 
              application/json: ./pet-examples.yaml
          responses: 
            '200': ./pet-examples.yaml
            '400': ./pet-errors-examples.yaml
```

adamaltman: Yes, you're right.

IgorKarpiuk: Changed ✅

adamaltman (approved):
You should also add docs. You can look at the other decorator docs and copy the page structure and change the details.

packages/core/src/utils.ts
```diff
@@ -153,6 +153,11 @@ export function readFileAsStringSync(filePath: string) {
   return fs.readFileSync(filePath, 'utf-8');
 }
 
+export function readAndParseFileSync<T>(filePath: string) {
```
tatomyr, line 156: Is it only for yaml and json files? If so, it worth renaming this functions to be explicit.

packages/core/src/decorators/common/media-type-examples-override.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const MediaTypeExamplesOverride: Oas3Decorator = ({ operationIds }) => {
+  function checkAndResolveRef<T>(node: any, resolver: ResolveFn): T | undefined {
+    if (!isRef(node)) {
+      return node;
+    }
+
+    const resolved = resolver<T>(node);
+    return resolved.error ? undefined : JSON.parse(JSON.stringify(resolved.node));
+  }
```
tatomyr, line 15: Please move this function out of the decorator.

tatomyr (approved):

tatomyr:
Left a few minor comments.

adamaltman (approved):

PR #902, opened 2022-10-10
fix: fix api overrides not picking some values from root config

### LINKED ISSUE #899: Any API-specific override resets all root settings
**Describe the bug**

Defining any API-specific override results in all non-customized props not receiving the root values.

**To Reproduce**
    ```yaml
        apis:
          repro@bug: 
            root: openapi.yaml
            decorators: {}
        extends:
          - recommended
    ```
2. And this OpenAPI file (without any Info license which will cause a warning with the `recommended` rules):
    ```yaml
    openapi: 3.1.0
    info:
      version: 1.0.0
      title: Example.com
      termsOfService: 'https://example.com/terms/'
      description: Sample.
    servers:
      - url: 'https://{tenant}/api/v1'
        variables:
          tenant:
            default: www
            description: Your tenant id
      - url: 'https://example.com/api/v1'
    paths: {}
    ```
3. Run this command with these arguments... `redocly lint`
4. See error (`Woohoo! Your OpenAPI definition is valid. 🎉` with no mention of any warning)

**Expected behavior**

I expected the above repro to say there is 1 warning.

I expect that `extends` defined at the root level of the configuration will apply to every API unless that API specifically overrides the `extends` property itself (even if it overrides a decorator as shown above).

For example, a specific override:

```yaml
        apis:
          repro@bug: 
            root: openapi.yaml
            decorators: {}
            extends: []
        extends:
          - recommended
```

**Logs**

n/a

**Redocly Version(s)**

- 1.0.0-beta.110
- 1.0.0-beta.109

It works in 1.0.0-beta.108 (although I needed to adjust the Redocly configuration file to reflect changes):
```yaml
apis:
  repro@bug: 
    root: openapi.yaml
    styleguide: 
      decorators: {}
styleguide: 
  extends:
    - recommended
```

**`Node.js` Version(s)**
16

**Additional context**

- Discovered when trying to reproduce #898

RomanHotsiy:
## What/Why/How?

The values were set to `undefined` by previous config processing in api override so it was overwriting all the root config values.

## Reference

Fixes #899 

## Testing

Added a test (this specific case was missing).

## Check yourself

- [x] Code is linted

## Security


packages/core/src/config/__tests__/config-resolvers.test.ts
```diff
@@ -464,4 +464,29 @@ describe('resolveConfig', () => {
...
+        rules: {
+          'operation-2xx-response': 'warn',
+        },
+      },
+    };
+
+    const { apis } = await resolveConfig(rawConfig, configPath);
+    expect(apis['petstore'].styleguide.rules).toBeDefined();
+    expect(apis['petstore'].styleguide.rules?.['operation-2xx-response']).toEqual('warn'); // form minimal ruleset
+  });
```
adamaltman, line 491: The `styleguide` object was eliminated in 109. 
```suggestion
  it('should default to the extends from the main config if no extends defined', async () => {
    const rawConfig: RawConfig = {
      apis: {
        petstore: {
          root: 'some/path',
          rules: {
            'operation-4xx-response': 'error',
          },
        },
      },
      extends: ['minimal'],
      rules: {
        'operation-2xx-response': 'warn',
      },
    };

    const { apis } = await resolveConfig(rawConfig, configPath);
    expect(apis['petstore'].rules).toBeDefined();
    expect(apis['petstore'].rules?.['operation-2xx-response']).toEqual('warn'); // form minimal ruleset
  });
```

adamaltman: I opened an issue about this: https://github.com/Redocly/redocly-cli/issues/904

adamaltman (approved):

AntonKozachuk (approved):

packages/core/src/config/__tests__/config-resolvers.test.ts
```diff
@@ -464,4 +464,29 @@ describe('resolveConfig', () => {
...
+        extends: ['minimal'],
+        rules: {
+          'operation-2xx-response': 'warn',
+        },
+      },
+    };
+
+    const { apis } = await resolveConfig(rawConfig, configPath);
+    expect(apis['petstore'].styleguide.rules).toBeDefined();
+    expect(apis['petstore'].styleguide.rules?.['operation-2xx-response']).toEqual('warn'); // form minimal ruleset
```
RomanHotsiy, line 490:
```suggestion
    expect(apis['petstore'].styleguide.rules?.['operation-2xx-response']).toEqual('warn'); // from minimal ruleset
```

PR #903, opened 2022-10-10
fix: detect api from the list by the root file match

RomanHotsiy:
## What/Why/How?

If `redocly-cli lint` was run with path to the file as argument it was not detecting proper overrides from `apis` config.

## Reference

Found it while fixing https://github.com/Redocly/redocly-cli/issues/899

## Testing

Added a unit test.

## Check yourself

- [x] Code is linted

## Security


RomanHotsiy:
> What happens if the same root file is used for multiple APIs:

The first match is used.

adamaltman (approved):

PR #906, opened 2022-10-10
refactor: rename four types

adamaltman:
Changed **types**:

- `PathsMap` to `Paths`
- `ResponsesMap` to `Responses`
- `EncodingsMap` to `EncodingMap`
- `SecuritySchemeFlows` to `OAuth2Flows`

## Reference

Closes #900 

## Testing

No code changes (only type names) -- existing tests should pass.

## Check yourself

- [x] Code is linted
- [ ] Tested with redoc/reference-docs/workflows
- [x] All new/updated code is covered with tests

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #894, opened 2022-10-03
feat: new assertion syntax 

roman-sainchuk:
## What/Why/How?
This is rough implementation of new assertion syntax according to ~[docs](https://github.com/Redocly/redocly-cli/pull/857)~ [new docs](https://github.com/Redocly/redocly-cli/pull/917).

## Reference

Also closes https://github.com/Redocly/redocly-cli/issues/915

## Check yourself

- [x] Code is linted
- [x] Tested with redoc/reference-docs/workflows
- [x] All new/updated code is covered with tests

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #917, opened 2022-10-23
docs: rework assertions into custom rules

adamaltman:
## What/Why/How?

- Rename assertions to custom rules. 
- Reshape the object (as shown in the docs).

## Reference

- #857 

## Check yourself

- [ ] Code is linted

## Security


docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
 
 
 Property | Type | Description
 -- | -- | --
-assert/my-assertion-name | [Assertion object](#assertion-object) | An assertion definition to enforce your custom API design standards. Add or edit your assertions in the configuration file. The assertion is a rule that starts with a `assert/` prefix followed by a unique assertion rule name. Assertion name is displayed in the lint log if the assertion fails. More than one assertion may be defined, and any assertion may have multiple asserts.
+my-custom-rule-name | [Custom rule object](#custom-rule-object) | Custom rules definitions enforce your custom API design standards. Add or edit your custom rules in the configuration file. A custom rule is a rule that starts with a `assert/` prefix followed by a unique rule name. Assertion names display in the lint log if the assertion fails. More than one assertion may be defined, and any custom rule may have multiple asserts.
```
RomanHotsiy, line 30: it doesn't have `assert/` prefix but the docs mention it.

docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
-subject | `string` \| [`string`] | **REQUIRED.** The [OpenAPI node type](#openapi-node-types) that the lint evaluates. Use with `context` for more control.
-property | `string` \| [`string`] \| null | Property name corresponding to the [OpenAPI node type](#openapi-node-types). If a list of properties is provided, assertions will evaluate against each property in the sequence. If not provided (or null), assertions will evaluate against the key names for the subject node type. See [property example](#property-example).
-context | [Context object](#context-object) | The context influences evaluation for assertions. When `matchParentKeys` or `excludeParentKeys` is used in the `context` object, it evaluates the specified subset of the subject type. The resolution of reference objects is done at the context level. If no context is provided, it evaluates the assertion for all instances of the given type. See [context example](#context-example).
+subject | [Subject object](#subject-object) | **REQUIRED.** Locates the [OpenAPI node type](#openapi-node-types) and possible properties and values that the [lint command](../commands/lint.md) evaluates. Use with `where` to filter further.
```
RomanHotsiy, line 36:
```suggestion
subject | [Subject object](#subject-object) | **REQUIRED.** Locates the [OpenAPI node type](#openapi-node-types) and possible properties and values that the [lint command](../commands/lint.md) evaluates. Use with `where` to narrow further.
```

docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
+assertions | [Assertion object](#assertion-object) | **REQUIRED.** Flags a problem when a defined assertion evaluates false. There are a variety of built-in assertions included. You may also create plugins with custom functions and use them as assertions.
+where | [Where object](#where-object) | Influences evaluation for assertions by only evaluating assertions when the `where` arguments evaluate to true. The `where` argument must be structured as a list of valid [where objects](#where-object). The `where` list is evaluated from top to bottom. When using the `where` list, it evaluates the filter subject specified with any filter assertions, and then continues to evaluate based on that subset of the values. The resolution of reference objects is done at the `where` level. See [where example](#where-example). The `where` evaluation itself does not result in any problems (it is used for filtering where to run assertions).
```
RomanHotsiy, line 38: We need to simplify and explain it better I think.

docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
-enum | [`string`] | Asserts a value is within a predefined list of values. See [enum example](#enum-example).
-pattern | `string` | Asserts a value matches a regex pattern. See [regex pattern example](#pattern-example).
+
+## Subject object
+
+Property | Type | Description
+-- | -- | --
+type | string |  **REQUIRED.** Locates the [OpenAPI node type](#openapi-node-types) that the [lint command](../commands/lint.md) evaluates.
+property | `string` \| [`string`] \| null | Property name corresponding to the [OpenAPI node type](#openapi-node-types). If a list of properties is provided, assertions evaluate against each property in the sequence. If not provided (or null), assertions evaluate against the key names for the subject node type. See [property example](#property-example).
+filterInParentKeys | [`string`] | The name of the subject's parent key that locates where assertions run. An example value given the subject `Operation` could be `filterInParentKeys: [get, put]` means that only `GET` and `PUT` operations are evaluated for the assertions.
```
RomanHotsiy, line 49: Examples are really useful here 👍.

docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
-maxLength | `integer` | Asserts a maximum length (exclusive) of a string or list (array). See [maxLength example](#maxlength-example).
-ref | `boolean \| string` | Asserts a reference object presence in object's property. A boolean value of `true` means the property has a `$ref` defined. A boolean value of `false` means the property has not defined a `$ref` (it has an in-place value). A string value means that the `$ref` is defined and the unresolved value must match the pattern (for example, `'/paths\/.*\.yaml$/'`). See [ref example](#ref-example).|
-`{pluginId}/{functionName}` | `object` | Custom assert defined in the plugin. This function will be called with options including the value. See [custom function example](#custom-function-example).|
+`{pluginId}/{functionName}` | `object` | Custom assert defined in the plugin. This function is called with options including the value. See [custom function example](#custom-function-example).
+
+## Where object
 
-## Context object
+The `where` object is part of a `where` list which must be defined in order from the root node.
```
RomanHotsiy, line 76:
```suggestion
The `where` object is part of a `where` list which must be defined in order from the root node. Intermediate nodes may be skipped.
```

docs/rules/custom-rules.md
```diff
@@ -16,49 +22,63 @@ rules:
...
 
 Property | Type | Description
 -- | -- | --
-type | `string` | **REQUIRED.** One of the [OpenAPI node types](#openapi-node-types).
-matchParentKeys | [`string`] | The list of parent object key names to evaluate with respect to the subject.
-excludeParentKeys | [`string`] | The list of parent object key names to not evaluate with respect to the subject.
-
-See the [context example](#context-example).
-
+whereSubject | Subject object | **REQUIRED.** Filters the subject further based on the definition.
```
RomanHotsiy, line 80: so you want to add prefix `where`? What is the reason for this? To minimize the confusion?

docs/rules/custom-rules.md
```diff
@@ -131,61 +157,99 @@ Notice we don't need to include `property` in this approach.
...
+  assertions:
+    enum:
+      - integer
+  where:
+    - whereSubject:
+        type: Parameter
+        property: name
+      whereAssertions:
+        enum:
+          - limit
```
RomanHotsiy, line 244: what do you think about using a shorter version of lists to make examples more compact?

```suggestion
assert/limit-is-integer:
  subject:
    type: Schema
    property: type
  assertions:
    enum: [integer]
  where:
    - whereSubject:
        type: Parameter
        property: name
      whereAssertions:
        enum: [limit]
```

RomanHotsiy (commented):
Left some comments.

There is one more aspect which needs explanation I think:

- if there is a property defined in the subject assertions run on the property value
- if there is only a type defined and no property, assertions run on the Subject's properties.
- some assertions can't be applied to a specific property, e.g. `mutuallyExclusive`

docs/rules/custom-rules.md
adamaltman: I don't like this approach because it has inconsistency and has some unwritten and uncommon rule.

docs/rules/custom-rules.md
adamaltman: I don't like it personally. Yes, it is to minimize confusion, but I would prefer without it.

adamaltman:
> if there is a property defined in the subject assertions run on the property value
> if there is only a type defined and no property, assertions run on the Subject's properties.

This is explained with examples too:

> Property name corresponding to the [OpenAPI node type](#openapi-node-types). If a list of properties is provided, assertions evaluate against each property in the sequence. If not provided (or null), assertions evaluate against the key names for the subject node type. See [property example](#property-example).

> some assertions can't be applied to a specific property, e.g. mutuallyExclusive

It is already there (each assertion that works with key names only mentions it):
> Asserts that listed properties (key names only) are mutually exclusive. See [mutuallyExclusive example](#mutuallyexclusive-example).

I also updated the docs based on the other feedback. Thank you!

docs/rules/custom-rules.md
adamaltman: I removed it in my most recent push.

docs/rules/custom-rules.md
RomanHotsiy: Okay, agreed.

RomanHotsiy:
> This is explained with examples too:

Great. I missed it. This looks good now 👍

docs/rules/custom-rules.md
```diff
@@ -131,61 +162,97 @@ Notice we don't need to include `property` in this approach.
...
-  - type: Response
-    matchParentKeys: ['201', '200']
-  subject: MediaTypesMap
-  disallowed: ['application/pdf']
+  where:
+  - subject:
+      type: Operation
+      filterByParentKeys: 
+        - put
+    assertions:
```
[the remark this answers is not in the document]
adamaltman: I would not want to have a `default: true` as a default because it is not in line with default values throughout our configuration file.

docs/rules/custom-rules.md
```diff
@@ -131,61 +162,97 @@ Notice we don't need to include `property` in this approach.
...
+    type: Schema
+    property: type
+  assertions:
+    const: integer
+  where:
+    - subject:
+        type: Parameter
+        property: name
+      assertions:
+        const: limit
```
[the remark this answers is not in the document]
adamaltman: Yes, that would be ideal:

```
    - subject:
        type: Parameter
        property: name
      assertions:
        const: limit
    - subject:
        type: Parameter
        property: in
      assertions:
        const: query
```

docs/rules/custom-rules.md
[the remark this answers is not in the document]
tatomyr: I see. Anyway, this particular example doesn't work at the moment since the `defined` assertion requires a single `property` to run against. Otherwise it falls into this error: https://github.com/Redocly/redocly-cli/pull/894/files#diff-8ca1f03a8561336a37c8a4f40e18491007a433f3144655ea5cde722aaf32e8f8R75-R79

docs/rules/custom-rules.md
```diff
@@ -131,61 +162,97 @@ Notice we don't need to include `property` in this approach.
...
-  context:
-  - type: Operation
-    matchParentKeys: [put]
-  - type: Response
-    matchParentKeys: ['201', '200']
-  subject: MediaTypesMap
-  disallowed: ['application/pdf']
+  where:
+  - subject:
+      type: Operation
```
adamaltman, line 213:
```suggestion
      type: Operation
      property: responses
```

docs/rules/custom-rules.md
```diff
@@ -131,61 +162,97 @@ Notice we don't need to include `property` in this approach.
...
+      filterByParentKeys: 
+        - put
+    assertions:
+      defined: true
+  - subject:
+      type: Response
```
adamaltman, line 219:
```suggestion
      type: Response
      property: description
```

docs/rules/custom-rules.md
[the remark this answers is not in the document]
adamaltman: I adjusted the samples. Also, I couldn't understand the logic of what runs on a key?

https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/common/assertions/asserts.ts#L17

For example, what does the `enum` assertion meant when it is run on a key? And what is as an example use case?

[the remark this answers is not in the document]
tatomyr: I see. It would be very hard to find out this fix for a newbie 😄 

When an assertion is supposed to run on keys set it means it could be applied to key names in a Node. Another case is when it could be applied to a particular value (then it requires a `property` to be defined). 

Example:

```yaml

subject: 
  type: PathsMap
assertions:
  enum: 
  - /pets
  - /store
```

which means you can only use the abovementioned keys.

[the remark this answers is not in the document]
adamaltman: FYI I renamed `PathsMap` to `Paths`. #906

docs/rules/custom-rules.md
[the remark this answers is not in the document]
tatomyr: I don't think it would be a quick fix. I'd left this as an improvement.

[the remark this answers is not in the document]
redzaj: I already have this problem and end up with need for creating custom plugin.

My case was to create assertion rule for specific names of query parameters. I wanted to pass defined list of possible names like: `q, offset, limit, sort` so they be consistent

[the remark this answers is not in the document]
RomanHotsiy: >What if we also want to apply this only to a Parameter which is located in: query for instance? 

Why. do not use an `enum` instead of `const`?

```
where:
  - subject:
      type: Parameter
      property: name
    assertions:
      enum:
        - limit
        - query
```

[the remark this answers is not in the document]
adamaltman: I think @RomanHotsiy example is incomplete. Try this with the released beta.112:

```yaml
assert/query-parameter-names:
  where:
     - subject:
          type: Parameter
          property: in
        assertions:
          const: query
  subject:
     type: Parameter
     property: name
  assertions:
     enum:
        - q
        - offset
        - limit
        - sort
```

adamaltman:
These docs should be released now because we released 112. 🎉 

I'll make one further revision to the docs within the next week.

RomanHotsiy (approved):

tatomyr (approved):

PR #922, opened 2022-11-02
fix: undefined process.cwd in browser environment

smeng9:
## What/Why/How?

Some of other libraries uses process.env in browser, making process defined, but left process.cwd undefined.
Thus breaking the code.

We should check for process.cwd before using it.

## Reference

Closes https://github.com/Redocly/redoc/issues/2055#issuecomment-1297661865

## Check yourself

- [x] Code is linted

## Security


packages/core/src/config/config.ts
```diff
@@ -50,7 +50,9 @@ function getIgnoreFilePath(configFile?: string): string | undefined {
       ? path.join(path.dirname(configFile), IGNORE_FILE)
       : path.join(configFile, IGNORE_FILE);
   } else {
-    return typeof process === 'undefined' ? undefined : path.join(process.cwd(), IGNORE_FILE);
+    return (typeof process === 'undefined' || typeof process.cwd === 'undefined')
```
tatomyr, line 53: Could you check if `isBrowser`  from `packages/core/src/env.ts` works for your case instead? I believe it might be much convenient.

smeng9 (commented):
Update check browser to use isBrowser

tatomyr (approved):

PR #923, opened 2022-11-04
refactor: do not add the recommended ruleset fallback when there is a config

tatomyr:
## What/Why/How?

Currently we provide the default fallback with the recommended ruleset if there's no extends explicitly set up in a Redocly config file or there's no config file at all.

We put a deprecation message some time ago, and now it's the highest time to finally change the approach: to extend with the default ruleset ONLY when there's no config provided.

~Also it would be great to remove the deprecated suggestions (like apiDefinitions &c.). -- DONE~

- [x] change default fallback behaviour
- [x] check configs for braking changes

## Check yourself

- [x] Code is linted

## Security


SmoliyY (approved):

adamaltman:
# Migration guide

The migration guide applies to `1.0.0-beta.113` and newer.

If you have an explicit extends declaration in your Redocly configuration file, you can skip this guide.
```yaml
extends:
  - recommended
```

## No configuration file

If you do not have a Redocly configuration file, then the prior behavior for running the `lint` command is as follows:

- From the CLI: defaults to `recommended` configuration.
- From Redocly Workflows: defaults to `minimal` configuration.

### Actions to take

Create a [Redocly configuration file](https://redocly.com/docs/cli/configuration/) named `redocly.yaml` and place it in the root of your repo.

To use the `recommended` configuration, add this to your [Redocly configuration file](https://redocly.com/docs/cli/configuration/)
```yaml
extends:
  - recommended
```

To use the `minimal` configuration, add this to your [Redocly configuration file](https://redocly.com/docs/cli/configuration/):

```yaml
extends:
  - minimal
```

## Configuration file missing `extends`

If you have a Redocly configuration file, it is missing an explicit `extends` declaration that selects a default configuration.
The behavior prior to `1.0.0-beta.113` is as follows:

- From the CLI:  TODO
- From Redocly Workflows:  TODO

### Actions to take

To use the `recommended` configuration, add this to your [Redocly configuration file](https://redocly.com/docs/cli/configuration/)
```yaml
extends:
  - recommended
```

To use the `minimal` configuration, add this to your [Redocly configuration file](https://redocly.com/docs/cli/configuration/):

```yaml
extends:
  - minimal
```

packages/core/src/config/config-resolvers.ts
```diff
@@ -41,25 +41,15 @@ export async function resolveConfig(rawConfig: RawConfig, configPath?: string):
...
-    extends: configExtends,
-    recommendedFallback,
-  };
 
   const apis = await resolveApis({
-    rawConfig: {
-      ...rawConfig,
-      styleguide: styleguideConfig,
-    },
+    rawConfig,
```
AlexVarchuk, line 46: Please, double check working resolving config with the new config format(flat config)

PR #931, opened 2022-11-11
chore: ignore discriminator error

AntonKozachuk:
## What/Why/How?
We would like to temporarily ignore an error regarding the discriminator usage.

## Reference
N/a

## Testing
Locally.

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security


RomanHotsiy (changes requested):
Any tests?

packages/core/src/rules/utils.ts
```diff
@@ -118,11 +118,14 @@ export function validateExample(
       }
     }
   } catch (e) {
-    report({
-      message: `Example validation errored: ${e.message}.`,
-      location: location.child('schema'),
-      from: location,
-    });
+    const discriminatorError = 'discriminator: requires oneOf or anyOf composite keyword';
+    if (e.message !== discriminatorError) {
```
tatomyr, line 122: What about `if (e.message === discriminatorError) { return }` ?

tatomyr (approved):

packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
```diff
@@ -438,4 +438,36 @@ describe('no-invalid-media-type-examples', () => {
       ]
     `);
   });
+
+  it('should not report if allOf used with discriminator', async () => {
```
RomanHotsiy, line 442: This test needs to be improved.

RomanHotsiy (approved):
The test need to be improved but let's merge and release it and then fix the test.

PR #960, opened 2022-12-08
feat: remove unused components in config

### LINKED ISSUE #945: Remove unused components in config

There is no way to designate that an API should be bundled with the remove-unused-components flag. This means the config alone is unreliable and can result in bundling with unused components even if the designer's intention is to force it to always bundle without unused components.

I would like a way to designate that remove-unused-components should be used.

Something like this. I understand it needs to run last.

```yaml
apis:
  core@latest:
    root: openapi/openapi.yaml
    decorators:
      filter-in:
        property: x-products
        value:
          - Core
      remove-unused-components: on
```

```yaml
apis:
  core@latest:
    root: openapi/openapi.yaml
    remove-unused-components: true
    decorators:
      filter-in:
        property: x-products
        value:
          - Core
```

I'm good with either.

**Additional context**

Happy to answer any questions.

AntonKozachuk:
## What/Why/How?
Allow enabling to remove unused components in config to use within the `bundle` command.

## Reference
Closes #945.

## Testing
Locally.

## Screenshots (optional)
N/a

## Check yourself

- [ ] Code is linted

## Security


tatomyr (approved):
Please also add a test without the `remove-unused-component` instructions so we can see the redundant part remains.

PR #961, opened 2022-12-08
hotfix: revert "feat: add vendor extensions to types"

RomanHotsiy:
Reverts Redocly/redocly-cli#958

This is a breaking change for some folks because of x-meta validation. We need to rethink if we want to validate `x-meta`.


RomanHotsiy:
I think I'll just remove x-meta validate instead.

PR #964, opened 2022-12-11
feat: implemeent a special VendorExtension type

RomanHotsiy:
## What/Why/How?

There is no good way to validate vendor extension with custom rules.
This PR adds a special type: `VendorExtension` which can be used for this purpose.

Kudos to @adamaltman for the idea.

Example:

```
  assert/vendor-extension:
    subject:
      type: VendorExtension
      property: prop
      filterInParentKeys: ['x-my-extension']
    severity: error
    assertions:
      enum: [foo, bar]
```

## Testing

Added e2e tests.

## Check yourself

- [x] Code is linted

## Security


packages/core/src/types/index.ts
```diff
@@ -63,6 +63,11 @@ export function mapOf(typeName: string) {
   };
 }
 
+export const VendorExtension: NormalizedNodeType = {
+  name: 'VendorExtension',
```
adamaltman, line 67: Do you think it should be `SpecificationExtension` instead of `VendorExtension`? The name changed in 3.0 and should therefore cover most of the usage?

RomanHotsiy: oh, I somehow missed this fact.

Yes, I think `SpecificationExtension` is better. What do you think about `SpecExtension` or just `Extension` though?

__tests__/commands.test.ts
```diff
@@ -13,7 +13,7 @@ expect.extend({
 callSerializer();
 
 describe('E2E', () => {
-  describe('lint', () => {
+  describe.only('lint', () => {
```
AntonKozachuk, line 16: This should be deleted probably.

__tests__/lint/assertions-vendor-extensions/.redocly.yaml
```diff
@@ -0,0 +1,36 @@
...
+    where:
+      - subject:
+          type: Info
+        assertions:
+          defined: true
+    severity: error
+    assertions:
+      enum: [foo, bar]
+
+  # built-in vendor extension should be processed as generic VendorExtension too
```
AntonKozachuk, line 28: Do we want to keep this?

__tests__/commands.test.ts
RomanHotsiy: Thanks 👍

__tests__/lint/assertions-vendor-extensions/.redocly.yaml
RomanHotsiy: Yes, I still want to get it working hopefully.

packages/core/src/rules/common/spec.ts
```diff
@@ -24,10 +24,12 @@ export const OasSpec: Oas3Rule | Oas2Rule = () => {
         }
         return;
       } else if (nodeType !== 'object') {
-        report({
-          message: `Expected type \`${type.name}\` (object) but got \`${nodeType}\``,
-          from: refLocation,
-        });
+        if (type !== SpecExtension) { // do not validate unknown extensions structure
```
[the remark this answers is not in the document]
RomanHotsiy: Both are needed. This one is for type check.

packages/core/src/types/oas2.ts
```diff
@@ -24,6 +24,7 @@ const Root: NodeType = {
     'x-ignoredHeaderParameters': { type: 'array', items: { type: 'string' } },
   },
   required: ['swagger', 'paths', 'info'],
+  extensionsPrefix: 'x-',
```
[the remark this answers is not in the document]
RomanHotsiy: I was thinking about it but decided to proceed like this for now.

__tests__/lint/assertions-vendor-extensions/.redocly.yaml
RomanHotsiy: Didn't get it working. I'll merge with this commented out for now. Hopefully we can fix it later.

AntonKozachuk (approved):

PR #985, opened 2023-01-05
fix: resolve nested refs inside spec extensions

AntonKozachuk:
## What/Why/How?
Resolve an issue with nested refs not resolved and throwing an error from the no-unresolved-refs rule.

## Reference
N/a

## Testing
Locally.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/rules/__tests__/no-unresolved-refs.test.ts
+++ b/packages/core/src/rules/__tests__/no-unresolved-refs.test.ts
@@ -196,4 +196,33 @@ describe('oas3 boolean-parameter-prefixes', () => {
 
     expect(replaceSourceWithRef(results, __dirname)).toMatchInlineSnapshot(`Array []`);
   });
+
+  it('should not report on nested refs inside specification extensions', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.0.0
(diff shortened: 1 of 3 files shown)


RomanHotsiy (approved):

packages/core/src/rules/__tests__/no-unresolved-refs.test.ts
```diff
@@ -196,4 +196,33 @@ describe('oas3 boolean-parameter-prefixes', () => {
 
     expect(replaceSourceWithRef(results, __dirname)).toMatchInlineSnapshot(`Array []`);
   });
+
+  it('should not report on nested refs inside specification extensions', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.0.0
```
tatomyr, line 203: Could you write also a test for 3.1? Just to be sure :)

tatomyr (approved):

packages/core/src/rules/__tests__/no-unresolved-refs.test.ts
AntonKozachuk: Yeah, of course.

PR #990, opened 2023-01-11
feat: add an option to include webhooks to operation-4xx-response rule

### LINKED ISSUE #955: Add an option to exclude webhooks from operation-4xx-response rule
In terms of our [operation-4xx-response/](https://redocly.com/docs/cli/rules/operation-4xx-response/) rule we treat webhooks the same as path items (as it's intended by the spec). 
However, webhooks ofter times are treated differently than path items so it makes sense to provide an option to exclude them from the rule. 

Add an option to exclude webhooks from the operation-4xx-response rule and similar ones.

Avoid applying the rule to the webhooks at all.

**Additional context**
https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1669127864676219

AntonKozachuk:
## What/Why/How?
In terms of our [operation-4xx-response/](https://redocly.com/docs/cli/rules/operation-4xx-response/) rule we treat webhooks the same as path items (as it's intended by the spec).
However, webhooks ofter times are treated differently than path items so it makes sense to provide an option to exclude them from the rule.

## Reference
Closes #955 

## Testing
Locally.

## Screenshots (optional)
N/a

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/docs/rules/operation-2xx-response.md
+++ b/docs/rules/operation-2xx-response.md
@@ -23,12 +23,22 @@ You can greatly improve the developer and user experience of your APIs by making
 |Option|Type|Description|
 |---|---|---|
 |severity|string|Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
+|validateWebhooks|[boolean]|Determines if responses inside webhooks are validated. Default `false`. |
 
 An example configuration:
 
 ```yaml
 rules:
   operation-2xx-response: error
+```
+
+The following example enables validation of responses inside webhooks:
+
+```yaml
+rules:
+  operation-2xx-response: 
+    severity: error
+    validateWebhooks: true
 ```
 
 ## Examples
--- a/docs/rules/operation-2xx-response.md
+++ b/docs/rules/operation-2xx-response.md
@@ -23,7 +23,7 @@ You can greatly improve the developer and user experience of your APIs by making
 |Option|Type|Description|
 |---|---|---|
 |severity|string|Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
-|validateWebhooks|[boolean]|Determines if responses inside webhooks are validated. Default `false`. |
+|validateWebhooks|boolean|Determines if responses inside webhooks are validated. Default `false`. |
 
 An example configuration:
 
--- a/packages/core/src/rules/utils.ts
+++ b/packages/core/src/rules/utils.ts
@@ -179,7 +179,7 @@ export function validateResponseCodes(
 
     return responseCodeRegexp.test(code);
   });
-  
+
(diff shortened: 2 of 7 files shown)


AntonKozachuk:
@adamaltman the new option has a `validateWebhooks` name. Please let me know if you have any opinions on this.

packages/core/src/rules/utils.ts
```diff
@@ -164,3 +164,20 @@ export function validateSchemaEnumType(
...
+) {
+  if (
+    !responseCodes.some((code) =>
+      codeType === '4XX' ? /4[Xx0-9]{2}/.test(code) : code === 'default' || /2[Xx0-9]{2}/.test(code)
+    )
+  ) {
+    report({
+      message: `Operation must have at least one \`${codeType}\` response.`,
+      location: { reportOnKey: true },
+    });
```
tatomyr, line 181: Would be easier to read if written in the following manner:

```ts

codeRange => {
  const containsNeededCode = responsesCode.some(code => code === codeRange || new RegExp(`^${codeRange[0]}[0-9][0-9]$`).test(code))
  if (!containsNeededCode) ...
}
```

RomanHotsiy (approved):
LGTM 👍

docs/rules/operation-2xx-response.md
```diff
@@ -23,6 +23,7 @@ You can greatly improve the developer and user experience of your APIs by making
 |Option|Type|Description|
 |---|---|---|
 |severity|string|Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
+|validateWebhooks|[boolean]|Determines if responses inside webhooks are validated. Default `false`. |
```
adamaltman, line 26:
```suggestion
|validateWebhooks|boolean|Determines if responses inside webhooks are validated. Default `false`. |
```
The brackets mean it is an array of boolean values.

packages/core/src/rules/utils.ts
```diff
@@ -164,3 +164,26 @@ export function validateSchemaEnumType(
...
+) {
+  const responseCodeRegexp = new RegExp(`^${codeRange[0]}[0-9Xx]{2}$`);
+
+  const containsNeededCode = responseCodes.some((code) => {
+    if (code === 'default' && codeRange === '2XX') {
+      return true;
+    }
+
+    return responseCodeRegexp.test(code);
+  });
```
tatomyr, line 181: This might be even better:

```suggestion
  const containsNeededCode = responseCodes.some((code) => 
    (codeRange === '2XX' && code === 'default') // It's OK to replace 2xx codes with the default
    || responseCodeRegexp.test(code));
```

tatomyr:
👍

PR #999, opened 2023-01-25
feat: new decorator info-override

RomanHotsiy:
## What/Why/How?

This decorator will merge/override additional properties into the info object.

With this decorator you cannot remove some properties from info. Do we need another decorator or some option in this to switch from "merge" to "complete replace" behaviour?

Thoughs?

## Testing

Added tests.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/__tests__/bundle/info-override/.redocly.yaml
+++ b/__tests__/bundle/info-override/.redocly.yaml
@@ -1,0 +1,8 @@
+apis:
+  main:
+    root: ./main.yaml
+
+decorators:
+  info-override:
+    title: Updated title
+    x-vendor: custom extension
--- a/__tests__/bundle/info-override/main.yaml
+++ b/__tests__/bundle/info-override/main.yaml
@@ -1,0 +1,24 @@
+openapi: 3.0.0
+servers:
+  - url: //petstore.swagger.io/v2
+    description: Default server
+info:
+  title: Example OpenAPI 3 definition.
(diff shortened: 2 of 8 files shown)


IgorKarpiuk (approved):

__tests__/bundle/info-override/.redocly.yaml
```diff
@@ -0,0 +1,8 @@
+apis:
+  main:
+    root: ./main.yaml
+
+decorators:
+  info-override:
+    title: Updated title
+    x-vendor: custom extension
```
IgorKarpiuk, line 8: Should  there be a descriptions like in snapshot ?

 ```yaml
 description: |
    This is a test API description.

    # Heading test

    Body [test](#) content.
```

RomanHotsiy: You are right

adamaltman: From the docs it looks like it should not need it. It overrides properties and leaves the ones that were not in the override alone, right?

adamaltman: Nevermind, the original and new one have different descriptions. 🤣

RomanHotsiy: Should be resolved now.

PR #1010, opened 2023-02-02
Update rule to include missing fields from RFC7807 specification

jeremyfiel:
- update rule docs

    - update tests

## Reference
updates #816 
cc @iosypov 

## Check yourself

- [x] Code is linted

## Security


RomanHotsiy:
Hey @jeremyfiel. Thanks for the PR.

I just checked the implementation of this rule and I find it a bit confusing (it was like that before). We'll need to rename some variables.

I wonder if every API that implements RFC7807 wants to define those non-required properties. I feel like `type` and `title` are the core ones and 99% of APIs implement them but not sure if it holds for other fields.

If we add these validations to the rule I would make them optional (using some flag) and disabled by default. What do you think?

jeremyfiel:
Per the spec, all 5 properties are optional so implementations can be all over the place. That makes it difficult to validate the object schema. I suppose maintaining `title` and `type` as required gives some guidance to use, at least, those two properties. 

in that case, the function `validateDefined...` usage here is probably not appropriate for all five properties and to your point, maybe adding them is unnecessary.  If anything, I recommend updating the docs to outline the properties of the RFC and indicate you only check for those two

tatomyr:
BTW, have you considered using assertions for that?
I believe a custom rule like the following could also be helpful (this is not complete though): 
```yaml
rules:
  assert/rfc:
    where: 
      - subject: 
          type: Response
          matchParentKeys: /4[Xx0-9]{2}/
        assertions:  
          defined: true
    subject: 
      type: SchemaProperties
    assertions: 
      defined: true
      enum: 
        - type
        - title
        - status
        - detail
        - instance
```

Another option would be to rewrite the rule to accept options with the required fields. E. g.:
```yaml
rules:
  rfc...:
    required:
    - type
    - status
```

jeremyfiel:
I didn't try any other method, this was mostly about updating the existing rule with the proper interpretation.

tatomyr:
I lean towards having the `requiredProperties` option to be able to define the needed fields:

```yaml
rules:
    operation-4xx-problem-details-rfc7807: 
       severity: error
       requiredProperties: [type, title]
```

Regarding the default behavior we have 3 options:
- require no fields by default (and maybe writing a warning)
- require all fields
- leave the current behaviour

Thoughts?

lornajane:
Having required and optional fields seems like the right option - it's common to see `type` and `title` and then some but not all of the others - and additional fields should always be permitted. Looking at @tatomyr 's example, users could, for example, configure all fields and set to a warning severity. Or they could use the settings there which match our current default. Can we proceed along these lines?

tatomyr:
Meanwhile, a new problem details standard has been released: https://datatracker.ietf.org/doc/html/rfc9457. It is supposed to replace RFC7807.

adamaltman:
I wish they waited to publish that RFC so they could get a number I could remember like RFC 9876. 🤣

PR #1029, opened 2023-03-22
chore: add rule for min length property and reworked assertionsFn 

### LINKED ISSUE #1011: Add schema rule for required string values to have a minLength specified
I would be beneficial to have a rule that checks that a minLength attribute is applied to all required string properties. Currently if a minLength is not supplied then an empty string is a valid value, which is not what is not the generally understood desired behaviour for required fields (i.e. required should mean "is present and has a value" not just "is present").

SmoliyY:
## What/Why/How?

- Add `required-min-length-string-type-property` for check if required string property contains `minLength` property.
- Reworked assertionsFn args. Now as third parameter it is receive `сtx` which contains -  original `ctx` and extended with properties, `rawValue` ,  and `baseLocation`

## Reference
Closes #1011

Corresponding PR for docs - https://github.com/Redocly/redocly-cli/pull/1030
## Testing
`required-min-length-string-type-property` rule

schema 

```yaml
schema:
    type: object
    required: 
      - b
      - a
      - k
    properties:
      a:
        type: string
        
      b:
        type: object
      k: 
        type: string
        minLength: 2
      c: 
        type: object
        required: 
          -  d
        properties: 
          d:
            type: string
            
```
results:

Screenshot 2023-03-23 at 13 46 38

Possibility to have context inside customFn . 

For example: 

```yaml

module.exports = {
    id: 'local',
    assertions: {
      checkWordsCount: (value, options, ctx) => {
        const words = value.split('');
   
        if(ctx.key === 'get' && !(words.length >= options.min)) {
          ....
       
      },
    },
  };
```

## Check yourself

- [x] Code is linted

## Security


adamaltman (commented):
It's missing docs.

adamaltman:
@SmoliyY I can't tell what happens if you have a required property but it is a type different from `string` such as `object` or `array`?

SmoliyY:
> I can't tell what happens if you have a required property but it is a type different from `string` such as `object` or `array`?

Nothing. Rule only works if property that required has type of `string`

> It's missing docs.

Regarding docs I am on it! I am going to add it in separate PR to avoid latency between merge(when docs will be published) and actual release.

RomanHotsiy (changes requested):

RomanHotsiy:
I also think we need to rename the rule: `required-min-length-string-type-property` doesn't match our style.

How about `string-property-min-length-required` or `string-property-min-length`?

RomanHotsiy:
cc @adamaltman ☝️

__tests__/lint/assertions-custom-function-options/plugin.js
```diff
@@ -1,19 +1,29 @@
 module.exports = {
   id: 'local',
   assertions: {
-    checkWordsStarts: (value, opts, location) => {
+    checkWordsStarts: (value, opts, assertionProperties) => {
```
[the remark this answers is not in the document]
SmoliyY: I think we can stick with name `assertionContext` instead. We already have our UserContext as `ctx` inside `assertionProperties`

SmoliyY:
> I also think we need to rename the rule: `required-min-length-string-type-property` doesn't match our style.
> 
> How about `string-property-min-length-required` or `string-property-min-length`?

Yes I have luck to achieve with such plugin - 
```js
module.exports = {
  id: 'local',
  assertions: {
    checkSchema: (_, opts, assertionContext) => {
      const name = opts.required;

      const rawValue = assertionContext.rawValue;
      const errors = [];

      if (rawValue?.required && rawValue?.properties) {
        const required = rawValue?.required;
        const properties = rawValue?.properties;

        for (const item of required) {
          if (properties[item] && properties[item].type === 'string') {
            if (!properties[item][name]) {
              errors.push({
                message: `Required property ${name} inside ${item} string property`,
                location: assertionContext.baseLocation,
              });
            }
          }
        }
      }
      return errors;
    },
  },
};

```

SmoliyY:
@adamaltman could you please take a look if `required-min-length-string-type-property` is fine for us or we should stick with other name for this rule?

packages/core/src/utils.ts
```diff
@@ -237,6 +237,8 @@ export function showErrorForDeprecatedField(
   );
 }
 
+export type OptionalIntersectionTypes<T, T2> = Partial<Omit<T, keyof T2>> & T2;
```
tatomyr, line 240: Could you leave a comment or simplify the type?

SmoliyY: It's a little bit complicated and hard to read, I'll remove this generic.

SmoliyY:
> Yes I have luck to achieve with such plugin 
With full context I've manage to create a better rule than previous. I've included it into tests.

tatomyr (approved):

RomanHotsiy (changes requested):

RomanHotsiy (approved):

PR #1030, opened 2023-03-23
docs: add docs for reworked assertionFn and new rule

SmoliyY:
## What/Why/How?
Updated docs for Assertions : 
- Add new property `assertionsProperties` that can be accessible in the `customFn`  instead of the Location. Now location is posted in the assertionsProperties as baseLocation propertie.

New rule: 'required-min-length-string-type-property'.
-  Requires that every property in the API definition with type `string` that are required also to have `minLength`

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/rules/custom-rules.md
+++ b/docs/rules/custom-rules.md
@@ -263,7 +263,7 @@ Property | Type | Description
 -- | -- | --
 value | `string` \| [`string`] | Value that appears at the corresponding location.
 options | `object` | Options that is described in the configuration file.
-location | `Location Object` | Location in the source document. See [Location Object](../resources/custom-plugins.md#location-object)
+assertionProperties | `object` | Assertion Properties contains 3 properties: `baseLocation`, `rawValue` and `ctx`. Base location (`baseLocation`) is corresponding to the Location in the source document, see [Location Object](../resources/custom-plugins.md#location-object). Raw value contains the original value that appears at at the corresponding location. Context (`ctx`) is corresponding to the [Context object](../resources/custom-plugins.md#the-context-object) that contains additional data and functionality.
 **Return**
 problems | [`Problem`] | List of problems. An empty list means all checks are valid.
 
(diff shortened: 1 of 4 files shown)


docs/rules/required-min-length-string-type-property.md
```diff
@@ -0,0 +1,90 @@
+# required-min-length-string-type-property
```
adamaltman, line 1: I wonder if this rule should be renamed. 

`required-string-property-missing-min-length`

docs/rules/required-min-length-string-type-property.md
```diff
@@ -0,0 +1,90 @@
...
+
+|OAS|Compatibility|
+|---|---|
+|2.0|✅|
+|3.0|✅|
+|3.1|✅|
+
+## API design principles
+
+The `minLength` property is useful in scenarios where you need to specify a minimum length requirement for a string or to check if the string is not empty, especially if the property is marked as `required`. It helps to prevent errors or unexpected behavior in your application that may result from having strings that are too short or even empty. By setting a minimum length, you can ensure that the string has enough characters to be useful in your application. Overall, using the `minLength` property for string type properties in OAS definitions is a best practice for ensuring data integrity and improving the overall quality of the API.
```
adamaltman, line 13:
```suggestion
The `minLength` keyword constrains string values. When a property of type `string` is `required`, defining the `minLength` helps prevent common mistakes, such as empty strings. Use the `minLength` property as a best practice to ensure data integrity and enhances the overall quality of the API.
```

docs/rules/custom-rules.md
```diff
@@ -263,7 +263,7 @@ Property | Type | Description
 -- | -- | --
 value | `string` \| [`string`] | Value that appears at the corresponding location.
 options | `object` | Options that is described in the configuration file.
-location | `Location Object` | Location in the source document. See [Location Object](../resources/custom-plugins.md#location-object)
+ctx | `object` | Context object corresponds to the [Context object](../resources/custom-plugins.md#the-context-object) and extended by 2 properties: `baseLocation`, and `rawValue`. Base location (`baseLocation`) corresponds to the location in the source document for current assertion. (See [Location Object](../resources/custom-plugins.md#location-object)). Raw value is the original value that appears at the corresponding assertion location.
```
adamaltman, line 266: We actually need a better example though (I think I gave a scenario to show).

adamaltman, line 266:
```suggestion
ctx | `object` | `ctx` object extends the [Context object](../resources/custom-plugins.md#the-context-object) with two properties: `baseLocation`, and `rawValue`. Base location (`baseLocation`) contains the location in the source document for current assertion. (See [Location Object](../resources/custom-plugins.md#location-object)). Raw value is the original assertion value.
```

adamaltman (changes requested):
Some thoughts about naming and docs suggestions.

docs/rules/required-min-length-string-type-property.md
SmoliyY: I've renamed rule in this [PR](https://github.com/Redocly/redocly-cli/pull/1039)

adamaltman (approved):

tatomyr (approved):

Issue #1018, opened 2023-03-04
Issue with joining path parameters


mibejjh:
**Describe the bug**

When joining multiple files, if there are multiple reference objects in the common path parameters, only the first one is kept and the other reference objects are removed.

For example, if you join files with the following paths object,

paths:
  /{a}/{b}:
    parameters:
      - $ref: "#/components/parameters/a"
      - $ref: "#/components/parameters/b"
you will get

paths:
  /{a}/{b}:
    parameters:
      - $ref: "#/components/parameters/a"

**To Reproduce**
3. Given this OpenAPI file(s)

- a.yaml
```
openapi: 3.0.0
info:
  title: Example API
  description: This is an example API.
  version: 1.0.0
servers:
  - url: https://example.com/api
paths:
  /users/{userId}/orders/{orderId}:
    parameters:
      - $ref: "#/components/parameters/userIdParam"
      - $ref: "#/components/parameters/orderIdParam"
    get:
      summary: Get an order by ID for a specific user
      responses:
        "200":
          description: OK
        "404":
          description: Not found
components:
  parameters:
    userIdParam:
      name: userId
      in: path
      description: ID of the user
      required: true
      schema:
        type: integer
    orderIdParam:
      name: orderId
      in: path
      description: ID of the order
      required: true
      schema:
        type: integer
```

- b.yaml
```
openapi: 3.0.0
info:
  title: Example API
  description: This is an example API.
  version: 1.0.0
servers:
  - url: https://example.com/api
paths:
  /users/{userId}/products/{productId}:
    parameters:
      - $ref: "#/components/parameters/userIdParam"
      - $ref: "#/components/parameters/productIdParam"
    get:
      summary: Get a product by ID for a specific user
      responses:
        "200":
          description: OK
        "404":
          description: Not found
  /users/{userId}/orders/{orderId}/items/{itemId}:
    parameters:
      - $ref: "#/components/parameters/userIdParam"
      - $ref: "#/components/parameters/orderIdParam"
      - $ref: "#/components/parameters/itemIdParam"
    get:
      summary: Get an item by ID for a specific order and user
      responses:
        "200":
          description: OK
        "404":
          description: Not found
components:
  parameters:
    userIdParam:
      name: userId
      in: path
      description: ID of the user
      required: true
      schema:
        type: integer
    productIdParam:
      name: productId
      in: path
      description: ID of the product
      required: true
      schema:
        type: integer
    orderIdParam:
      name: orderId
      in: path
[...]

PR #1034, opened 2023-03-27
fix: handle $ref in the joining path parameters

IgorKarpiuk:
## What/Why/How?

Handle $ref in the joining path parameters

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/1018

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -426,7 +426,6 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
 
         for (const pathParameter of joinedDef.paths[path]
           .parameters as Referenced<Oas3Parameter>[]) {
-
           // Compare $ref only if both are reference objects
           if (isRef(pathParameter) && isRef(parameter)) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/join.ts
```diff
@@ -418,14 +419,22 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
         joinedDef.paths[path].parameters = [];
       }
 
-      for (const parameter of pathItem.parameters as Oas3Parameter[]) {
+      for (const parameter of pathItem.parameters as (Oas3Parameter & { $ref: string })[]) {
```
tatomyr, line 422: Shouldn't it be `Oas3Parameter | { $ref: string }` instead of the `&`?
Can you make it more general? I mean, the same logic could be applied to other parts of a definition, not only to path parameters.

IgorKarpiuk: It looks that we don't need this logic anywhere else

tatomyr (approved):

Issue #992, opened 2023-01-18
@redocly/cli@1.0.0-beta.118: OASDecorator: context.resolve() cannot resolve external oas attributes if refs exist in array


afmhenry:
**The bug**

Introduction of SpecExtension has caused redocly plugin decorator context.resolve() to fail, when applied to nodes belonging to an external oas attributes (x-) containing an array. 

Can be reproduced since upgrading to @redocly/cli@1.0.0-beta.118
Can likewise be "resolved" by downgrading to @redocly/cli@1.0.0-beta.117

**To Reproduce**

1. unzip attached folder
2. `npm i @redocly/cli@1.0.0-beta.118`
3. `npx redocly bundle --config config.yaml oas-spec.yaml --output oas-spec-bundled.yaml`
4. Observed console output: the "x-bug" is not resolved, but "no-bug" is
```
x-bug node undefined
no-bug node { test: 'this', test1: 'succeeds on all versions' }
```
5. `npm i @redocly/cli@1.0.0-beta.117`
6. `npx redocly bundle --config config.yaml oas-spec.yaml --output oas-spec-bundled.yaml`
7. Both references are resolved in the console logs
```
x-bug node { test: 'this', test1: 'fails on versions .118+' }
no-bug node { test: 'this', test1: 'succeeds on all versions' }
```

**Expected behavior**
Can resolve field if it starts with "x-", even contained in an array entry. 

**Observed**
Cannot resolve ref in array, if it belongs to a key that starts with "x-"
Can resolve $ref in array if it doesn't belong to a key that starts with "x-"

**Full Code Scenario**
[redocly.zip](https://github.com/Redocly/redocly-cli/files/10444257/redocly.zip)

**Redocly Version(s)**
@redocly/cli@1.0.0-beta.117
@redocly/cli@1.0.0-beta.118

adamaltman:
Have you tried with with `1.0.0-beta.120`? There were some fixes related to specification extension type in that:

https://redocly.com/docs/cli/changelog/#100-beta120-2023-01-05

> Fixed an issue where `$refs` weren't resolved inside specification extensions.

afmhenry:
I have, 1.0.0-beta.118  and above all fail with the above given scenario

adamaltman:
Yes, this was fixed in `1.0.0-beta.120`.

afmhenry:
[...]

PR #1045, opened 2023-04-10
fix: resolve external oas attributes with if refs exist in array

IgorKarpiuk:
## What/Why/How?

Fixed a bug where context.resolve() did not work when applied to nodes belonging to external oas (x-) attributes that contain an array.

## Reference
Closes: https://github.com/Redocly/redocly-cli/issues/992
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/src/resolve.ts
+++ b/packages/core/src/resolve.ts
@@ -261,7 +261,7 @@ export async function resolveDocument(opts: {
       if (Array.isArray(node)) {
         const itemsType = type.items;
         // we continue resolving unknown types, but stop early on known scalars
-        if (type !== unknownType && itemsType === undefined) {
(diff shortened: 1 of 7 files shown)


packages/core/src/resolve.ts
```diff
@@ -261,7 +261,7 @@ export async function resolveDocument(opts: {
       if (Array.isArray(node)) {
         const itemsType = type.items;
         // we continue resolving unknown types, but stop early on known scalars
-        if (type !== unknownType && itemsType === undefined) {
+        if (type !== unknownType && type !== SpecExtension && itemsType === undefined) {
```
tatomyr, line 264: What about this?
```suggestion
        if (itemsType === undefined && type !== unknownType && type !== SpecExtension) {
```

tatomyr (approved):

Issue #641, opened 2022-04-13
`join` command regression: decorators/preprocessor are always executed


acote-coveo:
**Describe the bug**

Before [this commit](https://github.com/Redocly/openapi-cli/commit/b53e9e4acbd694ac4b0aa4c8f09fc4c5b6f95901), the join command wasn't bundling documents. By changing this behaviour, running a join now always execute the decorators/preprocessor. 

For us, it's a regression because we use decorators to remove internal documentation. So with the latest versions, we have no way to join ours different api files and keep the internal documentation conditionally.

**To Reproduce**

> openapi join file1.yaml file2.yaml

**Expected behavior**

It was expected that bundling wasn't part of the join command. From my understand, the bundling mechanism was only to resolve reference properly in the join command. Perhaps the join command could ignore the decorators/preprocessor from the config or we could support similar bundle parameters such as `--skip-decorators`

**Logs**
If applicable, add logs to help explain your problem.

**OpenAPI definition**

_.redocly.yaml_:

```yaml
lint:
  extends:
    - minimal
  plugins:
    - './build/openapi-plugins/decorators.js'
  decorators:
    cleanup/remove-hidden: error
    cleanup/remove-internal: error
```

**`openapi-cli` Version(s)**
beta-94

**`Node.js` Version(s)**
N/A -> Running in docker

**Additional context**

adamaltman:
That comment (with the bundling) was done to solve https://github.com/Redocly/openapi-cli/issues/339

There is a workaround you can do. In newer versions of OpenAPI-cli we support a new configuration file format.

```
apis:
  other@internal:
    root: ./other.yaml
  example@internal:
    root: ./openapi.yaml
  example@external:
    root: ./openapi.yaml
    lint:
      decorators:
        cleanup/remove-hidden: error
        cleanup/remove-internal: error
lint:
  plugins:
    - './build/openapi-plugins/decorators.js'
  extends:
     - minimal
```

In this case, you could use the `@internal` API for the join and the `@external` for some external documentation purpose.

Here is a lightweight example of using the same root file with different decorators:

https://github.com/redocly-demo/demo-remove-x-internal

lornajane:
Increasing the priority on this, unexpected processing of the API descriptions is not a good surprise to include with a join command, we should try to keep the commands as atomic as we can.

IgorKarpiuk:
[...]

Issue #1042, opened 2023-04-04
Separate commands for lint, bundle, decorate


lornajane:
**Is your feature request related to a problem? Please describe.**

As a user, I know which operations I want to perform, and in what order.
* I want to lint intentionally, not as a side effect of a bundle command
* I want to decorate either before or after linting
* By having the commands as a separate step, I can easily report (for example in CI) which part failed.

**Describe the solution you'd like**

Additional command `redocly decorate` and no additional operations (linting, decorating) as side-effects of other operations such as bundle.

We might also need a pre-processor command since that's another step that gets done by bundle.

**Describe alternatives you've considered**

Multiple config blocks for the same API with different names so that I can run one step and then another for each stage of the process to work around this limitation and try to restrict commands to a single operation at a time.

RomanHotsiy:
> I want to lint intentionally, not as a side effect of a bundle command

The bundle command doesn't run `lint` as a side effect. There is a special `--lint` flag for this. There is also a separate `lint` command.

> Additional command redocly decorate and no additional operations (linting, decorating) as side-effects of other operations such as bundle.

There is an issue with `redocly decorate` one.

We don't support decorators on unbundled definition. So decorate right now feels like a natural part of bundle. It will most likely need some efforts to support that. I need to check it.

lornajane:
Ah, OK. I think it would be great to have it separate if possible. Otherwise how about adding it as an opt-in, like lint? So you can `bundle --decorate` if you want both operations. Maybe `--pre-process` fits this as well. (See also: #641 )

RomanHotsiy:
It may be a bit non-intuitive when using toghether with config, but I'm not sure.

For more context, users can define config like that.

```
apis:
  api-name@internal:
    root: ./other.yaml
    decorators:
      some-decorator: error
```

Would it be clear that decorators won't run on bundle unless you add `--decorate` option?

lornajane:
I am not sure I'd expect the OpenAPI contents to be altered when running the `bundle` command, I'd just expect to get everything into a single file. So the decorators, which actually change the contents, are even more of a surprise than the linting. I expect bundle to just bundle. If I wanted to decorate, I'd say so. If we can't do it without bundling as well, I think that's OK, but it should still only happen when specifically requested/expected.

aviskase:
[...]

PR #1046, opened 2023-04-11
fix: do not execute decorators and preprocessors in join command

IgorKarpiuk:
## What/Why/How?
Don't execute decorators and preprocessors in join command

Add new options for join command:
- `--decorate` - Run decorators
- `--preprocess` - Run preprocessors

## Reference
Closes: https://github.com/Redocly/redocly-cli/issues/641 
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/__tests__/bundle/resolve-refs-from-x-attributes/.redocly.yaml
+++ b/__tests__/bundle/resolve-refs-from-x-attributes/.redocly.yaml
@@ -1,0 +1,7 @@
+apis:
+  main:
+    root: ./openapi.yaml
+decorators:
+  plugin/resolve-x: on
+plugins:
+  - "./plugin.js"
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/join.ts
```diff
@@ -92,6 +94,24 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
     )
   );
 
+  if (!argv.decorate) {
```
tatomyr, line 97: Taking into account [this](https://github.com/Redocly/redocly-cli/issues/1042#issuecomment-1503120311), how do you think about implementing this logic in the `bundle` command itself, so we can pass the `decorate` option there and avoid code duplication?

IgorKarpiuk: We can move this code in future. When we have concrete vision for that task

tatomyr (approved):

PR #1053, opened 2023-04-18
fix: sort top level oas keys

IgorKarpiuk:
## What/Why/How?

Sort top level OAS3 keys

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -94,17 +94,21 @@ export async function handleJoin(argv: JoinArgv, packageVersion: string) {
     )
   );
 
-
-  console.log('argv', argv);
   if (!argv.decorate) {
-    const decorators = new Set([...Object.keys(config.styleguide.decorators.oas3_0),
-      ...Object.keys(config.styleguide.decorators.oas3_1), ...Object.keys(config.styleguide.decorators.oas2)]);
+    const decorators = new Set([
+      ...Object.keys(config.styleguide.decorators.oas3_0),
+      ...Object.keys(config.styleguide.decorators.oas3_1),
+      ...Object.keys(config.styleguide.decorators.oas2),
+    ]);
     config.styleguide.skipDecorators(Array.from(decorators));
   }
 
   if (!argv.preprocess) {
-    const preprocessors = new Set([...Object.keys(config.styleguide.preprocessors.oas3_0),
-      ...Object.keys(config.styleguide.preprocessors.oas3_1), ...Object.keys(config.styleguide.preprocessors.oas2)]);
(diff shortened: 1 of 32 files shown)


packages/cli/src/utils.ts
```diff
@@ -391,3 +393,35 @@ export async function loadConfigAndHandleErrors(
     return new Config({ apis: {}, styleguide: {} });
   }
 }
+
+export function sorTopLevelKeysForOas(
+  document: Oas3Definition | Oas2Definition
+): Oas3Definition | Oas2Definition {
+  if ('swagger' in document) {
+    // do we need sort keys for oas2 ?
+    return document;
```
RomanHotsiy, line 402: yes, let's sort them too. I can help to define the order.

packages/cli/src/utils.ts
```diff
@@ -391,3 +393,35 @@ export async function loadConfigAndHandleErrors(
...
+    'x-webhooks',
+    'components',
+  ];
+  const result: any = {};
+  for (const key of orderedKeys as (keyof Oas3Definition)[]) {
+    if (document.hasOwnProperty(key)) {
+      result[key] = document[key];
+    }
+  }
+  return Object.assign(result, document);
```
RomanHotsiy, line 426:
```suggestion
  // merge any other top-level keys (e.g. vendor extensions)
  return Object.assign(result, document);
```

packages/cli/src/utils.ts
IgorKarpiuk: Okay, please share your variant of the order

RomanHotsiy: Let's use like here: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#swagger-object

But move `security`, `tags` and `externalDocs` after `produces`.

packages/cli/src/commands/join.ts
```diff
@@ -27,6 +27,7 @@ import {
   writeYaml,
   exitWithError,
   loadConfigAndHandleErrors,
+  sorTopLevelKeysForOas,
```
lornajane, line 30: Tiny thing: this should probably be `sortTopLevelKeysForOas`, for accuracy.

lornajane (commented):
This is looking good and does what I expected, except with `webhooks`. I think the order makes sense.

For some users, this is a change in behaviour, we should make it clear in the release notes what the feature is, and how they might see unexpected changes to an "unchanged" API description - and why that's a good thing, it'll always be in this new structure.

RomanHotsiy (approved):
LGTM, just make sure to include `webhooks` as suggested by @lornajane

Issue #1054, opened 2023-04-18
If no rules are configured lint passes 🎉 


adamaltman:
**Describe the bug**

A user was seeing no lint problems. However, `preview-docs` was crashing. The issue is their OpenAPI is not valid. Investigating it revealed they use a `redocly.yaml` configuration file with no `rules` and no `extends`.

Is it a bug?  I think so.

**To Reproduce**
1. Make and change into an empty directory.
2. Run `redocly lint`
3. See `Woohoo! Your OpenAPI definitions are valid. 🎉`

```
adam@adams-mbp Redocly % mkdir demo-empty; cd demo-empty; redocly lint
Woohoo! Your OpenAPI definitions are valid. 🎉
```

**Expected behavior**

I expect to see one of these messages:

- No rules were configured
- No APIs were provided

Or something like that.

**Logs**

- none

**OpenAPI definition**

- n/a

**Redocly Version(s)**

1.0.0-beta.123

**`Node.js` Version(s)**

v18.15.0

**Additional context**

- none

lornajane:
Tagging for GA since I think changing the output of the command is a fairly major thing and we definitely should make this change. Possibly related to #1036 which also concerns an invalid config file.

IgorKarpiuk:
In Case 2, should we show an error or a warning ?
An error can cause the lint to break for some users, even if the OpenAPI is correct. 
@adamaltman @lornajane @tatomyr what do you think ?

tatomyr:
Personally I'd vote for error. Something definitely should be fixed if someone is linting an empty definition or doing this against an empty ruleset.

lornajane:
I think warnings (sorry, I realise this is completely opposing advice!). There are a lot of contexts where we lint unexpectedly (or I have found it surprising), I think causing the program to exit with an error status could cause whole pipelines to fail, when we should offer the output as a warning so it can be observed/debugged, but not actually change the existing behaviour to exit with error.

adamaltman:
The cases where we lint is when:

- someone runs the `lint` command
- someone runs the `bundle` command with the `--lint` flag

Are there other cases?

adamaltman:
In my case, case 2 must be an error. This is not a valid OpenAPI file:

```yaml
openapi: 3.1.0
adam-was-here: true
```

None of our commands will work properly on that file.

IgorKarpiuk:
> Are there other cases ?

There seem to be only two cases

lornajane:
[...]

PR #1055, opened 2023-04-19
fix: show error if apis or rules not provided

IgorKarpiuk:
## What/Why/How?

 - show error if apis for lint not provided
 -  show error if  rules or extends not defined in the existing config file 

## Reference
Closes: https://github.com/Redocly/redocly-cli/issues/1054
## Check yourself

- [x] Code is linted

## Security


__tests__/bundle/info-description-override-error/.redocly.yaml
```diff
@@ -1,7 +1,8 @@
 apis:
   main:
     root: ./main.yaml
-
+rules:
```
tatomyr, line 4: Aren't these tests for testing `bundle`? I'd avoid adding rules here as they seems redundant.

IgorKarpiuk: Okay, i removed `--lint` option (where it is not needed) in bundle tests

packages/cli/src/utils.ts
```diff
@@ -453,3 +454,15 @@ function sortOas3Keys(document: Oas3Definition): Oas3Definition {
...
+
+export function checkIfRulesetExist(rules: typeof StyleguideConfig.prototype.rules) {
+  const ruleset = {
+    ...rules.oas2,
+    ...rules.oas3_0,
+    ...rules.oas3_0,
+  };
+
+  if (isEmptyObject(ruleset)) {
+    exitWithError('No rules were configured. Please configure the rules in the configuration file');
```
tatomyr, line 466: Should we exit with the error? I thought it had to be a warning.

lornajane: I think the "no rules" case could be a warning, since we do allow that configuration, but we should avoid showing the message with the :tada:  in it saying things are valid if we didn't actually validate anything.

adamaltman: We say this currently:

> Woohoo! Your OpenAPI definitions are valid. 🎉

We should say this maybe:

> ⚠️ No rules were configured. Learn how to configure rules: https://redocly.com/docs/cli/rules/

adamaltman: I think it should exit with an error only for the particular case when someone runs the lint command. If you are running the lint command and relying on the exit code, you should be aware that lint didn't actually run when you have no rules configured.

lornajane: I think the warning is more appropriate for some contexts, and we do allow a no-rules configuration as valid. The behaviour should be identical whether linting is happening with the lint command or as a side effect to bundle - and turning off rules to get a bundle of whatever the current state is could be a valid use case IMO.

lornajane: In my opinion, the fact that we have currently-passing pipelines is a big reason to stay with a warning. Breaking someone's pipelines makes work for our users and puts them off upgrading in future. I also don't think that no rules is always an error case - that's for the user to decide how to use the tools we give them; outputting the warning gives them the information they need.

adamaltman: I think this is the problem:

> In my opinion, the fact that we have currently-passing pipelines is a big reason to stay with a warning. Breaking someone's pipelines makes work for our users and puts them off upgrading in future. 

This argument doesn't work for me. Let's say a popular unit testing tool had a bug where they didn't report any errors with code. This allowed developers to ship broken code. Now, shouldn't they fix the issue so that tests resume working? (Which means pipelines will fail.) 

If they updated to Redocly CLI  1.0.0-beta.113 to 1.0.0-beta.125, then they are running lint without actually linting anything. This leads them to ship something invalid in production, which is much more devastating than a pipeline not passing. You don't need to lint to bundle the definition. If they were new to Redocly, they may think that it is doing something when it isn't doing anything. If they were pre- 1.0.0-beta.113 users who upgraded, then it used to do something and it doesn't do that any more.

If they are prior to 1.0.0-beta.113, then to keep a similar behavior we should show the error. I consider that we didn't throw an error a bug, and we are potentially causing damage to people who think they are linting when they are not linting.

RomanHotsiy: I agree that we should exit with the error. I can't remember a time when I was looking through success pipeline logs to see if there are any warnings 😅.

The tool is still in beta so I don't think this is too bad to break pipelines at this stage. I also believe we don't have too many users impacted by that.

Alternatively, we can keep the warning for now and delay the breaking change until GA. But I lean towards changing it now.

lornajane: Thanks for chiming in @RomanHotsiy . Let's implement it with errors.

__tests__/lint/extends-empty-array/.redocly.yaml
```diff
@@ -0,0 +1,5 @@
+apis:
+  main:
+    root: openapi.yaml
+
+extends: []
```
tatomyr, line 5: Could you add an empty new lines here and in other places for consistency?

PR #1058, opened 2023-04-25
feat: display message if new version available

IgorKarpiuk:
## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -13,6 +13,7 @@ import { handleLint } from './commands/lint';
 import { handleBundle } from './commands/bundle';
 import { handleLogin } from './commands/login';
 import { handlerBuildCommand } from './commands/build-docs';
+import { notifyUpdateCliVersion } from './utils';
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -13,6 +13,8 @@
     "strictFunctionTypes": true,
     "forceConsistentCasingInFileNames": true,
     "allowJs": false,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
```
RomanHotsiy, line 17: Why you need to enable this?

packages/cli/src/index.ts
```diff
@@ -428,4 +429,5 @@ yargs
   )
   .completion('completion', 'Generate completion script.')
   .demandCommand(1)
-  .strict().argv;
+  .strict()
+  .middleware([notifyUpdateCliVersion]).argv;
```
RomanHotsiy, line 433: Does it run before or after the command? Maybe we should run it after so not to slow down the commands?

PR #1062, opened 2023-04-27
feat: support rule/ as a prefix for configurable rules

### LINKED ISSUE #1061: Revisit assert/ naming for configurable rules

It can be confusing to call something an `assert` when it is really a rule that contains assertions

Name these with `rule/` prefixes instead

Live with the existing situation. We should retain the assert syntax as an alias for now anyway, so that anyone with existing rules can keep using what they have.

**Additional context**

Comes from a discussion in #1043

lornajane:
## What/Why/How?

This is for discussion as well as code review: What about using `rule/` as the prefix for our configurable rules. The assertions are the things inside the rule, and I think the naming contributes to the confusion around this feature.

I've tried to add support for changed syntax without breaking anything, but I'm not sure how to show the user a warning if they're using the old syntax (so we can turn that off at some future point).

**Changelog:** Change syntax for configurable rules to `rule/` instead of `assert/`. The old syntax still works, but is deprecated and will be removed in a future release.

Fixes #1061 

## Reference

https://redocly.com/docs/cli/rules/custom-rules/ (but also updated in https://github.com/Redocly/redocly-cli/pull/1043)

## Testing

Try creating configurable rules named `rule/my-rule-is-awesome` in `redocly.yaml`.

## Check yourself

- [ ] Code is linted

## Security


tatomyr:
> but I'm not sure how to show the user a warning if they're using the old syntax


If you remove [these lines](https://github.com/Redocly/redocly-cli/pull/1062/files#diff-2528117ea3cbe25dfbfd0583c9b11ba7baa7ef347184dfba32a73ec486aa1a5eR230-R232) it will show warninga in the Terminal and errors in VS Code like this (but for `assert/` instead of `rule/`):
Screenshot 2023-04-27 at 17 51 51

adamaltman (approved):
I like this syntax change.

lornajane:
@tatomyr I want the old syntax to keep working, but to warn the user that they should update it. We can then drop it in a future version some day.

lornajane:
When I do it this way @tatomyr , the warning doesn't explain what's wrong. I get an error about `severity`, which is a bit misleading. We should instead pick up the `assert/` and warn that it should be replaced with `rule/`

packages/core/src/types/redocly-yaml.ts
```diff
@@ -225,7 +225,10 @@ const ConfigRootTheme: NodeType = {
 const Rules: NodeType = {
   properties: {},
   additionalProperties: (value: unknown, key: string) => {
-    if (key.startsWith('assert/')) {
+    if (key.startsWith('rule/')) {
+      return 'Assert';
+    } else if (key.startsWith('assert/')) {
+      // keep the old assert/ prefix as an alias
```
tatomyr, line 231: Don't we want to highlight that this is not expected in the config (both in the terminal and VS Code extension)?

packages/core/src/config/config-resolvers.ts
```diff
@@ -411,7 +411,22 @@ function groupStyleguideAssertionRules({
...
+      `The 'assert/' syntax in ${ruleKey} is deprecated. Update your configuration to use 'rule/' instead. Examples and more information: https://redocly.com/docs/cli/rules/configurable-rules/\n`
+  );
+    }
+
+    if (
+      (ruleKey.startsWith('rule/') || ruleKey.startsWith('assert/')) &&
+      typeof rule === 'object' &&
+      rule !== null
+    ) {
       const assertion = rule as RawAssertion;
```
tatomyr, line 430: I'd move the firs `if` clause into the second one like this:

```suggestion

    if (
      (ruleKey.startsWith('rule/') || ruleKey.startsWith('assert/')) &&
      typeof rule === 'object' &&
      rule !== null
    ) {
        if (ruleKey.startsWith('assert/')) {
          logger.warn(
            `The 'assert/' syntax in ${ruleKey} is deprecated. Update your configuration to use 'rule/' instead. Examples and more information: https://redocly.com/docs/cli/rules/configurable-rules/\n`
          );
      }

      const assertion = rule as RawAssertion;
```

This way there's no need to duplicate all the condition.

packages/core/src/types/redocly-yaml.ts
lornajane: I don't know how to warn in this context. It is still valid, but not forever. It would be really good if it didn't autocomplete in the editor, but it isn't exactly an error either ...

tatomyr: If you want to avoid erroring in VS Code, it's ok for now. Having the deprecation messages should be enough.

lornajane:
@tatomyr Is there a way to update the snapshots? The additional warning is causing the outputs to not match, and so the tests fail.

tatomyr:
@lornajane the simplest way will be removing the failing ones an re-running e2e tests locally.

IgorKarpiuk (approved):

Issue #1066, opened 2023-04-24
Crash when `required-string-property-missing-min-length` is on and `required` is not array


duongdominhchau:
**Describe the bug**

Redoc crashes when `required-string-property-missing-min-length` is used and boolean is given to `required` (which expects an array).

Turning off that specific rule will make the crash disappear.

**Expected behavior**

An "invalid input" error should be displayed instead of crashing.

**Minimal reproducible OpenAPI snippet(if possible)**

```openapi
openapi: 3.1.0
info:
  version: 0.0.1
  title: BugDemo
components:
  schemas:
    BugDemo:
      type: object
      properties:
        id:
          type: string
          required: true # <--- This
```

**Screenshots**

![image](https://user-images.githubusercontent.com/56427890/233896679-1da3ed06-e59d-4164-87d9-9a11bc045446.png)

**Additional context**

RomanHotsiy:
@tatomyr take a look at this issue.
This is most likely same cause as https://github.com/Redocly/redocly-cli/issues/764

tatomyr:
The cause is not the same as in #764, but they are close enough. 

@duongdominhchau  the issue with the example above is that the `required` field must be an array (see the [docs](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties)).
In your case it probably should be like this (mind the indentation too):

```yaml
openapi: 3.1.0
info:
  version: 0.0.1
  title: BugDemo
components:
  schemas:
    BugDemo:
      type: object
      properties:
        id:
          type: string
      required: [id] # <--- This
```

But the error itself is quite misleading though, I agree.

tatomyr:
Should be fixed in `v1.0.0-beta.127`.

Issue #764, opened 2022-07-11
Incorrect error message


tatomyr:
**Describe the bug**
Sometimes we get the following error: 
```
  - Cannot read properties of null (reading 'allOf').
```
However, it isn't clear where exactly to look. Moreover, there may be no `allOf` keyword in the definition at all. 

**To Reproduce**
See the [conversation](https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1657544976863529).

**Expected behavior**
It should show a correct error message whatever it is.

**OpenAPI definition**
See [here](https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1657555994717109?thread_ts=1657544976.863529&cid=C02UHE7EL3E).

**Redocly Version(s)**
`1.0.0-beta.104`

**`Node.js` Version(s)**
`v16.14.0`

PR #1068, opened 2023-05-08
fix: incorrect error message on null nodes

tatomyr:
## What/Why/How?

Fixes the case when an openapi definition node is incorrect (and spec rule detected that) but still being processed because the ignored status has been lost (e.g. node is being referenced from another location).

## Reference

Closes https://github.com/Redocly/redocly-cli/issues/764
Closes https://github.com/Redocly/redocly-cli/issues/1066

## Check yourself

- [x] Code is linted

## Security


packages/core/src/walk.ts
```diff
@@ -118,7 +113,7 @@ export function walkDocument<T>(opts: {
 }) {
   const { document, rootType, normalizedVisitors, resolvedRefMap, ctx } = opts;
   const seenNodesPerType: Record<string, Set<any>> = {};
-  const seenRefs = new Set<any>();
```
RomanHotsiy, line 121: It was unused? 😮

RomanHotsiy (approved):
LGTM 👍

packages/core/src/walk.ts
tatomyr: Yep

IgorKarpiuk (approved):

Issue #324, opened 2021-07-12
Allow adding refs from pre-processors


konstantin-smolyakov-idt:
**Is your feature request related to a problem? Please describe.**
I would like to implement a custom plugin to conventionally add responses to each operation. For example, I would like to add 400 errors to all operations in my spec without having to specify it each time. 

**Describe the solution you'd like**

Let's imagine a configuration file with the following entries. It will mean that all operations can return 400 or 422. Potentially, this can be extended with a filter to include only certain operations by convention (e.g. a delete request might not return 400 error).
```
preprocessors:
    my-custom-pluging/conventional-response:
      '422':
        response:
          $ref: './common/errors/validation-error.response.yaml'
      '400':
        response:
          $ref: './common/errors/bad-request-error.response.yaml'
```
I've tried to implement a PoC for this, but it appears that references are all processed before any plugins are invoked. So I end up with `Can't resolve $ref` error messages from the bundler. 

**Describe alternatives you've considered**
The alternative will be to build a rule to enforce this. But this is not the same, because there is no way to auto-fix the issue and it adds some churn on the developer to remember to add all standard errors to operations. 

**Additional context**
I understand that this might not go well with the idea of reporting the precise error location in case other validators run on post-processed results. However, maybe it can be solved by restoring the post-processed tree somewhere in the temporary folder and pointing issues there.

adamaltman:
Have you considered writing a decorator to do this? I was thinking of a similar thing (common headers) recently (but as a decorator).

adamaltman:
Also, I suppose you could have these in your components, and reference the components? For example:
```
components:
  responses:
    validation-error:
```

And then you would inject `$ref: #/components/responses/validation-error` instead.

RomanHotsiy:
> I understand that this might not go well with the idea of reporting the precise error location in case other validators run on post-processed results

Yeah, preprocessors will break error reporting right now and this is not too simple to resolve the issue so we try to avoid them.
They will also break autocompletion in our vs code plugin, etc. 

We do in fact run "resolve" step before any preprocessor takes in to improve performance and to make our rules run synchronously. I believe we used to run resolve again after preprocessors run but looks it was removed, I think we can re-add it.

On the other hand, while I understand you want to reduce the duplication, note that you will lose some transparency.

konstantin-smolyakov-idt:
> Have you considered writing a decorator to do this? I was thinking of a similar thing (common headers) recently (but as a decorator).

[...]

PR #1081, opened 2023-05-15
feat: resolve $refs in preprocessors

tatomyr:
## What/Why/How?

> TODO: update docs

Resolve $refs after running preprocessors with the following option in **redocly.yaml**:

```yaml
...
resolve:
   resolveAfterTransformers: true
```

## Reference

Closes https://github.com/Redocly/redocly-cli/issues/324

## Check yourself

- [x] Code is linted

## Security


__tests__/bundle/bundle-lint-format/checkstyle-format-snapshot.js
```diff
@@ -9,7 +9,7 @@ exports[`E2E bundle lint format bundle lint: should be formatted by format: chec
 </checkstyle>
 
 ❌ Validation failed with 1 error.
-run \`openapi lint --generate-ignore-file\` to add all problems to the ignore file.
+run \`redocly lint --generate-ignore-file\` to add all problems to the ignore file.
```
RomanHotsiy, line 12: Could you extract those changes into separate PR plz?

tatomyr: Sure. Just stored everything in the draft in order not to lose the changes.

rotorgames:
@tatomyr will we be able to set it up in the redocly.config to make it work with the vscode extension?

packages/core/src/bundle.ts
```diff
@@ -129,6 +129,23 @@ export async function bundleDocument(opts: {
     });
   }
 
+  if (config.resolveAfterTransformers) {
```
RomanHotsiy, line 132: I was thinking about it a bit. Let's not have this config.

Let's enable it if only there are any preprocessors enabled.

We don't have too many users using preprocessors I believe.
We should just add a little warning somewhere in the docs that using preprocessors has a performance penalty.

packages/core/src/lint.ts
```diff
@@ -77,13 +77,35 @@ export async function lintDocument(opts: {
 
   const preprocessors = initRules(rules as any, config, 'preprocessors', oasVersion);
   const regularRules = initRules(rules as Oas3RuleSet[], config, 'rules', oasVersion);
-  const normalizedVisitors = normalizeVisitors([...preprocessors, ...regularRules] as any, types);
+
+  if (config.resolveAfterTransformers) {
```
RomanHotsiy, line 81: I think this is good to just have this piece copied over. Extracting into function may be tempting but this is easier to read I think.

👍

RomanHotsiy (commented):
Looks pretty good.

Left a few comments.

packages/core/src/bundle.ts
```diff
@@ -129,6 +129,23 @@ export async function bundleDocument(opts: {
...
+    // Make preliminary pass to be able to resolve refs defined in preprocessors in the next pass.
+    const preliminaryResolvedRefMap = await resolveDocument({
```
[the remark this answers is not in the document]
tatomyr: Could be. Let me rewrite it without the config option and see if that would be better.

tatomyr:
@rotorgames yes, I believe it will work in VS Code extension by default.

packages/core/src/bundle.ts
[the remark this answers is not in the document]
tatomyr: Yeah, using `let` makes more sense in this context. Refactored.

IgorKarpiuk (approved):

Issue #1063, opened 2023-04-27
Ignoring a rule also ignores completely unrelated rules


jankodera-ps:
**Describe the bug**
`--skip-rule` seems to skip all rules instead of just the rule that is specified to be skipped.

**To Reproduce**
1. Given this [repro.yaml](https://gist.github.com/jankodera-ps/5a580bb371256f38c93fb8a7ffeefa2e)
2. Run `redocly lint repro.yaml`
4. See two errors (this is expected): missing scopes (`spec` rule), missing 4XX response (`operation-4xx-response` rule).
5. Run `redocly lint --skip-rule=operation-4xx-response repro.yaml`
6. Get `Woohoo! Your OpenAPI definitions are valid.`

**Expected behavior**
Step 6 should return one error, from `spec` rule.

**Redocly Version(s)**
1.0.0-beta.125

**`Node.js` Version(s)**
19.9.0

Orest-Yastremskyy:
Hi @jankodera-ps,

I believe that this issue occurs because of the wrong order of parameters passed to the lint command.

What I mean is if you pass the API definition file first and additional parameters after (e.g. ` redocly lint repro.yaml --skip-rule=operation-4xx-response` instead of `redocly lint --skip-rule=operation-4xx-response repro.yaml`), everything works fine. In case you pass additional parameters first, you will get the invalid validation of the API definition:

![image](https://user-images.githubusercontent.com/96121440/235080777-217d7cc9-c268-450b-b318-d99fabf1b431.png)


Hope it helps.

jankodera-ps:
Okay, happy to see it can be easily fixed, but have to say this behavior is definitely unexpected.

tatomyr:
The issue here is that the `--skip-rule` option expects an array, so it recognises the positional argument as a part of the array. This is a known [issue](https://github.com/yargs/yargs/issues/1848) in **yargs** library, which we use here for parsing the arguments. 

By the way, the behaviour will change after this MR: https://github.com/Redocly/redocly-cli/pull/1055 (it will fail with a message that no APIs were provided).

One option to fix this is to use the parser configuration setting `'greedy-arrays': false` (which I believe will eventually become **yargs** default). However, it will affect our current users' scripts.

jankodera-ps:
[...]

PR #1083, opened 2023-05-16
fix: avoid different result depending on option arguments position

tatomyr:
## What/Why/How?

Changes greedy arrays option to false in **yargs**.
This eliminates dependency on position of arguments, e.g. those commands will produce the same result: 

```bash
redocly lint --skip-rule=operation-4xx-response openapi.yaml

redocly lint openapi.yaml --skip-rule=operation-4xx-response
```

To pass an array of values you will have to explicitly point to the option, e.g.: 

```bash 
redocly lint --skip-rule=operation-4xx-response --skip-rule=operation-2xx-response openapi.yaml
```

The change affects **ALL** options behaviour: `--skip-rule`, `--skip-decorator`, `--skip-preprocessor`, `--extends` and `--files`.

Additional motivation is that **yargs** is going to make this behaviour default in next versions ([link](https://github.com/yargs/yargs-parser/blob/main/README.md#greedy-arrays)).

## Reference

Closes https://github.com/Redocly/redocly-cli/issues/1063

Related docs: [link](https://github.com/Redocly/redocly-cli/pull/1084)

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -19,6 +19,7 @@ const version = require('../package.json').version;
 yargs
   .version('version', 'Show version number.', version)
   .help('help', 'Show help.')
+  .parserConfiguration({ 'greedy-arrays': false })
   .command(
     'stats [api]',
     'Gathering statistics for a document.',


RomanHotsiy (approved):

RomanHotsiy:
Make sure to add a short explanation to the Changelog.

lornajane:
I think we might also want to update the usage/help output, but would it be better for us to do that in a follow up pull request? I'd be happy for it to be included here, or captured as a separate issue (tagged GA).

RomanHotsiy:
> I think we might also want to update the usage/help output

I feel like it should be done in this pull-request. I don't have a strong opinion here. Either works 😄

tatomyr:
> I think we might also want to update the usage/help output

@lornajane  What exactly output do you mean?

roman-sainchuk (approved):

lornajane:
I mean the output of `redocly lint --help`

tatomyr:
> I mean the output of redocly lint --help
Oh, it gets generated automatically by yargs, I believe.

Issue #1032, opened 2023-03-24
Make `push` command clearer in usage and help text


lornajane:
**Is your feature request related to a problem? Please describe.**

The existing `push` command is a bit confusing with arguments named `maybeX` and some positionals that are optional.

**Describe the solution you'd like**

I'd like a more precise and friendly command. As a logged in user I should be able to do:

```
redocly push awesome-api.yaml --organization AwesomeCorp --name AwesomeApi --version v1.3
```

The API (API name or file/path to API description) is optional in the first parameter position, and we have this pattern across the other commands too, so I think it's a convention to keep.

The current command uses a "destination", which should be a named parameter rather than positional (since both positionals are optional). We don't use destination anywhere else that I can find, and it's a composite where the docs show the user how to find the API name, and the version, and add the correct punctuation (optionally with the organization as well). I think it's simpler just to ask for the things we actually want and that they know - the API name and desired version.

With no arguments at all, all APIs configured will be pushed to their existing destinations: `redocly push`

The current branch name is already supported as a parameter `--branch` but this has no help text to explain it, we should add that. IT would be used like this: `redocly push myAwesomeAPI --branch cool-feature-7`

The example above uses the organization ID from the configuration file, the API root, name and version as set in the configuration file.

To push a new version of an existing API:  `redocly push myAwesomeAPI --version v1.44`

**Describe alternatives you've considered**

Keeping the current setup is an option, but #265 and #1031 indicate that the destination piece is difficult to get right - with separate arguments we can be more specific about what is wrong when a user makes a mistake, and potentially allow people to quote the strings on the command line to accommodate spaces/symbols, which is why I think this suggestion might be easier to use.

**Additional context**

It's a breaking change - which is OK in beta but let's work hard at writing really good changelogs and error messages (I'm happy to help with this), to help users understand what should be different. We could also consider releasing one version that silently supports the old syntax, to give people a chance to move over, and drop it in a later release.

**Remaining questions**

What happens if the version isn't specified? Does the tool push to a new version, or the last known one?

How does the `upsert` parameter fit in with this? The `upsert` parameter currently has no help text (and the docs assume I know what an upsert is in this context) - so let's fix that! If I push to an existing API version without setting the `upsert`, would the push fail?

If users need to know their Organization, API name and versions, could we make those available from command line as a quick lookup/listing? I don't think we know how many people have this already in a configuration file.

[...]

PR #1084, opened 2023-05-16
docs: make greedy arrays false in yargs

tatomyr:
## What/Why/How?

Docs for https://github.com/Redocly/redocly-cli/pull/1083

## Check yourself

- [ ] Code is linted

## Security


docs/commands/lint.md
```diff
@@ -22,19 +22,19 @@ redocly lint --version
...
---help | boolean | Show help.
---lint-config | string | Specify the severity level for the configuration file. <br/> **Possible values:** `warn`, `error`, `off`. Default value is `warn`.
---max-problems | integer | Truncate output to display the specified [maximum number of problems](#max-problems).
---skip-preprocessor | array | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.
---skip-rule | array | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.
---version | boolean | Show version number.
+| Option                 | Type     | Description                                                                                                                        |
+| ---------------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| apis                   | [string] | Array of API definition filenames that need to be linted. See [the Apis section](#apis) for more options.                          |
```
lornajane, line 27: Should this be `api` and not an array if we only accept a single argument? And the description perhaps making it more clear that it's just a string.

docs/commands/join.md
```diff
@@ -40,19 +41,19 @@ redocly join --version
...
---without-x-tag-groups | boolean | Skip automated `x-tagGroups` creation. See the [without-x-tag-groups](#without-x-tag-groups) section.
---version | boolean | Show version number.
---output, -o | string | Name for the joined output file. Defaults to `openapi.yaml`. **If the file already exists, it's overwritten.**
+| Option                             | Type     | Description                                                                                                                                                                                              |
+| ---------------------------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| apis                               | [string] | **REQUIRED.** 1. Array of paths to API definition files that you want to join. At least two input files are required.<br />2. A wildcard pattern to match API definition files within a specific folder. |
```
lornajane, line 46: This one is a bit different to the others, because we do allow multiple positional arguments for this specific command. Is it accurate to call it an array? Or should we instead mention that this argument should repeat as many times as needed?

docs/commands/bundle.md
```diff
@@ -169,15 +168,15 @@ All other information is omitted.
 You may want to skip specific preprocessors, rules, or decorators upon running the command.
 
 ```bash Skip preprocessors
-redocly bundle --skip-preprocessor=discriminator-mapping-to-one-of,another-example
+redocly bundle --skip-preprocessor=discriminator-mapping-to-one-of --skip-preprocessor=another-example
 ```
 
 ```bash Skip rules
-redocly bundle --skip-rule=no-sibling-refs,no-parent-tags
```
RomanHotsiy, line 176: I wander if we can support backward-compatibility here? What do we receive as a value for `skip-rule` arg from the parser? `no-sibling-refs,no-parent-tags`? Maybe we can split it manually?

tatomyr: It never worked this way :(

docs/commands/lint.md
tatomyr: It accepts multiple apis, so I guess the description is correct.

tatomyr: It accepts multiple apis, so I guess the description is correct.

docs/commands/bundle.md
```diff
@@ -24,27 +24,26 @@ redocly bundle --version
...
---skip-rule | [string] | Ignore certain rules. See the [Skip preprocessor, rule, or decorator section](#skip-preprocessor-rule-or-decorator).
---version | boolean | Show version number.
-
+| Option                     | Type     | Description                                                                                                                                                                                                                                                     |
+| -------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| apis                       | [string] | List of API root definition filenames or names assigned in the `apis` section of your Redocly configuration file. Default values are all names defined in the `apis` section within your configuration file.                                                    |
```
adamaltman, line 29: Should the values be comma separated?

tatomyr: I think they shouldn't. The most common behaviour for positional arrays is simply separating items by spaces.

docs/commands/join.md
tatomyr: I don't think we allow. The only command that accepts multiple positionals is `push`.

docs/commands/lint.md
lornajane: I think it is more correct to say that we accept one API argument, supplied multiple times. So the usage should be something like:

```
redocly lint [options ...] apis ...
```

If there is a space in it, it's a new input parameter. At least, that's what I'd expect from using other CLI tools.

docs/commands/join.md
lornajane: I think the "array" here is the `api` argument multiple times. The `push` command should be getting an update to remove the multiple positionals (at least officially, even if we still support the old syntax)

docs/commands/bundle.md
lornajane: Agree that there should not be commas in CLI inputs. It's more expected to do repeated arguments

docs/commands/join.md
tatomyr: I'd suggest updating the docs regarding `push` command in scope of this ticket: [Make push command clearer in usage and help text#1032](https://github.com/Redocly/redocly-cli/issues/1032)

docs/commands/bundle.md
adamaltman: This is a positional argument (not a named parameter). I suppose it is unlikely anybody will use the names of other options as their API names.

docs/commands/join.md
adamaltman: @tatomyr the `bundle` command above indicates it accepts multiple positional arguments too (and I'm aware people do use that).

docs/commands/lint.md
```diff
@@ -14,27 +14,28 @@ To learn more about preprocessors and rules, refer to the [custom plugins](../re
...
---help | boolean | Show help.
---lint-config | string | Specify the severity level for the configuration file. <br/> **Possible values:** `warn`, `error`, `off`. Default value is `warn`.
---max-problems | integer | Truncate output to display the specified [maximum number of problems](#max-problems).
---skip-preprocessor | array | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.
---skip-rule | array | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.
---version | boolean | Show version number.
+| Option                 | Type     | Description                                                                                                                             |
+| ---------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| api                    | string   | An API definition filename that need to be linted. Multiple entires could be specified. See [the Apis section](#apis) for more options. |
```
adamaltman, line 28: How are multiple entries specified?
```suggestion
| api                    | string   | An API definition filename that need to be linted. Multiple entries could be specified. See [the Apis section](#apis) for more options. |
```

I don't understand why all the commands have an `<apis>` positional and now the lint command is changing to not have that?

adamaltman: Where exactly does the `#apis` link go to?

tatomyr: `#apis` goes here: https://redocly.com/docs/cli/commands/lint/#apis (it corresponds with the line 42 below).

docs/commands/join.md
tatomyr: @adamaltman Yes, it does. Are we discussing the `bundle` or `join` or `push` command here? I've got lost 😄

lornajane:
I think the scope of this changing is creeping quite a bit. Let's take the discussion about `api` vs `apis` to a separate issue, since that's mostly a user interface and docs update and it would apply to all commands, and concentrate updates only on the parameters that were changed in this pull request. I'm happy to help with the followup issue.

docs/commands/lint.md
lornajane: I realised that this should be a wider conversation, beyond the scope of this pull request.

lornajane (approved):
Thanks for fixing this, and for being patient with us all with extra questions :)

PR #1089, opened 2023-05-19
chore: add performance benchmark with hyperfine

tatomyr:
## What/Why/How?

Added performance benchmarking using `hyperfine`. 

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -10,19 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Cache dependencies
-        uses: actions/cache@v3
+      - uses: actions/setup-node@v3
         with:
-          path: ~/.npm
-          key: npm-${{ hashFiles('package-lock.json') }}
-          restore-keys: |
-            npm-${{ hashFiles('package-lock.json') }}
-            npm-
-      - name: Environments
-        run: |
-          echo
-          echo Using Node $(node -v), NPM $(npm -v)
-          echo
+          node-version: 18
+          cache: 'npm'
       - name: Install Dependencies
         run: npm ci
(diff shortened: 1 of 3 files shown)


.github/workflows/performance.yaml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.npm
+          key: npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            npm-${{ hashFiles('package-lock.json') }}
+            npm-
```
RomanHotsiy, line 20: no need to use it anymore, use setup-node@v3 action which has cache built-in and which is more convenient:

https://github.com/actions/setup-node#caching-global-packages-data

.github/workflows/performance.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Environments
+        run: |
+          echo
+          echo Using Node $(node -v), NPM $(npm -v)
+          echo
+      - name: Install Dependencies
+        run: npm ci
+      - name: Install External
+        run: npm i -g hyperfine @redocly/cli@latest
```
RomanHotsiy, line 29: should it compare with the latest or with the version from main?

I think latest is fine for now but would be great to check-out main branch and compare it to it somehow in the future.

tatomyr: Yes, comparing to main would be better, but I haven't found a simple and reliable solution to this yet. Will check how this approach works first, and then we can decide how to evolve it.

RomanHotsiy (approved):

IgorKarpiuk (approved):

SmoliyY (approved):

Issue #779, opened 2022-07-27
Update outdated version warning in output


adamaltman:
**Is your feature request related to a problem? Please describe.**

There is no good way to alert users to new updates in the CLI tool.

**Describe the solution you'd like**

I'd like part of the CLI output to check and show the current version versus the running version. If the current version is different, to show this output... something like

```bash
A new version of Redocly CLI (1.0.0-beta.105) is available.
Update now: npx @redocly/cli@latest upgrade
Changelog: {TODO}
```

Optimal would be an integrated changelog information and show every change between the version installed and the current version.

**Describe alternatives you've considered**

None yet.

**Additional context**

Consider also the JSON output format so that it can be parsed and programmatically displayed in Workflows.

RomanHotsiy:
`redoc-cli` team just opened a PR for this too: https://github.com/Redocly/redoc/pull/2100/files. We can use same implementation.

tatomyr:
Had to revert this as lint fails in Workflows.

IgorKarpiuk:
This is fails in Workflows because webpack doesn't work with `import-lazy`
- We can fork [this repo](https://github.com/yeoman/update-notifier)
- Or we can use [this fork](https://github.com/datacrafts-io/update-notifier-webpack), but it has not been updated for some time
- We can create our own solution 

@tatomyr @RomanHotsiy

lornajane:
For the target audience using the CLI either themselves at the terminal, or in a CI setup, the version information and upgrade prompt are really valuable. However since some of our commands output JSON or other formats, would we consider sending this to `stderr`, or maybe only fitting the output to a subset of commands? And would either of those approaches help with the problem we encountered previously where Workflows tripped up on the extra output?

PR #1090, opened 2023-05-19
feat: create update notifier for cli

IgorKarpiuk:
## What/Why/How?

Create update notifier for cli and display message if new version available 

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/779
## Screenshots (optional)
Знімок екрана 2023-05-22 о 10 33 09

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/update-version-notifier.ts
```diff
@@ -0,0 +1,98 @@
...
+    const padding = ' '.repeat(maxLength - cleanColors(line).length);
+    const extraSpaces = index !== 0 ? ' '.repeat(SPACE_TO_BORDER) : '';
+    return `${extraSpaces}${chalk.yellow('║')}  ${line}${padding}  ${chalk.yellow('║')}`;
+  }
+};
+
+const isNeedsToBeChecked = (): boolean => {
+  try {
+    // Last check time is stored as a timestamp in a file in the OS temp folder
+    const lastCheckFile = path.join(os.tmpdir(), TIMESTAMP_FILE);
```
tatomyr, line 77: Please verify this will work on Win.

packages/cli/src/update-version-notifier.ts
```diff
@@ -0,0 +1,98 @@
...
+import fetch from 'node-fetch';
+import { cleanColors } from './utils';
+
+const { version, name } = require('../package.json');
+
+const TIMESTAMP_FILE = 'redocly-cli-check-version';
+const SPACE_TO_BORDER = 4;
+
+// 12 hours
+const INTERVAL_TO_CHECK = 1000; // * 60 * 60 * 12;
```
tatomyr, line 15: Don't forget to clean up here and in other places.

packages/cli/src/update-version-notifier.ts
```diff
@@ -0,0 +1,96 @@
+import * as os from 'os';
+import * as path from 'path';
+import * as fs from 'fs';
+import * as semver from 'semver';
```
SmoliyY, line 4: Is it make sense to import all module, instead of only needed methods?

packages/cli/src/update-version-notifier.ts
```diff
@@ -0,0 +1,96 @@
...
+        return getLineWithPadding(line, index);
+      })
+      .join('\n')}
+    ${yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║')}
+    ${yellow('╚' + border + '╝')}
+  `;
+
+  process.stderr.write(banner);
+
+  function getLineWithPadding(line: string, index: number): string {
```
SmoliyY, line 66: For me it is a little bit confusing to have function declaration here, I would better move this function outside, and pass `maxLength` as argument.

packages/cli/src/update-version-notifier.ts
IgorKarpiuk: Nope, Thanks

SmoliyY (approved):

tatomyr:
@IgorKarpiuk could you also check if it runs in a TEST environment, so the notifier won't get triggered in that case?

SmoliyY:
> @IgorKarpiuk could you also check if it runs in a TEST environment, so the notifier won't get triggered in that case?

Maybe it is better to rename `TEST` on `REDOCLY_TEST` because as we know some other libraries also relay on such env variable and our tests can work in not appropriate way. We have  this issue with prev implementation of the same feature

SmoliyY (approved):

tatomyr:
@SmoliyY I believe `NODE_TEST` is the conventional env variable.

tatomyr (approved):

tatomyr:
Actually, we might only want to skip this in CI environment. Not sure if `NODE_ENV` makes too much sense.

PR #1097, opened 2023-05-24
Add MLC for link checking

lornajane:
Ignores relative links since we can only check those from the main
website project.

## What/Why/How?

It's difficult to know when we're breaking links in our docs - this will automatically check on each pull request

## Reference

Uses MLC (https://github.com/becheran/mlc)

And its GitHub action (https://github.com/marketplace/actions/markup-link-checker-mlc)

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/.github/workflows/mlc-link-check.yaml
+++ b/.github/workflows/mlc-link-check.yaml
@@ -1,0 +1,17 @@
+name: Check links
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - main
+jobs:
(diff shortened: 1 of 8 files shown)


lornajane:
@adamaltman and @HCloward you might have opinions on this additional docs tool. I can add it to the other repos if we think it's a good fit - I just added it here as a good small target to start with :)

adamaltman (approved):

tatomyr (approved):

tatomyr (approved):

PR #1102, opened 2023-05-25
Refresh lint article

lornajane:
Rescue some changes from #665 that are worth keeping, and signpost the
rules documentation for configuration examples rather than having that
information duplicated here.

## Check yourself

- [ ] Code is linted

## Security


docs/commands/lint.md
```diff
@@ -102,18 +102,7 @@ By default, the CLI tool looks for the [Redocly configuration file](/docs/cli/co
...
 ```
 
-### Extend configuration
-
-The `--extends` option allows you to extend the existing configuration. This option accepts one of the following values: `minimal`, `recommended`, `all`. Each of the values is a base set of rules that the lint command will use. You can further modify this set in cases when you want to have your own set of rules based on the existing one, including particular rules that cover your specific needs.
-
-:::warning Important
-
-When you run the `lint` command without a configuration file, it uses the `extends: recommended` by default.
-However, if you have a configuration file, but it doesn't include any rules or extends configuration, the `lint` command shows an error.
```
adamaltman, line 112: Isn't this part important to understanding the behavior of the command?

docs/commands/lint.md
```diff
@@ -191,6 +231,8 @@ redocly lint --max-problems 200
 < ... 2 more problems hidden > increase with `--max-problems N`
 ```
 
+The default value for `max-problems` is 100.
```
adamaltman, line 234: We usually report default values in the summary tables (should this move into line 34?).

docs/commands/lint.md
```diff
@@ -173,12 +214,11 @@ redocly lint --format=checkstyle
 </checkstyle>
 ```
 
-In this format, `lint` uses the [Checkstyle](https://checkstyle.org/) XML report format.
 Due to the limitations of this format, only file name, line, column, severity,
 and rule ID (in the `source` attribute) are included. All other information is
 omitted.
 
-### Max problems
+### Limit the problem count
```
adamaltman, line 221: If you change this, change the corresponding deep anchor link on line 34.

docs/commands/lint.md
```diff
@@ -156,8 +145,60 @@ openapi/core.yaml:
 
 In this format, `lint` shows the file name, line number, and column where the problem occurred. However, the output is compressed and omits other contexts and suggestions.
 
+##### JSON
+
+It can be useful to get the output in JSON format to be processed by other tools.
```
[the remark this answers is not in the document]
lornajane: Good point, I'll add a warning. I think the JSON output is more likely to be useful than linting multiple APIs in one command is, so it should be documented.

docs/commands/lint.md
lornajane: I don't think this section belongs here in the CLI command reference. We have the information in the rules section.

docs/commands/lint.md
lornajane: Ahhh, there's a few of these! Thanks!

docs/commands/lint.md
lornajane: I'll move it.

docs/commands/lint.md
```diff
@@ -31,18 +31,18 @@ redocly lint --version
...
-| --max-problems         | integer  | Truncate output to display the specified [maximum number of problems](#max-problems).                                              |
+| --max-problems         | integer  | Truncate output to display the specified [maximum number of problems](#max-problems). Default value is 100.                        |
 | --skip-preprocessor    | [string] | Ignore certain preprocessors. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.                       |
 | --skip-rule            | [string] | Ignore certain rules. See the [Skip preprocessor or rule section](#skip-preprocessor-or-rule) below.                               |
 | --version              | boolean  | Show version number.                                                                                                               |
 
 ## Examples
 
-### Apis
+### <a id="apis"></a>Lint APIs
```
adamaltman, line 41: Do you need the anchor tag? Won't it be calculated that way automatically?

adamaltman (approved):

docs/commands/lint.md
adamaltman: Oh, it is changing to Lint APIs (I missed the word Lint).

Issue #682, opened 2022-05-11
Not correct validation $ref in properties


AlexVarchuk:
**Describe the bug**
The lint command should mark this schema as invalid.

**To Reproduce**
1. run the lint command with the example of the definition

**Expected behavior**
Should report a problem with $ref. $ref cannot use in `properties` like in the example

**OpenAPI definition**
_openapi.yaml_
```
paths:
  /pet:
    get:
      operationId: example
      summary: summary example
      responses:
        '200':
          description: example description
          content:
            application/json:
              schema:
                type: object
                properties:
                  $ref: './nestedRef.yaml'
                  name:
                    type: string

```
_nestedRef.yaml_
```
total:
  description: Total data number
  type: number
  example: 1000
pages:
  description: Total pages
  type: number
  example: 5
```

**Redocly Version(s)**
1.0.0-beta.97

**Node.js Version(s)**
15

lornajane:
We should change this behaviour with our upcoming major release.

PR #1112, opened 2023-05-31
feat: create rule for $ref validation

IgorKarpiuk:
## What/Why/How?

Add new rule to restrict the use of $ref

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/682

## Screenshots (optional)
Знімок екрана 2023-05-31 о 13 02 12

## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/oas3/spec-ref-validation.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const SpecRefValidation: Oas3Rule = () => {
+  function validateNoRef(node: unknown, { report, rawLocation }: UserContext) {
+    if (isRef(node)) {
+      report({
+        message: 'Invalid usage of $ref',
+        location: rawLocation.child('$ref').key(),
+      });
+    }
+  }
```
tatomyr, line 13: Do we need a separate function? Why don't put it simply in place [here](https://github.com/Redocly/redocly-cli/pull/1112/files?diff=unified&w=1#diff-7a126114b978592806f48c748aecefa06259fc940e41d8f3dcdc410db5380b44R44)?

packages/core/src/rules/oas3/spec-ref-validation.ts
```diff
@@ -0,0 +1,50 @@
+import { Oas3Rule, VisitFunction } from '../../visitors';
+import { isRef } from '../../ref-utils';
+import { UserContext } from '../../walk';
+
+export const SpecRefValidation: Oas3Rule = () => {
+  function validateNoRef(node: unknown, { report, rawLocation }: UserContext) {
+    if (isRef(node)) {
+      report({
+        message: 'Invalid usage of $ref',
```
tatomyr, line 9: What about changing the message to `Field $ref is not expected here.`?

tatomyr (commented):
Is this rule only applicable to OAS3? Can we have a similar one for OAS2?

Also, you are only assessing nodes that are not allowed to contain refs. What if a $ref is somewhere inside these nodes? Wouldn't it be better to go through all nodes and report a problem when the place is not explicitly allowed to contain refs?

IgorKarpiuk:
@tatomyr I believe we can use it for both

packages/core/src/rules/common/spec-ref-validation.ts
```diff
@@ -0,0 +1,28 @@
...
+    'Header',
+    'SecurityScheme',
+    'Link',
+    'Callback',
+    'PathItem',
+  ];
+
+  return {
+    any(_node, { report, rawNode, rawLocation, type }) {
+      if (!nodesToValidate.includes(type.name) && isRef(rawNode)) {
```
tatomyr, line 20: It's hard to read :) Could you avoid negation in the statement?

IgorKarpiuk: I think it's better to have one condition here

packages/core/src/rules/common/spec-ref-validation.ts
```diff
@@ -0,0 +1,30 @@
...
+    'Header',
+    'SecurityScheme',
+    'Link',
+    'Callback',
+    'PathItem',
+  ];
+
+  return {
+    any(_node, { report, rawNode, rawLocation, type }) {
+      const isShouldCheck = !nodesToSkip.includes(type.name);
```
tatomyr, line 20:
```suggestion
      const shouldCheck = !nodesToSkip.includes(type.name);
```

packages/core/src/rules/common/spec-ref-validation.ts
```diff
@@ -0,0 +1,30 @@
...
+
+      if (isShouldCheck && isRef(rawNode)) {
```
tatomyr, line 22:
```suggestion
      if (shouldCheck && isRef(rawNode)) {
```

tatomyr (approved):
Suggested one small improvement.
Overall looks good!

packages/core/src/rules/common/spec-ref-validation.ts
RomanHotsiy, line 1: can we still rename this rule? I would prefer `spec-strict-refs` or something esle.

`spec-ref-validation` is not consistent with our other rule names. Also, `validation` is redundant in lint rule context.

cc @IgorKarpiuk @tatomyr @adamaltman

IgorKarpiuk: Okay, I will rename

tatomyr: I like `spec-strict-refs`.

lornajane: I do too! Good catch @RomanHotsiy

Issue #1104, opened 2023-05-25
The regular expression "^[a-zA-Z0-9\.\-_]+$" should only be applied to fixed-fields under component


ZyzychO:
Examples map keys defined in components must adhere to a pattern: '^[a-zA-Z0-9._-]+$'. (Source: https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml)

However, no such restrictions are defined on the examples map defined inside of a media-type object.

```
openapi: 3.0.0
info:
  title: Sample API
  description: Optional 
  version: 0.1.9

paths:
  /users:
    get:
      summary: Returns a list of users.
      description: Optional extended description in CommonMark or HTML.
      responses: 
        '200': 
          description: OK
          content: 
            application/json: 
              schema: 
                type: object
              examples:
                My valid example:
                  value: 
                    foo: bar
                A_valid_example:
                  value: 
                    baz: quz   
```
As you can see on the example above, the name of the example `My valid example` will be defined as invalid by rule:  **spec-components-invalid-map-name**.
<img width="1063" alt="Screenshot 2023-05-26 at 00 06 16" src="https://github.com/Redocly/redocly-cli/assets/106024790/4edf960d-9744-407b-ad67-de295770f1d6">
[example-yaml.yaml.zip](https://github.com/Redocly/redocly-cli/files/11569417/example-yaml.yaml.zip)

**Expected behavior**

rule: **spec-components-invalid-map-name** should apply only to the fixed fields under the components, this restriction should not apply to examples mapped inside of a media-type object

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_alhxo)

IgorKarpiuk:
@ZyzychO We added this to resolve issue https://github.com/Redocly/redocly-cli/issues/959. Do we need to change this behavior?

@tatomyr

tatomyr:
@IgorKarpiuk I believe we should comply with the specification. If someone wants additional restrictions, they may use a custom rule.

adamaltman:
@IgorKarpiuk that example is correct and the behavior should stay the same for components. The sample file provided for #959 though didn't include inline examples defined within paths.

```yaml
components: { it should check here }
paths: { it should not apply here }
```

IgorKarpiuk:
[...]

PR #1115, opened 2023-05-31
fix: apply regular expression only to fixed fields under components

IgorKarpiuk:
## What/Why/How?

 Apply regular expression `^[a-zA-Z0-9\.\-_]+$` only to fixed fields under components

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/1104

## Check yourself

- [x] Code is linted

## Security


packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -215,7 +215,7 @@ describe('Oas3 spec-components-invalid-map-name', () => {
     expect(results).toMatchInlineSnapshot(`Array []`);
   });
 
-  it('should report about invalid keys inside nested examples', async () => {
+  it('should does not report about invalid keys inside nested examples', async () => {
```
tatomyr, line 218:
```suggestion
  it('should not report invalid keys inside nested examples', async () => {
```

tatomyr (commented):
A tiny correction. The rest looks valid to me.

tatomyr (approved):

PR #1121, opened 2023-06-05
feat: Remote plugins

pboos:
### What
Add support for remote plugins

### Why
Because they were not supported yet and that way one can use the same custom plugin in multiple places without copy paste. 😄 

### How
Using fetch to load the plugin and temporarily storing it locally.

In `.redocly.yaml` one can specify the plugin like this now:
```
plugins:
  - 'https://raw.githubusercontent.com/your-company/redocly-plugins/main/redocly-plugin.js'
```

### Limitations
No support for protected URLs.

## Testing

Added 2 tests:
- Verify old behaviour (loading from local file) still works
- Verify new behaviour (loading from url) works by mocking node-fetch

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/core/__tests__/utils.ts
+++ b/packages/core/__tests__/utils.ts
@@ -45,7 +45,7 @@ export const yamlSerializer = {
   },
 };
 
-export function makeConfigForRuleset(
+export async function makeConfigForRuleset(
   rules: Oas3RuleSet,
   plugin?: Partial<Plugin>,
   version: string = 'oas3'
@@ -55,7 +55,7 @@ export function makeConfigForRuleset(
   Object.keys(rules).forEach((name) => {
     rulesConf[`${ruleId}/${name}`] = 'error';
   });
-  const plugins = resolvePlugins([
+  const plugins = await resolvePlugins([
     {
       ...plugin,
       id: ruleId,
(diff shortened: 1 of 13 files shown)


pboos:
Thank you for the quick response!

> * security, CLI should ask user if they want to run untrusted code from external plugin. We can save last-code hash and avoid asking it every time
Totally agree. This could be good for local runs. But this would break CI, so for usage in CI there should be a flag to disable this check. This will require storing it that hash somewhere. Is there already tools in the codebase for this? If yes, can you point me to it?

> * browser usage, we use openapi-core in the browser env in a few other projects and we try to avoid node usage in this package (while we still have some), we need to test if your PR doesn't break anything for us
Yes, that should be the way to go. I saw it in the code I worked with that this is the goal and not yet possible either in the plugin code. So for now I went without the support for that. If you have ideas how to make that possible, let me know and I could try to make it work.

> Please, allow us some time to get to this. We're now focused on preparing 1.0.0 release. We'll get to this right after. Thanks for your understanding.
Great to here there will finally be a 1.0.0 release 🥳 

As we would love to use remote plugins @getyourguide we might just build a docker image with these changes on our internal docker registry. Because we want to use the same rules in over 20 repositories. So copy pasting the plugin everywhere is not a good option for us.

lornajane:
Thanks @pboos for the pull request and the discussion. After internal discussion, we decided we're not ready to support this functionality and need to think more about how it would work in all the different scenarios that plugins can be used.

I'm therefore closing the pull request, and recommend that users fetch the plugins themselves using existing tools, and then use those plugins with Redocly CLI.

PR #1127, opened 2023-06-09
AsyncAPI linting support

### LINKED ISSUE #1126: Plans on supporting AsyncAPI?
Currently we are using Redocly CLI for bundling and linting our OpenAPI spec files.
We also providing some of our guidelines as linter rules: https://github.com/otto-de/api-guidelines/#api-linting

In the future we might want to lint our AsyncAPI spec files and also add custom linter rules.

Are you planning to support [AsyncAPI](https://www.asyncapi.com/)?

Thanks! I'd love to hear back from you.

RomanHotsiy:
## What/Why/How?

(description edited by @lornajane as the idea takes shape!)

The goal is to add linting support for AsyncAPI files to Redocly. Initially this will be partial AsyncAPI v2.6.0 support, and feedback is very welcome.

Progress:

- [x] Channel bindings
- [x] Server bindings
- [x] Operation bindings
- [x] Message bindings
- [x] Configurable rules
- [x] Documentation

Examples of AsyncAPI files can be found in the AsyncAPI project: https://github.com/asyncapi/spec/tree/master/examples

## Reference

Fixes #1126 

## Check yourself

- [ ] Code is linted

## Security


lornajane:
@tatomyr I think this is ready for your review. I see the build is failing, but I'm not sure if I should be pushing prettier fixes to files I didn't change ...

@TaylorKrusen if you want to have a look at the guide I wrote, it would be good to get some feedback on that.

packages/core/src/oas-types.ts
```diff
@@ -1,57 +1,99 @@
...
   if (typeof root !== 'object') {
     throw new Error(`Document must be JSON object, got ${typeof root}`);
   }
 
-  if (!(root.openapi || root.swagger)) {
-    throw new Error('This doesn’t look like an OpenAPI document.\n');
-  }
+  // if (!(root.openapi || root.swagger)) {
+  //   throw new Error('This doesn’t look like an OpenAPI document.\n');
+  // }
```
tatomyr, line 53: Please remove the comments.

packages/core/src/oas-types.ts
```diff
@@ -1,57 +1,99 @@
...
+  if (root.asyncapi && root.asyncapi.startsWith('2.')) {
+    return SpecVersion.Async2;
+  }
+
+  if (root.asyncapi) {
+    throw new Error(`Unsupported AsyncAPI Version: ${root.asyncapi}`);
   }
 
-  throw new Error(`Unsupported OpenAPI Version: ${root.openapi || root.swagger}`);
+  throw new Error(`Unsupported specification`);
```
tatomyr, line 84: Could you add a unit test that covers these cases?

packages/core/src/config/config-resolvers.ts
```diff
@@ -151,6 +151,9 @@ export function resolvePlugins(
         if (pluginModule.rules.oas2) {
           plugin.rules.oas2 = prefixRules(pluginModule.rules.oas2, id);
         }
+        if (pluginModule.rules.async2) {
+          plugin.rules.async2 = prefixRules(pluginModule.rules.async2, id);
+        }
       }
       if (pluginModule.preprocessors) {
         if (!pluginModule.preprocessors.oas3 && !pluginModule.preprocessors.oas2) {
```
tatomyr, line 159: Shouldn't we allow also async preprocessors? 

```suggestion
        if (!pluginModule.preprocessors.oas3 && !pluginModule.preprocessors.oas2 && !pluginModule.preprocessors.async2) {
```

tatomyr: For some reason I cannot select the line 161, but it also needs to be updated, I believe.

tatomyr: Similar on lines 144-145 and 177-178.

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Lint AsyncAPI
```
TaylorKrusen, line 2: These files are synced and published in the [Redocly CLI documentation](https://redocly.com/docs/cli/), correct? 
If so, adding SEO information here will make this guide more discoverable through search.
Recommendation:
```
---
seo:
  title: Lint AsyncAPI with Redocly CLI
  description: Unlock powerful linting capabilities for AsyncAPI documents. Use the Redocly CLI to enforce basic validation, configure rules, or even build custom plugins for AsyncAPI.
---
```

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+rules:
+  spec: error
+```
+
+The empty `extends` element instructs Redocly CLI not to use any existing
+rulesets, but to emit an error if the `spec` rule finds any problem. This rule
+checks that the document structure matches what is expected by the AsyncAPI
+specification.
+
+With this configuration file, and your AsyncAPI description file, run the linting command:
```
TaylorKrusen, line 36: Are there any public AsyncAPI examples we can direct folks to in case they want to try it out but don't have their own AsyncAPI document ready to go? 
(I found this [blog post](https://www.asyncapi.com/blog/understanding-asyncapis), but not a stand-alone repo)

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+
+```
+redocly lint asyncapi.yaml
+```
+
+The output describes any structural problems with the document, or reports that it is valid.
```
TaylorKrusen, line 42: A distinction between warnings and errors would be useful here.

"The default linting output includes both errors and warnings. Errors are structural problems that invalidate your AsyncAPI document. Warnings are typically problems that impact quality. The linter will report that your spec is valid if there are no errors (even if there are warnings)."

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## AsyncAPI rules
+
+To expand the linting checks for an AsyncAPI description, start by enabling
+some of the built-in rules. The currently-supported rules are:
+
+* `info-contact`: the `Info` section must contain a valid `Contact` field.
```
TaylorKrusen, line 50: Nit: trailing '.' on this bullet, but not on others. Either is fine as long as it's the same throughout the list.

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+We expect the list to expand over time, so keep checking back - and let us know
+if you have any requests by [opening an issue on the GitHub
+repo](https://github.com/Redocly/redocly-cli/issues).
+
+To use a rule in your own linting setup, add the rule name to the `rules`
+configuration section, and declare the severity level (either `error`, `warn`
+or `off`). Here's an example of a rules block where a missing contact section
+causes a warning, and a tag without a description triggers an error:
+
+```yaml
```
TaylorKrusen, line 66: Could we also add an example that makes use of the empty `extends: []` above?

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+* `jms`
+* `sns`
+* `solace`
+* `sqs`
+* `stomp`
+* `redis`
+* `mercure`
+
+If you're using other protocols, you can still use Redocly CLI to lint an
+AsyncAPI description, but the details of those protocol bindings aren't
```
TaylorKrusen, line 128: Do we also want users to submit requests for protocols they'd like to see coverage for? Would it be possible for users (or maybe vendors) to submit PRs adding support for new protocols?

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Lint AsyncAPI
+---
+
+# Lint AsyncAPI with Redocly CLI
```
TaylorKrusen, line 5: I know it's a bit outside the scope of this document, but I think a concise blurb about _what_ the AsyncAPI is (or what it's used for) would go a long way here. Even just a sentence or two with outbound links would be useful.

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+
+In addition to providing lint functionality for multiple OpenAPI formats, Redocly CLI also has support for AsyncAPI.
+Redocly CLI supports the following linting approaches with AsyncAPI documents:
+
+* AsyncAPI document validation, including full binding validation for [supported protocols](#supported-protocols).
+* Supported versions:
+  - AsyncAPI 2.6
```
TaylorKrusen, line 12:
```suggestion
  - [AsyncAPI 2.6](https://www.asyncapi.com/docs/reference/specification/v2.6.0)
```

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
+---
```
TaylorKrusen, line 1: Q -- will the linebreaks I'm seeing in the markdown resolve to the published docs? (just checking in case it's a copy / paste formatting issue or something similar)
 
Screenshot 2023-09-08 at 4 14 41 PM

docs/guides/lint-asyncapi.md
```diff
@@ -0,0 +1,133 @@
...
+* Supported versions:
+  - AsyncAPI 2.6
+  - earlier versions in the 2.x family may also validate successfully
+* Built-in rules for checking common standards requirements (see the [list of AsyncAPI rules](#asyncapi-rules)).
+* [Configurable rules](../rules/configurable-rules.md) so that you can build your own rules following common patterns
+* [Custom plugins](../custom-plugins/index.md) for advanced users that need additional functionality
+
+## Lint an existing AsyncAPI file
+
+Redocly CLI takes its settings from a `redocly.yaml` configuration file. Below
```
TaylorKrusen, line 20: Does the `redocly.yaml` config also [reference file location](https://redocly.com/docs/cli/configuration/#example-of-apis-configuration) in the same way? In the past I've needed to use that approach to reference multiple OpenAPI files in different locations. That might be outside the scope of this guide, but I thought I'd check.

docs/guides/lint-asyncapi.md
lornajane: No, markdown only renders new paragraphs after a blank line. I softwrapped this text so that it's easier to see the diff when things change (we don't have a standard approach for this yet)

packages/core/src/config/config-resolvers.ts
lornajane: @RomanHotsiy can you comment on this one please?

packages/cli/src/utils.ts
```diff
@@ -245,6 +245,7 @@ export function handleError(e: Error, ref: string) {
     case SyntaxError:
       return exitWithError(`Syntax error: ${e.message} ${e.stack?.split('\n\n')?.[0]}`);
     default: {
+      console.log(e);
```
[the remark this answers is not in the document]
lornajane: good catch. Done!

packages/core/src/oas-types.ts
lornajane: Removed.

docs/guides/lint-asyncapi.md
lornajane: Yes, these docs serve both as local markdown docs in the repo, and we also publish them to our main docs site. We don't have SEO entries on any of our pages, not even the main intended entry points. I've opened #1255 to capture this.

docs/guides/lint-asyncapi.md
lornajane: Yes, the file locations are the same everywhere as far as I know.

docs/guides/lint-asyncapi.md
lornajane: We do need something like this, but it doesn't belong here in a how-to guide, and I'm not adding any other AsyncAPI content at this point. If it means that only people who know what AsyncAPI is can use it now - that's probably correct. I think if we get some feedback, we'll create a whole AsyncAPI docs section with the concepts and examples and stuff (and probably split out the reference sections from this guide, they are also in the wrong place here really but they will get us started) and the do a formal announcement.

lornajane: TL;DR - really insightful and valid point, and I'm going to ignore it !!

docs/guides/lint-asyncapi.md
lornajane: Good suggestion, I added a link to the examples in the AsyncAPI project repo

docs/guides/lint-asyncapi.md
lornajane: Good catch! I am bad at this.

docs/guides/lint-asyncapi.md
lornajane: This is hard because we don't have rulesets for AsyncAPI yet, I've shipped the code with a recommended ruleset but it only includes the `spec` rule, so that will probably change, and then we'll break stuff if people have used our examples. I think it needs to stay like this for now, and when we have rulesets it would be worth adding something then.

docs/guides/lint-asyncapi.md
lornajane: I didn't call for it because it's kind of a PITA to create right now, and really it would be better for vendors to produce custom plugins, but until we fix [#1223](https://github.com/Redocly/redocly-cli/issues/1223) that's less useful. It's the three most recently added protocols that are missing I believe - and I'd really love to be encouraging more community contribution but we need to release something first...

docs/guides/lint-asyncapi.md
lornajane: This should exist somewhere, but not in a specific guide. If we don't have it, let's add it!

packages/core/src/config/config-resolvers.ts
RomanHotsiy: We don't have to.

We can always start from "do not allow" and allow later. But I don't see any problems with allowing them.

lornajane: Got it. I've added the support @tatomyr described, I can't see a reason not to add this either.

packages/core/src/oas-types.ts
lornajane: Thanks! Could you take a look at the tests I added? It looks okay locally, but should it have updated the snapshot files? It didn't .... I might need some help to get the build to pass.

docs/guides/lint-asyncapi.md
adamaltman: We do have a standard approach: https://sembr.org/

packages/core/src/oas-types.ts
tatomyr: Resolved.

packages/core/src/types/asyncapi.ts
```diff
@@ -0,0 +1,1141 @@
...
+
+const ChannelMap: NodeType = {
+  properties: {},
+  additionalProperties: 'Channel',
+};
+
+const ChannelBindings: NodeType = {
+  properties: {},
+  allowed() {
+    // allow all supported values, not all have deep linting
```
tatomyr, line 38: Looks like a TODO comment. If it is, please use the `TODO:` notation.

packages/core/src/types/asyncapi.ts
```diff
@@ -0,0 +1,1141 @@
...
+const ExternalDocs: NodeType = {
+  properties: {
+    description: { type: 'string' },
+    url: { type: 'string' },
+  },
+  required: ['url'],
+};
+
+const SecurityRequirement: NodeType = {
+  // done
```
tatomyr, line 82: There are comments like this here and there. Please remove what is not needed anymore.

packages/core/src/types/asyncapi.ts
```diff
@@ -0,0 +1,1141 @@
...
+};
+
+const Message: NodeType = {
+  properties: {
+    messageId: { type: 'string' },
+    headers: 'Schema',
+    payload: 'Schema', // TODO strictly this does not cover all cases
+    correlationId: 'CorrelationId',
+
+    schemaFormat: { type: 'string' }, // todo
```
tatomyr, line 207: Please use the `TODO:` notation and describe what needs to be done here.

packages/core/src/types/asyncapi.ts
tatomyr: There are several comments like this.

packages/core/src/types/asyncapi.ts
```diff
@@ -0,0 +1,1141 @@
...
+      case 'openIdConnect':
+        return ['type', 'openIdConnectUrl', 'description'];
+      default:
+        return ['type', 'description'];
+    }
+  },
+  extensionsPrefix: 'x-',
+};
+
+// --- Per-protocol node types
```
tatomyr, line 592: Let's move those to a different files (maybe inside `async/` folder).

packages/core/src/bundle.ts
```diff
@@ -224,13 +214,23 @@ export function mapTypeToComponent(typeName: string, version: OasMajorVersion) {
         default:
           return null;
       }
+    case SpecMajorVersion.Async2:
+      switch (typeName) {
+        case 'Schema':
+          return 'schemas';
+        // TODO + verify
```
tatomyr, line 221: What exactly needs to be verified here?

__tests__/lint/async2/redocly.yaml
```diff
@@ -0,0 +1,38 @@
...
+  streetlights-kafka:
+    root: ./streetlights-kafka.yml
+  streetlights-mqtt:
+    root: ./streetlights-mqtt.yml
+  streetlights-operation-security:
+    root: ./streetlights-operation-security.yml
+  websocket-gemini:
+    root: ./websocket-gemini.yml
+
+extends: []
```
tatomyr, line 35: We don't need to write this explicitly as Redocly CLI doesn't extend by default anymore when there is a config file. Just clarifying.

tatomyr (commented):
@lornajane  I updated the tests and renamed the `majorSpecVersion` function into `getMajorSpecVersion`. Also left a few more questions and suggestions.

P.S. Please rebase this against the main branch.

packages/core/src/types/asyncapi.ts
lornajane: I really wanted to put these in their own files, can you maybe move one to show how to do it and bring it into this structure correctly? @Roman didn't have time to help on this and I wasn't sure ...

packages/core/src/types/asyncapi.ts
lornajane: It's not a TODO, it's an explanation. We support all the possible protocol values here, but not all of them have linting for that protocol. This is so that we won't produce errors if we get an API description that we can't verify in detail, but we aren't guaranteeing detailed support for them all.

Does this make sense, and do we need to do anything else?

packages/core/src/bundle.ts
RomanHotsiy: I wanted to double check if async api has schemas, I believe it does so it can be removed.

packages/core/src/types/asyncapi.ts
RomanHotsiy: Let's keep it in this file for now. It can be always moved.

packages/core/src/bundle.ts
lornajane: Thanks! Removed

packages/core/src/types/asyncapi.ts
lornajane: Expanded this and a couple of other comments

packages/core/src/types/asyncapi.ts
lornajane: Oops, left over from the draft version! Thanks for spotting, removed them.

__tests__/lint/async2/redocly.yaml
lornajane: Got it. I will remove it - and I it was in the guide as well, so this is a really good tip.

packages/core/src/types/asyncapi.ts
tatomyr: Oh, thanks for the explanation! I just wanted to be sure that we're using TODO:s consistently.

tatomyr (approved):
I'm approving the MR, but as it now contains my changes too, I'd like someone else from the team to approve it as well. 🙂

SmoliyY (approved):

.changeset/quick-emus-punch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Added support for linting AsyncAPI v2 files, so that a wider range of API descriptions can use the Redocly linting workflow
```
tatomyr, line 6:
```suggestion
Added support for linting AsyncAPI v2 files, so that a wider range of API descriptions can use the Redocly linting workflow.
```

PR #1130, opened 2023-06-13
Docs: fix example of command to combine files

aviskase:
## What/Why/How?
The example command uses `bundle` in the documentation about combining OpenAPI files. I guess the correct one should be `join`.

## Check yourself

- [ ] Code is linted

## Security


tatomyr (approved):
Thank you for spotting this!

PR #1131, opened 2023-06-13
Docs for migrating from Spectral to Redocly CLI

lornajane:
Add a migration guide for users migrating from Spectral. Includes a best-effort list of rules, to help with the translation process. This probably will need lots of updates as we get feedback from migrating users.

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/docs/guides/migrate-from-spectral.md
+++ b/docs/guides/migrate-from-spectral.md
@@ -59,3 +59,69 @@ rules:
 
 It is also possible to configure additional rules for specific APIs using the [APIs object](../configuration.md#apis-object) to set per-API rules (or exceptions!).
 
+
+### Redocly rules and Spectral equivalents
+
+Included here is an attempt to map the simliar-but-not-identical naming of rules between the tools. If you spot anything that needs adding or updating, please [tell us](https://github.com/redocly/redocly-cli/issues/new)?
+
+
+| Spectral rules                         | Redocly rules                                 |
+|----------------------------------------|-----------------------------------------------|
+| `duplicated-entry-in-enum`             |                                               |
+| `info-contact`                         | `info-contact`                                |
+| `info-description`                     |                                               |
+| `info-license`                         | `info-license`                                |
+| `license-url`                          | `info-license-url`                            |
+| `no-$ref-siblings`                     |                                               |
+| `no-eval-in-markdown`                  |                                               |
+| `no-script-tags-in-markdown`           |                                               |
+| `oas3-api-servers`                     | `no-empty-servers`                            |
+| `oas3-examples-value-or-externalValue` | `no-example-value-and-externalValue`          |
+| `oas3-operation-security-defined`      | `security-defined`                            |
+| `oas3-parameter-description`           | `parameter-description`                       |
+| `oas3-schema`                          | `spec`                                        |
+| `oas3-server-not-example.com`          | `no-server-example.com`                       |
+| `oas3-server-trailing-slash`           | `no-server-trailing-slash`                    |
+| `oas3-unused-component`                | `no-unused-components`                        |
+| `oas3-valid-media-example`             | `no-invalid-media-type-examples`              |
+| `oas3-valid-schema-example`            | `no-invalid-schema-examples`                  |
+| `openapi-tags`                         |                                               |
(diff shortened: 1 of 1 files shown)


docs/guides/migrate-from-spectral.md
```diff
@@ -0,0 +1,127 @@
...
+
+rules:
+  security-defined: off
+  rule/naming:
+    subject:
+      type: Operation
+      property: summary
+    assertions:
+      pattern: /test/    
+    message: "Operation must not include the word test"
```
RomanHotsiy, line 57: I think this rule checks the opposite (if the operation summary includes the word test).

lornajane: Fixed it. Good catch :)

RomanHotsiy (commented):
I think this is missing the important point about the custom rules, this is how most of the real spectral rulesets are composed.

It would be hard to compose a migration guide but I think we should mention those and link to the corresponding docs.

docs/guides/migrate-from-spectral.md
```diff
@@ -0,0 +1,131 @@
+# Migrate to Redocly from Spectral
+
+[Spectral](https://stoplight.io/open-source/spectral) offers similar linting capabilities to Redocly CLI and the rest of the Redocly tools. This guide lays out the differences so you can switch tools if you want to.
```
adamaltman, line 3:
```suggestion
[Spectral](https://stoplight.io/open-source/spectral) offers similar linting capabilities to Redocly CLI. This guide lays out the differences so you can switch tools if you want to.
```

I wonder if that "and the rest of the Redocly tools" adds anything? Also, I'm not sure what other capability it would be similar to besides the CLI lint command?

docs/guides/migrate-from-spectral.md
```diff
@@ -0,0 +1,131 @@
...
+### Redocly rules and Spectral equivalents
+
+Included here is an attempt to map the simliar-but-not-identical naming of rules between the tools. If you spot anything that needs adding or updating, please [tell us](https://github.com/redocly/redocly-cli/issues/new)?
+
+
+| Spectral rules                         | Redocly rules                                 |
+|----------------------------------------|-----------------------------------------------|
+| `duplicated-entry-in-enum`             |                                               |
+| `info-contact`                         | `info-contact`                                |
+| `info-description`                     |                                               |
```
adamaltman, line 72: I wonder if it would be helpful to include configurable rule equivalents, like:

```yaml
rules:
  rule/info-description:
    subject:
      type: Info
      property: description
    assertions:
      defined: true
```

no-eval:

```yaml
rules:
  rule/no-eval:
    subject:
      type: any
      property: description
    assertions:
      notPattern: 'eval('
```
no-script
```yaml
rules:
  rule/no-script-tag:
    subject:
      type: any
      property: description
    assertions:
      notPattern: '<script'
```

openapi-tags:

```yaml
rules:
  rule/openapi-tags:
    subject:
      type: Tag # or maybe TagList ?
    assertions:
      defined: true
```

operation-tags

```yaml
rules:
  rule/operation-tags:
    subject:
      type: Operation
      property: tags
    assertions:
      defined: true
      maxLength: 1 # I guess we already have `operation-singular-tag`
```

It looks like there are 7 of them in total that don't exist in Redocly rules? Or perhaps open an issue and link to it where people can 👍 if they want a built-in equivalent rule where a simple configurable rule isn't possible? I posted a few up above (untested).

docs/guides/migrate-from-spectral.md
lornajane: The linting capabilities in Redocly CLI, Workflows, and whatever else are similar to what Spectral offers. I was trying to make the point that it is not just the CLI, but all the Redocly tools, if people adopt all of them.

docs/guides/migrate-from-spectral.md
lornajane: I love this and will move it out into another issue so we can publish this basic post soon, and work on expanding it!

lornajane:
Followup issue is #1177 . I'll also add signposting to the configurable rules to the document. Thank you both for picking up on this omission!

RomanHotsiy (approved):
🎉 

Thanks for "Configurable and extensible rules" section.

PR #1133, opened 2023-06-15
feat: updated cli config accroding to portal config

### LINKED ISSUE #1076: Update Redocly config validation
Redocly CLI should validate the universal Redocly config correctly. We have to update the type tree to be in sync with JSONSchema we use to validate the config in Portal.

Skavengerr:
## What/Why/How?
Fixes Issue #1076 by updating the Redocly CLI configuration, which was copied from the portal config.

## Reference

Fixes #1076

## Check yourself

- [x] Code is linted

## Security


packages/core/src/types/redocly-yaml.ts
```diff
@@ -151,12 +152,338 @@ const RootConfigStyleguide: NodeType = {
...
+    off: { type: 'boolean', default: false },
+    position: { type: 'string', enum: ['first', 'last', 'replace', 'off'], default: 'first' },
+    strictExamples: { type: 'boolean', default: false },
+    errorIfForcedExampleNotFound: { type: 'boolean', default: false },
+    description: { type: 'string' },
+  },
+};
+
+
+export const ConfigRoot = {
```
tatomyr, line 482: Why removed the type?

```suggestion
export const ConfigRoot: NodeType = {
```

packages/core/src/types/redocly-yaml.ts
```diff
@@ -151,12 +152,338 @@ const RootConfigStyleguide: NodeType = {
...
   properties: {
     organization: { type: 'string' },
-    apis: 'ConfigApis',
     ...RootConfigStyleguide.properties,
-    theme: 'ConfigRootTheme',
+    theme: { type: 'object', default: {} },
```
tatomyr, line 486: I'd prefer extending 'ConfigRootTheme' type if it's possible. BTW, do we still use it somewhere?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -172,6 +499,17 @@ const ConfigRoot: NodeType = {
         type: 'string',
       },
     },
+    redirects: { type: 'object', additionalProperties: redirectConfigSchema, default: {} },
+    licenseKey: { type: 'string' },
+    apis: apiConfigSchema,
```
tatomyr, line 504: Why not extending the 'ConfigApis' type?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -54,7 +55,7 @@ const builtInRulesList = [
   'scalar-property-missing-example',
   'spec-components-invalid-map-name',
   'required-string-property-missing-min-length',
-  'spec-strict-refs',
+  'spec-ref-validation',
```
tatomyr, line 58: We renamed this in the previous release. Why it's reverted back?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -141,7 +142,7 @@ const ConfigStyleguide: NodeType = {
   },
 };
 
-const RootConfigStyleguide: NodeType = {
+export const RootConfigStyleguide: any = {
```
tatomyr, line 145: Could you avoid  the `any` type?

tatomyr (changes requested):

packages/core/src/types/redocly-yaml.ts
tatomyr: I still see the wrong name. It should be `'spec-strict-refs'` instead.

__tests__/lint/spec-strict-refs/snapshot.js
```diff
@@ -1,46 +1,21 @@
...
-18 | type: object
-19 | properties:
-20 |   $ref: './props.yaml'
-   |   ^^^^
-21 |   name:
-22 |     type: string
-
-Error was generated by the spec-strict-refs rule.
-
+exports[`E2E lint spec-strict-refs 1`] = `
```
tatomyr, line 3: Why this has been changed?

tatomyr (changes requested):

__tests__/lint/spec-strict-refs/snapshot.js
IgorKarpiuk: @tatomyr `E2E lint spec-ref-validation 1` from old snapshot. and this one  just changed location from [here](https://github.com/Redocly/redocly-cli/pull/1133/files?diff=split&w=0#diff-3fdf21f2551fc727676e37674367dfa62ed3434527db5aecfe04a9daf270a696L42)

packages/core/src/types/redocly-yaml.ts
```diff
@@ -141,40 +142,6 @@ const ConfigStyleguide: NodeType = {
   },
 };
 
-const RootConfigStyleguide: NodeType = {
```
tatomyr, line 144: Can we get this and `ConfigRoot` back in **redocly-yaml.ts** and move [this line](https://github.com/Redocly/redocly-cli/pull/1133/files?diff=split&w=1#diff-2528117ea3cbe25dfbfd0583c9b11ba7baa7ef347184dfba32a73ec486aa1a5eR186) to the **config-root.ts** file instead?

packages/core/src/types/config-root.ts
```diff
@@ -0,0 +1,376 @@
+import { NodeType } from '.';
```
tatomyr, line 1: I suggest renaming this file into **external-config-schemas.ts** or something like that.

tatomyr (commented):
Please keep the existing **redocly-yaml.ts** with the minimal changes and move the extra config schemas in the separate file.

PR #1134, opened 2023-06-16
Feature: Add `component-name-unique` rule

pboos:
## What/Why/How?
Adds the following rule: `component-name-unique`

This will allow checking for unique component names for
- parameter
- requestBody
- response
- schema

This helps avoid problems with generators. For example if combining multiple specs into one where the same component names are used.

## Testing
- [x] Add tests for each rule.

## Check yourself

- [x] Code is linted

## Security


tatomyr:
Hi @pboos,

Thanks for your contribution! I'm curious if this could be one rule instead of four? I'm thinking of something like this:

```yaml
unique-names:
  parameter: error
  requestBody: warn
```

What do you think? 
cc @RomanHotsiy, @lornajane

pboos:
> Hi @pboos,
> 
> Thanks for your contribution! I'm curious if this could be one rule instead of four? I'm thinking of something like this:
> 
> ```yaml
> unique-names:
>   parameter: error
>   requestBody: warn
> ```
> 
> What do you think? cc @RomanHotsiy, @lornajane

@tatomyr that definitely would be possible. I originally had it that way locally but thought it might be nicer to separate them. Wasn't aware of this type of configuration being possible :). I will see if I can merge them and do that configuration. Should not be too hard to do (once I figure out how that works).

RomanHotsiy (commented):
Hey @pboos!

Amazing work! Thanks for your contribution.

I would like to suggest some alternatives to discuss because I'm not sure if the names of the rules are clear enough. For example `parameter-name-unique` is a bit ambiguous:
- it can mean "unique within operation"
- it can mean unique globally
- it can mean unique component name (your case).

I have a few ideas on how to potentially improve it:

- rename rules to somehow add `component` into them, not sure how to name them specifically though
- create a single rule: `component-name-unique` which can accept type of components to enforce but should enforce all the components to be unique by default (isn't it what is usually needed?)
- we can create a assertion so someone can author custom rules using `unique` assertion, see more here: https://redocly.com/docs/cli/rules/configurable-rules/


What do you think?

pboos:
> I have a few ideas on how to potentially improve it:
> 
> * rename rules to somehow add `component` into them, not sure how to name them specifically though
> * create a single rule: `component-name-unique` which can accept type of components to enforce but should enforce all the components to be unique by default (isn't it what is usually needed?)
> * we can create a assertion so someone can author custom rules using `unique` assertion, see more here: https://redocly.com/docs/cli/rules/configurable-rules/
> 
> What do you think?

@RomanHotsiy thank you for the feedback! Very much appreciated.

I have updated the code now to just be one rule `component-name-unique` that allows configuration the severity for each component type. The only side effect is that the test file now is very huge as I test all the separate ones plus one additional combined case. But this might be acceptable.

Let me know what you think of the new code.

packages/core/src/rules/common/component-name-unique.ts
```diff
@@ -0,0 +1,158 @@
...
+      leave(root: Oas3Definition | Oas2Definition, ctx: UserContext) {
+        components.forEach((value, key, _) => {
+          if (value.size > 1) {
+            const component = getComponentFromKey(key);
+            const definitions = Array.from(value)
+              .map((v) => `- ${v}`)
+              .join('\n');
+
+            const problem: Problem = {
+              message: `${component.typeName} '${component.componentName}' is not unique. It is defined at:\n${definitions}`,
```
tatomyr, line 107: How do you feel about this message?

```suggestion
              message: `Component '${component.typeName}/${component.componentName}' is not unique. It is defined at:\n${definitions}`,
```

packages/core/src/rules/common/component-name-unique.ts
```diff
@@ -0,0 +1,158 @@
...
+  Oas3Definition,
+  Oas3Parameter,
+  Oas3RequestBody,
+  Oas3Response,
+  Oas3Schema,
+  OasRef,
+} from '../../typings/openapi';
+import { Oas2Definition } from '../../typings/swagger';
+
+const TYPE_NAME_SCHEMA = 'Schema';
```
tatomyr, line 13: Why do you need this mapping? For me it looks like it only adds additional complexity 🙂

packages/core/src/rules/common/component-name-unique.ts
```diff
@@ -0,0 +1,158 @@
...
+    const [typeName, componentName] = key.split('/');
+    return { typeName, componentName };
+  }
+
+  function addComponentFromAbsoluteLocation(typeName: string, absoluteLocation: string): void {
+    const componentName = getComponentNameFromAbsoluteLocation(absoluteLocation);
+    addFoundComponent(typeName, componentName, absoluteLocation);
+  }
+
+  const rule: Oas3Visitor = {
```
tatomyr, line 82: If this rule is only applicable to OAS3, please move it into the `rules/oas3/` folder.

packages/core/src/rules/common/component-name-unique.ts
```diff
@@ -0,0 +1,158 @@
...
+
+          addComponentFromAbsoluteLocation(
+            typeName,
+            resolvedRef.location.absolutePointer.toString()
+          );
+        }
+      },
+    },
+    Root: {
+      leave(root: Oas3Definition | Oas2Definition, ctx: UserContext) {
```
tatomyr, line 98: I noticed inconsistency with [this line](https://github.com/Redocly/redocly-cli/pull/1134/files#diff-e086c7501a57ca548a3dfb56a69a297d4733596de3b0da7cf8ebf6080f25bb53R82). Shouldn't it be like this: 
```suggestion
      leave(root: Oas3Definition, ctx: UserContext) {
```

tatomyr (commented):
Thanks for the update. Left a couple of comments.

packages/core/src/rules/common/component-name-unique.ts
```diff
@@ -0,0 +1,158 @@
...
+            const component = getComponentFromKey(key);
+            const definitions = Array.from(value)
+              .map((v) => `- ${v}`)
+              .join('\n');
+
+            const problem: Problem = {
+              message: `${component.typeName} '${component.componentName}' is not unique. It is defined at:\n${definitions}`,
+            };
+
+            const openName = getOptionNameForTypeName(component.typeName);
```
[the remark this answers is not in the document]
pboos: Should have been `optionName` 🤦 . Or even better `optionComponentName`. Renamed it now.

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,113 @@
...
+
+An example configuration:
+
+```yaml
+rules:
+  schema-name-unique:
+    schema: error
+    parameter: off
+    response: warn
+    requestBody: warn
```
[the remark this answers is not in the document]
pboos: No worries. Makes sense. Updated in code (all the places)

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,113 @@
...
+| schema      |string| Possible values: `off`, `warn`, `error`. Default: not set. |
+| parameter   |string| Possible values: `off`, `warn`, `error`. Default: not set. |
+| response    |string| Possible values: `off`, `warn`, `error`. Default: not set. |
+| requestBody |string| Possible values: `off`, `warn`, `error`. Default: not set. |
+
+An example configuration:
+
+```yaml
+rules:
+  schema-name-unique:
```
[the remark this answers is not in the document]
pboos: Ups, this one clearly got forgotten. Updated code now (also the other one i forgot as well).

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Examples
+
+
+Given this configuration:
+
+```yaml
+rules:
+  schema-name-unique: error
```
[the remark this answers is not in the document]
pboos: done

packages/core/src/rules/common/component-name-unique.ts
pboos: Not 100% sure it is right with the `/`. As normally it is `parameters/ParameterOne` but with the code currently it would be `Parameter/ParameterOne`.

What do you think about `Parameter component 'ParameterOne' is no unique...` ? Have pushed it with this change for now.

The other way would be to use the mapping I have to convert `Parameter` into `parameters` and then combining with `/` will work. But it becomes again a bit more complex. WDYT?

pboos: Actually went with `parameters/ParameterOne` now (pushed) 😄. The output is nicer for the user.

packages/core/src/rules/common/component-name-unique.ts
pboos: Moved to oas3 folder now.

packages/core/src/rules/common/component-name-unique.ts
pboos: Done 👍

pboos:
> Thanks for the update. Left a couple of comments.

@tatomyr Thank you for the comments. I have updated the PR accordingly and would be glad for a further review.

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,113 @@
...
+          type: string
+```
+
+## Relates rules
+
+- [no-unused-components](./no-unused-components.md)
+
+## Resources
+
+- [Rule source](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/component-name-unique.ts)
```
pboos, line 113: @tatomyr the `lintcheck` doesn't like this due to the link not yet existing [[link to failure](https://github.com/Redocly/redocly-cli/actions/runs/5347202202)]. But it will exist once this PR gets merged. How was this done for the other rules that link to their own file?
Only other ways I see:
1. Link it to latest commit in the PR (not so nice, but nicer than not having it)
2. Remove the link (not that nice, as it is good to have it same as the others)

Okay with whatever is the preferred way from you.

For now I will remove it just so that the next run will pass...

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,109 @@
+# component-name-unique
+
+Verifies component names are unique.
+
+|OAS|Compatibility|
+|---|---|
+|2.0|✅|
```
adamaltman, line 7: Is it compatible with 2.0? It looks from the file organization that is is for 3.0, and 3.1?

docs/rules/component-name-unique.md
```diff
@@ -0,0 +1,109 @@
...
+components:
+  schemas:
+    FooSchema:
+      type: object
+      properties:
+        otherField:
+          type: string
+```
+
+### Example of a **correct** schema files
```
adamaltman, line 83:
```suggestion
### Example of **correct** schema files
```

adamaltman (commented):
Thanks @pboos for your contribution.

🎉

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -0,0 +1,157 @@
...
+    typeNames.push(TYPE_NAME_PARAMETER);
+  }
+  if (options.responses !== 'off') {
+    typeNames.push(TYPE_NAME_RESPONSE);
+  }
+  if (options.requestBodies !== 'off') {
+    typeNames.push(TYPE_NAME_REQUEST_BODY);
+  }
+
+  function getOptionComponentNameForTypeName(typeName: string): string | null {
```
RomanHotsiy, line 41: Let's move function definitions after the return statement so they don't interrupt the code flow.

docs/rules/component-name-unique.md
pboos: Removed 2.0 compatibility now in the documentation 👍

packages/core/src/rules/oas3/component-name-unique.ts
pboos: Good idea. I moved them after the return statement. And those that don't depend on the `components` map I moved outside of the function to the end of the file.

pboos:
@RomanHotsiy Thank you for the comment. What else is needed to get this merged? :) As I can't runt he pipelines nor merge it myself :).

RomanHotsiy:
@tatomyr let's review and merge it? It looks good to me 👍

packages/core/src/rules/common/component-name-unique.ts
tatomyr: I wonder if [mapTypeToComponent](https://github.com/Redocly/redocly-cli/blob/0ce9cab59a72fec7ae43e103a79207e4fd6eef67/packages/core/src/bundle.ts#L191) could help with the mapping 🤔

docs/rules/component-name-unique.md
tatomyr: Don't worry, we'll add the link ourselves! 

Just in case, we usually create two separate merge requests - one for the rule itself, and one for the docs (I've recently updated our [Contributing guide](https://github.com/Redocly/redocly-cli/blob/main/.github/CONTRIBUTING.md#built-in-rules-changes) regarding the flow).

tatomyr (approved):
I like the changes, thank you @pboos for your contribution! 
I'm going to merge this.

Issue #1044, opened 2023-04-05
Improve product information and opt-out


lornajane:
**Is your feature request related to a problem? Please describe.**

We have a little anonymised data about how the command is used, but want to add more detail, for example whether a config file is used. We should also check what version a user has and notify if they are out of date.

Users should be able to opt-out of sending us this data, for example by setting an environment variable. This should be clearly documented (for example in the main README).

**Describe the solution you'd like**

Commands and switches (but no user-specific information)

**Describe alternatives you've considered**

Do nothing! Always an option, but currently we can't do much with the data and there's no opt-out.

RomanHotsiy:
@tatomyr DM me if you need to discuss it in more details 🙏

PR #1136, opened 2023-06-20
chore: refactor CLI commands (use command wrapper)

tatomyr:
## What/Why/How?

- Refactored the CLI to use a command wrapper. The wrapper helps organise code by extracting repetitive things (loading the config, sending telemetry) from the command handlers themselves.
- Introduced telemetry.
- Implemented opt-outing through the env variable `REDOCLY_TELEMETRY=off` and Redocly config: `telemetry: off`.

Naming is under consideration. Feel free to make a suggestion.

> TODO: merge related docs after releasing this changes: https://github.com/Redocly/redocly-cli/pull/1139

## Reference

Closes https://github.com/Redocly/redocly-cli/issues/1044 

## Check yourself

- [x] Code is linted

## Security


packages/core/src/redocly/registry-api.ts
```diff
@@ -8,6 +8,7 @@ import type {
 import type { AccessTokens, Region } from '../config/types';
 import { DEFAULT_REGION, DOMAINS } from '../config/config';
 import { isNotEmptyObject } from '../utils';
+
```
SmoliyY, line 11:
```suggestion
```

packages/cli/src/utils.ts
```diff
@@ -227,30 +224,31 @@ export function pluralize(label: string, num: number) {
 
 export function handleError(e: Error, ref: string) {
   switch (e.constructor) {
+    case HandledError: {
+      throw e;
```
SmoliyY, line 228: What is that for? Did not get it) And why we do not wrap it in `exitWithError` ?

packages/cli/src/wrapper.ts
```diff
@@ -0,0 +1,137 @@
...
+import { lintConfigCallback } from './commands/lint';
+import type { CommandOptions } from './types';
+
+export function commandWrapper<T extends CommandOptions>(
+  commandHandler: (argv: T, config: Config, version: string) => Promise<void>
+) {
+  return async (argv: Arguments<T>) => {
+    let code: 0 | 1 = 0;
+    let hasConfig;
+    let telemetry;
```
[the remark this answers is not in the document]
tatomyr: It adds some complexity as we will have to define this default two times.

packages/cli/src/utils.ts
tatomyr: When we face a HandledError instance we know that it has been already handled by someone else. So we simply re-throw the error forward.

tatomyr: For other cases we want to exit with some pretty error message.

packages/cli/src/utils.ts
```diff
@@ -227,30 +224,31 @@ export function pluralize(label: string, num: number) {
...
     case SyntaxError:
       return exitWithError(`Syntax error: ${e.message} ${e.stack?.split('\n\n')?.[0]}`);
     default: {
-      process.stderr.write(
-        red(`Something went wrong when processing ${ref}:\n\n  - ${e.message}.\n\n`)
-      );
-      process.exit(1);
-      throw e;
+      console.error(e.stack);
+      exitWithError(`Something went wrong when processing ${ref}:\n\n  - ${e.message}.`);
```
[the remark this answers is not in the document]
roman-sainchuk: this is the last `case` so is not necessary to return something unless result of `exitWithError` might be used by caller

packages/cli/src/utils.ts
```diff
@@ -373,7 +371,7 @@ export function printUnusedWarnings(config: StyleguideConfig) {
 
 export function exitWithError(message: string) {
```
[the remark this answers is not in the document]
tatomyr: I had thought about this, but I cannot come up with a better name. Moreover, we kind of 'preparing' to exit in this function. Could you maybe suggest a better alternative?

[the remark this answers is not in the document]
SmoliyY: In case if we do not exit a I would suggest something common - like `processErrorMessage`.  Or `printMessageWithError`.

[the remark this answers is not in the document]
IgorKarpiuk: Actually we `exit` by throwing an error 😅 . 
The name `exitWithError` looks okay for me.
I can suggest the name `terminateWithError` or `terminateWithErrorMessage`

packages/cli/src/wrapper.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+      const config: Config = (await loadConfigAndHandleErrors({
+        configPath: argv.config,
+        customExtends: argv.extends as string[] | undefined,
+        region: argv.region as Region,
+        files: argv.file as string[] | undefined,
+        processRawConfig: lintConfigCallback(argv as T & Record<string, undefined>, version),
+      })) as Config;
+      telemetry = config.telemetry;
+      hasConfig = !!(config.styleguide && !config.styleguide.recommendedFallback);
```
IgorKarpiuk, line 27: Why do we need to check the `styleguide` field here? Can it be an empty object if there is no configuration?

SmoliyY (approved):

packages/cli/src/__tests__/commands/join.test.ts
```diff
@@ -12,7 +13,7 @@ describe('handleJoin fails', () => {
   colloreteYellowMock.mockImplementation((string: string) => string);
 
   it('should call exitWithError because only one entrypoint', async () => {
-    await handleJoin({ apis: ['first.yaml'] }, 'cli-version');
+    await handleJoin({ apis: ['first.yaml'] }, {} as any, 'cli-version');
```
RomanHotsiy, line 16: Why `as any`?

RomanHotsiy: I see, this is tests

RomanHotsiy (approved):
LGTM 👍

IgorKarpiuk (approved):

PR #1137, opened 2023-06-21
feat: update push command

IgorKarpiuk:
## What/Why/How?

- Add new options `--organization` and `--destination`.
- Mark `maybeApiOrDestination` and `maybeDestination` as deprecated
- Rename `job-id` option and mark `batch-id` as deprecated

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/1032

## Screenshots (optional)
Знімок екрана 2023-06-26 о 17 18 49

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/__tests__/commands/push.test.ts
+++ b/packages/cli/src/__tests__/commands/push.test.ts
@@ -32,7 +32,7 @@ describe('push', () => {
         destination: '@org/my-api@1.0.0',
         branchName: 'test',
         public: true,
-        'batch-id': '123',
+        'job-id': '123',
         'batch-size': 2,
       },
       ConfigFixture as any
@@ -54,7 +54,7 @@ describe('push', () => {
     });
   });
 
-  it('fails if batchId value is an empty string', async () => {
+  it('fails if jobId value is an empty string', async () => {
     await handlePush(
       {
         upsert: true,
@@ -62,7 +62,7 @@ describe('push', () => {
         destination: '@org/my-api@1.0.0',
         branchName: 'test',
         public: true,
-        'batch-id': ' ',
+        'job-id': ' ',
         'batch-size': 2,
       },
       ConfigFixture as any
@@ -79,7 +79,7 @@ describe('push', () => {
         destination: '@org/my-api@1.0.0',
         branchName: 'test',
         public: true,
-        'batch-id': '123',
+        'job-id': '123',
         'batch-size': 1,
       },
       ConfigFixture as any
@@ -89,14 +89,9 @@ describe('push', () => {
   });
 
   it('push with --files', async () => {
-    // (loadConfigAndHandleErrors as jest.Mock).mockImplementation(({ files }) => {
-    //   return { ...ConfigFixture, files };
-    // });
-
     const mockConfig = { ...ConfigFixture, files: ['./resouces/1.md', './resouces/2.md'] } as any;
 
-    //@ts-ignore
-    fs.statSync.mockImplementation(() => {
+    (fs.statSync as jest.Mock).mockImplementation(() => {
       return { isDirectory: () => false, size: 10 };
     });
 
@@ -131,7 +126,7 @@ describe('push', () => {
         destination: 'test@v1',
         branchName: 'test',
         public: true,
-        'batch-id': '123',
+        'job-id': '123',
         'batch-size': 2,
       },
       ConfigFixture as any
@@ -139,9 +134,7 @@ describe('push', () => {
 
     expect(exitWithError).toBeCalledTimes(1);
     expect(exitWithError).toBeCalledWith(
-      `No organization provided, please use the right format: ${yellow(
-        '<@organization-id/api-name@api-version>'
-      )} or specify the 'organization' field in the config file.`
+      `No organization provided, please use --organization option or specify the 'organization' field in the config file.`
     );
   });
 
@@ -154,7 +147,7 @@ describe('push', () => {
         destination: 'my-api@1.0.0',
         branchName: 'test',
         public: true,
-        'batch-id': '123',
+        'job-id': '123',
         'batch-size': 2,
       },
       mockConfig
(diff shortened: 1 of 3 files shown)


packages/cli/src/__tests__/commands/push.test.ts
```diff
@@ -157,7 +155,7 @@ describe('push', () => {
     });
   });
 
-  it('push should work if destination not provided and api in config is provided', async () => {
+  it('push should fail if destination not provided and api in config is provided', async () => {
```
tatomyr, line 158: Please split it into 2 tests.

packages/cli/src/__tests__/commands/push.test.ts
```diff
@@ -170,22 +168,25 @@ describe('push', () => {
...
 
-    expect(redoclyClient.registryApi.pushApi).toBeCalledTimes(1);
+    expect(exitWithError).toBeCalledTimes(1);
+    expect(exitWithError).toHaveBeenLastCalledWith(
+        `No destination provided, please use --destination option to provide destination.`
+    );
   });
 
-  it('push should fail if destination and apis not provided', async () => {
+  it('push should fail if apis not provided', async () => {
```
tatomyr, line 181: Why removed the wording?

packages/cli/src/commands/push.ts
```diff
@@ -342,12 +352,32 @@ export const transformPush =
...
+      );
+    } else if (maybeApiOrDestination && DESTINATION_REGEX.test(maybeApiOrDestination)) {
+      process.stderr.write(
+          yellow(
+              'Deprecation warning: Do not use the first parameter as a destination. Please use a separate --destination and --organization options instead.\n\n'
+          )
+      );
+      api = undefined;
+      destination = maybeApiOrDestination;
+    }
```
tatomyr, line 377:
```suggestion
    let api, destination;
    if (maybeDestination) {
      process.stderr.write(
          yellow(
              'Deprecation warning: Do not use the second parameter as a destination. Please use a separate --destination and --organization instead.\n\n'
          )
      );
      api = maybeApiOrDestination;
      destination = maybeDestination;
    } else if (maybeApiOrDestination && DESTINATION_REGEX.test(maybeApiOrDestination)) {
      process.stderr.write(
          yellow(
              'Deprecation warning: Do not use the first parameter as a destination. Please use a separate --destination and --organization options instead.\n\n'
          )
      );
      destination = maybeApiOrDestination;
    }
```

packages/cli/src/commands/push.ts
```diff
@@ -35,12 +35,13 @@ type PushArgs = {
   destination?: string;
   branchName?: string;
   upsert?: boolean;
-  'batch-id'?: string;
+  'job-id'?: string;
```
tatomyr, line 38: I'm not certain about renaming. Do we want to keep the old name for backward compatibility?

RomanHotsiy: yes.

IgorKarpiuk: @RomanHotsiy should we also add `job-size` instead `batch-size` ?

RomanHotsiy: Let's keep batch size I think.

packages/cli/src/commands/push.ts
```diff
@@ -84,9 +87,15 @@ export async function handlePush(argv: PushOptions, config: Config): Promise<voi
     );
   }
 
-  if (batchId && !batchId.trim()) {
+  if (!name && api) {
```
tatomyr, line 90: Could you left a comment on why you check this case?

lornajane:
The usage instruction `redocly push --help` isn't correct. It still lists the "maybe" parameters - these should be removed. There should be a single positional. I think the deprecated `--batch-id` should also be hidden (I will include it in the migration guides for anyone coming from an earlier version, which means it'll show up in the search results on the docs if people are looking for it).

lornajane:
Please capture a followup issue against the Workflows project to update the push instructions - the old commands work, but the output warnings so updating the instructions will give a better user experience (and probably fewer support tickets).

packages/cli/src/commands/push.ts
```diff
@@ -34,12 +34,13 @@ export type PushOptions = {
   destination?: string;
   branchName?: string;
   upsert?: boolean;
-  'batch-id'?: string;
+  'job-id'?: string;
   'batch-size'?: number;
   region?: Region;
   'skip-decorator'?: string[];
```
lornajane, line 40: I know we didn't introduce this in this change - but why is the skip-decorator option here? I assume the push command doesn't do anything other than push ... or does it also lint/bundle/decorate/something else ?

IgorKarpiuk: Because we bundle a definition before push it

lornajane: Ah, right. And bundle still also decorates as a side effect, is that right? Do we need `skip-rule`, `skip-preprocessor` and the other options also?

IgorKarpiuk: We definitely do not need `skip-rules`, but I am not sure regarding `skip-preproccesors` because  we recommend avoid preprocessing https://redocly.com/docs/cli/resources/custom-plugins/#preprocessors

tatomyr:
The related changes in Workflows: https://github.com/Redocly/workflows/pull/2644

@lornajane FYI

IgorKarpiuk:
> The usage instruction `redocly push --help` isn't correct. It still lists the "maybe" parameters - these should be removed. There should be a single positional. I think the deprecated `--batch-id` should also be hidden (I will include it in the migration guides for anyone coming from an earlier version, which means it'll show up in the search results on the docs if people are looking for it).

Updated usage instruction  for push command please check

tatomyr (approved):

lornajane:
We are missing help text for the branch, organization, destination, and upsert parameters. Is it simplest for me to just add those quickly?

tatomyr:
@lornajane please go ahead!

lornajane (approved):
Thank you, this is so much better :) Everything works as expected as far as I can tell.

lornajane:
(hang on, I'll just fix that build failure, sorry!)

PR #1143, opened 2023-06-27
chore: improve data obfuscation

tatomyr:
## What/Why/How?

Improving data obfuscation (better files/folder/organizations/apis handling).
Detecting CI/Docker environments.

## Testing

Tested on lab6

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,6 @@ RUN npm cache clean --force && rm -rf /build
 
 WORKDIR /spec
 
-ENTRYPOINT [ "openapi" ]
+ENTRYPOINT [ "redocly" ]
+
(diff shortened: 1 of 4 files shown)


lornajane:
Suggested additional obfuscation
* remove the full data of the `-o`/`--output` parameter, but keep the file extension (I'm curious about the yaml/json split for bundle, the docs are html but it's still useful)
* change the `apis` data - we're using arrays with `***.yaml` but showing where a config alias is used. It would be better to have `file-yaml`, `file-json`, `api-name`, `api-name@version` and so on - literally these values, masking what the user actually supplied
*

IgorKarpiuk (approved):
LGTM

PR #1152, opened 2023-07-05
fix: correctly validate schema `type: null`

tatomyr:
## What/Why/How?

This fixes wrong type validation (broken in **beta.111**).

Since then, we didn't recognise `type: null` being an error:

```yaml

WrongType:
     type: null

```

Whilst the `null` should be a string:

```yaml

CorrectType:
     type: "null"
```

## Reference

The original MR: https://github.com/Redocly/redocly-cli/pull/891

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/__tests__/lint/oas3.1/openapi.yaml
+++ b/__tests__/lint/oas3.1/openapi.yaml
@@ -228,7 +228,7 @@ components:
           type: 'null'
         tag:
           type:
-            - null
+            - 'null'
             - string
             - integer
           examples:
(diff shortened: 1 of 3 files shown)


lornajane (commented):
This does what I expect, but I think we need to fix the OpenAPI example in the `lint/oas3.1` tests to get the build to pass

tatomyr:
@lornajane sure. Just thinking whether I should change the definition or the snapshot.

Issue #1146, opened 2023-06-28
redocly/cli lint hangs for a while before exiting


philipnbbc:
**Describe the bug**

Running version >= `v1.0.0-beta.127` (`latest` is `v1.0.0-beta.129`) of `redocly/cli` Docker results in the validation completing and then the process sits there for a while before exiting. The wait time can be significant (30 sec) for larger OpenAPI spec files.

**To Reproduce**

A small OpenAPI file, `example.yaml`:
```yaml
openapi: 3.0.3
info:
  title: Test
paths:
  /:
    get:
      responses:
        '200':
```

1. Given the OpenAPI example file shown above
2. Run `docker run --rm -v $PWD:/spec redocly/cli lint example.yaml`
3. It outputs "run `redocly lint --generate-ignore-file` to add all problems to the ignore file." and then sits there for a couple seconds
4. Run `docker run --rm -v $PWD:/spec redocly/cli:v1.0.0-beta.126 lint example.yaml` and it exits immediately

Using a larger OpenAPI file results in a longer and more significant hang time.

**Expected behavior**

It shouldn't, or it is unclear why it would hang for a while before exiting.

**Redocly Version(s)**

Issue is with Docker image `redocly/cli` version >= `v1.0.0-beta.127` (`latest` == `v1.0.0-beta.129`)

philipnbbc:
Tried something... It turns out it hangs when behind a proxy. I assume it is attempting to make a HTTP request?

Can that be behavior be disabled? Switching telemetry off doesn't make a difference: `docker run --rm -e REDOCLY_TELEMETRY=off -v $PWD:/spec redocly/cli lint example.yaml`

lornajane:
Thanks for the update, although you've disproved my working hypothesis which was that the telemetry could cause a problem despite our testing on it! We'll take a look and keep you posted.

tatomyr:
Hi @philipnbbc, 
Could you check if setting env variable `CI=true` solves this?

philipnbbc:
> Could you check if setting env variable `CI=true` solves this?

I ran `docker run --rm -e CI=true -v $PWD:/spec redocly/cli lint example.yaml` and the issue remains.

The output of tcpdump showed DNS queries for `api.redocly.com`.

tatomyr:
Which version is this? Could you also switch off the telemetry? My last guess is this is a combo 😅

philipnbbc:
[...]

PR #1161, opened 2023-07-07
fix: handle aborted request in docker and remove redundant call

IgorKarpiuk:
## What/Why/How?

Handle aborted request in docker and remove redundant call in telemetry

## Reference

Related: https://github.com/Redocly/redocly-cli/issues/1146

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/update-version-notifier.ts
+++ b/packages/cli/src/update-version-notifier.ts
@@ -35,7 +35,7 @@ const isNewVersionAvailable = (current: string, latest: string) => compare(curre
 const getLatestVersion = async (packageName: string): Promise<string> => {
   const latestUrl = `http://registry.npmjs.org/${packageName}/latest`;
   const response = await fetch(latestUrl);
-  const info = await response.json();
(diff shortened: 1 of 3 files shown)


packages/cli/src/update-version-notifier.ts
```diff
@@ -35,7 +35,7 @@ const isNewVersionAvailable = (current: string, latest: string) => compare(curre
 const getLatestVersion = async (packageName: string): Promise<string> => {
   const latestUrl = `http://registry.npmjs.org/${packageName}/latest`;
   const response = await fetch(latestUrl);
-  const info = await response.json();
+  const info = await response?.json();
   return info.version;
```
tatomyr, line 39: If there can be no response, there could be no info, thus info.version will also fail.

IgorKarpiuk: This is not fail because we handle errors. But this is good catch, Thanks.

I updated this.

PR #1173, opened 2023-07-12
fix: use proxy env variables and abort-controller package to prevent hanging behind proxies

tatomyr:
## What/Why/How?

While working behind proxy,  HTTP_PROXY or HTTPS_PROXY environment variables might be used to detect this and skip  unnecessary http calls. 
Also, added `abort-controller` package to ensure consistent usage throughout different nodeJS versions.  

## Reference

Related to https://github.com/Redocly/redocly-cli/issues/1146

## Testing

IN PROGRESS. Trying to replicate the original issue...

## Check yourself

- [x] Code is linted

## Security


lornajane:
I'm closing this as we're not actively working on it. If we get more replication cases, we should revisit.

PR #1174, opened 2023-07-13
feat: run custom rules without plugin-Id

IgorKarpiuk:
## What/Why/How?

Run rules from plugins even if the `plugin-id ` is not specified (Search by name)

The warning from the config file will still be displayed if the `plugin- id` is not specified
## Screenshots (optional)
Screenshot 2023-07-13 at 11 15 19

## Check yourself

- [ ] Code is linted

## Security

### DIFF
--- a/packages/core/src/config/config-resolvers.ts
+++ b/packages/core/src/config/config-resolvers.ts
@@ -449,26 +449,53 @@ function groupStyleguideAssertionRules({
 
 function registerCustomAssertions(plugins: Plugin[], assertion: AssertionDefinition) {
   for (const field of keysOf(assertion.assertions)) {
-    const [pluginId, fn] = field.split('/');
 
-    if (!pluginId || !fn) continue;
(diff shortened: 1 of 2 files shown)


lornajane:
Are we building this in response to an issue or a feature request somewhere? I'd prefer that we require clear, qualified names for the rules rather than trying to infer something that could lead to clashing names (and make it more difficult for us to introduce new built-in rules later).

tatomyr:
@lornajane the idea is to simplify using custom rules as someone can forget to prefix rules with id and we still can resolve the intended rule. If I'm not mistaken, it works the following way: if we find a built-in rule, we use it; if we don't - we look for a rule in plugins, and if it matches exactly one - we use it; otherwise, we throw an error. While I understand the idea, I don't particularly like the changes, so I'd vote to close this.

lornajane:
We discussed this further with the team and @RomanHotsiy will open an issue with more detail, but we'll close this PR for now.

PR #1176, opened 2023-07-17
chore: v1.0.0-rc

tatomyr:
## What/Why/How?

Release  v1.0.0-rc

## Check yourself

- [ ] Code is linted

## Security


lornajane:
Could we call it `1.0.0-rc1` please? Just in case we need another RC before we're sure ...

tatomyr:
@lornajane no problem, but I believe it should contain a dot then, I believe: `1.0.0-rc.1`. At least, according to the [semver docs](https://semver.org/#spec-item-11). Am I correct?

tatomyr:
Updated

lornajane (approved):
Exciting times!

PR #1182, opened 2023-07-19
fix: resolve problem with docker build

IgorKarpiuk:
## What/Why/How?

Resolve problem with building docker image. And update smoke tests for testing it 

## Check yourself

- [x] Code is linted

## Security


tatomyr (approved):
Thanks!
It would be also great to write a smoke check after publishing to npm/docker.

PR #1211, opened 2023-08-04
chore: enable chagesets

tatomyr:
## What/Why/How?

This introduces [Changesets](https://github.com/changesets/changesets) flow.

Each feat/fix PR should also contain a changeset (to create one, run `npx changeset`).

After merging it into **main**, a PR for the release will be created. It will include version updates and local CHANGELOG.md files for each package. Also, an entry like the one below gets added to the **docs/changelog.md** via [a script](https://github.com/Redocly/redocly-cli/pull/1211/files#diff-b21b06900294248b9a9a7ca103a26c5fb0d10bd47cb5abb74cfdd68a761d5c69):

```
## 1.0.1 (2023-08-04)

### Patch Changes

- Add Changesets.
- Updated @redocly/openapi-core to v1.0.1.

```

Merging that PR will trigger the release procedure. After successful publishing, a notice will be sent to Slack.

## Testing

Tested as much as possible locally and in GHA (not tested the actual release).
The release and notification flow to be tested during the next release.

## Check yourself

- [x] Code is linted

## Security


.github/workflows/release.yaml
```diff
@@ -150,3 +175,20 @@ jobs:
...
+    needs:
+      - release
+      - post-release-smoke-checks
+    if: needs.release.outputs.published == 'true'
+    name: Post the Release Message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Post to a Slack channel
+        id: slack
+        uses: slackapi/slack-github-action@v1.23.0
```
RomanHotsiy, line 189: We may need to add a small sleep but maybe not relevant here.

RomanHotsiy (commented):
LGTM 👍

.github/workflows/release.yaml
tatomyr: I'm going to check if this works. If not, I'll add the timeout next time.

IgorKarpiuk (approved):

.github/workflows/release.yaml
```diff
@@ -17,14 +20,52 @@ jobs:
...
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: RomanHotsiy/changesets-action@v1
+        with:
+          # This expects you to have a script called release which does a build for your packages and calls changeset publish
+          publish: npm run release
+          commit: 'chore: 🔖 release new versions'
```
roman-sainchuk, line 50: Would be great to see release version in commit message

roman-sainchuk: Is changeset able to set git tags ?

roman-sainchuk (approved):

.github/workflows/release.yaml
tatomyr: I believe changesets automatically adds tags during publishing.

PR #1220, opened 2023-08-08
remove unused components was skipping schemas with a top-level allOf

### LINKED ISSUE #1209: remove-unused-components skips schemas that have a top-level `allOf`
**Describe the bug**

Given an OpenAPI document that contains something along these lines:

```
openapi: "3.0.0"
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Used'
components:
  schemas:
    Unused:
      allOf:
        - type: object
          properties:
            one:
              type: string
        - type: object
          properties:
            two:
              type: string
    Used:
      type: object
      properties:
        link:
          type: string
        child:
          $ref: '#/components/schemas/Two'
```

It would be expected that bundle --remove-unused-components would remove the Unused schema since it isn't used.

However, the current behavior is that after running bundle with --removed-unused-components, Unused is not removed. This _appears_ to be intentional (see https://github.com/Redocly/redocly-cli/blob/cfa31330b04f22e7b855143b2eef0d040b0873fa/packages/core/src/rules/oas3/remove-unused-components.ts#L64) but I have not found any reference to _why_ it was intentional.

**To Reproduce**

1. Copy the OpenAPI content above to a file.
2. Run redocly bundle <file.yaml> -o output.yaml --remove-unused-components
3. Check the output.yaml file.

**Expected behavior**

An unused component gets removed regardless of whether it has an `allOf` or not.

**OpenAPI definition**

Have only tested with 3.0 but the same `if` statement is present in the `oas2` rule.

**Redocly Version(s)**

1.0.0 and main.

**`Node.js` Version(s)**

16.17.0, 16.18.0, and 18.16.0

jedelson-pagerduty:
fixes #1209

## What/Why/How?

Details #1209 but essentially a schema like:

```
Unused:
  allOf:
    - type: object
      properties:
        one:
          type: string
    - type: object
      properties:
        two:
          type: string
```

will never be removed

## Testing

Added a unit test for this case. Have also tested this change on a real-world document with this situation.

## Check yourself

- [X] Code is linted

## Security

### DIFF
--- a/.changeset/light-games-clap.md
+++ b/.changeset/light-games-clap.md
@@ -1,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Fixed an issue where remove-unused-components would not remove schemas which had a top-level allOf and improved detection of unused schemas in no-unused-components rule.
--- a/packages/core/src/rules/oas3/remove-unused-components.ts
+++ b/packages/core/src/rules/oas3/remove-unused-components.ts
@@ -12,10 +12,11 @@ export const RemoveUnusedComponents: Oas3Rule = () => {
   function registerComponent(
     location: Location,
     componentType: keyof Oas3Components,
-    name: string
+    name: string,
+    used = false
   ): void {
     components.set(location.absolutePointer, {
-      used: components.get(location.absolutePointer)?.used || false,
+      used: components.get(location.absolutePointer)?.used || used,
       componentType,
       name,
     });
@@ -65,10 +66,13 @@ export const RemoveUnusedComponents: Oas3Rule = () => {
       },
     },
     NamedSchemas: {
-      Schema(schema, { location, key }) {
-        if (!schema.allOf) {
-          registerComponent(location, 'schemas', key.toString());
-        }
+      Schema(schema, { location, key, resolve }) {
+        registerComponent(
+          location,
+          'schemas',
+          key.toString(),
+          schema.allOf && schema.allOf.some((v) => v.$ref && resolve(v).node?.discriminator)
+        );
       },
     },
     NamedParameters: {
--- a/packages/core/src/rules/oas2/__tests__/remove-unused-components.test.ts
+++ b/packages/core/src/rules/oas2/__tests__/remove-unused-components.test.ts
@@ -1,5 +1,6 @@
 import { outdent } from 'outdent';
-import { parseYamlToDocument, replaceSourceWithRef, makeConfig } from '../../../../__tests__/utils';
+import { cloneDeep } from 'lodash';
+import { parseYamlToDocument, makeConfig } from '../../../../__tests__/utils';
 import { bundleDocument } from '../../../bundle';
 import { BaseResolver } from '../../../resolve';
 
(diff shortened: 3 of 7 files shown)


packages/core/src/rules/oas3/remove-unused-components.ts
```diff
@@ -65,10 +65,8 @@ export const RemoveUnusedComponents: Oas3Rule = () => {
       },
     },
     NamedSchemas: {
-      Schema(schema, { location, key }) {
-        if (!schema.allOf) {
```
tatomyr, line 69: Could you check if that will correctly work for discriminators?

jedelson-pagerduty:
@tatomyr I've updated the code to deal with using discriminator from a referenced allOf. Still not 100% sure this is case that drew the original concern.

I should still add a test for the rule change but I'm less familiar with how rules are tested (vs. decorators) but wanted to push this up to see if this was what you were referring to.

jedelson-pagerduty:
@tatomyr I've added the rule test. To summarize:

* If an "allOf" schema references some other schema **and** that schema has a `discriminator` the "allOf" schema is considered used (even though it may not actually be used... it is _potentially_ used). `no-unused-components` will not create an issue and `remove-unused-components` will not remove it.
* If an "allOf" schema references some other schema **and** none of the referenced schemas have a `discriminator`, the "allOf" schema is considered used following the normal usage rules (i.e. is ref'd somewhere else).
* If an "allOf" schema doesn't reference other schemas, the "allOf" schema is considered used following the normal usage rules (i.e. is ref'd somewhere else).

I believe this addresses the original FIXME.

jedelson-pagerduty:
Open questions:

* Should this be a `minor` change instead of a `patch`?
* Should I create two separate changeset files?

tatomyr:
> Should this be a minor change instead of a patch?
Should I create two separate changeset files?

Please use minor (replace the existing one).

jedelson-pagerduty:
> Please use minor (replace the existing one).

Done

.changeset/light-games-clap.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Fixed an issue where remove-unused-components would not remove schemas which had a top-level allOf and improved detection of unused schemas in no-unused-components rule.
```
adamaltman, line 6: @tatomyr do we need this in a conventional commit type of format? (`fix: `)

adamaltman:
@tatomyr I'm curious why this is a `minor` instead of a `patch`? I thought a bug fix would be a `patch`? Is this bugfix not considered backwards compatible perhaps?

@jedelson-pagerduty overall great contribution 🎉 . Thank you so much for your contribution.

.changeset/light-games-clap.md
tatomyr: No, this is a description that goes to the changelog. We decided to follow the existing changelog format.

tatomyr:
@adamaltman perhaps we decided this is more of an improvement rather than a fix. I don’t recall the exact reason though. 
@IgorKarpiuk could you take a look and finalise this once you’re done with the current release?

packages/core/src/rules/oas3/remove-unused-components.ts
```diff
@@ -65,10 +66,13 @@ export const RemoveUnusedComponents: Oas3Rule = () => {
...
-          registerComponent(location, 'schemas', key.toString());
-        }
+      Schema(schema, { location, key, resolve }) {
+        registerComponent(
+          location,
+          'schemas',
+          key.toString(),
+          schema.allOf && schema.allOf.some((v) => v.$ref && resolve(v).node?.discriminator)
```
tatomyr, line 74: I'm afraid this won't work correctly with transient allOf-discriminator pair. Consider this example:

```yaml
openapi: '3.0.0'
info: 
  title: X
  description: X
  version: 1.0.0

components: 
  schemas: 
    BaseVehicle: 
      type: object
      description: Vehicle
      discriminator:
        propertyName: powerSource
      properties:
        powerSource:
          description: How is the vehicle powered.
          type: string
          example: pedaling

    Vehicle:
      allOf: 
        - $ref: "#/components/schemas/BaseVehicle"
        - type: object 
          properties:
            vehicleType:
              description: The type of vehicle.
              type: string
              example: bicycle

    PedaledVehicle:
      allOf:
        - $ref: "#/components/schemas/Vehicle"
        - type: object
          description: Pedaled Vehicle
          properties:
            range:
              description: The 95th percentile range of a trip in kilometers.
              type: integer
              example: 100
```

The `PedaledVehicle` will be considered unused despite `Vehicle` resolves to `BaseVehicle` which actually contains the `discriminator`. I'm not sure this is a common use case but technically it looks like a valid one.

tatomyr, line 74:
```suggestion
          schema.allOf?.some((v) => v.$ref && resolve(v).node?.discriminator)
```

.changeset/light-games-clap.md
tatomyr, line 6: Could you rephrase the description so it emphasises the improvement rather than the fix?

tatomyr (commented):
Hi @jedelson-pagerduty! Sorry that the review took so long  😅  

I believe your approach is correct. However, there are some corner cases when we are removing actually used components (which is worse than leaving unused ones). I found one. Do you see anything else? Maybe @lornajane has more to suggest?

packages/core/src/rules/oas3/remove-unused-components.ts
jedelson-pagerduty: @tatomyr Appreciate the feedback! And yes, I suppose that is valid in the general case.

I'm wondering if this is fruitless -- I am trying to apply a very strict definition of "used" that works for my use case (in this example, I would actually expect all those schemas to be removed) but wouldn't be generally applicable. 

I can look at taking another run at this (although it might be a few days) but I am starting to think a different approach is necessary.

tatomyr:
@jedelson-pagerduty do you have any updates on this? 

FYI we've recently moved the `remove-unused-components` decorator, so that's why this MR got the conflicts 😅.

tatomyr:

PR #1222, opened 2023-08-08
Custom plugin documentation

lornajane:
## What/Why/How?

Our custom plugins are brilliant, but the documentation is unclear and hard to follow. This pull request attempts to outline the available features, and add separate documentation for each thing, as well as clearer overviews on why each one is useful.

## Reference

https://redocly.com/docs/cli/resources/custom-plugins/

## Check yourself

- [ ] Code is linted

## Security


lornajane:
The various guides for decorators and custom rules need some updates (and some of this is in core now too, which is a great problem to have!), I'll do those separately since this is already plenty to review.

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
+## Configuration in plugins
```
HCloward, line 1:
```suggestion
# Configuration in plugins
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+
+Plugins can include configuration, under the `configs` key. This is
```
HCloward, line 3:
```suggestion
You can bundle rule configurations under the `configs` key in your plugin code. These configurations are
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+useful when your plugin declares multiple rules or decorators, and you want to
```
HCloward, line 4:
```suggestion
useful when you declare multiple rules or decorators in your plugin, and you want to
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+define within the plugin how to combine them.  Each plugin can have multiple
+bundles of configuration, and the user can use any of them in their
```
HCloward, line 6:
```suggestion
bundles of rule configurations, and the user can use any of them in their
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+configuration file.
+
+
+Here's an example of configuration in a plugin, defining two bundles:
```
HCloward, line 10:
```suggestion
Here's an example plugin configuration, defining two rule bundles:
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+      rules: {
+        'operation-id-not-test': 'off',
+        'boolean-parameter-prefixes': 'error',
+      },
+    }
+  }
+};
+```
+
+With the example plugin above id as `my-local-plugin`, the `all` and `minimal`
```
HCloward, line 32:
```suggestion
In the example, the plugin id is `my-local-plugin` and the `all` and `minimal` rule
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+configurations become available in the user's configuration file as
```
HCloward, line 33:
```suggestion
configurations are available in the user's configuration file as
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+`"my-local-plugin/all"` and `"my-local-plugin/minimal"`, respectively. Use
+these configurations by adding them under `extends:` in `redocly.yaml`.
```
HCloward, line 35:
```suggestion
these configurations by adding them under `extends:` in your `redocly.yaml` configuration file.
```

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+
+```yaml
+extends:
+  - my-local-plugin/all
+```
+
+Plugins can include rules, decorators and preprocessors in their configuration bundles.
```
HCloward, line 42:
```suggestion
Plugins can include rules, decorators, and preprocessors in their configuration bundles.
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
+# Decorators in plugins
+
+Decorators transform API descriptions, by adding, removing, or changing elements of the document. Before you build your own decorators:
+
+* [learn about Redocly decorators](../decorators.md)
+* [check the list of built-in decorators](../decorators.md#list-of-decorators)
+
+If you didn't find an existing decorator that fit your needs, then adding a decorator in a plugin is a good way to proceed.
```
HCloward, line 8:
```suggestion
If you can't find an existing decorator that fits your needs, then you can add a decorator in a plugin.
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+
+:::warning Preprocessors
+
+Decorators and preprocessors are the same in structure, but preprocessors are run _before_ linting, and decorators are run after. We always recommend using decorators where possible, since the document might not be valid or structured as expected if the linting step hasn't run yet.
+
+:::
+
+## Plugin structure
+
+In order to create a preprocessor or decorator, the object that is exported from your module has to conform to the following interface:
```
HCloward, line 18:
```suggestion
To create a preprocessor or decorator, the object that is exported from your module has to conform to an interface such as the following example:
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+    oas3: {
+      "decorator-id": () => {
+        // ...
+      }
+    }
+  }
+}
+```
+
+Each decorator or preprocessor is a function that returns an object. The object's keys are the node types in the document, and each of those can contain any or all of the `enter()`, `leave()` and `skip()` functions for that node type. More information and examples on the [visitor pattern page](./visitor.md).
```
HCloward, line 40:
```suggestion
Each decorator or preprocessor is a function that returns an object. The object's keys are the node types in the document, and each of those can contain any or all of the `enter()`, `leave()` and `skip()` functions for that node type. Find more information and examples on the [visitor pattern page](./visitor.md).
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+        if(target.description) {
+          target.description = "✨ " + String(target.description);
+        }
+      }
+    },
+  }
+};
+```
+
+Decorators use the [visitor pattern](./visitor.md) to run an operation on every node in the document. In this example, when the code does `leave()` of an `Operation` node, it checks if the node (passed as `target` in this example) has a description, and updates it if it does.
```
HCloward, line 65:
```suggestion
Decorators use the [visitor pattern](./visitor.md) to run an operation on every node in the document. In this example, when the code executes the `leave()` function on the `Operation` node, it checks if the node (passed as `target` in this example) has a description, and updates it if it does.
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Further examples of custom decorators
+
+See some more examples of decorators:
+- Follow our [replace-servers-url tutorial](../guides/replace-servers-url.md).
+- Change your [OAuth2 token URL](../guides/change-token-url.md).
+
+## Preprocessors
+
+Preprocessors follow the same structure and operation as decorators, but they are run before the validation/linting step. This makes them brittle because the document may not be valid, and the extra processing step can cause performance impacts. We recommend looking for alternative approaches to preprocessing.
```
HCloward, line 156:
```suggestion
Preprocessors follow the same structure and operation as decorators, but they are run before the validation/linting step. Running before the validation/linting step makes them brittle because the document may not be valid, and the extra processing step can cause performance impacts. We recommend looking for alternative approaches to preprocessing.
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
+# Decorators in plugins
+
+Decorators transform API descriptions, by adding, removing, or changing elements of the document. Before you build your own decorators:
+
+* [learn about Redocly decorators](../decorators.md)
```
HCloward, line 5:
```suggestion
* [Learn about Redocly decorators](../decorators.md)
```

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+* [check the list of built-in decorators](../decorators.md#list-of-decorators)
```
HCloward, line 6:
```suggestion
* [Check the list of built-in decorators](../decorators.md#list-of-decorators)
```

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,113 @@
...
+Here is a basic example of a nested visitor:
+
+```js
+function ExampleRule() {
+  const seen = {};
+  return {
+    Operation: {
+      // skip: (value, key) => ... // if needed
+      // enter(operation) {} // if needed
+      Schema(schema, ctx, parents) {
```
tatomyr, line 64: Wouldn't it be better to always explicitly use the `enter` visitor in our documentation? I remember, named visitor functions confused me when I saw them first time. It still looks strange to me that the same thing could be an object or a function  😅

lornajane: Ah, good spot. I didn't update this bit (just moved it to a new file) but I agree an object would be clearer. I'll switch the example.

lornajane: @tatomyr can you take another look? I tried to replace this existing example with an equivalent. It seems to work the same but would be good to get your eyes on it too

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+  };
+}
+
+```
+
+The `ctx` object here holds all the context, which can be used to give more situation-aware functionality to the rules you build. This is one of the main use cases for custom rules. The `report()` method is used to give information to return to the user if the node being visited doesn't comply with the rule. You can read the [context](#the-context-object) and [location](#location-object) sections for more information.
+
+Adding this as part of a plugin requires you to add it to the `rules` part of the plugin object, under the relevant document type. The example rule here is intended to be used with OpenAPI, so the plugin code in `plugins/my-rules.js` is as follows:
+
+```
```
HCloward, line 39:
```suggestion
```js
```

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+The context object contains additional functionality that is helpful for rules to do their jobs. As the name implies, the context object contains information that is relevant to the context of the rule. The context object has the following properties:
+
+- `location` - current location in the source document. See [Location Object](#location-object)
+- `parentLocations` - mapping of parent node to its location (only for nested visitors)
+- `type` - information about current type from type tree
+- `parent` - parent object or array
+- `key` - key in parent object or array
+- `oasVersion` specific OAS minor version of current document (can be `oas2`, `oas3` or `oas3_1`).
+
+The context object also offers some additional functionality that
```
HCloward, line 77: This is odd - perhaps this was added as a mistake?

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+
+Validate your OpenAPI document with `redocly lint openapi.yaml`. With this rule enabled as shown, any `operationId` fields that are "test" cause the validation step to emit a warning. You can use the example here as a basis for building your own rules, and many rules can be included in a single plugin.
+
+## Object references
+
+### The context object
+
+The context object contains additional functionality that is helpful for rules to do their jobs. As the name implies, the context object contains information that is relevant to the context of the rule. The context object has the following properties:
+
+- `location` - current location in the source document. See [Location Object](#location-object)
```
HCloward, line 70:
```suggestion
- `location` - Current location in the source document. See [Location Object](#location-object).
```

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+- `parentLocations` - mapping of parent node to its location (only for nested visitors)
+- `type` - information about current type from type tree
```
HCloward, line 72:
```suggestion
- `type` - Information about the current type from the type tree.
```

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+- `parent` - parent object or array
```
HCloward, line 73:
```suggestion
- `parent` - Parent object or array.
```

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,116 @@
...
+- `key` - key in parent object or array
```
HCloward, line 74:
```suggestion
- `key` - Key in parent object or array.
```

docs/custom-plugins/extended-types.md
```diff
@@ -0,0 +1,74 @@
+# Type extensions in plugins
+
+Redocly CLI in its core has a type tree which defines the structure of the API definition.
+Redocly CLI then uses it to do type-aware traversal of OpenAPI Document.
```
HCloward, line 4:
```suggestion
Redocly CLI then uses it to do a type-aware traversal of an OpenAPI definition.
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+title: Custom plugins in Redocly
+redirectFrom:
+  - /docs/cli/custom-rules/
+  - /docs/cli/resources/custom-rules/
+  - /docs/cli/resources/custom-plugins/
+---
+
+# Custom plugins
+
+Custom plugins are a powerful way to extend Redocly. Use of custom plugins requires requires in-depth knowledge of the document type (such as OpenAPI) that you are working with, JavaScript coding skills, and the plugin interface.
```
HCloward, line 11:
```suggestion
Custom plugins are a powerful way to extend Redocly. Use of custom plugins requires in-depth knowledge of the document type (such as OpenAPI) that you are working with, JavaScript coding skills, and the plugin interface.
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+Custom plugins are recommended for advanced users who have exhausted the built-in options available, and who can develop JavaScript extensions themselves.
+Each plugin is a JavaScript module which can export custom rules, configuration, preprocessors, decorators or type tree extensions.
```
HCloward, line 13:
```suggestion
Each plugin is a JavaScript module which can export custom rules, configurations, preprocessors, decorators or type tree extensions.
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+:::Info
+For many users, the highly [configurable rules](../rules/configurable-rules.md) covers their needs without needing a custom plugin.
+:::
+
+Custom plugins use the [visitor pattern](./visitor.md) to traverse the document structure and apply the rules and decorators (and preprocessors if you have them) to each element. Rules defined in custom plugins may need additional context available, and this can be achieved using the [nested visitor approach](./visitor.md#nested-visitors) (please note that nested visitors are not available for decorators or preprocessors).
+
+## Extend Redocly with custom plugins
+
+The custom plugin functionality offers multiple ways to extend the built-in functionality of Redocly:
```
HCloward, line 23:
```suggestion
Custom plugins can extend the built-in functionality of Redocly in the following ways:
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+- **Write your own custom rules in JavaScript**. If your API standards include elements that can't be expressed by our existing rules, including configurable ones, you can code your own rules. See the [documentation on writing your own rules in custom plugins](./custom-rules.md).
```
HCloward, line 25:
```suggestion
- **Write your own custom rules in JavaScript**. If your API standards include elements that can't be expressed by our existing rules, including configurable ones, you can code your own rules. See the [documentation on writing your own rules in custom plugins](./custom-rules.md) for more information.
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+  - plugins/my-best-plugin.js
+  - plugins/another-plugin.js
+```
+
+The paths are relative to the configuration file location. Where there are multiple features that happen at the same time, such as decorators, the plugins loaded first in the `plugins` section are processed first.
+
+
+### Plugin structure
+
+The minimal plugin should export an `id` string that is used to refer to the contents of the plugin in configuration:
```
HCloward, line 66:
```suggestion
The minimal plugin should export an `id` string that is used to refer to the contents of the plugin in the `redocly.yaml` configuration file:
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+```js
+module.exports = {
+  id: 'my-local-plugin'
+}
+```
```
HCloward, line 72: Should we also include a sample `redocly.yaml` file with the id used?

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+<a id="oas-major-versions"></a>
+## Supported formats
+
+Everything that is exported from plugin relates to one of the supported document formats, such as OpenAPI v3. It is done by exporting an object containing a key-value mapping from a document format and version (`oas2` or `oas3` are supported) to an extension object (rules, preprocessors, decorators).
```
HCloward, line 77:
```suggestion
Everything that is exported from a plugin relates to one of the supported document formats, such as OpenAPI v3. Plugins work by exporting an object containing a key-value mapping from a document format and version (`oas2` or `oas3` are supported) to an extension object (rules, preprocessors, decorators).
```

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+Before processing the definition document Redocly CLI detects the document format and applies corresponding set of extensions.
```
HCloward, line 79:
```suggestion
Before processing the definition document, Redocly CLI detects the document format and applies a corresponding set of extensions.
```

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+To understand how this applies to your API description, think of the document as a tree structure. The top level elements are entries like `info` and `components`. To examine the `description` field in the `info` section, the visitor goes to the `info` node first, then on to the `description` node. This pattern is repeated all over the document as the visitor pattern works its way around the document tree.
+
+<a id="visitor-object"></a><a id="format-of-visitor-object"></a>
+
+## Structure of the visitor object
+
+In your plugin, create a JavaScript visitor object, and describe the functionality required for each type of node.
+
+Redocly CLI calls `enter()` while going down the tree and `leave()` going up the tree after processing the node and its children.
+`skip()` predicate is called and if it returns `true` the node is ignored for this visitor.
```
HCloward, line 14: This statement is confusing to me. When is the skip predicate called?

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+
+Top level **visitor functions** run only once for each node.
+If the same node is referenced by the $ref multiple times,
+top-level **visitor functions** are executed only once for this node.
+
+This works fine for most context-free rules which check basic things. If you need contextual info you should use [nested visitors](#nested-visitors).
+
+## Nested visitors
+
+Visitor object (if it is not `any` or `ref`) can define [nested visitors](#nested-visitors).
```
HCloward, line 52:
```suggestion
The visitor object (if it is not `any` or `ref`) can define [nested visitors](#nested-visitors).
```

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+    }
+  };
+}
+```
+
+The `Schema` **visitor function** is called by Redocly CLI only if the Schema Object is encountered while traversing a tree while the Operation Object is **entered**.
+
+As the third argument, `enter()` in a **nested visitor object** accepts the `parents` object with corresponding parent nodes as defined in the **visitor object**.
+
+<div class="attention"> It will be executed only for the first level of Schema Object.</div>
```
HCloward, line 74:
```suggestion
<div class="attention"> It will be executed only for the first level of the Schema Object.</div>
```

docs/decorators.md
```diff
@@ -4,22 +4,27 @@ redirectFrom:
 ---
 # Decorators
 
-All Redocly CLI decorators are listed on this page.
+Decorators are a way of changing an API description during bundling. This can be useful for many use cases:
```
HCloward, line 7:
```suggestion
Decorators are a way of changing an API description during bundling. This updating during bundling can be useful for the following use cases:
```

docs/custom-plugins/custom-rules.md
lornajane: Apparently my brain stalled mid-sentence? Thanks for catching this

docs/custom-plugins/index.md
HCloward: So it might be an American/English thing, because from what I can tell configuration can be made plural with an "s" in American English and it is fine. I do not have particularly strong feelings either way though, so your choice.

docs/custom-plugins/index.md
lornajane: I'm not sure it's that useful because this section is about how to structure the plugin, and how it should be used is different for each type of thing you put in your plugin, so I put the examples in each specific page.

docs/custom-plugins/visitor.md
lornajane: Let me have another go. We visit each node in turn and call whichever of these methods is defined.

docs/custom-plugins/index.md
lornajane: If it  seems off to you, it probably seems wrong to other US English speakers so I'm accepting the suggestion. Happy to hear that either would be valid though!

docs/custom-plugins/custom-config.md
```diff
@@ -0,0 +1,51 @@
...
+extends:
+  - my-local-plugin/all
+```
+
+Plugins can include rules, decorators, and preprocessors in their configuration bundles.
+
+:::info Rulesets
+
+If your configuration includes only Redocly built-in rules and decorators, try
+using a [ruleset](..//rules.md#rulesets) to achieve this (it's simpler than a
```
adamaltman, line 47:
```suggestion
using a [ruleset](../rules.md#rulesets) to achieve this (it's simpler than a
```
?

docs/custom-plugins/custom-decorators.md
```diff
@@ -0,0 +1,158 @@
...
+        if(target.description) {
+          target.description = "✨ " + String(target.description);
+        }
+      }
+    },
+  }
+};
+```
+
+Decorators use the [visitor pattern](./visitor.md) to run an operation on every node in the document. In this example, when the code executes the `leave()` function on the `Operation` node, it checks if the node (passed as `target` in this example) has a description, and updates it if it does.
```
adamaltman, line 65: question: Do you have to use `target` or can it be replaced with any string, such as `op`? I'm not clear for example if this also works:

```js
module.exports = OperationSparkle;

function OperationSparkle() {
  console.log("adding sparkles ... ");
  return {
    Operation: {
      leave(op) {
        if(op.description) {
          op.description = "✨ " + String(op.description);
        }
      }
    },
  }
};
```

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,117 @@
...
+- `parent` - Parent object or array.
+- `key` - Key in the parent object or array.
+- `oasVersion` - Specific OAS minor version of the current document (can be `oas2`, `oas3` or `oas3_1`).
+
+The context object also offers some additional functionality to resolve references and to return information about a problem to the user. The methods available are as follows:
+
+- `resolve(node)` - Synchronously dereferences `$ref` node to its value. Works only with `$refs` from the original document. If you need to resolve a reference from another source, you can use the optional second parameter: `resolve(node, from: string)`.
+- `report(descriptor)` - Reports a problem in the definition and returns information to the user. See [Report rule context](#report-rule-context) for more information.
+
+<a id="context-report"></a>
```
adamaltman, line 82: suggestion: Remove this anchor because this is a new page and there are no pre-existing links.
```suggestion
```

docs/custom-plugins/extended-types.md
```diff
@@ -0,0 +1,74 @@
+# Type extensions in plugins
+
+Redocly CLI in its core has a type tree which defines the structure of the API definition.
+Redocly CLI then uses it to do a type-aware traversal of an OpenAPI definition.
```
adamaltman, line 4: question: Is the plan to update this from saying "OpenAPI" once this merges? https://github.com/Redocly/redocly-cli/pull/1127

docs/custom-plugins/extended-types.md
```diff
@@ -0,0 +1,74 @@
...
+const XMetaData = {
+  properties: {
+    lifecycle: { type: 'string'},
+    'owner-team': { type: 'string'},
+  },
+  required: ['lifecycle']
+};
+```
+
+Note the quotes around the `owner-team` key since it contains a hyphen `-`. These fields are strings but you can also define other datastructures if you need to represent something more complex. The [built-in type definitions](https://github.com/Redocly/redocly-cli/tree/main/packages/core/src) are a good place to look for examples.
```
adamaltman, line 36:
```suggestion
Note the quotes around the `owner-team` key since it contains a hyphen `-`. These fields are strings but you can also define other data structures if you need to represent something more complex. The [built-in type definitions](https://github.com/Redocly/redocly-cli/tree/main/packages/core/src) are a good place to look for examples.
```

docs/custom-plugins/extended-types.md
```diff
@@ -0,0 +1,74 @@
...
+    lifecycle: production
+    owner-team: Engineering/Integrations
+```
+
+Define the type extension for the new `x-metadata` section, in this example, we make the `lifecycle` field required.
+
+```js
+const XMetaData = {
+  properties: {
+    lifecycle: { type: 'string'},
```
adamaltman, line 29:
```suggestion
    lifecycle: { type: 'string', enum: ['development', 'staging', 'production']},
```
suggestion: make it a little more than a basic string

docs/custom-plugins/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+The minimal plugin should export an `id` string that is used to refer to the contents of the plugin in the `redocly.yaml` configuration file:
+
+```js
+module.exports = {
+  id: 'my-local-plugin'
+}
+```
+
+<a id="oas-major-versions"></a>
```
adamaltman, line 74:
```suggestion
```
Suggestion: remove anchor because there are no links to this document or anchor that I could find. Maybe I'm  searching the wrong place?

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
+# Visitor pattern
+
+_Visitor_ is a design pattern that allows operations to be performed on individual elements in a more complex structure. Redocly uses it as the basis for custom plugins where every node in a document must be evaluated, for example when applying rules or decorators.
+
+To understand how this applies to your API description, think of the document as a tree structure. The top level elements are entries like `info` and `components`. To examine the `description` field in the `info` section, the visitor goes to the `info` node first, then on to the `description` node. This pattern is repeated all over the document as the visitor pattern works its way around the document tree.
+
+<a id="visitor-object"></a><a id="format-of-visitor-object"></a>
```
adamaltman, line 7:
```suggestion
```
Suggestion: Remove -- I didn't find any anchor links to either of these anchors.

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Structure of the visitor object
+
+In your plugin, create a JavaScript visitor object, and describe the functionality required for each type of node.
+
+Redocly CLI calls `enter()` while going down the tree and `leave()` going up the tree after processing the node and its children.
```
adamaltman, line 13: idea: I'm usually not for diagrams or illustrations, because they can be hard to maintain. I could see a diagram adding some value here though. 

```
Root enter() --> Info enter() --> License enter() --> ... (turn around) ...  ↓
    --> License leave() --> Info leave() --> Root leave()
```

I don't think this is required to do (I mean to leave it as an idea for the future if we find people need more help with the concept).

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+If the `skip()` predicate is defined and returns `true` the node is ignored for this visitor.
```
adamaltman, line 14: question: Does this result in all children nodes being ignored too? I'm trying to understand the reason to skip something versus omitting it from the plugin.

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+- `ref` - visitor is called on $ref nodes.
+
+
+## Visitors execution and $ref
+
+Top level **visitor functions** run only once for each node.
+If the same node is referenced by the $ref multiple times,
+top-level **visitor functions** are executed only once for this node.
+
+This works fine for most context-free rules which check basic things. If you need contextual info you should use [nested visitors](#nested-visitors).
```
adamaltman, line 48:
```suggestion
This works fine for most context-free rules. If you need contextual info you should use [nested visitors](#nested-visitors).
```
suggestion: Not sure what is the line between basic and not basic? It seems like contextual info is the line, so maybe it is more clear with that removed because the next sentence addresses the context.

docs/custom-plugins/visitor.md
```diff
@@ -0,0 +1,111 @@
...
+    }
+  };
+}
+```
+
+The `Schema` **visitor function** is called by Redocly CLI only if the Schema Object is encountered while traversing a tree while the Operation Object is **entered**.
+
+As the third argument, `enter()` in a **nested visitor object** accepts the `parents` object with corresponding parent nodes as defined in the **visitor object**.
+
+<div class="attention"> It will be executed only for the first level of the Schema Object.</div>
```
adamaltman, line 74:
```suggestion
:::attention
It is executed only for the first level of the Schema Object.
:::
```

adamaltman (approved):
Great 🎉 

I left some suggestions, questions, and ideas.

docs/custom-plugins/custom-rules.md
```diff
@@ -0,0 +1,117 @@
...
+Each rule is a function that returns an object with methods that Redocly CLI calls to "visit" nodes while traversing the definition document. The object keys are the node types that are encountered in the document. In this simple example, the custom plugin holds a rule that fails if any `operationId` is set to "test".
+
+To keep the plugin code manageable, each rule can go in its own file. This example is in `plugins/rules/opid-not-test.js`:
+
+```js
+module.exports = OperationIdNotTest;
+
+function OperationIdNotTest() {
+  return {
+    Operation(operation, ctx) {
```
tatomyr, line 22: One more place with implicit usage of  the visitor func.

tatomyr (approved):
A last comment. Overall looks good.

docs/custom-plugins/custom-decorators.md
lornajane: yes, I tried to use "target" in an attempt to make it a bit more copy/paste friendly, but called out that it was just an example, so I think it's clear.

docs/custom-plugins/extended-types.md
lornajane: I think we'll need to update a few things, yes! I intentionally made things more vague in a few places but at this time, I think the type extension feature is really only useful for OpenAPI.

docs/custom-plugins/custom-rules.md
lornajane: Thanks! Fixed it :)

PR #1226, opened 2023-08-14
Fix/wrong location in no trailing slash rule

tatomyr:
## What/Why/How?

Fixed the wrong location pointer when reporting on the 'no-path-trailing-slash' rule. Previously it was pointing on a value instead of a key when using $refs.

## Check yourself

- [x] Code is linted

## Security


adamaltman (approved):
Thank you.

IgorKarpiuk (approved):

PR #1234, opened 2023-08-17
chore: replace stale dependencies

IgorKarpiuk:
## What/Why/How?

- Set minimal `Node.js` version to 14
- Remove unsupported `assert-node-version` and `glob-promise` dependencies
- Replace `portfinder` with `get-port-please` package

## Reference
https://github.com/orgs/Redocly/projects/12/views/5?pane=issue&itemId=35308489
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -130,32 +130,6 @@ jobs:
           node-version: 14
       - run: bash ./__tests__/smoke/run-smoke.sh "npm i redoc redocly-cli.tgz" "npm run"
 
-  run-smoke--npm--node-12:
-    needs: prepare-smoke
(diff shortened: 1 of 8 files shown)


SmoliyY (approved):

robotadam:
Hey there; sorry for commenting on an old PR, but I was taking a look after seeing a notice about the old version of semver required here. It looks like `assert-node-version` was vendored into the new ts file, but not removed from `package.json`, so it's still included in the dependency tree: https://www.npmjs.com/package/@redocly/cli?activeTab=dependencies

RomanHotsiy:
Thanks @robotadam!

@IgorKarpiuk ☝️

PR #1239, opened 2023-08-20
docs: rename to API description for consistency

### LINKED ISSUE #1237: CLI Help improvements
Here's what I'm finding as I go. Will send a pull request with changes. 

1. CLI Help uses the terms `descriptions` and `definition` interchangeably which is confusing for users who don't know they are the same thing.

```
Commands:
  redocly stats [api]                       Show statistics for an API
                                            description.
  redocly split [api]                       Split an API definition into a
                                            multi-file structure.
  redocly join [apis...]                    Join definitions [experimental].
  redocly push [api] [maybeDestination]     Push an API definition to the
  [maybeBranchName]                         Redocly API registry.
  redocly lint [apis...]                    Lint definition.
  redocly bundle [apis...]                  Bundle definition.
  redocly login                             Login to the Redocly API registry
                                            with an access token.
  redocly logout                            Clear your stored credentials for
                                            the Redocly API registry.
  redocly preview-docs [api]                Preview API reference docs for the
                                            specified definition.
  redocly build-docs [api]                  Produce API documentation as an HTML
                                            file
  redocly completion                        Generate completion script.
```

2. Commands explain what not why.

> Login to the Redocly API registry with an access token.

Can this talk about what functionality I would gain or what use cases I would solve from doing that instead? 

> Generate completion script.

What does this mean?

3. Split command requires outDir but doesn't declare that in the help

```
> redocly split api/openapi.bundled.yaml
redocly split [api]

Split an API definition into a multi-file structure.

Positionals:
  api  API definition file that you want to split            [string] [required]

Options:
  --version    Show version number.                                    [boolean]
  --help       Show help.                                              [boolean]
  --outDir     Output directory where files will be saved.   [string] [required]
  --separator  File path separator used while splitting. [string] [default: "_"]
  --config     Path to the config file.                                 [string]


Missing required argument: outDir
```

4. bundle -o is missing documentation

```
  -o, --output                                                      	[string]
```

adamaltman:
## What/Why/How?

- Find/replace "API description" with "API definition"

## Reference

- Fixes #1237

## Testing

- n/a

## Check yourself

- [x] Code is linted

## Security


adamaltman:
I wonder why the node-18 smoke tests failed? I don't think this changes anything that would impact that. @tatomyr can you please check it?

IgorKarpiuk:
@adamaltman There is some problem with resolving dependencies. We know about it and trying to fix.

lornajane:
Thanks @IgorKarpiuk, can you link the ticket for the dependencies thing so we can track it?

IgorKarpiuk:
Here is the ticket https://github.com/Redocly/redocly-cli/issues/1240

It should not block this PR  you can merge it if needed.

lornajane:
@adamaltman  I believe that "API description" would be the more correct term, perhaps we should find and replace the inverse of this? "API description" is used in more places I think (this isn't very scientific, but for example https://learn.openapis.org/ and most of the other OpenAPI outputs use it). I also think that "document" could be a valid way to refer to the openapi file.

adamaltman:
Yes, it seemed to be called definition more in 3.0 vs 3.1. 

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#openapi-document

Although they still use definition in the intro:

> An OpenAPI definition can then

RomanHotsiy:
I would switch everywhere to "OpenAPI document" then.

Here is a relevant issue from OpenAPI spec repo: https://github.com/OAI/OpenAPI-Specification/issues/2007

[Learn OpenAPI ](https://learn.openapis.org/) seems to interchangeably use "OpenAPI document", "OpenAPI description document" or "OpenAPI description". I would stick to one variant.

philsturgeon:
There is an API Description, which is an API Description Document that uses a specific API Description Format.

OpenAPI is an *API Description Format*, so it does not say OpenAPI Description Document, that would be redundant. It generally calls the document an OpenAPI Document.

So its either an OpenAPI Document, or API Description Document if you're trying to be format agnostic. 

Finally sometimes it's not really a document, more of a text box or representation of an API in some other state, so API Description is generally more than adequate in that case and all other cases mentioned above. 

tldr IMO just call it API Description everywhere :D

adamaltman:
I renamed it everywhere from "definition" to "description" -- it affects a wider set of files. Please review @Redocly/dark-side

tatomyr:
@adamaltman there are few things are missed in the code. If you don't mind, I'll push my updates.

adamaltman:
Go ahead @tatomyr , thank you!

IgorKarpiuk (approved):

.changeset/three-moose-cheer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Renamed API definition into API description for consistency.
```
lornajane, line 6:
```suggestion
Renamed API definition to API description for consistency.
```

lornajane:
There are some places where we also used "definition" rather than "API definition" as a shorthand way to refer to the API descriptions. The existing changes are very good though, thanks @adamaltman ! I'll push some additions shortly.

lornajane (approved):
All looks really good!

I've requested that we edit the "WooHoo" message to be API rather than OpenAPI - @tatomyr is handling that.

One tiny correction on the changelog wording.

Otherwise this is very good and a nice step towards making our meaning clearer throughout the docs.

lornajane (approved):
Thanks for updating the command output!

PR #1258, opened 2023-09-12
Add `ignoreCase` for `tags-alphabetical` rule

justnero:
## What/Why/How?

For human perception, case doesn't really make a huge difference, especially when trying to find an item on the list. We used to have a custom assertion to do the same thing, but with an `ignoreCase` option to achieve the goal (see referenced PR).

## Reference

https://github.com/Rebilly/api-definitions/pull/1590

## Testing

Test cases were added to cover the option.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.changeset/forty-onions-deliver.md
+++ b/.changeset/forty-onions-deliver.md
@@ -1,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Added ignoreCase option for tags-alphabetical rule
--- a/.changeset/forty-onions-deliver.md
+++ b/.changeset/forty-onions-deliver.md
@@ -1,6 +1,6 @@
 ---
-'@redocly/openapi-core': patch
-'@redocly/cli': patch
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
 ---
 
-Added ignoreCase option for tags-alphabetical rule
+Added `ignoreCase` option for `tags-alphabetical` rule.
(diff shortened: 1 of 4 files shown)


RomanHotsiy:
@tatomyr we try to use `false` as a default value across all our configuration. So it's better to rename it to `caseSensitive` or `matchCase`.

justnero:
Won't it be a breaking change, though?

RomanHotsiy:
Oh, yes, I think you're right. I didn't check the PR. Our current behavior is to be case-sensitive.

I think we should keep it as is, @tatomyr.

tatomyr:
Ok then, let's keep it.

tatomyr (approved):
Thanks for the contribution @justnero!

.changeset/forty-onions-deliver.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Added ignoreCase option for tags-alphabetical rule
```
tatomyr, line 6: Oh, actually I think this should be a minor change, as this adds a new functionality. @justnero what do you think?

```suggestion
---
'@redocly/openapi-core': minor
'@redocly/cli': minor
---

Added `ignoreCase` option for `tags-alphabetical` rule.
```

justnero: Agreed, wasn't quite sure about this one.

tatomyr (approved):

PR #1259, opened 2023-09-13
fix: update redocly.yaml validation

tatomyr:
## What/Why/How?

Redocly config validation update. 
Copied and pasted validation schemas from the Portal, and added additional 'rules' validation (`metadata-schema` and `custom-fields-schema`). 

## Testing

Tested on Portal playground's and Rebilly redocly.yaml files.

## Check yourself

- [x] Code is linted

## Security


lornajane:
Are we asking docs in this pull request, or capturing in an issue what we need to add later? Link the issue if its separate

vital4u4ok (approved):

Oprysk (approved):

AlexVarchuk (approved):

tatomyr:
@lornajane I don't think there's something to be documented here. We are only catching up with the Portal config, no changes to the CLI config itself.

SmoliyY (approved):
It also will be great to see some e2e test for validation of config. 
At least for cases with `metadata-schema` and `custom-fields-schema`

lornajane:
Got it. I thought we were also adopting the rules and would need docs. Now I see it's just the validation

tatomyr:
@SmoliyY I see your point, but those don't make sense in the CLI itself as they are not quite rules and are only related to the Portal.

PR #1261, opened 2023-09-14
Lets talk about API Style Guides

philsturgeon:
## What/Why/How?

Updating README to talk about API Style Guides, to be consistent with how other tools talk about it. 

If this approach is good we can update the rest of the docs to follow along the same lines, downplaying "API Standards" (the word standard is already massively overloaded in the world of APIs) and hype up Rulesets are a technical implementation of an API Style Guide (Rulesets focused on helping you design and build better APIs) or OpenAPI Style Guides (Rulesets helping you write better OpenAPI, but doesn't really tell you how to write the API itself.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Redocly CLI
 
-[@Redocly](https://redocly.com) CLI is your all-in-one OpenAPI sidekiq. It can help you build, manage, improve, and quality-check your OpenAPI descriptions, all of which comes in handy for various phases of the API Lifecycle. Create API Style guides to make API governance easy, and publish beautiful API reference documentation. Supports OpenAPI 3.1, 3.0 and OpenAPI 2.0 (legacy Swagger).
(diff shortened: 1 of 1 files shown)


philsturgeon:
@lornajane 🙌🏻

lornajane:
I need to think more about this. I don't use the style guide language because it sounds a bit vague and optional, like "try to avoid passive voice in prose". It's more like coding standards or strict typing in how fundamental adopting this approach is. I'm not terribly worried about what other people call things, but I can see that "standard" is doing a lot of work!

lornajane:
Could we settle on something like:

API governance: the practice, process and tools of making APIs more excellent on purpose

Guidelines: the written word version of the standard expected from this API. These are the lines that will guide us to success.

Ruleset: programmatic checks that we run against an API.

Governance is the umbrella term, and includes both the guidelines and the rulesets... could this work and be understood?

philsturgeon:
Sure, its nice and consistent.

lornajane (approved):
Thanks @philsturgeon !

PR #1270, opened 2023-09-22
fix: improve detection accuracy on no-server-example.com rule

### LINKED ISSUE #1269: no-server-example.com
**Describe the bug**

https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/no-server-example.com.ts#L6

This only detects an exact match `localhost` or `example.com` but the intention is for it to match https://example.com and http://foo.example.com too (etc...).

**To Reproduce**

**Expected behavior**

Match: https://example.com and https://foo.example.com.

Perhaps we can change it to this:
```js
const pattern = /^(.*[\./])?(example\.com|localhost)/;

if (pattern.test(server.url)) {
    // ... rest of code ...
}
```

**Logs**

n/a

**OpenAPI description**

n/a

**Redocly Version(s)**

Latest (1.2)

**`Node.js` Version(s)**

18

**Additional context**

adamaltman:
## What/Why/How?

Improves the detection accuracy for the `no-server-example.com` rule.

## Reference

Closes #1269 

## Testing

Added tests and had to adjust multiple tests which used variants of example.com.

## Check yourself

- [ ] Code is linted

## Security


RomanHotsiy (approved):

adamaltman:
I'm having timeouts often when running tests in my local environment. I need help updating these snapshots or the various test server URLs to not be example.com or localhost.

tatomyr (approved):

Issue #1268, opened 2023-09-21
Join command fails to resolve nested references if used in multiple files


MatthiasGabriel:
**Describe the bug**

redocly join seems to have issues with resolving nested references if the object itself is referenced in multiple files.
If only the object is referenced only in one file the resolving works perfectly fine.

**To Reproduce**

I uploaded a showcase to https://github.com/MatthiasGabriel/redocly-join-showcase

The following command fails
`redocly join --without-x-tag-groups .\api.yaml .\foo\get\Get.yaml .\foo\create\Create.yaml .\foo\update\Update.yaml .\foo\list\List.yaml`

`[1] foo\shared\NestedFoo.yaml:4:5 at #/properties/bar

Can't resolve $ref

2 | properties:
3 |   bar:
4 |     $ref: './Bar.yaml'
5 |   fooBar:
6 |     type: string

Error was generated by the bundler rule.

❌ Errors encountered while bundling C:\Users\Matthias\Repositories\raiffeisen\devlan2\redocly-test\foo\create\Create.yaml: join will not proceed.
`

**Expected behavior**

**Logs**

-

**OpenAPI description**

-

**Redocly Version(s)**

1.2.0

**`Node.js` Version(s)**

v20.4.0

**Additional context**

All "single" endpoints itself get rendered correctly.
E.g. `redocly join --without-x-tag-groups .\api.yaml .\foo\get\Get.yaml`

If NestedFoo is replaced by SimpleFoo everything renders correctly.

Our original problem appeared when using subtypes (SubFoo), however I think the NestedFoo shows the problem a bit simpler.

tatomyr:
Thanks for reporting the bug!

unional:
I think I have a similar issue, which in my case involves referencing "sibling" references:

```yaml
# api-a/openapi.yaml
paths:
  /somepath:
    $ref: paths/somepath.yaml

# api-a/paths/somepath.yaml
get:
  ...snip...
  responses:
    "400":
      $ref: ../../common-spec/component/responses/problem.400.yaml

# common-spec/component/responses/problem.400.yaml
...snip...
content:
  application/problem+json:
    schema:
      # fail with "Can't resolve $ref here"
      $ref: ../schemas/problem.400.yaml
```

IgorKarpiuk:
Should be fixed in `1.2.1` version

MatthiasGabriel:
Hi @tatomyr  @IgorKarpiuk 

[...]

PR #1291, opened 2023-10-09
fix: use separate resolvers for each file in join

IgorKarpiuk:
## What/Why/How?
Fixed an issue with nested refs in the join command

The problem occurs when two files have a reference to one file, and this file contains another references

When we resolve multiple documents with one `BaseResolver`, we share a common cache. When we try to resolve an already cached document, we get a reference to the node in the cache
The bundle visitor mutates the node, which can be a reference to the cache . Therefore, for the next document, the values in the cache will have already been changed and the node may be invalid for this document, since it will contain a reference to the component section, not to the file.

Solution: Use a separate `BaseResolver` (with its own cache) for each file during  `bundle`

## Reference
 https://github.com/Redocly/redocly-cli/issues/1268
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.changeset/dull-ducks-beam.md
+++ b/.changeset/dull-ducks-beam.md
@@ -3,4 +3,4 @@
 '@redocly/cli': patch
 ---
 
-Fixed an issue with nested refs in the join command
+Fixed an issue with nested refs in the `join` command.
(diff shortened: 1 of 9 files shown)


.changeset/dull-ducks-beam.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Fixed an issue with nested refs in the join command
```
tatomyr, line 6:
```suggestion
Fixed an issue with nested refs in the join command.
```

SmoliyY (approved):

PR #1293, opened 2023-10-11
chore: add casing option for sentence case

TaylorKrusen:
## What/Why/How?
Added a new casing option, `Sentencecase`. Primary use case is for operation summaries. 

## Check yourself

- [x] Code is linted

## Security


packages/core/src/types/redocly-yaml.ts
```diff
@@ -298,6 +298,7 @@ const AssertionDefinitionAssertions: NodeType = {
         'MACRO_CASE',
         'COBOL-CASE',
         'flatcase',
+        'Sentencecase'
```
RomanHotsiy, line 301: should we rename it to `Sentence case` (with a space in it)?

I also wander if we should add `Title Case` too for symmetry.

lornajane:
I'll be honest: I'm not in favour of adding this and having to support it as part of our core functionality. The case checking works pretty well for how we use it now (mostly operationIds and URLs), but moving to prose checking is a whole additional world that I would not want to prioritise over our existing plans. It's not just the additional requirements of proper nouns, but would we also need to support markdown validation and formatting if we add this for prose fields. Don't get me started on title case ....

My advice: use a configurable rule for your use case, and let's save more-than-a-single-value validations for another day.

RomanHotsiy:
> My advice: use a configurable rule for your use case, and let's save more-than-a-single-value validations for another day.

But this PR adds implementation for configurable rules exactly. We have a `casing` assertion which has some predefined casing options. This PR just adds one or two more. Nothing else. It's not opinionated about any kind of specific prose fields.

Yes, we can use configurable rule with regexp but then I don't see why we should keep other casings? All of them can be replaced with regexp too.

lornajane:
The other formats are for single variables or paths, they don't have spaces in. These new cases are for fields that have written words, not just one value. It's a very different use case.

TaylorKrusen:
I think I'm in favor of doing this through a configurable rule rather than adding it to core. My reasoning:
- Sentence case checking via Regex is only predictable for summaries. Other fields may need capitalization for proper nouns or have multiple sentences.
- It's achievable using the tools we've already built-in, so using them to accomplish it (and teaching others to do the same) may have greater benefit longer term. 

Planning to close this PR in the next 24 hours unless I hear otherwise.

RomanHotsiy:
> Sentence case checking via Regex is only predictable for summaries. Other fields may need capitalization for proper nouns or have multiple sentences.

This makes sense 👍. Thanks.

lornajane:
Closing as discussed.

Issue #1205, opened 2023-08-02
Extend `split` command to optionally produce JSON output


lornajane:
**Is your feature request related to a problem? Please describe.**

When splitting an existing JSON-format API description, the resulting files are in YAML with no option to keep working with JSON.

**Describe the solution you'd like**

An option to get the split files as JSON

IgorKarpiuk:
@lornajane do you think we need the same for `join` command ?

As an option, we can add an option `--ext` like in  `bundle` command

lornajane:
I didn't realise we didn't have this on join! Yes please, although it could go into a separate issue if required.

About the formats: I think `--ext` is quite confusing, could we use:
 - the output filetype (so that we don't write yaml into `api.json`)
 - keep the same type as the first input file (if there is no output file so we're writing to stdout)

lornajane:
Fixed in https://github.com/Redocly/redocly-cli/pull/1305

PR #1305, opened 2023-10-19
feat: extend split and join commands to produce JSON output

IgorKarpiuk:
## What/Why/How?
 
Extend `split` and `join` commands to optionally produce JSON output.

### Split

Determine the output format by the extension of the input file, if the extension is not supported use `yaml`
Supported file extensions: `json`, `yaml`, `yml`

### Join 

Determine the output format by the extension of the file from `output` option, if output option is not provided use extension of first entry point, if extension not supported use `yaml`.

Supported file extensions: `json`, `yaml`, `yml`

## Reference

Related: https://github.com/Redocly/redocly-cli/issues/1205

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.changeset/plenty-mugs-suffer.md
+++ b/.changeset/plenty-mugs-suffer.md
@@ -2,4 +2,4 @@
 '@redocly/cli': minor
 ---
 
-Extended `split` and `join` commands to optionally produce JSON output.
+Add JSON output support to the `split` and `join` commands.
--- a/.changeset/plenty-mugs-suffer.md
+++ b/.changeset/plenty-mugs-suffer.md
@@ -1,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+Extended `split` and `join` commands to optionally produce JSON output.
--- a/.changeset/plenty-mugs-suffer.md
+++ b/.changeset/plenty-mugs-suffer.md
@@ -1,5 +1,5 @@
 ---
-'@redocly/cli': patch
+'@redocly/cli': minor
 ---
 
 Extended `split` and `join` commands to optionally produce JSON output.
--- a/docs/commands/join.md
+++ b/docs/commands/join.md
@@ -14,9 +14,9 @@ With Redocly CLI, you can solve this problem by using the `join` command that ca
 
 To easily distinguish the origin of OpenAPI objects and properties, you can optionally instruct the `join` command to append custom prefixes to them.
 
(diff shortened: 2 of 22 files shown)


.changeset/plenty-mugs-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': patch
```
tatomyr, line 2:
```suggestion
'@redocly/cli': minor
```

packages/cli/src/__tests__/commands/join.test.ts
```diff
@@ -80,7 +81,7 @@ describe('handleJoin fails', () => {
     );
   });
 
-  it('should call writeYaml function', async () => {
+  it('should call writeFileByExt function', async () => {
```
tatomyr, line 84: Do you mean this?

```suggestion
  it('should call writeToFileByExtension function', async () => {
```

packages/cli/src/utils.ts
```diff
@@ -219,6 +234,27 @@ export function writeYaml(data: any, filename: string, noRefs = false) {
...
+
+  if (process.env.NODE_ENV === 'test') {
+    process.stderr.write(content);
+    return;
+  }
+  fs.mkdirSync(dirname(filename), { recursive: true });
+  fs.writeFileSync(filename, content);
+}
+
+export function getAndValidateFileExtension(fileName: string): string {
```
tatomyr, line 248: I wonder if it makes sense to be more precise here:

```suggestion
export function getAndValidateFileExtension(fileName: string): 'yaml' | 'yml' | 'json' {
```

IgorKarpiuk: Make sense, Thank you

docs/commands/join.md
roman-sainchuk, line 1: Dont we need to update docs for split command as well ?

roman-sainchuk (approved):

.changeset/plenty-mugs-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': minor
+---
+
+Extended `split` and `join` commands to optionally produce JSON output.
```
lornajane, line 5:
```suggestion
Add JSON output support to the `split` and `join` commands.
```

lornajane (approved):
Tweak for the changelog language, but otherwise this looks good to me, I can't see where we would update the split command docs, the feature works as I expected it to, and I'm excited to be able to share this with the customer when we merge and release :)

PR #1308, opened 2023-10-23
chore: update typescript to v5 and synchronise predefined rulesets through strict typing

tatomyr:
## What/Why/How?

When adding rules, we have to export them from [index files](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-f277132551c50a77a6ebd53914aa50dc43d628d8675938b5769cc7a3eab1f0c1R56), add into our [predefined rulesets](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-f197d2a38d8cd9f2d09464deb9747226062f3ad06e30c0e77fcb6f8138f31719R3), and list in the [config validation file (redocly-yaml.ts)](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-2528117ea3cbe25dfbfd0583c9b11ba7baa7ef347184dfba32a73ec486aa1a5e). Currently they are not completely in sync with each other. Introducing strict typing helps to fix the discrepancies.

However, TS v4 works inconsistently, so I updated it to v5. This caused update/installation of several other packages (including Jest and ESLint). 

> [!NOTE] 
Most of the updated test files are merely a clean-up after upgrading to Jest v29. Exceptions are: [load.test.ts](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-79f56024a65109d10c766475e388ab5e6f998bc4a3697af2f03a9dcf0f8a6a56R175), [config-resolvers.test.ts.snap](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-d32d701bdd9d26b88498888c8c1c010798b3f526f1ee1609a4d44eb9fd2dcf99R1) and [this snapshot](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-d07667a6aabcd49c1059ea8906e66de4fee9049d6e308d25d0750afbd8ed99b8).

Nothing should change for users.

**Open questions:** 
- ~Should the previously omitted rules use severity '**off**' or '**error**' in the [**all**](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-b4ab3024ec27a1473c25552e3747d9a56224631682812b8bdd180c3affeb0c74R42) ruleset?~ -> 'error'
- Are we going to create a separate **async2** section in our rulesets like we have for exclusive OAS2/3 rules? (See [here](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-f197d2a38d8cd9f2d09464deb9747226062f3ad06e30c0e77fcb6f8138f31719R3)). cc @lornajane 

## Check yourself

- [x] Code is linted

## Security


packages/core/src/config/types.ts
```diff
@@ -37,16 +43,24 @@ export type PreprocessorConfig =
...
   extends?: string[];
   doNotResolveExamples?: boolean;
   recommendedFallback?: boolean;
 
-  rules?: Record<string, RuleConfig>;
-  oas2Rules?: Record<string, RuleConfig>;
-  oas3_0Rules?: Record<string, RuleConfig>;
-  oas3_1Rules?: Record<string, RuleConfig>;
+  rules?: T extends 'built-in'
+    ? { [key in BuiltInCommonRuleId | BuiltInCommonOASRuleId]: RuleConfig }
```
tatomyr, line 53: This is how we can enforce our built-in rulesets to be in sync with the rules that we validate in [redocly-yaml.ts](https://github.com/Redocly/redocly-cli/pull/1308/files#diff-2528117ea3cbe25dfbfd0583c9b11ba7baa7ef347184dfba32a73ec486aa1a5eR7).

lornajane:
I think the async2 rules (and in future async3) should have the same sections as the oas2/3 ones do, does that answer the question?

Also I have a question about where you removed all the `off` rules. Does this make them show up as unused as a warning in the command output? The Async2 linting without any parameters produces more output than I expected and I have been meaning to look if that's because I haven't configured all the other rules to be `off`.

packages/core/src/walk.ts
```diff
@@ -185,7 +198,7 @@ export function walkDocument<T>(opts: {
 
       const anyEnterVisitors = normalizedVisitors.any.enter;
       const currentEnterVisitors = anyEnterVisitors.concat(
-        normalizedVisitors[type.name]?.enter || []
+        (normalizedVisitors[type.name]?.enter as NormalizeVisitor<VisitorNode<any>[]>) || []
```
IgorKarpiuk, line 201: Can we use `unknown` instead of `any` here ?
```suggestion
        (normalizedVisitors[type.name]?.enter as NormalizeVisitor<VisitorNode<unknown>[]>) || []
```

packages/core/src/config/types.ts
IgorKarpiuk: Maybe we can add a separate type for this, something like

```ts
type RuleMap<T, V> = T extends 'built-in' ? { [key in V]: RuleConfig } : Record<string, RuleConfig>;
```

and use it 

```ts
  rules?: RuleMap<T, BuiltInCommonRuleId | BuiltInCommonOASRuleId>;
  oas2Rules?: RuleMap<T, BuiltInOAS2RuleId>;
  oas3_0Rules?: RuleMap<T, BuiltInOAS3RuleId>;
  oas3_1Rules?: RuleMap<T, BuiltInOAS3RuleId>;
```

tatomyr:
> I think the async2 rules (and in future async3) should have the same sections as the oas2/3 ones do

Ok, I'll add `channels-kebab-case` and `no-channel-trailing-slash` to the rulesets. Which severity should it be for minimal/recommended? 

> The Async2 linting without any parameters produces more output than I expected 

@lornajane this is partially related to my question :) I believe it's because the `rules` section in the rulesets implies that the linter runs ALL of them for any API docs type (OAS2/3 and Async2). If we want to show only appropriate ones, we have to reconsider the `rules` to only contain a really COMMON subset of rules, and move all that don't match into separate sections.

lornajane:
I think only `spec` (error) and `operation-operationId` (warn) rules make sense for the recommended ruleset for AsyncAPI, the others are rather a matter of taste (and the way AsyncAPI does tags is quite weird so probably those rules need some attention anyway). We may adjust this over time as we learn how people use the rules in their own projects.

roman-sainchuk (approved):
LGTM 👍

IgorKarpiuk (approved):
👍

PR #1309, opened 2023-10-24
fix: use same host for websocket server as for http server when previewing locally

f3ng3r:
## What/Why/How?
Today the WS server used for hot reloading is initialized without specifying a host. But in the `hot.js` file, `127.0.0.1` is hardcoded as the host it should connect to. This breaks in some cases, such as when the websocket server binds to ipv6, and therefore isn't available on `127.0.0.1`.

This PR changes it so that the host provided via `--host` or `-h` is passed on to the WS server as well, not only to the http server. It also adds this host as a global variable, alongside the `wsPort` variable, so that `hot.js` can pick up the correct host when connecting to the websocket; instead of the hardcoded host today.

## Testing
No tests seem to exist for the preview-server functionality and I have not added any. Instead I have tested manually with a variety of commonly used hosts (`::`, `0.0.0.0`, `::1`, `localhost`, `127.0.0.1`) to confirm that they work.

## Check yourself

- [X] Code is linted

## Security

### DIFF
--- a/.changeset/strange-insects-jog.md
+++ b/.changeset/strange-insects-jog.md
@@ -1,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+The --host/-h argument provided for the preview http server, is now also used by the ws server.
--- a/.changeset/strange-insects-jog.md
+++ b/.changeset/strange-insects-jog.md
@@ -2,4 +2,4 @@
 '@redocly/cli': patch
 ---
 
-The --host/-h argument provided for the preview http server, is now also used by the ws server.
+The `--host/-h` argument in the `preview-docs` command is now also used by the WebSocket server for hot reloading.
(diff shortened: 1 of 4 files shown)


tatomyr:
Hi @f3ng3r! 
Thank you for your contribution! Will take a look at this shortly.

.changeset/strange-insects-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+The --host/-h argument provided for the preview http server, is now also used by the ws server.
```
tatomyr, line 5:
```suggestion
The `--host/-h` argument in the `preview-docs` command is now also used by the WebSocket server for hot reloading.
```

IgorKarpiuk (approved):
LGTM

Could you check if any docs need to be updated?

Issue #529, opened 2022-01-25
Working behind proxy


joaonlopes:
**Is your feature request related to a problem? Please describe.**

We need our continuous integration to upload the openapi yaml to the Redocly API registry, but we always get time out, we think the reason for this timeout is that we are behind an enterprise proxy.

`FetchError: request to https://api.redoc.ly/registry failed, reason: connect ETIMEDOUT 34.238.30.35:44
    at ClientRequest.<anonymous> (C:\ (...) \AppData\Roaming\npm\node_modules\@redocly\openapi-cli\node_modules\node-fetch\lib\index.js:1491:11)
    at ClientRequest.emit (node:events:390:28)
    at TLSSocket.socketErrorListener (node:_http_client:447:9)
    at TLSSocket.emit (node:events:390:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  type: 'system',
  errno: 'ETIMEDOUT',
  code: 'ETIMEDOUT'
}`

Authorization failed. Please check if you entered a valid API key.

**Describe the solution you'd like**

We would like openapi cli working behind a proxy server.

**Describe alternatives you've considered**

We did a quick modification to your code in _registry-api.ts_ to handle the proxy connection:

```
const HttpsProxyAgent = require('https-proxy-agent');
private async request(path = '', options: RequestInit = {}, region?: Region) {
    const headers = Object.assign({}, options.headers || {}, { 'x-redocly-cli-version': version });
    if (!headers.hasOwnProperty('authorization')) { throw new Error('Unauthorized'); }
const proxyAgent = new HttpsProxyAgent('XXX:XXX/');
    const response = await fetch(`${this.getBaseUrl(region)}${path}`,  Object.assign({}, options, { headers }, { agent: proxyAgent}) );
    if (response.status === 401) { throw new Error('Unauthorized'); }
    if (response.status === 404) {
      const body: RegistryApiTypes.NotFoundProblemResponse = await response.json();
      throw new Error(body.code);
    }
    return response;
  }
```

(see: https://www.scrapingbee.com/blog/proxy-node-fetch/)

This seems to work for the login command but still timesout for the push command:

`
FetchError: request to (...) host failed, reason: connect ETIMEDOUT 52.216.108.107:443
[...]

PR #1321, opened 2023-11-08
feat: basic push-status command implementation and extended push command 

roman-sainchuk:
## What/Why/How?

Now `push` command also applied for BH push. 
Syntax for push command:

```sh
REDOCLY_AUTHORIZATION=<api key> \ 
npm run cli -- push \ 
    --organization <orgId> \
    --project <projectId> \ 
    --mount-path <blue harvest mount path> \
    --branch <branch where files are pushed from> \
    --author 'Author Name <author-email@mail.com>' \
    --message <message files are committed with> \
    --files <folders or/and files>  \
    --domain 'http://localhost'
```

or

```sh
REDOCLY_AUTHORIZATION=<api key> \ 
REDOCLY_DOMAIN='http://localhost' \
npm run cli -- push  \ 
    --organization <orgId> \
    --project <projectId> \ 
    --mount-path <blue harvest mount path> \
    --branch <branch where files are pushed from> \
    --author 'Author Name <author-email@mail.com>' \
    --files <folders or/and files> \
    --message <message files are committed with> \
```

Syntax for `push-status` command: 

```sh
REDOCLY_AUTHORIZATION=<api key> \ 
npm run cli -- push-status  <pushId> \
    --organization <orgId> \
    --project <projectId> \ 
    --domain 'http://localhost' \
    --wait
```

or 

```sh
REDOCLY_AUTHORIZATION=<api key> \ 
REDOCLY_DOMAIN='http://localhost' \
npm run cli -- push-status  <pushId>
    --organization <orgId> \
    --project <projectId> \ 
    --wait
```

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/push-bh.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const [name, email] = author.split('<');
+
+  return {
+    name: name.trim(),
+    email: email.replace('>', '').trim(),
+  };
+}
+
+function collectFilesToPush(files: string[]): FileToUpload[] {
```
tatomyr, line 111: Do we preserve nested folder structure? There could be references based on paths.

lornajane:
We need to discuss the naming of this command, either to somehow reuse the `push` command, or to use `[product] push`, or to pick a different word altogether. Adding the suffix won't be clear to users.

roman-sainchuk:
> We need to discuss the naming of this command, either to somehow reuse the `push` command, or to use `[product] push`, or to pick a different word altogether. Adding the suffix won't be clear to users.

Agree, for now it's not clear how to properly distinguish which command to run when you hit `push`. We discussed `[product] push` approach yesterday with @tatomyr, but looks like this approach will confuse users even more.

I'm also thinking about [docker approach with contexts](https://docs.docker.com/engine/context/working-with-contexts/). Example:

`redocly context create bh-eu --domain 'https://auth.beta.eu.redocly.com/' --description 'EU Region'`

And then:

`redocly context use bh-eu`
and
`redocly push ...`

@lornajane , @RomanHotsiy , @tatomyr what do you think?

roman-sainchuk:
The default context will be hardcoded as `https://app.beta.redocly.com/`, but looks like it is breaking change

RomanHotsiy:
Yeah, I think the suffix is a temporary solution not to get blocked. I don't have an answer now. Let me think a bit.

packages/cli/src/commands/push-bh.ts
roman-sainchuk: Yes, we preserve it relative to entry points
e.g `push ../../folder --mount-path /docs/remotes`

Lets say content of the `../../folder` is next:

```treeview
├── app
│   ├── index.html
├── openapi.yaml
└── some-ref.yaml
``` 

The `/docs/remotes` (in BlueHarvest) will have the same content

tatomyr: 👍

lornajane:
It's fine to work on it with the suffix, but we can't merge until we figure out a better approach. I think we'll end up with other commands that also have the product names in, so maybe adopting that extra piece is the least confusing option.

I'm not in favour of introducing the context as it's quite complex and since the push command is mostly intended to run in CI, it should be a repeatable single command.

roman-sainchuk:
> either to somehow reuse the push command

This is not an option because the syntax and options are totally different, also it will be confusing for user to understand what options should be provided to satisfy required arguments for chosen product (e.g. `push --help`)

lornajane:
Perhaps we can pick a different command name, to keep things very clear and use that language consistently. How about `upload` or `commit` ? So far we have only decided on options we can't use, so I'm trying to find things that could be ruled in rather than ruled out, if that makes sense.

packages/cli/src/blue-harvest/api-keys.ts
```diff
@@ -0,0 +1,30 @@
+import { resolve } from 'path';
+import { homedir } from 'os';
+import { existsSync, readFileSync } from 'fs';
+
+const TOKEN_FILENAME = '.redocly-config.json';
+
+function isNotEmptyObject(obj: any) {
```
tatomyr, line 7: Please use the one from `core/src/utils`.

adamaltman:
@roman-sainchuk I dislike the context proposal a lot. I like this one a lot:
```
   REDOCLY_DOMAIN='http://localhost' \
```

packages/cli/src/blue-harvest/api-keys.ts
roman-sainchuk: I cannot, should I export it from there?

roman-sainchuk:
@adamaltman me too, but on order to have the same command name and proper distinguishing between contexts (workflows/BH) we need to pass it even when you want to get the hint, e.g.:

```sh
REDOCLY_DOMAIN='http://localhost' redocly push --help
```

otherwise it will print  workflows version of the `push` syntax

packages/cli/src/blue-harvest/api-keys.ts
tatomyr: It is already being exported: https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/utils.ts#L177

packages/core/src/redocly/index.ts
```diff
@@ -8,6 +8,8 @@ import { RegionalToken, RegionalTokenWithValidity } from './redocly-client-types
 import { isNotEmptyObject } from '../utils';
 import { colorize } from '../logger';
 
+export * from './blue-harvest';
```
tatomyr, line 11: I think this naming will be confusing for external contributors.

packages/cli/src/cms/commands/__tests__/push-status.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { PushStatusResponse } from '@redocly/openapi-core';
+
+const remotes = {
+  getPushStatus: jest.fn(),
+  getRemotesList: jest.fn(),
+};
+
+jest.mock('@redocly/openapi-core', () => ({
+  ...jest.requireActual('@redocly/openapi-core'),
+  BlueHarvestApiClient: jest.fn().mockImplementation(function (this: any, ...args) {
```
adamaltman, line 11: There shouldn't be blue harvest in code. It should be called Reunite.

packages/cli/src/cms/domains.ts
```diff
@@ -0,0 +1,11 @@
+const DEFAULT_DOMAIN = 'https://app.beta.redocly.com';
```
adamaltman, line 1:
```suggestion
const DEFAULT_DOMAIN = 'https://app.cloud.redocly.com';
```

tatomyr:
I suggest moving the `redocly/cloud` (and possibly all `redocly`) folder to the CLI package from core. The intention is to use this command in CI/CD pipelines, so it looks like we can make the `openapi-core` bundle smaller if we don't include this things there.

packages/core/src/redocly/cloud/api-client.ts
```diff
@@ -0,0 +1,199 @@
...
+  async getPush({
+    organizationId,
+    projectId,
+    pushId,
+  }: {
+    organizationId: string;
+    projectId: string;
+    pushId: string;
+  }) {
+    const response = await fetchWithTimeout(
```
tatomyr, line 151: Please take into consideration the possibility of the `push` command working behind a proxy. Please consult to the [original issue](https://github.com/Redocly/redocly-cli/issues/529) and the suggested [fork](https://github.com/joaonlopes/redocly-cli-behind-proxy). 
You might also find this closed [MR](https://github.com/Redocly/redocly-cli/pull/1173) useful in that context.

packages/cli/src/index.ts
```diff
@@ -146,23 +147,75 @@ yargs
...
+          domain: { description: 'Specify a domain.', alias: 'd', type: 'string' },
+          wait: {
+            description: 'Wait for build to finish.',
+            type: 'boolean',
+          },
+          'max-execution-time': {
+            description: 'Maximum execution time in seconds.',
+            type: 'number',
+            default: 600,
+          },
```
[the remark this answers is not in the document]
IgorKarpiuk: If the user passes `--wait` without a value, it will overwrite the default value. and I'm not sure it's a good idea to have two types for one option (`boolean` and `number`)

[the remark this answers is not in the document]
tatomyr: Could it be an optional `number` instead?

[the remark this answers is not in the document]
roman-sainchuk: I think we wanna mix unrelated things 
How will it be used in case we make it optional `number` ? 
Passing `--wait` without argument will not be possible I think, so what we have to pass when we wanna wait until end of deployments?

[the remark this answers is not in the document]
tatomyr: > we wanna wait until end of deployments?

Is there a reason why we don't want to use some default  timeout (a big one) in that case? Is it practical to allow operations without any time limit?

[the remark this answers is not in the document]
IgorKarpiuk: I think that the deployment process will depend on the size of the portal, so it is worthwhile to provide the ability to change this value. The time limit protects us from endless waiting if something goes wrong, so It seems that it is always better to have a time limit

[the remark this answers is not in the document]
tatomyr: >  better to have a time limit

So it looks like those options don't make too much sense one without the other.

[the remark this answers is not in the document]
IgorKarpiuk: >So it looks like those options don't make too much sense one without the other.

This is a fact

packages/cli/src/update-version-notifier.ts
```diff
@@ -2,7 +2,7 @@ import { tmpdir } from 'os';
 import { join } from 'path';
 import { existsSync, writeFileSync, readFileSync, statSync } from 'fs';
 import { compare } from 'semver';
-import fetch from './fetch-with-timeout';
+import fetch from '@redocly/openapi-core/lib/fetch-with-timeout';
```
tatomyr, line 5: Is it imported from the core?

packages/cli/src/cms/api/api-client.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    if (response.ok) {
+      return responseBody as T;
+    }
+
+    throw new Error(responseBody.title || response.statusText);
+  }
+
+  async getDefaultBranch(organizationId: string, projectId: string) {
+    const response = await fetch(
```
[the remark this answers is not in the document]
IgorKarpiuk: Because we expect an error if something goes wrong, but `fetchWithTimeout` catch the errors and just return `undefined`

packages/core/src/redocly/cloud/api-client.ts
IgorKarpiuk: I added `abort-controller` package and moved new client to `cli` package. Also created issue to move old [client and handle proxy](https://github.com/Redocly/redocly-cli/issues/1404)

packages/cli/src/index.ts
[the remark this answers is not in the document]
tatomyr: If you leave this as it is, please make `wait` required for `max-execution-time`.

packages/cli/src/utils.ts
```diff
@@ -659,3 +662,29 @@ export function checkForDeprecatedOptions<T>(argv: T, deprecatedOptions: Array<k
...
+
+export function findAndApplyPushHandler(argv: PushArguments) {
+  if (argv.project && argv['mount-path']) {
+    if (!argv.message || !argv.author || !argv.branch) {
+      process.stdout.write(
+        'Error: message, author and branch are required for push command to the CMS'
+      );
+      return;
+    }
+    commandWrapper(handleCMSPush)({
```
tatomyr, line 674: I'd leave the `commandWrapper` in the index file.

packages/cli/src/index.ts
```diff
@@ -446,6 +566,11 @@ yargs
           description: 'Project directory.',
           default: '.',
         },
+        config: {
```
tatomyr, line 569: Does the `preview` command needs the `config` property? Should we add it in this MR?

tatomyr (commented):
Thank you @IgorKarpiuk! Looks much better now. Left one minor comment.

packages/cli/src/index.ts
IgorKarpiuk: I think this was added during conflict resolution, `PreviewProjectOptions` contain the `config` option, so it should be there.

IgorKarpiuk: Removed

tatomyr (approved):

tatomyr:
Please don't forget to add a changeset (I believe it's needed here).

PR #1324, opened 2023-11-10
docs: Use a single sidebar for all Redocly CLI content

lornajane:
## What/Why/How?

The decorators had their own sub sidebar which meant those pages missed linking users to the rest of our content and could be disorientating. This pull request removes the extra sidebar, the decorators are already listed in the top-level sidebar for this project.

## Check yourself

- [x] Code is linted

## Security


tatomyr:
@lornajane could you attach a screenshot? I cannot figure out what has changed.

adamaltman (approved):

adamaltman:
@tatomyr it changes the sidebar behavior. You can see a preview of the website in the checks there is a link.

image

lornajane:
@tatomyr is right, the live site does the same thing as the preview at this point. But before we moved to the new platform, the decorators had a weird sidebar and I never understood why. I had deleted it on my local version, and now I can't see that it does anything any more so I removed it. As long as you think the current behaviour is okay, I think we're good.

tatomyr (approved):
Thanks for the explanation 👍

Issue #1317, opened 2023-11-02
New rule to enforce array parameters format


tatomyr:
**Is your feature request related to a problem? Please describe.**

There's no built-in rule to enforce explicit descriptions of array-type parameters (in particular, having `style` and `explode` properties set). It's possible to define a configurable rule to require those fields for all types of parameters, but they don't really make sense for non-array ones. 

**Describe the solution you'd like**

A rule similar to [boolean-parameter-prfixes](https://redocly.com/docs/cli/rules/boolean-parameter-prefixes/) that enforces explicit description of  parameters of array type, so properties `style` and `explode` are required:

```yaml
components:
  parameters:
    ArrayParameter:
       in: query
       style: form      # <-- required
       explode: true  # <-- required
       schema:
           type: array # <-- either explicitly defined as an array
           items: ...     # <-- or implicitly by having the `items` field

       
```

Naming to be discussed, maybe `explicit-parameter-format`? 

It also could have some configuration, like the following:

```yaml
rules:
  explicit-parameter-format: 
     severity: error
     in: 
        - header
        - query
```

**Describe alternatives you've considered**

Writing a custom rule or requiring `style` and `explode` fields to be set for all types of parameters via a configurable rule -- each option has its drawbacks.

**Additional context**

Related discussion: https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1698886138140729

RomanHotsiy:
I raised it to p2: the impact is minimal but the confidence is high and the implementation is very easy.

RomanHotsiy:
Raising to p1. It's getting important for one of our customers.

lornajane:
Rule naming: how about `array-parameter-fields` since it requires additional fields on the array params, and the rule that applies only to boolean params is named `boolean-parameter....`

RomanHotsiy:
I like the `array-parameter-` prefix. How about `array-parameter-serialization`?

lornajane:
The `style` parameter is about serialization but I'm not sure that `explode` is for that purpose, or perhaps not only for that purpose - and so I ended up with the more vague `fields` suffix. The relevant specification section: https://spec.openapis.org/oas/v3.1.0#parameter-object

I don't feel strongly either way about using `serialization`

RomanHotsiy:
I believe `explode` is about serialization too. Check this out: https://swagger.io/docs/specification/serialization/

tatomyr:
Resolved in v1.5.0.

PR #1338, opened 2023-11-22
feat: create new rule to enforce array parameters format

IgorKarpiuk:
## What/Why/How?

Create new rule `array-parameter-serialization` rule to enforce array parameters format.

This Rule enforces to define  fields `explode` and `style` for parameter if param's schema type is `array` or has `items` or `preffixItems`

## Testing
https://github.com/Redocly/redocly-cli/issues/1317
## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/.changeset/silver-singers-glow.md
+++ b/.changeset/silver-singers-glow.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added new rule `array-parameter-serialization` to enforce array parameters format.
--- a/.changeset/silver-singers-glow.md
+++ b/.changeset/silver-singers-glow.md
@@ -3,4 +3,4 @@
 "@redocly/cli": minor
 ---
 
-Added new rule `array-parameter-serialization` to enforce array parameters format.
+Added new rule `array-parameter-serialization` to require that serialization parameters `style` and `explode` are present on array parameters.
--- a/docs/rules/array-parameter-serialization.md
+++ b/docs/rules/array-parameter-serialization.md
@@ -105,6 +105,5 @@ paths:
 
 ## Resources
 
-- [Rule source for OAS 2.0](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas2/boolean-parameter-prefixes.ts)
-- [Rule source for OAS 3.0 and 3.1](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/boolean-parameter-prefixes.ts)
+- [Rule source for OAS 3.0 and 3.1](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/array-parameter-serialization.ts)
 - [OpenAPI Parameter](https://redocly.com/docs/openapi-visual-reference/parameter/) docs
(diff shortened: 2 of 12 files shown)


lornajane:
Could we get a changeset please?

The rule name doesn't need `-rule` in it, but the rest is good :)

.changeset/silver-singers-glow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added new rule `array-parameter-serialization` to enforce array parameters format.
```
lornajane, line 6:
```suggestion
Added new rule `array-parameter-serialization` to require that serialization parameters `style` and `explode` are present on array parameters.
```

lornajane (commented):
Minor changelog suggestion, I haven't tried the feature itself yet

docs/rules/array-parameter-serialization.md
```diff
@@ -0,0 +1,110 @@
...
+- [configurable rules](./configurable-rules.md)
+- [boolean-parameter-prefixes](./boolean-parameter-prefixes.md)
+- [no-invalid-parameter-examples](./no-invalid-parameter-examples.md)
+- [parameter-description](./parameter-description.md)
+- [operation-parameters-unique](./operation-parameters-unique.md)
+
+## Resources
+
+- [Rule source for OAS 2.0](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas2/boolean-parameter-prefixes.ts)
+- [Rule source for OAS 3.0 and 3.1](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/boolean-parameter-prefixes.ts)
```
tatomyr, line 109: Aren't those rather related rules than resources, are they?

docs/rules/array-parameter-serialization.md
```diff
@@ -0,0 +1,110 @@
...
+          explode: true
+          schema:
+            type: array
+            items:
+              type: string
+```
+
+## Related rules
+
+- [configurable rules](./configurable-rules.md)
```
tatomyr, line 100: Is it related? In what way?

lornajane:
I think I broke something! The following snippet does return as valid, which is correct:

```
      parameters:
        - name: f
          in: query
          style: label
          explode: true
          schema:
            type: array
            items:
              type: string
```

However if the `explode` field is set to `false` (it's boolean, so it might well be set to false!) then I get the error `Parameter `f` should have `style` and `explode ` fields`. Could you take another look?

docs/rules/array-parameter-serialization.md
IgorKarpiuk: Just wrong links, fixed, thanks

docs/rules/array-parameter-serialization.md
IgorKarpiuk: I believe yes, we created this rule from similar configurable rules.

IgorKarpiuk:
Thanks @lornajane I updated the code and added it to test

tatomyr (approved):

IgorKarpiuk:
@lornajane here is the test https://github.com/Redocly/redocly-cli/pull/1338/files#diff-4cf43b18ff2bdc2e4cdc694067da4ae9be20137568cc0196702c01433b0e0ebaR161

lornajane (approved):
Working as expected now (that failing link check is because we can't publish docs at the moment, so feel free to merge regardless)

PR #1339, opened 2023-11-23
Fix schema type of AsyncAPI operation tags

shishkin:
## What/Why/How?

Tags type on operation and operation trait wasn't using available TagList type that other elements used.

## Reference

https://www.asyncapi.com/docs/reference/specification/v2.6.0#tagObject

## Testing

Added an example of operation with tags for linting.

## Screenshots (optional)

N/A

## Check yourself

- [x] Code is linted

## Security


lornajane:
The tests are here: https://github.com/Redocly/redocly-cli/tree/main/__tests__/lint/async2 - but they're really just the examples supplied by the AsyncAPI project to make sure we're not breaking anything.

Tags do not work the same way in OpenAPI (where they are defined at the top level and then used in the operations) and in AsyncAPI (where they are more freeform within each operation or trait), which is why the objects aren't re-used. Could you say something about your use case and if you saw a problem with the AsyncAPI linting?

shishkin:
Yes, I tried to lint a valid AsyncAPI spec and got a bunch of errors like this:

```
{
  "ruleId": "spec",
  "severity": "warn",
  "message": "Expected type `string` but got `object`.",
  "location": [
    {
      "source": {
        "absoluteRef": "/",
        "body": "..."
      },
      "pointer": "#/channels/my-channel/publish/tags/0",
      "reportOnKey": false
    }
  ],
  "suggest": []
}
```

The AsyncAPI spec I linked in the PR doesn't separate tag definition and usage. An example of an operation with tags from the spec is as follows:

```yaml
operationId: registerUser
summary: Action to sign a user up.
description: A longer description
security:
  - petstore_auth:
    - write:pets
    - read:pets
tags:
  - name: user
  - name: signup
  - name: register
message:
  headers:
    type: object
    properties:
      applicationInstanceId:
        description: Unique identifier for a given instance of the publishing application
        type: string
  payload:
    type: object
    properties:
      user:
        $ref: "#/components/schemas/userCreate"
      signup:
        $ref: "#/components/schemas/signup"
bindings:
  amqp:
    ack: false
traits:
  - $ref: "#/components/operationTraits/kafka"
```

shishkin:
I've added a test that passes with this change, but fails if the change is reverted.

lornajane (approved):
Thanks very much for the pull request, the explanation and the test. It took us a while to verify, so thanks for your patience!

PR #1345, opened 2023-12-01
feat: resolve refs in config

tatomyr:
## What/Why/How?

This allows using $ref in config files like this:

```yaml
# redocly.yaml
seo:
  $ref: seo.yaml
rbac: 
  $ref: rbac.yaml
```

⚠️  When using the `push` command with $ref's in config file, user should explicitly upload all referenced files in the config using `--files` option.

- [x] Resolve $refs in config files
- [x] Add basic linting support
- [x] Extend config node types to be linted
- [x] Add basic support for VS Code extension
- [x] Check compatibility with other tools (scout and portal)
- [ ] ~Look into https://github.com/Redocly/redoc/issues/740#issuecomment-1670740699~ (is not related; better to be handled in a separate MR)

Also fixed a couple of issues with resolving refs and formatting the output.

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/1219

## Check yourself

- [x] Code is linted
- [x] Tested with redoc/reference-docs/workflows (internal)
- [x] All new/updated code is covered with tests

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #1364, opened 2023-12-21
docs: explain using refs in a config file

tatomyr:
## What/Why/How?

Added docs for $refs in the config file: https://github.com/Redocly/redocly-cli/pull/1345.

Merge this alongside [the release](https://github.com/Redocly/redocly-cli/pull/1361).

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -231,7 +231,7 @@ Therefore, only the header from the first match is used in the request.
 ### Split up the configuration file
 
 As your config file grows, you may want to split it into multiple parts.
-That is possible by using references in a config similar to how they are used in OpenAPI descriptions:
+Splitting a config file is possible by using references in a config similar to how they are used in OpenAPI descriptions:
 
 ```yaml
 extends:
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -227,3 +227,23 @@ resolve:
 
 The first match takes priority when a URL matches multiple patterns.
 Therefore, only the header from the first match is used in the request.
+
+### Split up the configuration file
+
+As your config file grows, you may want to split it into multiple parts.
+That is possible by using references in a config similar to how they are used in OpenAPI descriptions:
+
+```yaml
+extends:
+  - recommended
+
+theme:
+  openapi:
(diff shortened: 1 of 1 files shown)


docs/configuration/index.md
```diff
@@ -227,3 +227,23 @@ resolve:
 
 The first match takes priority when a URL matches multiple patterns.
 Therefore, only the header from the first match is used in the request.
+
+### Split up the configuration file
+
+As your config file grows, you may want to split it into multiple parts.
+That is possible by using references in a config similar to how they are used in OpenAPI descriptions:
```
HCloward, line 234:
```suggestion
Splitting a config file is possible by using references in a config similar to how they are used in OpenAPI descriptions:
```

docs/configuration/index.md
```diff
@@ -227,3 +227,23 @@ resolve:
...
+
+theme:
+  openapi:
+    $ref: ./openapi-theme.yaml
+  mockServer:
+    $ref: ./mockserver.yaml
+```
+
+{% admonition type="attention" %}
+When using the `push` command with `$ref`s in a config file, all referenced files should explicitly upload using the `--files` option
```
HCloward, line 248:
```suggestion
When using the `push` command with a config file that includes `$ref`s , all referenced files are explicitly uploaded using the `--files` option.
```

HCloward (approved):
Made a couple of suggestions, otherwise looks good to me.

docs/configuration/index.md
tatomyr: @HCloward would the following work instead?

```suggestion
When using the `push` command with a config file that includes `$ref`s, all referenced files should be explicitly uploaded using the `--files` option.
```

HCloward: yep, that works - why "should" though -  is there a possibility it won't work for some reason?

tatomyr: Not really. It means that users should manually specify all the parts that need to be uploaded in the `--files` option. Do you think "are" fits here better?

tatomyr: Ok, thanks for the clarification!

Issue #474, opened 2021-12-15
Support specifying hostname for preview-docs command


Nadock:
**Is your feature request related to a problem? Please describe.**
We run all our tooling in Docker containers, except the `openapi-cli` because we can't tell the `preview-docs` command to bind to `0.0.0.0` instead of `127.0.0.1` inside the container.

**Describe the solution you'd like**
Add a `-h`/`--host` argument to the `preview-docs` command, similar to the `-p`/`--port` argument that already exists.

PR #1367, opened 2023-12-22
feat: add project preview command

volodymyr-rutskyi:
## What/Why/How?

Add a `preview` command that allows to run a preview of Redocly projects.
The command can accept the `product` to use for preview as an argument, deduce it fro the project's `package.json`, or use Realm by default.

The product package is executed via NPX.

The supported products are: 
* Redoc: `@redocly/redoc`
* Revel: `@redocly/revel`
* Reef: `@redocly/reef`
* Realm: `@redocly/realm`
* Redoc + Revel: `@redocly/redoc-revel`
* Redoc + Reef: `@redocly/redoc-reef`
* Revel + Reef: `@redocly/revel-reef`

## Reference

Docs PR: https://github.com/Redocly/redocly-cli/pull/1384

## Screenshots (optional)

image

## Check yourself

- [ ] Code is linted

## Security


packages/cli/src/index.ts
```diff
@@ -410,6 +412,36 @@ yargs
...
+          type: 'string',
+          choices: ['redoc', 'revel', 'reef', 'realm', 'redoc-revel', 'redoc-reef', 'revel-reef'],
+        },
+        plan: {
+          type: 'string',
+          choices: PRODUCT_PLANS,
+          default: 'enterprise',
+          description: '',
+        },
+        port: {
```
tatomyr, line 430: I don't see the `port` to be used anywhere. Is it intentional? 

Do we also need the `host` option similar to [this](https://github.com/Redocly/redocly-cli/blob/main/packages/cli/src/index.ts#L431)?
See [the issue](https://github.com/Redocly/redocly-cli/issues/474) for the `preview-docs` command.

packages/cli/src/commands/preview-project/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (!isValidProduct(product)) {
+    process.stderr.write(`Invalid product ${product}`);
+    throw new Error(`Project preview launch failed`);
+  }
+
+  const productName = PRODUCT_NAMES[product];
+  const packageName = PRODUCT_PACKAGES[product];
+
+  process.stdout.write(`\nLaunching preview of ${productName} ${args.plan} using NPX\n\n`);
```
tatomyr, line 21:
```suggestion
  process.stdout.write(`\nLaunching preview of ${productName} ${plan} using NPX\n\n`);
```

packages/cli/src/commands/preview-project/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+  'redoc-revel': '@redocly/redoc-revel',
+  'revel-reef': '@redocly/revel-reef',
+  'redoc-reef': '@redocly/redoc-reef',
+  redoc: '@redocly/redoc',
+  revel: '@redocly/revel',
+  reef: '@redocly/reef',
+};
+
+export const PRODUCT_NAMES: { [key in Product]: string } = {
+  redoc: 'Redoc',
```
tatomyr, line 14: Since the command makes `redoc` available for preview, among other options, it looks like there's no need for the `preview-docs` command anymore. Are we going to deprecate it at some point? If yes, when do we want to add the deprecation message?

packages/cli/src/index.ts
```diff
@@ -410,6 +412,36 @@ yargs
...
+          alias: 'p',
+          type: 'number',
+          description: 'Preview port.',
+        },
+        'project-dir': {
```
tatomyr, line 435: Does it make sense to point to a specific file instead of a directory for `redoc`?

packages/cli/src/commands/preview-project/constants.ts
volodymyr-rutskyi: This is a little confusing, but for now the `redoc` that can be previewed with the `preview-docs` command is not the current open source Redoc. The `preview-docs` command will stay relevant for a some more time.

packages/cli/src/index.ts
volodymyr-rutskyi: This is not the open source Redoc and this one can only accept a directory.

packages/cli/src/index.ts
volodymyr-rutskyi: Nice catch. Added use of the port argument. 

The `0.0.0.0` address is not allowed in the project preview, so there's no need for `host` option.

tatomyr:
This looks good. However, the solution seems to be fragile. I'm afraid of the future situation when a user has an older version of Redocly CLI with one config structure, and the `preview` command pulls the latest version of Realm with a different config schema. Am I getting that right? Can we avoid the situation somehow?

lornajane:
If the user has the other package dependency installed, could that package be used instead of npx? I'm thinking of how we evolve changes that affect both packages, and how users can pin versions if they need to (either for software dependency process reasons, or because they aren't ready to accept a change in our newer releases).

I'm not sure if the errors are coming through to the CLI correctly. If you supply a directory that is not a directory, it says it's launching the preview and then exits.

lornajane:
A few comments on the command design:
- The project directory argument is strange, could we use `--sourceDir` instead, or something else?
- Is the directory required? It looks like it's not, but then the default (the current directory I think) should be included in the help output.
- What's the default value for the optional product parameter? (I think realm? Let's just add it to the help output)
- Please expand the command description to say what it does, it's too terse IMO.

I don't see any documentation for the preview command, but it should be included with the pull request, especially given the confusing naming with our existing `preview-docs` command, and explanations added to both pages. We'll also need to add some information about how the preview command is run since it's different to our other commands.

volodymyr-rutskyi:
@lornajane thank you for the extensive feedback!

> If the user has the other package dependency installed, could that package be used instead of npx? I'm thinking of how we evolve changes that affect both packages, and how users can pin versions if they need to (either for software dependency process reasons, or because they aren't ready to accept a change in our newer releases).

If the user has one of the product packages in `package.json`, then that package will be used. Then, if the user has a specific version in `node_modules`, NPX will launch that version. 

However, if the user specifies the product argument, that product's package will be used ignoring what's in `package.json` (version from `node_modules will still be used if available).

---

> I'm not sure if the errors are coming through to the CLI correctly. If you supply a directory that is not a directory, it says it's launching the preview and then exits.

The CLI preview command only serves as an interface for launching previews using product NPM packages, all of the output from those packages is shown and the console is also interactive. The behavior you described seems to be the fault of the packages that are being launched and should be fixed on that side. I will look into that.

---

> I don't see any documentation for the preview command, but it should be included with the pull request, especially given the confusing naming with our existing preview-docs command, and explanations added to both pages. We'll also need to add some information about how the preview command is run since it's different to our other commands.

I was told that the docs should come in a separate PR because the docs are published right after PR merge and the CLI changes go through a different release process. I have the docs as a WIP and will open a PR and link it to this one later today or on Monday morning.

---

> What's the default value for the optional product parameter? (I think realm? Let's just add it to the help output)

By default if the argument is not specified it tries to find a `package.json` file in the project directory and see if one of the packages is present there. If not, it defaults to `realm`. That info will be in the docs and I'll try to make it clearer in the help output.

volodymyr-rutskyi:
> This looks good. However, the solution seems to be fragile. I'm afraid of the future situation when a user has an older version of Redocly CLI with one config structure, and the `preview` command pulls the latest version of Realm with a different config schema. Am I getting that right? Can we avoid the situation somehow?

This should not be a problem. The `preview` does not rely on Redocly CLI for config. Right now, product NPM packages handle config schema by themselves and in the future they might handle config using the core package pinned in their dependencies, but should not rely on the CLI that launches them.

volodymyr-rutskyi:
Adjusted some naming and descriptions and linked a [PR with documentation for the command](https://github.com/Redocly/redocly-cli/pull/1384).
@lornajane

tatomyr (approved):

Issue #1245, opened 2023-08-24
Incorrect validation error "Example validation errored: schema with key or id ... already exists"


ehuelsmann:
In the attached YAML schema [API.yaml.gz](https://github.com/Redocly/redocly-cli/files/12433360/API.yaml.gz), the following error is consistently reported:

Example validation errored: schema with key or id "/tmp/tmp.odmvjRGLjU/API.yaml#/components/schemas/commonId" already exists.

1406 | application/json:
1407 |   schema:
1408 |     $ref: '#/components/schemas/Warehouse'
1409 |   examples:
1410 |     validWarehouse:

referenced from API.yaml:1407:15 at #/paths/\~1products\~1warehouses\~1{id}/get/responses/200/content/application\~1json 

Warning was generated by the no-invalid-media-type-examples rule.

API.yaml: validated in 242ms

However, the schema component id "commonId" appears exactly once in the specification. There is also no overlap between the "allOf" elements in the type being validated.

**To Reproduce**

1. Unpack the attached API.yaml
2. Run px @redocly/cli lint API.yaml

Will show the error above.

**Expected behavior**

I'm expecting the file to validate without errors. (There are no problems in the examples of this file as far as I can verify.)

**Logs**

None.

**OpenAPI definition**

Spec 3.0.3.

**Redocly Version(s)**

$ npx @redocly/cli --version
1.0.0-beta.125

**`Node.js` Version(s)**

$ node --version
v18.16.0

**Additional context**

None that I know of.

Orest-Yastremskyy:
Hi @ehuelsmann 

Thank you for reaching out.

We were able to replicate this issue and it appears to be a bug.

Please note that we will notify you as soon as the issue is resolved.

tatomyr:
Hi @ehuelsmann, the fix is already available in Redocly CLI 1.7. Could you confirm the issue has been resolved?

ehuelsmann:
@tatomyr, just verified with Redocly CLI 1.8: the problem is indeed gone!

Thank you so much!

PR #1377, opened 2024-01-02
fix: false positive errors when validating examples with refs

tatomyr:
## What/Why/How?

This solves an edge case validation error when there are multiple references in media type examples and the corresponding schemas.

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/1245

## Check yourself

- [x] Code is linted

## Security


.changeset/small-rules-vanish.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed false positive errors when validating media type examples that contain references.
```
lornajane, line 6: So that I understand this properly: we fixed a problem where no errors were reported - but the examples were wrong? So users might see errors without changing their API description, is that correct? Was it specifically when the changes went across files?

tatomyr: >  So users might see errors without changing their API description, is that correct?

Actually vice versa. We show an error when there's no error (due to wrong references resolving). After the update, false errors should be gone.

lornajane: OK let's update the changelog entry to describe what the user might experience.

tatomyr, line 6:
```suggestion
Fixed false positive errors when validating media type examples. Now, the linter won't error on valid examples that contain references. 
```

.changeset/small-rules-vanish.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed false positive errors when validating media type examples. Now, the linter won't error on valid examples that contain references.
```
adamaltman, line 6:
```suggestion
Fixed false positive errors with valid media type examples that contain references.
```
I think we should avoid "Now," in the changelogs and docs in general. Now, what do you think about that? 😆

tatomyr: Let's have Roman's review on the solution firstly. It might turn out we don't need any changeset at all 🤣

RomanHotsiy (approved):
I checked the ajv code and our code again.

This looks like it should not break anything! 🎉 

Let's merge.

.changeset/small-rules-vanish.md
RomanHotsiy: I checked it. LGTM.

lornajane: I'm sorry, I did already edit this changeset once, and while I agree that it's better without the "Now", I think it lost some context. Combining the two versions, how about this?
```suggestion
Fixed a problem where the linter incorrectly returned an error for valid examples that contain references.
```

lornajane (approved):
Really happy to get this sorted out. Thanks!

.changeset/small-rules-vanish.md
RomanHotsiy: I'm sure pretty sure he is good to 😄

Go ahead and accept the suggestion + merge.

PR #1382, opened 2024-01-05
feat: added filename extension support for more x-codeSamples languages

shahednasser:
## What/Why/How?

The documentation for [x-codeSamples](https://redocly.com/docs/api-reference-docs/specification-extensions/x-code-samples/) mentions that `lang` names are supported based on [this list](https://github.com/github-linguist/linguist/blob/master/lib/linguist/popular.yml). However, when it comes to writing the examples in their own files, the `langToExt` utility function only supports a subset of languages.

So, for example, if I use the lang `TypeScript`, the language isn't checked for in the `langToExt` function, which results in the file being names `getundefined`, as the extension file name is undefined.

This PR adds support for all the languages mentioned by the [languages list](https://github.com/github-linguist/linguist/blob/master/lib/linguist/popular.yml), and additionally adds support for `tsx` language. As part of our project, we needed the support for it.

## Check yourself

- [x] Code is linted

## Security


.changeset/cold-moles-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Added filename extension support for more x-codeSamples languages
```
tatomyr, line 5:
```suggestion
Added filename extension support for more `x-codeSamples` languages.
```

packages/cli/src/utils.ts
```diff
@@ -146,6 +146,27 @@ export function langToExt(lang: string) {
     javascript: '.js',
     js: '.js',
     python: '.py',
+    c: '.c',
+    'c++': '.cpp',
+    coffeescript: '.litcoffee',
+    css: '.css',
```
tatomyr, line 152: I'm not sure we need css (and probably some other languages, like html) here.

tatomyr (commented):
Thank you for contributing! Left a few comments. Also please make sure to update the appropriate test case.

packages/cli/src/utils.ts
shahednasser: I've removed `css`, `html`, and `dm`. Let me know if I should remove anything else as well.

shahednasser:
> Also please make sure to update the appropriate test case.

I've made changes to the test at `packages/cli/src/__tests__/utils.test.ts`.

.changeset/cold-moles-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Added filename extension support for more `x-codeSamples` languages
```
tatomyr, line 5:
```suggestion
Added filename extension support for more `x-codeSamples` languages.
```

tatomyr (approved):
LGTM. Thank you again!

PR #1383, opened 2024-01-05
chore: move TypeScript type dependencies to dev dependencies

fxOne:
## What/Why/How?

Currently, `@types/node` is a dependency in the `@redocly/openapi-core` package. As these are only TypeScript types, they are not needed in production and can be moved to a dev dependency.

Also, there is already a `@types/node@17.0.31` in the root package. My question would be whether the dependency there should be sufficient and whether the dependency should be changed to a more up-to-date version.

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12894,7 +12894,6 @@
       "license": "MIT",
       "dependencies": {
         "@redocly/ajv": "^8.11.0",
-        "@types/node": "^14.11.8",
         "colorette": "^1.2.0",
         "js-levenshtein": "^1.1.6",
         "js-yaml": "^4.1.0",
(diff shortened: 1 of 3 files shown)


tatomyr:
Thank you for spotting this on!
You absolutely can try updating the types to a newer version unless there're massive issues with that.

tatomyr:
@lornajane no we don't. If @fxOne won't update this, I'll merge this MR as it is.

fxOne:
@tatomyr I have updated `@types/node` to the latest version.

tatomyr (approved):
Thank you @fxOne for improving Redocly CLI! 😄

PR #1386, opened 2024-01-10
docs: Use Museum API instead of Petstore examples

### LINKED ISSUE #1374: Use Museum API as our example API description
**Describe the bug**

We use petstore as an example in many places, but this it outdated and our new [Museum API](https://github.com/Redocly/museum-openapi-example) would be better.

Replace the usage in CLI docs (currently in these files):
- [ ] ~commands/push.md~ complex command, only used for workflows
- [x] commands/bundle.md
- [x] commands/join.md
- [x] commands/lint.md
- [x] decorators/media-type-examples-override.md
- [x] installation.md

lornajane:
## What/Why/How?

Fixes #1374 

## Reference

https://github.com/Redocly/museum-openapi-example/

## Check yourself

- [x] Code is linted

## Security


docs/commands/lint.md
```diff
@@ -111,9 +88,7 @@ If you try to execute the `lint` command without apis when your project doesn't
 
 By default, the CLI tool looks for the [Redocly configuration file](../configuration/index.md) in the current working directory. Use the optional `--config` argument to provide an alternative path to a configuration file.
 
-```bash
-redocly lint --config=./another/directory/config.yaml
-```
+`redocly lint --config=./another/directory/config.yaml`
```
adamaltman, line 91: Isn't it better as before as a code block? (it is copy/paste-able as above, and people can also report problems with the code snippet as above)

lornajane: I think the general consensus was that really the whole page needs refactoring, but that this way is fine for now (and better than the wall of code samples). Happy to open a followup issue, but I'd like to focus on the example updates in this PR.

RomanHotsiy: I may look better (which is a matter of taste 💅) but I find it way harder to copy this way.
It's also inconsistent as some pages use code snippets and this one uses inline blocks.

If we decide to keep it as is can we open a ticket for refactoring this page?

lornajane: We definitely can! There's a lot more we could do if we wanted to invest some serious time in all the command pages IMO. Readability was my main aim on this pass, I think copying the config is more valuable than copying a short command that a user will need to edit anyway, so I picked this setup as the least worst option.

lornajane: I've added #1429 so we don't lose sight of this.

docs/commands/bundle.md
```diff
@@ -90,98 +90,6 @@ By default, the CLI tool looks for the Redocly configuration file in the current
...
-📦 Created a bundle for pet.yaml at bundled/pet.yaml 35ms.
-```
-
-{% /tab  %}
-{% /tabs  %}
-In this format, `bundle` uses the [Checkstyle](https://checkstyle.org/) XML report format.
-Due to the limitations of this format, the output _only_ includes the filename, line, column, severity,
-and rule ID (in the `source` attribute).
-All other information is omitted.
-
```
TaylorKrusen, line 184: Just to confirm, did we remove these formats from the CLI?

lornajane: There's an open ticket to drop linting from the bundle command, which is what this is the output for. So I figured we'd be dropping this section soon.

TaylorKrusen (approved):
+1 on the preference for code blocks because it's easier to copy. But it looks like we created an issue to update the code formatting, so this one LGTM.

docs/commands/lint.md
```diff
@@ -44,50 +44,31 @@ The `lint` command behaves differently depending on how you pass apis to it and
...
-{% tabs %}
-{% tab label="Command" %}
-
-```bash
-redocly lint
-```
-
-{% /tab  %}
-{% tab label="Configuration file" %}
+You can omit apis completely when executing the `lint` command to check all APIs defined in the configuration file. Run `redocly lint` with noarguments to lint all defined APIs; an example `redocly.yaml` file is shown below:
```
HCloward, line 71:
```suggestion
You can omit apis completely when executing the `lint` command to check all APIs defined in the configuration file. Run `redocly lint` with no arguments to lint all defined APIs; an example `redocly.yaml` file is shown below:
```

docs/commands/lint.md
```diff
@@ -245,127 +208,80 @@ redocly lint --format=json
...
-{% tabs %}
-{% tab label="Command" %}
 
-```shell
-redocly lint openapi/petstore.yaml --generate-ignore-file
-```
-
-{% /tab  %}
-{% tab label="Output" %}
+`redocly lint museum-with-errors.yaml --generate-ignore-file` runs the lint command and add all the errors to an ignore file.
```
HCloward, line 247:
```suggestion
`redocly lint museum-with-errors.yaml --generate-ignore-file` runs the lint command and adds all the errors to an ignore file.
```

docs/decorators/media-type-examples-override.md
```diff
@@ -1,109 +1,216 @@
...
-          '400':
-            application/json: ./pet-errors-examples.yaml
+            application/json: ./opening-hours-examples.yaml
 ```
 
+Replace a `requestBody` media type example by using the `request` key in `redocly.yaml`.
+The `examples` section in the API description is wholly replaced by the contents of the file you reference.
+
+Replace a response media type example for a specific status by setting `responses` and then the desired status in your configuration file.
+The `examples` section for that response status is replaced by the contents fo the file specified.
```
HCloward, line 37:
```suggestion
The `examples` section for that response status is replaced by the contents of the file specified.
```

HCloward (approved):

Issue #1314, opened 2023-10-25
Supporting `x-tagGroups` in `join` command


unional:
**Describe the problem to be solved**

Currently, the `join` command creates a spec that uses the filename as the `x-tagGroups`.

This cause all tags are added to the same `openapespec` group for the following:

```sh
# /packages/api-a/open_api/openapespec.yaml
# /packages/api-b/open_api/openapespec.yaml

redocly join packages/api-a/open_api/openapispec.yaml packages/api-b/open_api/openapispec.yaml -o public/openapi.yaml

# or
redocly join packages/*/open_api/*.yaml -o public/openapi.yaml
```

**Describe the solution you'd like**

Use `info.title` of each file as the group name

**Describe alternatives you've considered**

add a flag to specify which prop to use.

**Additional context**

tatomyr:
Thanks for the suggestion @unional!

tatomyr:
Hey @unional, could you check if using [this option](https://redocly.com/docs/cli/commands/join/#prefix-tags-with-info-prop) covers your case?

unional:
> Hey @unional, could you check if using [this option](https://redocly.com/docs/cli/commands/join/#prefix-tags-with-info-prop) covers your case?

Nope. That changes the tags not the tag groups.

lornajane:
It's very difficult to pick a prefix solution here that works for everyone - although we do have the option to not create the groups which I think is important. I agree that `info.title` would be (or should be) a more API-specific group label and would like us to make this change, but we need to take care with special characters in it to create a useful value for the tag groups.

Also we'll need to make it very clear in the changelog that this behaviour is changing - minor version bump at a minimum. I know the join command is still marked experimental but we do see quite a bit of usage!

unional:
I observe another issue related to this.
Without the `x-tagGroups` support for the `join` command, I can' control the order or filter out specific tags from the join doc.

Able to filter out tags are especially important as we can add the next version of API to the documentation during development while they are not exposed in the final doc.

lornajane:
Thanks for your comment @unional , but in general it's better to open a new issue than to comment on a closed one where additional conversation might not be seen.

For this additional issue, perhaps you can use the filter-out decorator after joining to produce the output of the combined API descriptions without the tagged endpoints? There's more information about the filter-out decorator in the docs: https://redocly.com/docs/cli/decorators/filter-out/

Issue #1203, opened 2023-08-01
Join command creates x-tagGroups that don't match extension schema and fails linting


johnrengelman:
**Describe the bug**

When using the `join` command with automatic `x-tagGroup` generation, the cli populates a `description` property on the object using the description from the api that was joined https://github.com/Redocly/redocly-cli/blob/f7a6a036ec8dba9f0b526504d9ad3cd4550f0572/packages/cli/src/commands/join.ts#L324-L343

The resulting document fails when evaluated with the `lint` command, because `description` is not a valid property for `x-tagGroups` elements https://github.com/Redocly/redocly-cli/blob/f7a6a036ec8dba9f0b526504d9ad3cd4550f0572/packages/core/src/types/oas3.ts#L35-L41

**To Reproduce**

1. `npx @redocly/cli join <api1.yml> <api2.yml> -o joined.yml`
2. `npx @redocly/cli lint joined.yml`

**Expected behavior**

**Logs**

```
[1] bundle.yml:5742:5 at #/x-tagGroups/0/description

Property `description` is not expected here.

5740 |     .....<snip>

Error was generated by the spec rule.
```

**OpenAPI definition**

`.redocly.yml` -
```
extends:
  - recommended
rules:
  no-invalid-media-type-examples:
    severity: error
    allowAdditionalProperties: true
  operation-operationId-unique: warn
```

**Redocly Version(s)**

1.0.0

`v16.16.0`

lornajane:
Thanks for the clear bug report. If you don't need the tag groups, you can work around this for now by adding `--without-x-tag-groups` to your command until we get this sorted. More information here: https://redocly.com/docs/cli/commands/join/#without-x-tag-groups

johnrengelman:
Yup - only issue with that option is that there is no way to add x-tagGroups when using join because they are ignored from the sub-APIs.

lornajane:
I'm not sure why we're adding the description when creating an `x-tagGroup`, and we don't actually support this in our own [explanation of the extension](https://redocly.com/docs/api-reference-docs/specification-extensions/x-tag-groups/). I think the right thing to do here is to stop adding descriptions to tag groups when joining API descriptions.

Related: an open issue about using `info.title` rather than the filename to name the tag groups #1314

PR #1394, opened 2024-01-16
feat: removed descriptions and added info.title as group names for x-tagGroup  in Join command

o-bbk:
## What/Why/How?
Removed descriptions adding for x-tagGroups for the `join` command as they are not supported and cause errors on linting.

## Reference
Resolves https://github.com/Redocly/redocly-cli/issues/1203
Resolves https://github.com/Redocly/redocly-cli/issues/1314

## Check yourself

- [x] Code is linted

## Security


.changeset/grumpy-toys-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Removed descriptions adding for x-tagGroups for the `join` command.
```
lornajane, line 5: Please include something about why this change was made, and the differences that the user may observe.

lornajane (dismissed):

tatomyr (approved):

lornajane:
Feature does work perfectly though - just the changelog needs some work :)

.changeset/grumpy-toys-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/cli": patch
+---
+
+- Removed descriptions adding for x-tagGroups for the `join` command. Descriptions in x-tagGroups are not supported and cause errors on linting.
+- `info.title` is now used as a name in x-tagGroups instead of a file name for the `join` command to make possible the joining of files with the same names.
```
lornajane, line 6: Will existing users see changed behaviour? Is it configurable?

adamaltman: What if the `info.title`'s are the same?

tatomyr: Use info versions as well? Is it a valid case to join different versions of the same API descriptions? If even that fails, I'd expect a warning/error. Alternatively, we can add indexes (like Title_1, Title_2 and so on), but this behaviour might be confusing.

lornajane: If the titles are the same, they'll only get one `x-tagGroups` entry. Let's link to the [info-override decorator](https://redocly.com/docs/cli/decorators/info-override/) in the join command docs so that the user can adjust one or all of the titles to suit their needs before joining.

malis42: This docs change will be resolved in a separate PR here https://github.com/Redocly/redocly-cli/pull/1422

tatomyr (approved):

.changeset/grumpy-toys-reply.md
HCloward, line 6:
```suggestion
- Updated `info.title` to be used as a name in x-tagGroups instead of a file name for the `join` command, so you can now join files with the same names.
```

malis42: Fixed, thanks!

HCloward (approved):

.changeset/grumpy-toys-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/cli": patch
```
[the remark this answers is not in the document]
malis42: Changed to minor as requested, thanks!

PR #1407, opened 2024-01-24
fix: extra whitespace after update notifier

lornajane:
## What/Why/How?

We did a new release yesterday and I noticed that when I get the release banner, I also get slightly indented output from whichever command I was using.

## Screenshots (optional)
Screenshot from 2024-01-24 15-46-35

## Check yourself

- [x] Code is linted

## Security

### DIFF
--- a/packages/cli/src/utils/update-version-notifier.ts
+++ b/packages/cli/src/utils/update-version-notifier.ts
@@ -64,27 +64,27 @@ const renderUpdateBanner = (current: string, latest: string) => {
   const maxLength = Math.max(...messageLines.map((line) => cleanColors(line).length));
 
   const border = yellow('═'.repeat(maxLength + SPACE_TO_BORDER));
+  const extraSpaces = ' '.repeat(SPACE_TO_BORDER);
 
-  const banner = `
-    ${yellow('╔' + border + '╗')}
-    ${yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║')}
-    ${messageLines
-      .map((line, index) => {
-        return getLineWithPadding(maxLength, line, index);
-      })
-      .join('\n')}
-    ${yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║')}
-    ${yellow('╚' + border + '╝')}
-`;
-
+  const banner = [
+    '',
+    extraSpaces + yellow('╔' + border + '╗'),
+    extraSpaces + yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
+    messageLines.map(getLineWithPadding(maxLength, extraSpaces)).join('\n'),
+    extraSpaces + yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
+    extraSpaces + yellow('╚' + border + '╝'),
+    '',
+    '',
+  ].join('\n');
   process.stderr.write(banner);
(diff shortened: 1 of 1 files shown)


packages/cli/src/update-version-notifier.ts
```diff
@@ -75,7 +75,7 @@ const renderUpdateBanner = (current: string, latest: string) => {
       .join('\n')}
     ${yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║')}
     ${yellow('╚' + border + '╝')}
-  `;
+`;
```
tatomyr, line 78: Oh, the indentation itself is correct. A better solution would be to use `outdent`, but it's in the dev deps, and I'm reluctant to move it into the actual dependencies. I believe we can improve this by not using the template string literals, like the following:

```ts
  const banner = [
    '',
    yellow('╔' + border + '╗'),
    yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
    messageLines
      .map((line, index) => {
        return getLineWithPadding(maxLength, line, index);
      })
      .join('\n'),
    yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
    yellow('╚' + border + '╝'),
    '',
  ].join('\n');
```

but I need someone to test this :) 
Or I can test that later.

Anyway, @IgorKarpiuk is currently moving this file so I'd wait until this is merged: https://github.com/Redocly/redocly-cli/pull/1321

IgorKarpiuk: @tatomyr It does not work  🙂
Maybe we can just add `\n`

tatomyr: We can, but it's strange it doesn't.

tatomyr: This should do the trick: 

```ts

const renderUpdateBanner = (current: string, latest: string) => {
  const messageLines = [
    `A new version of ${cyan('Redocly CLI')} (${green(latest)}) is available.`,
    `Update now: \`${cyan('npm i -g @redocly/cli@latest')}\`.`,
    `Changelog: https://redocly.com/docs/cli/changelog/`,
  ];
  const maxLength = Math.max(...messageLines.map((line) => cleanColors(line).length));

  const border = yellow('═'.repeat(maxLength + SPACE_TO_BORDER));
  const extraSpaces = ' '.repeat(SPACE_TO_BORDER);

  const banner = [
    '',
    extraSpaces + yellow('╔' + border + '╗'),
    extraSpaces + yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
    messageLines.map(getLineWithPadding(maxLength, extraSpaces)).join('\n'),
    extraSpaces + yellow('║' + ' '.repeat(maxLength + SPACE_TO_BORDER) + '║'),
    extraSpaces + yellow('╚' + border + '╝'),
    '',
    '',
  ].join('\n');
  process.stderr.write(banner);
};

const getLineWithPadding =
  (maxLength: number, extraSpaces: string) =>
  (line: string): string => {
    const padding = ' '.repeat(maxLength - cleanColors(line).length);
    return `${extraSpaces}${yellow('║')}  ${line}${padding}  ${yellow('║')}`;
  };

```

tatomyr:
Updated the MR. Here's the screenshot:

image

IgorKarpiuk (approved):

Issue #1392, opened 2024-01-12
Path ending with # breaks path resolution


maininformer:
**Describe the bug**

When the `$ref` ends with a `#` I receive: `Can't resolve $ref: ENOENT: no such file or directory`, but removing the `#` fixes the issue

**To Reproduce**
To reproduce set these up:
```
.
├── openapi.yml
└── schema
    └── get
        └── response.json

```

`openapi.yml`
```                                                                X
 openapi: '3.0.2'
 info:
   title: API Title
   version: '1.0'
 servers:
   - url: https://localhost:80/api/v1
 paths:
   /awesomeEndpoint:
     get:
       summary: summary
       responses:
         '200':
           description: OK
           content:
             application/json:
               schema:
                 $ref: './schema/get/response.json#'
```

`response.json`
```
 {
     "$schema": "http://json-schema.org/draft-2020-12/schema#",
     "title": "Get",
     "type": "array",
     "items": {
         "type": "string"
     }
 }
```

then run:
```
$ redocly bundle openapi.yml
```

**Expected behavior**

Success.

**Logs**

```
└[~/Documents/redocly]> redocly bundle openapi.yml
bundling openapi.yml...
[1] openapi.yml:17:17 at #/paths/~1awesomeEndpoint/get/responses/200/content/application~1json/schema

Can't resolve $ref: ENOENT: no such file or directory '/Users/maininformer/Documents/redocly/schemas/get/response.json'
```

**Redocly Version(s)**

1.6.0

**`Node.js` Version(s)**

18.8.0

**Additional context**

Removing the `#` from the end of the `$ref` resolves the issue:

```
 openapi: '3.0.2'
 info:
   title: API Title
   version: '1.0'
 servers:
   - url: https://localhost:80/api/v1
 paths:
   /awesomeEndpoint:
     get:
       summary: summary
       responses:
         '200':
           description: OK
           content:
             application/json:
               schema:
                 $ref: './schema/get/response.json'
```

tatomyr:
Thanks for reporting this! Please notice that it also won't error if you add the slash after the hash:
```yaml
$ref: './schema/get/response.json#/'
```

PR #1408, opened 2024-01-25
fix: hash ending path resolution fails

maininformer:
## What/Why/How?

Currently if a `$ref` ends with `#`, meaning the document root, the code breaks because it cannot find the path. For instance:

```
schema: path/to/schema#
```
This fixes https://github.com/Redocly/redocly-cli/issues/1392

## Reference

[json schema](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#idExamples)
[Issue](https://github.com/Redocly/redocly-cli/issues/1392)

## Testing

E2E test [here](https://github.com/maininformer/redocly-cli/tree/fix/hash-ending-path-resolution-fails/__tests__/bundle/reference-ending-in-hash) 

## Has code been changed?

## Security


packages/core/src/ref-utils.ts
```diff
@@ -43,7 +43,8 @@ export function escapePointer<T extends string | number>(fragment: T): T {
 }
 
 export function parseRef(ref: string): { uri: string | null; pointer: string[] } {
-  const [uri, pointer] = ref.split('#/');
+  const [uri, pointer_] = ref.split('#');
```
tatomyr, line 46: I don't like the naming though. @malis42 could you suggest a better solution?

tatomyr (commented):
Thank you @maininformer for your contribution and the original issue! The solution looks correct, but I'd like to make it a bit nicer.

packages/core/src/ref-utils.ts
malis42: I think we should do a small refactor in this area and reuse already existing function which is defined below the `parseRef` function, like this:

```ts
export function parseRef(ref: string): { uri: string | null; pointer: string[] } {
  const [uri, pointer = ''] = ref.split('#');

  return {
    uri: uri || null,
    pointer: parsePointer(pointer),
  };
}

export function parsePointer(pointer: string) {
  return pointer.split('/').map(unescapePointer).filter(isTruthy);
}
```

.changeset/soft-planets-watch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed an issue where `$ref`s ending in `#` (instead of `#/`) would break the application.
```
lornajane, line 6: Do we silently handle this situation, or provide a better error message when we see it?

HCloward (approved):

tatomyr (approved):
LGTM. Thanks again @maininformer! Your changes will become available in the next release.

PR #1423, opened 2024-02-01
chore: do not update domain env variable from code

IgorKarpiuk:
## What/Why/How?

Do not update `REDOCLY-DOMAIN` env variable from code, use static variable instead

## Has code been changed?

## Security


IgorKarpiuk:
Should be fixed in https://github.com/Redocly/redocly-cli/pull/1433

PR #1425, opened 2024-02-01
fix: update `push-status` messages

IgorKarpiuk:
## What/Why/How?

Fix the wording and eliminate uppercase for messages in both the `push-status` and `push` commands.

## Has code been changed?

## Security

### DIFF
--- a/packages/cli/src/cms/commands/push-status.ts
+++ b/packages/cli/src/cms/commands/push-status.ts
@@ -178,13 +178,13 @@ function displayDeploymentAndBuildStatus({
       spinner.stop();
       return process.stdout.write(
         `${colors.green(
-          `🚀 ${buildType.toLocaleUpperCase()} deployment succeeded.`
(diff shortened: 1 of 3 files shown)


packages/cli/src/cms/commands/push-status.ts
```diff
@@ -177,15 +177,15 @@ function displayDeploymentAndBuildStatus({
     case 'success':
       spinner.stop();
       return process.stdout.write(
-        `${colors.green(
-          `🚀 ${buildType.toLocaleUpperCase()} deployment succeeded.`
-        )}\n${colors.magenta('Preview URL')}: ${colors.cyan(previewUrl!)}\n`
+        `${colors.green(`🚀 ${buildType.toLocaleUpperCase()} deploy success.`)}\n${colors.magenta(
```
adamaltman, line 180: Can we avoid the upper casing here and below too?

tatomyr (approved):

Issue #401, opened 2021-10-08
`join` command with `--prefix-components-with-info-prop` looses additional schemes in security requirement


simon-spinner:
**Describe the bug**
A security requirement of an operation may specify multiple security schemes that all must be fulfilled. When using `join` with `--prefix-components-with-info-prop` only the first security scheme gets copied to the joined document

**To Reproduce**
1. Given no `.redocly.yaml` file
2. And this OpenAPI file(s)
openapi1.yaml:
```
openapi: 3.0.2
info:
  title: Spec1
  version: 1.0.0
servers:
  - url: 'https://localhost:8080/api'
paths:
  /example:
    get:
      responses:
        '200':
          description: OK
      security:
        - scheme1: []
          scheme2: []      
components:
  securitySchemes:
    scheme1:
      type: http
      scheme: basic
    scheme2:
      type: http
      scheme: basic
```
openapi2.yaml:
```
openapi: 3.0.2
info:
  title: Spec2
  version: 1.0.0
servers:
  - url: 'https://localhost:8080/api'
paths: {}
```

3. Run this command with these arguments... `openapi join --prefix-components-with-info-prop title openapi1.yaml openapi2.yaml`
4. See the resulting joined spec:

```
openapi: 3.0.2
info:
  title: Spec1
  version: 1.0.0
servers:
  - url: 'https://localhost:8080/api'
paths:
  /example:
    get:
      responses:
        '200':
          description: OK
      security:
        - Spec1_scheme1: []
      tags:
        - openapi1_other
tags:
  - name: openapi1_other
    x-displayName: other
x-tagGroups:
  - name: openapi1
    tags:
      - openapi1_other
components:
  securitySchemes:
    Spec1_scheme1:
      type: http
      scheme: basic
    Spec1_scheme2:
      type: http
      scheme: basic

```

**Expected behavior**
The operation `get /example` should have 2 security schemes: `Spec1_scheme1` and `Spec1_scheme2`. But it only contains `Spec1_scheme1`.

**`openapi-cli` Version(s)**
1.0.0-beta.54

**`Node.js` Version(s)**
v16.8.0

tatomyr:
Hi @simon-spinner! 
We've just released v1.9.1 with the fix. Could you check if it works for you?

PR #1439, opened 2024-02-15
fix: join command security schemas fix

malis42:
## What/Why/How?
Based on [this link ](https://swagger.io/docs/specification/authentication/) we know that security schemas support several authentication types. The security section lets you combine the security requirements using logical OR and AND to achieve the desired result.

Currently we support only logical OR, when security schema is defined using logical AND, `join` command only returns first record from array of objects.

## Reference
Fixes https://github.com/Redocly/redocly-cli/issues/401

## Testing
After running `join --prefix-components-with-info-prop title openapi1.yaml openapi2.yaml` with specs provided in https://github.com/Redocly/redocly-cli/issues/401 we get correct outcome:
image

More complicated examples are covered in E2E tests.

## Has code been changed?
Yes

## Check yourself

- [x] Code is linted

## Security


packages/cli/src/commands/join.ts
```diff
@@ -756,8 +756,17 @@ function addComponentsPrefix(description: string, componentsPrefix: string) {
 function addSecurityPrefix(security: any, componentsPrefix: string) {
   return componentsPrefix
     ? security?.map((s: any) => {
-        const key = Object.keys(s)[0];
-        return { [componentsPrefix + '_' + key]: s[key] };
+        // Check if security contains more than one object to determine if its logical 'OR' or 'AND'
+        if (Object.keys(s).length > 1) {
```
tatomyr, line 760: Do we need this condition? Looks like this would work without the `else` clause for any number of object keys:

```ts

          const joinedSecuritySchema = {};
          for (const [key, value] of Object.entries(s)) {
            Object.assign(joinedSecuritySchema, { [componentsPrefix + '_' + key]: value });
          }
          return joinedSecuritySchema;
```

malis42: Fixed!

IgorKarpiuk (approved):
LGTM

tatomyr (approved):
LGTM, thanks!

.changeset/little-camels-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Fix an issue when using `join` command resulted in generating incorrect schema, if security schema used logical AND for security requirements.
```
HCloward, line 5:
```suggestion
If the security schema uses logical AND for security requirements, the `join` command generates the correct schema, as expected.
```

HCloward (approved):

Issue #1440, opened 2024-02-16
Join with --prefix-components-with-info-prop does not produce any output since version 1.9.0


simon-spinner:
**Describe the bug**
redocly-cli crashes on join if schema property contains a null value

**To Reproduce**

1. Given the two OpenAPI file(s)

openapi1.yaml
```yaml
openapi: '3.0.2'
info:
  title: API1
  version: '1.0'
servers:
  - url: https://api.server.test/v1
paths:
  {}
components:
  schemas:
    Schema1:
      type: string
```

openapi2.yaml
```yaml
openapi: '3.0.2'
info:
  title: API2
  version: '1.0'
servers:
  - url: https://api.server.test/v1
paths:
  {}
components:
  schemas:
    Schema2:
      type: string
      default: null
      nullable: true
```

3. Run `redocly join --prefix-components-with-info-prop title openapi1.yaml openapi2.yaml`
4. No output is provided

**Expected behavior**

Join works as it did before version 1.9.0 was released.

**Redocly Version(s)**

1.9.0

tatomyr:
Taking a look. Thanks for reporting!

tatomyr:
@simon-spinner could you confirm the issue got resolved in v1.9.1?

PR #1444, opened 2024-02-19
fix: join command failing on null values with --prefix-components-with-info-prop option

tatomyr:
## What/Why/How?

Since 1.9.0 `join` fails when used with the --prefix-components-with-info-prop option and a schema contains `null` value.

This PR resolves the issue by moving the crawl [visitor](https://github.com/Redocly/redocly-cli/pull/1444/files#diff-e52f3b6820e3de0cd0ec76de555265a5139350fdfa3cd210402edeaa9fdd56eeR151) out of the loop. 

Also, did a small refactoring of `split` command (reusing shared things with `join`, types correction).

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/1440

## Has code been changed?

## Security

### DIFF
--- a/.changeset/spotty-pets-doubt.md
+++ b/.changeset/spotty-pets-doubt.md
@@ -1,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Fixed an issue when the `join` command was failing to process schemas containing `null` values with the `--prefix-components-with-info-prop` option.'
--- a/.changeset/spotty-pets-doubt.md
+++ b/.changeset/spotty-pets-doubt.md
@@ -2,4 +2,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed an issue when the `join` command was failing to process schemas containing `null` values with the `--prefix-components-with-info-prop` option.'
+Fixed a problem where the `join` command did not process schemas containing `null` values when the `--prefix-components-with-info-prop` option was used.'
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -801,30 +801,33 @@ async function validateApi(
   }
 }
 
-function crawl(object: any, visitor: any) {
+export function startsWithComponents(node: string) {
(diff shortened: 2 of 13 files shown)


.changeset/spotty-pets-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Fixed an issue when the `join` command was failing to process schemas containing `null` values with the `--prefix-components-with-info-prop` option.'
```
lornajane, line 5:
```suggestion
Fixed a problem where the `join` command did not process schemas containing `null` values when the `--prefix-components-with-info-prop` option was used.'
```

lornajane (commented):
Minor grammatical update.

Might be worth adding a comment about why this was the answer? I see the bug report, I see some changes to dependencies, but it's not obvious to me (or future-us) why this is the fix.

tatomyr:
@lornajane updated the description.

lornajane (approved):
changeset is good

packages/cli/src/commands/join.ts
```diff
@@ -801,30 +803,19 @@ async function validateApi(
...
       for (const name of Object.keys(mapping)) {
-        if (isString(mapping[name]) && mapping[name].startsWith(`#/${COMPONENTS}/`)) {
-          mapping[name] = mapping[name]
+        const mappingPointer = mapping[name];
+        if (typeof mappingPointer === 'string' && startsWithComponents(mappingPointer)) {
+          mapping[name] = mappingPointer
             .split('/')
-            .map((name: string, i: number, arr: []) => {
+            .map((name, i, arr) => {
               return arr.length - 1 === i && !name.includes(componentsPrefix)
```
simon-spinner, line 819: This check would better be `!name.startsWith(componentsPrefix)` so that if someone coincidentally uses the component prefix somewhere inside the name, the prefix is added nonetheless

Issue #1441, opened 2024-02-16
Path containing # breaks path resolution


avanbremen:
**Describe the bug**

When the `$ref` contains `#` I receive: `Can't resolve $ref: ENOENT: no such file or directory`.

Similar to issue #1392.

Most likely due to [change](https://github.com/Redocly/redocly-cli/compare/%40redocly/cli%401.8.2...%40redocly/cli%401.9.0#diff-25c0338c90db3454ba7548187ad0db4e1e288ffca3c1d6ef65f45a442e728074) in `ref-utils.ts`.

Tested with release [@redocly/cli@1.9.0](https://github.com/Redocly/redocly-cli/releases/tag/%40redocly%2Fcli%401.9.0).

**To Reproduce**

For example:

`openapi.yml`

```yaml
openapi: 3.1.0

info:
  title: My API
  version: 1.0.0
  description:
    $ref: './info#description.md'
```

`info#description.md`

```markdown
# Description

Redocly is awesome.
```

**Expected behavior**

Success.

**Logs**

```
Can't resolve $ref: ENOENT: no such file or directory '/home/runner/work/openapi-specification/openapi-specification/src/v1/info'
```

**OpenAPI description**

**Redocly Version(s)**

`1.9.0`

**Additional context**

Similar to issue #1392.

GitHub Actions runner version: `2.313.0`.

jaredLunde:
This downstream issue is also related to the change to `ref-utils.ts`
https://github.com/drwpow/openapi-typescript/issues/1542#issuecomment-1947507415

tatomyr:
Hey guys. Prepared a fix. Feel free to leave your comments: https://github.com/Redocly/redocly-cli/pull/1445.

tatomyr:
@avanbremen, @jaredLunde could you confirm the issue got resolved in v1.9.1?

jaredLunde:
@tatomyr this fixed the issue for me 😄 thank you!

avanbremen:
@tatomyr tested with latest ([1.10.1](https://github.com/Redocly/redocly-cli/releases/tag/%40redocly%2Fcli%401.10.1)), the issue is now solved. Thanks!

PR #1445, opened 2024-02-19
fix: parse $refs containing the hash symbol correctly

tatomyr:
## What/Why/How?

Since the last release, $refs has been resolved incorrectly when a file names contain the hash character. This was due to splitting URIs by '#' instead of '#/'. 

This PR reverts this change while still allowing URIs to end with '#'.

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/1441

Corresponding docs: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#idExamples

## Has code been changed?

## Security

### DIFF
--- a/.changeset/slimy-planes-deliver.md
+++ b/.changeset/slimy-planes-deliver.md
@@ -3,4 +3,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed resolving $refs. Previously it was failing resolve file names that contain the hash symbol.
+Fixed resolving $refs. Previously, it was failing to correctly resolve file names that contain the hash symbol.
--- a/.changeset/slimy-planes-deliver.md
+++ b/.changeset/slimy-planes-deliver.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed resolving $refs. Previously it was failing resolve file names that contain the hash symbol.
--- a/.changeset/slimy-planes-deliver.md
+++ b/.changeset/slimy-planes-deliver.md
@@ -3,4 +3,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed resolving $refs. Previously, it was failing to correctly resolve file names that contain the hash symbol.
+Fixed a bug with resolving $refs to file names that contain the hash symbol.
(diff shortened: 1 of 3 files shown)


tatomyr:
Hi @maininformer! I've updated your change a bit. Could you check if it still works for you?

IgorKarpiuk (approved):

.changeset/slimy-planes-deliver.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed resolving $refs. Previously, it was failing to correctly resolve file names that contain the hash symbol.
```
lornajane, line 6:
```suggestion
Fixed a bug with resolving $refs to file names that contain the hash symbol.
```

lornajane (approved):

maininformer:
> Hi @maininformer! I've updated your change a bit. Could you check if it still works for you?

Hey @tatomyr, I have an e2e test for this; if it passes I'm happy, it's called `reference-ending-in-hash`. See [here](https://github.com/Redocly/redocly-cli/tree/main/__tests__/bundle/reference-ending-in-hash)

Maybe unrelated, but I see in your changes: 

```
  const [uri, pointer = ''] = ref.split('#/');
  return {
    uri: (uri.endsWith('#') ? uri.slice(0, -1) : uri) || null,
```

I am not sure if `uri` will ever end with `#`, unless there is a typo; because we are splitting it by `#/` here. Example:

```
> 'someUri#/pointer'.split('#/')
['someUri', 'pointer']
```

maininformer:
@tatomyr sorry for the confusion, I was referring to this line:

```
uri: (uri.endsWith('#') ? uri.slice(0, -1) : uri) || null,
```

After running `const [uri, pointer = ''] = ref.split('#/');` on a ref, the uri will not end in `#` anymore, because we have split on `#/`. Meaning, we will never reach `uri.slice(0, -1)`.

Unless we have something like: `someUrl##/somepointer`

tatomyr:
> After running const [uri, pointer = ''] = ref.split('#/');

if `ref = './info.json#'` then after splitting we will recieve `uri = './info.json#', pointer = ''` (because there's no `#/` splitter in the `ref`). Then, the `uri.endsWith('#')` statement will be `true`. Does that make sense?

maininformer:
Ah I see. Smart! Thanks for explaining.

Issue #477, opened 2021-12-20
lint: non-existing object properties listed as required are ignored


davidkuridza:
**Describe the bug**
The `openapi-cli` ignores non-existing object properties which are listed as required. Anything can be listed under the `required:` tag and the `lint` will not return any errors.

**To Reproduce**

1. Use the example [petstore.yaml](https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v2.0/yaml/petstore.yaml) (the behaviour is the same with 2.0, 3.0 and 3.1)
2. Add a non-existing or rename an existing required property:
```diff
diff --git a/petstore.yaml b/petstore.yaml
index 5f41fe0..531cbbf 100644
--- a/petstore.yaml
+++ b/petstore.yaml
@@ -78,6 +78,7 @@ definitions:
     required:
       - id
       - name
+      - non-existing-property
     properties:
       id:
         type: integer
```

3. Run `openapi lint petstore.yaml --format stylish`
4. The output is:
```console
No configurations were defined in extends -- using built-in recommended configuration by default.

validating petstore.yaml...
petstore.yaml:
  2:1   warning  info-description        Info object should contain `description` field.
  5:3   warning  info-license-url        License object should contain `url` field.
  29:7  warning  operation-4xx-response  Operation must have at least one `4xx` response.
  47:7  warning  operation-4xx-response  Operation must have at least one `4xx` response.
  66:7  warning  operation-4xx-response  Operation must have at least one `4xx` response.

petstore.yaml: validated in 26ms

Woohoo! Your OpenAPI definition is valid. 🎉
You have 5 warnings.
```

5. The exit code:
```console
$ echo $?
0
```

**Expected behavior**
There should be an error saying the required property is not defined, for example:
```
"non-existing-property" is present in required but not defined as property in definition "Pet"
```

**`openapi-cli` Version(s)**
```console
$ openapi --version
1.0.0-beta.73
```

**`Node.js` Version(s)**
```console
$ node --version
v17.2.0
```

**Additional context**
- The behaviour is the same with all OpenAPI versions. 
- There is no `.redocly.yaml` present, the default out-of-the-box `openapi-cli` is used.
- The same can be observed using the latest NPM and Docker versions.

adamaltman:
I don't think this is a bug. This seems like a feature request? 

By default the objects allow additional properties. So by example it is technically correct (unless the additionalProperties was set to false -- in which case, it would be a bug).

[...]

PR #1446, opened 2024-02-19
docs: update the no-required-schema-properties-undefined rule description

tatomyr:
## What/Why/How?

It might be unclear that the rule is case sensitive. Added an example to illustrate that.

## Reference

See the comment: https://github.com/Redocly/redocly-cli/issues/477#issuecomment-1943467337

## Has code been changed?

## Security

### DIFF
--- a/docs/rules/no-required-schema-properties-undefined.md
+++ b/docs/rules/no-required-schema-properties-undefined.md
@@ -85,7 +85,7 @@ schemas:
         example: doggie
 ```
 
-The rule works case-sensitive, which means a property `name` will be invalid in combination with the string `Name` in the `required` list:
+The rule is case-sensitive, which means a property `name` does not match the string `Name` in the `required` list:
 
 ```yaml
 schemas:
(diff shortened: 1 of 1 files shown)


tatomyr:
@lornajane do you recall any recent changes to the vale linting process?

lornajane:
Not very recently - #1390 but it was still correctly failing (if that makes sense!) after that merged.

docs/rules/no-required-schema-properties-undefined.md
```diff
@@ -74,15 +74,28 @@ schemas:
...
-          example: doggie
+      id:
+        type: integer
+        format: int64
+      name:
+        type: string
+        example: doggie
+```
+
+The rule works case-sensitive, which means a property `name` will be invalid in combination with the string `Name` in the `required` list:
```
lornajane, line 88:
```suggestion
The rule is case-sensitive, which means a property `name` does not match the string `Name` in the `required` list:
```

lornajane:
Let's definitely not hold up this PR for Vale, I'll pick it up separately

lornajane (approved):
Much clearer, thanks!

PR #1454, opened 2024-02-28
chore: use real opanapi file in an example for clarity

tatomyr:
## What/Why/How?

Someone can copy and paste the command from our example and it won't work. 
Kudos to @lesyk-lesyk for spotting this.

## Has code been changed?

## Security


lornajane (approved):
Makes sense to me!

PR #1455, opened 2024-02-29
fix: replace spaces in components prefixes with underscores

tatomyr:
## What/Why/How?

When joining with `--prefix-components-with-info-prop=title` it's possible that the title contains spaces. That produces $refs with spaces (which is at least confusing). This replaces the spaces with underscores.

## Has code been changed?

## Security


.changeset/yellow-eels-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Fixed a problem where the `join` command with `--prefix-components-with-info-prop` was creating $refs with spaces by replacing them with underscores.
```
lornajane, line 5:
```suggestion
Added inflection to the `join` command so that `--prefix-components-with-info-prop` replaces spaces with underscores to create less confusing $refs.
```

lornajane (approved):
I didn't run the command, but looking at the updated snapshots, I think this is a good approach. Suggested an improvement to the changelog entry as I think its more of a feature than a fix.

PR #1464, opened 2024-03-01
docs: update Vale config naming and setup

lornajane:
Rename the Vale configuration to make it clear this is the Redocly Docs ruleset, and fix one error that was found in the current docs set. Also simplify config (this started as an investigation to a missed error, the extra config doesn't seem to be needed).

## Has code been changed?

## Security


lornajane:
One of the problems here is that Vale runs on all files, but filters the problems to report only on changed files since we have `filter: files` in our configuration. Change to `filter: nofilter` to run across all files, including those without updates in this pull request. (this note really only here to remind me what I did when this confuses me next time!)

HCloward (approved):

Issue #874, opened 2022-09-15
remove-unused-components rule is not processed transitively


cliffred:
**Describe the bug**
Given the following snippet:
```yaml
Customer:
  type: object
  properties:
    id:
      type: string
    username:
      type: string
    address:
      $ref: '#/components/schemas/Address'
```

Assume that Customer is unused and Customer is the only component with a reference to Address. On linting only Customer is shown as an unused component. But once that one is removed Address becomes unused. 

The same goes for `redocly bundle input.yaml --remove-unused-components -o output.yaml`. You need multiple passes to remove all unused components.

I think it would make sense to process this rule recursively to not need multiple passes to remove all unused components.

**To Reproduce**
`redocly lint https://petstore3.swagger.io/api/v3/openapi.json`
It shows among others _Component: "Customer" is never used._
Which after removal will result in Address never being used.

**Expected behavior**
Show all components are not referenced, whether it's directly or indirectly.

**Redocly Version(s)**
1.0.0-beta.109

RomanHotsiy:
Yes, I believe we mark `Address` as used when we traverse `components`.
An easy fix may be to disable marking usage from `components` 🤔 .

adamaltman:
Also a remove-unused-components issue here: #944

jazanne:
this is pretty frustrating... i had to write a script that keeps rerunning command until all unused components are actually removed

tatomyr:
Tangentially related: https://github.com/Redocly/redocly-cli/issues/1209

PR #1472, opened 2024-03-05
fix: remove unused components transitively

fb55:
## What/Why/How?

Removes components if their parent paths were removed. Loops until no more changes were made.

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/874

## Testing

Tests were updated to cover the new logic.

## Check yourself

## Security


fb55:
@tatomyr Please have a look!

tatomyr:
@fb55 sure thing, will take a look!

tatomyr (changes requested):
Sorry for the delay! 

Although I like the approach, there's one major issue with that solution to be fixed. 
Also, I left a few style change suggestions for the sake of readability (also applicable to the OAS2 variant). 

And one more thing. Whatever changes here should also apply to the corresponding rule [no-unused-components](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/no-unused-components.ts). 
 
Again, thank you for your efforts. I hope you'll find my comments useful 🙂.

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -49,17 +56,32 @@ export const RemoveUnusedComponents: Oas3Decorator = () => {
...
-              delete root.components[componentType];
+        const removedPaths: string[] = [];
+
+        let lastRemoveCount = 0;
+        do {
+          lastRemoveCount = 0;
+          for (const [path, { used, name, componentType }] of components) {
+            const isUsed = used?.some(
+              (location) =>
+                !removedPaths.some((removed) => location.absolutePointer.startsWith(removed))
```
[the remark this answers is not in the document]
fb55: Great catch! Adding a check for a trailing slash.

fb55:
Thanks for the review, addressed all comments!

> And one more thing. Whatever changes here should also apply to the corresponding rule [no-unused-components](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/no-unused-components.ts).

Do you mean introducing the additional data structures for tracking where a component is used?

tatomyr:
> Do you mean introducing the additional data structures for tracking where a component is used?

Actually, at a second glance, I'm not sure we should apply the same logic in the rule itself... We're not removing transitive components but rather checking the current state. 

Let me have another look.

packages/core/src/decorators/oas2/remove-unused-components.ts
```diff
@@ -16,15 +16,44 @@ export const RemoveUnusedComponents: Oas2Decorator = () => {
...
+  function removeUnusedComponents(root: Oas2Definition, removedPaths: string[]): number {
+    const removedLengthStart = removedPaths.length;
+
+    for (const [path, { usedIn, name, componentType }] of components) {
+      const used = usedIn?.some(
+        (location) =>
+          !removedPaths.some(
+            (removed) =>
+              location.absolutePointer.startsWith(removed) &&
+              (location.absolutePointer.length === removed.length ||
```
tatomyr, line 34: Isn't this the same as the following?
```
location.absolutePointer === removed ||
location.absolutePointer.startsWith(removed + '/') 

```

packages/core/src/decorators/oas2/remove-unused-components.ts
```diff
@@ -16,15 +16,44 @@ export const RemoveUnusedComponents: Oas2Decorator = () => {
...
       componentType,
       name,
     });
   }
 
+  function removeUnusedComponents(root: Oas2Definition, removedPaths: string[]): number {
+    const removedLengthStart = removedPaths.length;
+
+    for (const [path, { usedIn, name, componentType }] of components) {
+      const used = usedIn?.some(
```
tatomyr, line 29:
```suggestion
      const used = usedIn.some(
```

Is the optional chaining still needed?

tatomyr (changes requested):
WIth the latest changes it doesn't seem to remove transitive references anymore 😅.

packages/core/src/decorators/oas2/remove-unused-components.ts
fb55: Yes, but it isn't comparing the strings twice & doesn't create a string copy. We are running through this loop a lot, so making it performant seems important.

packages/core/src/decorators/oas2/remove-unused-components.ts
fb55: It is not!

fb55:
> WIth the latest changes it doesn't seem to remove transitive references anymore 😅

Tests are passing, and the codebase I am looking at (~1000 auto-generated components with ~20% unused) is not showing any unused component errors after processing. But the reported number of removed components has gone down from ~220 to ~160 — not quite sure about what happened here.

tatomyr:
> Tests are passing

Indeed they are. Sorry, mixed up with another errors 🙂.

packages/core/src/decorators/oas3/__tests__/remove-unused-components.test.ts
```diff
@@ -168,4 +168,146 @@ describe('oas3 remove-unused-components', () => {
...
+        components:
+          schemas:
+            Unused: # <-- this will be removed correctly
+              type: integer
+            Transitive: # <-- this will be removed correctly
+              type: object
+              properties:
+                link:
+                  $ref: '#/components/schemas/Unused'
+            Used: # <-- this will be removed INCORRECTLY because the pointer starts with a previously removed path
```
tatomyr, line 258: Could you change the comment? It *will not* be removed after the fix 🙂.

packages/core/src/decorators/oas2/__tests__/remove-unused-components.test.ts
```diff
@@ -152,4 +152,77 @@ describe('oas2 remove-unused-components', () => {
...
+                '200':
+                  schema:
+                    $ref: '#/definitions/Used'
+              operationId: listPets
+              summary: List all pets
+        definitions:
+          Unused:
+            enum:
+              - 1
+              - 2
```
tatomyr, line 176: Could you also clean up the tests from the needless fields like that one?

fb55: This was copied from other test cases, and I am hesitant to include a larger cleanup here.

fb55:
@tatomyr Anything I can help with to make this move forward?

fb55:
Realised there was a comment I had not addressed, just did so!

packages/core/src/decorators/oas2/remove-unused-components.ts
fb55: Added a comment!

lornajane (approved):
Changelog is good (approval from @Redocly/technical-writers)

tatomyr (approved):
Great job @fb55! 🔥 
Your changes will be available in the next release.

fb55:
@tatomyr Thanks for shepherding this to the finish line!

Issue #661, opened 2022-04-27
Two schemas are referenced with the same name but different content. Renamed Item to Item-2.


colinmollenhour:
If using subdirectories to organize a large number of components and some of these components share the same file basename (but exist in different subdirectories) you get warnings like this:

```
[6] paths/beta/rating/merchantRates.yaml:59:7 at #/get/parameters/3

Two schemas are referenced with the same name but different content. Renamed fields to fields-5.
```

This also comes up with models like `models/Order/Item.yaml` and `models/Shipment/Item.yaml` because both files are named `Item` so they are merged in as `Item` and `Item-2` and so-on.

**Example spec**

`openapi.yaml`
```
openapi: 3.0.3
info:
  title: test
paths:
  '/api/global/beta/rating/rateGroups':
    $ref: 'paths/beta/rating/rateGroups.yaml'
  '/api/global/beta/rating/merchantRates':
    $ref: 'paths/beta/rating/merchantRates.yaml'
```

`paths/beta/rating/rateGroups.yaml`
```
get:
  summary: List all Rate Groups
  description: foo
  parameters:
    - $ref: ../../../components/parameters/rate/group/fields.yaml
```

`paths/beta/rating/merchantRates.yaml`
```
get:
  summary: List all Merchant Rates
  description: foo
  parameters:
    - $ref: ../../../components/parameters/merchant/rate/fields.yaml
```

**Expected behavior**
The refs are valid so ideally there should not be such warnings and ideally it should not require the refs to be declared explicitly in the top level file as a workaround as that is much more tedious than just directly referencing the files where they are needed.

**`openapi-cli` Version(s)**
1.0.0-beta.73

tatomyr:
Hi @colinmollenhour,
Thank you for reporting. I agree it might feel like the behaviour is not too obvious. We might consider implementing this in future. 
However, you can safely skip these warnings, I believe. Or alternatively you can also try using different file names.

adamaltman:
These warnings are for your benefit (not because there is an issue with your definitions) -- they could be posed as "informational" instead of "warning". It is letting you know we cannot use the same file name when we move the reference from a remote reference to the components section in the bundled file. If you want it to be completely dereferenced, you could use the dereferenced flag.

However, I think there could be a better way for us to come up with schema names. For example, in the case of a conflict, we append `-2`, `-3`, `-etc.` In your example above, we could try something like using the folder name such as `Shipment_Item`.

colinmollenhour:
[...]

PR #1490, opened 2024-03-19
feat: adds useSchemaTitleAsRef in the bundler

kennethaasan:
## What/Why/How?

First of all, thank you for creating redoc/redocly. We already have quite a big codebase with schemas used in AsyncAPI specs, and now we want to use the redoc documentation and redocly bundler for some of our OpenAPI specs. This seems to be the only blocker for us.

## Reference

Please see #661. We're not able to use the bundler because it uses the file names instead of the schema titles as references.

## Testing

Added a test in `packages/core/src/__tests__/bundle.test.ts`

## Check yourself

## Security

### DIFF
--- a/.changeset/five-snails-divide.md
+++ b/.changeset/five-snails-divide.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+adds useSchemaTitleAsRef in the bundler which uses the schema title as ref instead of the file name
--- a/packages/core/src/__tests__/__snapshots__/bundle.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/bundle.test.ts.snap
@@ -233,6 +233,36 @@ components:
 
 `;
 
+exports[`bundle should bundle external refs and use schema title if useSchemaTitleAsRef is true 1`] = `
+openapi: 3.1.0
+paths:
+  /football/match:
+    get:
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FootballMatch'
+  /basketball/match:
+    get:
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BasketballMatch'
+components:
+  schemas:
+    FootballMatch:
+      title: FootballMatch
+      type: object
+    BasketballMatch:
+      title: BasketballMatch
+      type: object
+
+`;
+
 exports[`bundle should bundle external refs and warn for conflicting names 1`] = `
 openapi: 3.0.0
 paths:
--- a/packages/core/src/__tests__/bundle.test.ts
+++ b/packages/core/src/__tests__/bundle.test.ts
@@ -74,6 +74,16 @@ describe('bundle', () => {
     expect(res.parsed).toMatchSnapshot();
   });
 
+  it('should bundle external refs and use schema title if useSchemaTitleAsRef is true', async () => {
+    const { bundle: res, problems } = await bundle({
+      config: new Config({} as ResolvedConfig),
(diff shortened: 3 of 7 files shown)


lornajane:
Thanks @kennethaasan! I think we need to consider if title makes the most sense in the widest number of situations. I am concerned about two things:
 - in general, we try not to add configuration options since it's very easy to quickly get too many to be useful
 - assuming that the title field is present and unique is not a given, repurposing data might not solve the problem for most users.

kennethaasan:
Hey @lornajane, thanks for taking a look. I saw #1496, but for us title makes more sense than the path. What if we combine the two and use title if it exists, and path as a fallback?

Regarding configuration options, I think we can make this (1. title, 2. path) the default behavior. Based on what I see in #1496, the current approach with file name is the least best option.

What do you think? The only alternative we have is to make our own bundler which I would like to avoid.

adamaltman:
@kennethaasan you are proposing a breaking change. @lornajane's PR is also a breaking change (albeit less of one).

kennethaasan:
Yes, but it doesn't have to be. Initially, I proposed adding it as a configuration option. I can imagine having it as a configuration option makes sense because different users might want different names.

lornajane:
How about adding an extension field to indicate to `bundle` how to name the schema, to avoid letting the tool guess the name in situations where it can't guess correctly? I said that I didn't think it was great to re-use `title`, but  `x-component-schema-name` or something would be very specific, wouldn't break anything for anyone not using it, and would give the control that I think you're asking for @kennethaasan .

(let's not finalise that naming without discussion and checking if there's anything already in use that does the same thing)

We can also consider improving the name-guessing and reduce the warnings as separate ideas.

kennethaasan:
Do you mean you must set that extension field in each schema then? We already have quite a big code base with JSON schemas used for AsyncAPI where we bundle with the title, and we want to re-use those schemas when bundling OpenAPI specs and use the title there as well. Setting an extension in each schema will not work for us because it's too much boilerplate and duplicated work. I'm happy to create our own bundler, but I was hoping to use this one :/

tatomyr:
@adamaltman I don't think it's a breaking change. 

@kennethaasan please correct me if I'm wrong, but you've merely added the `useSchemaTitleAsRef` option (default to `false`) to the `bundle` function which I consider a minor change. However, if we were to add a new option, I believe users should be able to add that option through the CLI interface. Also, that needs to be documented.

Another question, what behaviour you would expect if the options are set to true but there's no `title` in a schema? (I'd not mix titles and filenames since there may be collisions between those two either). What approach to choose when titles contain special symbols (e.g. spaces, slashes &c.)?

adamaltman:
@tatomyr why wouldn't it be a breaking change?

We have clients that rely on the `bundle` to generate SDKs, etc...

tatomyr:
> why wouldn't it be a breaking change? (I'm talking about changing the default behavior, not a configuration option. We have clients that rely on the bundle to generate SDKs, etc...)

I mean that the default bundler behaviour is not changing as far as I can judge based on the code changes.

lornajane:
@kennethaasan (going back a few steps in the discussion, hopefully this isn't too disjointed). If we added a specific extension field and used that for the name if it's present, that's not a breaking change. If you already have the information in place in `title` or a similar field, I'd suggest you add a custom plugin to create the new field automatically before bundling.

Let me know if an example would be useful.

kennethaasan:
Sorry for the late response (I've been on holiday). If you think almost no one needs this, I agree with @adamaltman that this should perhaps not be added here. We will instead create and maintain our own bundler.

kennethaasan:
Closing because we made our own bundler that re-uses lots of your bundler.

Issue #1356, opened 2023-12-13
Create a command for linting Redocly config itself


tatomyr:
**Is your feature request related to a problem? Please describe.**

There's no dedicated command to lint a Redocly config file as the `lint` command does for an API description file to ensure the config is valid. It could be especially useful to have one when working on the config changes.

**Describe the solution you'd like**

A command named `lint-config` (or similarly) that will exit with non-zero code and output problems if there are some in the config.

**Describe alternatives you've considered**

Redocly config gets validated each time a command gets executed, so technically it's possible to use any command to achieve something similar, like the following: `redocly bundle --lint-confg=error`.

PR #1491, opened 2024-03-19
feat: command for linting Redocly config itself

DmitryAnansky:
## What/Why/How?

Added a dedicated command `redocly check-config` to lint a Redocly config file as the lint command does for an API description file to ensure the config is valid.

```bash
redocly check-config
redocly check-config [--config=<path>]
```

## Reference

Closes [#1356](https://github.com/Redocly/redocly-cli/issues/1356)

## Check yourself

## Security


__tests__/verify-config/invalid-config--lint-config-off/snapshot.js
```diff
@@ -0,0 +1,6 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E verify-config test with option: { dirName: 'invalid-config--lint-config-off', option: 'off' } 1`] = `
+
+
```
tatomyr, line 5: Does this option even makes sense? Maybe we can restrict the command to not accept the `off` severity?

__tests__/helpers.ts
```diff
@@ -12,6 +12,7 @@ type CLICommands =
   | 'login'
   | 'logout'
   | 'preview-docs'
+  | 'verify-config'
```
tatomyr, line 15: I'm not sure about the name though. Most likely, we don't want to name it `lint-config` as we already have such an option (so it will be confusing to use the same for a command name), but `verify-config` also doesn't sound right to me. What are other options? `config`, `validate-config`, `check-config`, ...?

packages/cli/src/wrapper.ts
```diff
@@ -11,7 +11,7 @@ import { lintConfigCallback } from './commands/lint';
 import type { CommandOptions } from './types';
 
 export function commandWrapper<T extends CommandOptions>(
-  commandHandler: (argv: T, config: Config, version: string) => Promise<void>
+  commandHandler?: (argv: T, config: Config, version: string) => Promise<void>
```
tatomyr, line 14: I'd better go with a dummy command handler (e.g. `() => {}`) instead of making it optional.

packages/cli/src/utils/getCommandNameFromArgs.ts
```diff
@@ -0,0 +1,9 @@
+import { Arguments } from 'yargs';
+
+export function getCommandNameFromArgs(argv: Arguments | undefined): string | number {
+  if (!argv || !argv._ || !argv._.length) {
+    return '';
+  }
+
+  return argv._[0];
```
tatomyr, line 8: Would this also work? 

```suggestion
  return argv?._?.[0] ?? '';
```

DmitryAnansky: Can be shorter, applied.

packages/cli/src/wrapper.ts
DmitryAnansky: Works both ways, changed to suggested.

__tests__/verify-config/invalid-config--lint-config-off/snapshot.js
DmitryAnansky: `off` removed

__tests__/helpers.ts
DmitryAnansky: Yeah, `lint-config` was already in use, and that is the reason I have chosen the command name such as `verify-config`. I don't have a strict preference in this naming, so can be `validate-config` or smth else.
Let's discuss it to finalize.

cc: @lornajane , @RomanHotsiy , @adamaltman

.changeset/tasty-waves-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Add `verify-config` command to validate a Redocly configuration file.
```
tatomyr, line 5:
```suggestion
Added `verify-config` command to validate a Redocly configuration file.
```

packages/cli/src/wrapper.ts
```diff
@@ -10,8 +10,14 @@ import {
...
 
+type CommandHandler<T extends CommandOptions> = (
+  argv: T,
+  config: Config,
+  version: string
+) => Promise<void> | void;
+
 export function commandWrapper<T extends CommandOptions>(
-  commandHandler: (argv: T, config: Config, version: string) => Promise<void>
+  commandHandler: CommandHandler<T> = async () => {}
```
tatomyr, line 20: Is there a need to specify the default function?

__tests__/helpers.ts
lornajane: I think `check-config` could be more clear, and we can encourage users to run it as a CI "check" as well.

docs/commands/verify-config.md
```diff
@@ -0,0 +1,37 @@
+# `verify-config`
+
+## Introduction
+
+This is a dedicated command to lint a Redocly config file as the lint command does for an API description file to ensure the config file is valid.
```
lornajane, line 5:
```suggestion
Use this command to check that everything in a Redocly configuration file is valid and in the expected format.
Adding this check before using the configuration file with other commands can catch any problems at an early stage.
This command uses the same mechanism as our API linting to match a file against an expected data structure.
```

packages/cli/src/wrapper.ts
DmitryAnansky: Its workaround is to allow an empty function param `()=>{}`.
I prefer variant with optional parameter and check if we need to call `commandHandler`

DmitryAnansky: Or maybe we can change type to 
```
type CommandHandler<T extends CommandOptions> =
  | ((argv: T, config: Config, version: string) => Promise<void>)
  | ((argv: T) => void);
```

lornajane (commented):
Our lint and bundle commands output which file(s) they are linting/bundling, could this one do the same? (I know it should be obvious, but it's one more clue to a user that might help them to help themselves!)

DmitryAnansky:
> Our lint and bundle commands output which file(s) they are linting/bundling, could this one do the same? (I know it should be obvious, but it's one more clue to a user that might help them to help themselves!)

It does the same linting before running those commands. So the output is the same when errors occur.
Changes here only allow to have separate commands to self-validate configuration and not proceed to any other actions.
I only added an additional `config is valid` output in the case we run `verify-config` to avoid no message in the success case.

lornajane:
But in the success case: successful bundle says what file it bundled. Successful lint says what file it linted. Successful check-config doesn't mention a file.

lornajane:
The deprecation warning notices come out after the success thing. I think everything is correct here, but that the success message should be the last thing listed.

I'm seeing:
```
✅  Your config is valid.
The 'features.openapi' field is deprecated. Use theme.openapi instead. Read more about this change: https://redocly.com/docs/api-registry/guides/migration-guide-config-file/#changed-properties
```

lornajane:
Overall: This is in really good shape! I think we should rename the command, but it works like a charm! I haven't been able to find a config file in any of my testing directories or the cookbook that it doesn't correctly either lint or report a problem with. Impressive :)

docs/commands/verify-config.md
DmitryAnansky: Applied, thx

__tests__/helpers.ts
DmitryAnansky: @RomanHotsiy, @adamaltman any obections?

DmitryAnansky: 'check-config' it is.
I will rename it soon. Thx

tatomyr (approved):
Thanks!

lornajane (approved):
Super happy to have this feature! Thank you @DmitryAnansky

PR #1496, opened 2024-03-21
feat: Add path-aware prefixes instead of numeric suffixes when same-named schemas are found during bundling

lornajane:
## What/Why/How?

Related to #661 , a possible improvement to avoiding name clashes in schema naming.

## Check yourself

## Security


adamaltman (commented):
This is an idea, but it's not one that all of our customers would like and I'm not sure it should be the default behavior.

lornajane:
Proposal not pursued and no longer needed.

Issue #1250, opened 2023-09-07
Drop linting option from bundle command


lornajane:
**Is your feature request related to a problem? Please describe.**

The bundle command includes some confusing options (format) that can apply to the linting part of the command. This also produces confusing output because bundle should output a single API description, and linting outputs some error/warning feedback (or the sucess message).

**Describe the solution you'd like**

Since we see very little usage of bundle combined with lint, please drop the option to add linting to the bundle command, and encourage users to run these steps separately (will need docs updates as well as tools changes).

I think this means we need to:
- [ ] Remove `--format` option (this relates to problem reporting)
- [ ] Remove `--max-problems` (this relates to problem reporting)
- [ ] Remove `--skip-rule` (rules aren't used when bundling)
- [ ] Remove `--lint`
- [ ] Remove `--extends` (rulesets aren't used when bundling)
- [ ] Update related documentation for bundle command
- [x] Add a guide on linting and then bundling (we'll edit this if we add command chaining support) #1385 
- [ ] If a user users any of the removed options, let them know to use the commands separately and link to the guide.

**Describe alternatives you've considered**

Require filenames only. But command chaining is an expected feature of CLI tools, so I think we should support it and adjust the tools accordingly.

tatomyr:
Related: https://github.com/Redocly/redocly-cli/issues/1397.

PR #1503, opened 2024-03-25
feat: drop linting option from bundle command

DmitryAnansky:
## What/Why/How?

Drop linting option from `bundle` command

- [x]  Remove --format option (this relates to problem reporting)
- [x]  Remove --max-problems (this relates to problem reporting)
- [x]  Remove --skip-rule (rules aren't used when bundling)
- [x]  Remove --lint
- [x]  Remove --extends (rulesets aren't used when bundling) **IT IS DESIDED TO KEEP IT**
- [x]  Update related docmentation for bundle command
- [x]  Add a guide on linting and then bundling (we'll edit this if we add command chaining support)
- [x]  If a user users any of the removed options, let them know to use the commands separately and link to the guide.

## Reference

Closes [#1250](https://github.com/Redocly/redocly-cli/issues/1250)
## Check yourself

## Security


DmitryAnansky:
@tatomyr / @lornajane 
I am not sure we need to remove `--extends` option from the `bundle` command.
It is actually used in codebase inside pre-bundle (actually pre-any command) `commandWrapper`.
Its applied in the auto-linting of the configuration

```
      const config: Config = (await loadConfigAndHandleErrors({
        configPath: argv.config,
        customExtends: argv.extends as string[] | undefined,
        region: argv.region as Region,
        files: argv.files as string[] | undefined,
        processRawConfig: lintConfigCallback(argv as T & Record<string, undefined>, version),
```
Please let me know if you have any strong arguments against keeping `--extends` option in the `bundle` command.

DmitryAnansky:
@lornajane 

Could you please let me know if we need some extra docs changes, related to this task point
`Add a guide on linting and then bundling (we'll edit this if we add command chaining support)` ?

I have found another guide that was added previously `docs/guides/lint-and-bundle.md` and it looks like it is ready to use.

lornajane:
Indeed, that guide entry is incomplete, I'm not sure how we didn't update the original details in the issue, but I've done that now.

packages/cli/src/index.ts
```diff
@@ -493,8 +470,46 @@ yargs
           choices: ['warn', 'error', 'off'] as ReadonlyArray<RuleSeverity>,
           default: 'warn' as RuleSeverity,
         },
+        format: {
+          hidden: true,
+          deprecated: true,
+        },
+        lint: {
```
DmitryAnansky, line 477: Need to keep those options here if we want notify users, otherwise strict mode will force `help` display.

packages/cli/src/index.ts
```diff
@@ -493,8 +470,46 @@ yargs
...
+      const LINT_AND_BUNDLE_DOCUMENTATION_LINK =
+        'https://redocly.com/docs/cli/guides/lint-and-bundle/#lint-and-bundle-api-descriptions-with-redocly-cli';
+
+      DEPRECATED_OPTIONS.forEach((option) => {
+        if (argv[option]) {
+          process.stdout.write(
+            `Option --${option} is no longer supported. Please use separate commands, as described in the ${LINT_AND_BUNDLE_DOCUMENTATION_LINK}.`
+          );
+          process.stdout.write('\n');
+          process.exit(1);
```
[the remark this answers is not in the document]
lornajane: This is a great question and I don't know the right answer! I hate to break people's pipelines but we probably need to do what you've done here so they realise that action is needed on their side. Otherwise we're not linting when they think we are, and that could store up problems later.

.changeset/gorgeous-countries-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Removed linting options from the `bundle` command.
```
lornajane, line 5:
```suggestion
Removed lint support from the bundle command to support a wider range of use cases. Users should update to [run lint and bundle separately](https://redocly.com/docs/cli/guides/lint-and-bundle/).
```

packages/cli/src/index.ts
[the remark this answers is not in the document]
lornajane: How about:
 - return the wording and link from the deprecation warning as an error
 - then show the usage information as we would for any other unknown command

lornajane:
Thank you @DmitryAnansky !! I made some minor comments, but this change is very good and works well in my local CLI context. What I'm not sure about it how to test which of our own dependencies use this feature and whether we'll end up breaking something in one of our cloud platforms.

docs/commands/bundle.md
```diff
@@ -20,32 +20,23 @@ redocly bundle --version
...
+| --help                     | boolean  | Show help.                                                                                                                                                                                                                                                      |
+| --keep-url-references, -k  | boolean  | Keep absolute url references.                                                                                                                                                                                                                                   |
+| --lint-config              | string   | Specify the severity level for the configuration file. <br/> **Possible values:** `warn`, `error`, `off`. Default value is `warn`.                                                                                                                              |
```
adamaltman, line 33: Is this still relevant?

DmitryAnansky: Yes, this `lint-config` settings is still relevant.
It is used for command wrapper checks.

packages/cli/src/index.ts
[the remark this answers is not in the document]
DmitryAnansky: Looks good for me.
I have updated this part, so the output would look like =>
Screenshot 2024-03-27 at 15 15 53

packages/cli/src/index.ts
```diff
@@ -16,7 +17,7 @@ import { handlerBuildCommand } from './commands/build-docs';
 import { cacheLatestVersion, notifyUpdateCliVersion } from './utils/update-version-notifier';
 import { commandWrapper } from './wrapper';
 import { version } from './utils/update-version-notifier';
-import type { Arguments } from 'yargs';
+import { Arguments, showHelp } from 'yargs';
```
tatomyr, line 20: Please import the showHelp separately.

tatomyr (commented):
~Please also check how it works in the VS Code extension.~ Sorry, no need checking in the extension as there's no changes in the openapi-core package.

tatomyr (approved):

lornajane (approved):

Issue #1397, opened 2024-01-17
Remove combination operations from `join`


lornajane:
Join has CLI options to also lint, decorate, and preprocess. These operations all have their own commands (or they're part of bundle), they default to false.

Remove these features from `join` before it gets its stable release, users should combine the commands themselves in the order that makes sense for their use case. Users can join in one operation, then lint (or lint and then join, depending on their needs. Or both!), or join and then bundle to apply the transforms. The join command should just join.

PR #1507, opened 2024-03-26
feat: remove --lint, --preprocess, and --decorate options from the join command

DmitryAnansky:
## What/Why/How?

Remove options from the `join` command:

- [x]  lint
- [x] decorate
- [x] preprocess

## Reference

Closes [#1397](https://github.com/Redocly/redocly-cli/issues/1397)

## Check yourself

## Security


.changeset/eight-fireants-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Removed combination operations like `lint`, `decorate` and `preprocess` from the `join` command.
```
HCloward, line 5:
```suggestion
Removed combination operations like `lint`, `decorate` and `preprocess` from the `join` command, so users can combine the commands in whatever order works best for their use case.
```

lornajane (approved):
Looks good and locally seems to behave as I would expect. Thanks!

.changeset/eight-fireants-provide.md
DmitryAnansky: Will apply variant suggested by @lornajane.

tatomyr:
I don't reckon this PR completely resolves the original issue. 
@lornajane, besides the command line options, decorators and preprocessors can be configured through a config file (i.e. in the `apis` section). They will still be executed. Did you mean only to remove the options or the entire possibility to fire decorators/preprocessors with the `join` command?

lornajane:
Ah, yes. The join command should not do any preprocessing at all, for any reason. That functionality should be only in bundle.

DmitryAnansky:
> Ah, yes. The join command should not do any preprocessing at all, for any reason. That functionality should be only in bundle.

So in current changes we are skipping default one
```

 const preprocessors = new Set([
    ...Object.keys(config.styleguide.preprocessors.oas3_0),
    ...Object.keys(config.styleguide.preprocessors.oas3_1),
    ...Object.keys(config.styleguide.preprocessors.oas2),
  ]);
  config.styleguide.skipPreprocessors(Array.from(preprocessors));
```

packages/cli/src/commands/join.ts
```diff
@@ -72,15 +64,13 @@ export type JoinOptions = {
   'lint-config'?: RuleSeverity;
 };
 
-export async function handleJoin(argv: JoinOptions, config: Config, packageVersion: string) {
+export async function handleJoin(argv: JoinOptions, config: Config) {
```
tatomyr, line 67: I believe the `packageVersion` is still needed as it should be used [here](https://github.com/Redocly/redocly-cli/pull/1507/files#diff-3d0894febbd7923f278d468139041923d786a33f480741eabb8d465b9040eb4cR135) instead of the document version (most likely that's a mistake).

lornajane:
My perspective is more about how the user sees the feature. It makes sense that `join` would bundle behind the scenes - but does the user know or care about that? I didn't at all expect that a `join` command would also perform our `transform` operations, it's fine to bundle and resolve the refs, but I don't think it's useful for decorators to be applied here. Quite possibly people will need to make some decorator changes as well, either before or after joining, but we have that functionality and they can built it into their pipelines in a way that's clear.

lornajane:
Updated to add: let's signpost the decorators from the join command docs.

packages/cli/src/commands/join.ts
DmitryAnansky: updated, please take a look

tatomyr (approved):
Looks good. Thanks!

lornajane:
"Signpost" literally a post with a sign on it!! But in documentation terms, it means to help users to find the direction, by adding information and a link about something to "signpost" them to another thing. In this instance, you can translate my request as "let's add some information and a link from the join documentation to the decorators documentation"

DmitryAnansky:
> "Signpost" literally a post with a sign on it!! But in documentation terms, it means to help users to find the direction, by adding information and a link about something to "signpost" them to another thing. In this instance, you can translate my request as "let's add some information and a link from the join documentation to the decorators documentation"

Added similar console notification messages as for bundle deprecated options.
@lornajane , please take a look.

lornajane (approved):
Pushed a docs change, hope you don't mind. It looks good.

PR #1509, opened 2024-03-29
Ignore validating example with externalValue against schema

xeptore:
## What/Why/How?

Currently, when an example object has `externalValue` property [`no-invalid-media-type-examples`](https://redocly.com/docs/cli/rules/no-invalid-media-type-examples/) lint rule fails as it is validating the _example_ object (the parent object) against the schema definition.

I could assume there are couple of ways to handle this case. However, as there is already another lint rule ([`no-example-value-and-externalValue`](https://redocly.com/docs/cli/rules/no-example-value-and-externalValue/)), which is kind of related to this one, and its default severity level causes either one of `value`, or `externalValue` properties to exist for a specific example object, I decided to go with the simple exclusion condition, which basically deactivates the validation for example objects that has `externalValue` set. I also intentionally didn't implement the logic to verify the specified URL set for `externalValue` to check whether the target resource actually matches the schema definition or not (by issuing HTTP request for example) as it's both complex to implement properly due to various cases including relative URLs, and for external resources being unpredictable by nature, e.g., the server might be unavailable at the time of linting, or returing a temporary failure due to network issues.

## Reference

I faced the issue while I was trying to add file examples for request, and response bodies of an API according to the official documentation ([here](https://swagger.io/docs/specification/adding-examples/#external)).

## Testing

Relevant tests are updated to cover this scenario.

## Screenshots (optional)

It can be tested by disabling the condition I added to the lint rule.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
+++ b/packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
@@ -474,4 +474,149 @@ describe('no-invalid-media-type-examples', () => {
 
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
   });
+
+  it('should not report if only externalValue is set', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.0.0
+        paths:
+          /pet:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          a:
+                            type: string
+                          b:
+                            type: number
+                      examples:
+                        first:
+                          externalValue: "https://example.com/example.json"
+      `,
+      'foobar.yaml'
+    );
+
+    const results = await lintDocument({
+      externalRefResolver: new BaseResolver(),
+      document,
+      config: await makeConfig({ 'no-invalid-media-type-examples': 'error' }),
+    });
+
+    expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
+  });
+
+  it('should not report if value is valid and externalValue is also set', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.0.0
+        paths:
+          /pet:
+            get:
+              responses:
+                '200':
+                  content:
+                    application/json:
+                      schema:
+                        type: object
+                        properties:
+                          a:
+                            type: string
+                          b:
+                            type: number
+                      examples:
+                        first:
+                          externalValue: "https://example.com/example.json"
+                          value:
+                            a: "A"
+                            b: 0
+      `,
+      'foobar.yaml'
+    );
+
+    const results = await lintDocument({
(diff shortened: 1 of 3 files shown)


packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -35,6 +35,9 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
             location = isMultiple ? resolved.location.child('value') : resolved.location;
             example = resolved.node;
           }
+          if (Object.prototype.hasOwnProperty.call(example, 'externalValue')) {
```
tatomyr, line 38: Hi @xeptore. Thanks for bringing this up! 

When both `value` and `externalValue` are present, the proposed solution does not validate the `value` at all. Theoretically, someone could disable the `no-example-value-and-externalValue` rule and that would prevent the wrong value from being caught. 

My suggestion is to check if the `value` actually exists, and if so, validate it. Otherwise, just skip the validation:

```suggestion
          if (isMultiple && typeof example.value === 'undefined') {
```

Does that make sense?

xeptore: Sure, committed in 728ebac2168c9004093de27cde2909026e9adc3e.

packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
```diff
@@ -458,6 +460,8 @@ describe('no-invalid-media-type-examples', () => {
                       examples:
                         first:
                           value: {}
+                        second:
```
tatomyr, line 463: Could you also create a separate test for the case instead of extending the existing ones? I'm afraid that the descriptions might not be too helpful. 
Please don't forget to add the case when both `externalValue` and `value` (correct and incorrect) are defined.

tatomyr (commented):
Left one more comment

packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
xeptore: Done in 31758ac4a22e214c6667ba41717cffe9867b917b.

xeptore:
> Also, please add a patch [changeset](https://github.com/Redocly/redocly-cli/blob/main/CONTRIBUTING.md#pull-request-guidelines) for both `core` and `cli` packages describing the changes. Thanks!

Done in 82869150e676cf1d59d35e2f09cc6faf86027474.

> P.S. Please don't forget to run prettier 🙂

Done in cf02cec879c8e9e557fae2f2f7093f802741da80.

lornajane (approved):
Looks good to me, but will need @tatomyr approval as well (I'm really only here for @Redocly/technical-writers )

tatomyr:
Thank you @xeptore for spending your time on this! 👍

xeptore:
> Thank you @xeptore for spending your time on this! 👍

Happy to help!
Cheers

Issue #1223, opened 2023-08-08
TypeExtension isn't available to configurable rules


lornajane:
**Describe the bug**

I used a custom plugin and added a new type to my type tree. I expected to be able to use the type as a subject in a configurable rule, but I get the error ``type` can be one of the following only: "any", "Root", "Tag", "TagList", ...` as if my new type isn't available. I think the type is correctly added because the `spec` rule picks up if the structure of the data isn't correct.

**To Reproduce**

1. With a custom plugin `plugins/type-extension.js` that contains the following code:
```
const XMetaData = {
  properties: {
    lifecycle: { type: 'string'},
    'owner-team': { type: 'string'},
  },
  required: ['lifecycle']
};

module.exports = {
  id: 'type-extension',
  typeExtension: {
    oas3(types) {
      newTypes = {
        ...types,
        XMetaData: XMetaData,
        Info: {
          ...types.Info,
          properties: {
            ...types.Info.properties,
            'x-metadata': 'XMetaData',
          }
        }
      };
      return newTypes;
    }
  }
};

```

1. Given this `redocly.yaml` file

```
extends: []

plugins:
  - 'plugins/type-extension.js'

rules:
  spec: warn
  rule/metadata-lifecycle:
    subject:
      type: XMetaData
      property: lifecycle
    assertions:
      enum: ["alpha", "beta", "production", "deprecated"]
```

1. And an OpenAPI file with this (snippet from the start of the file)

```
  x-metadata:
    lifecycle: production
    owner-team: Engineering/Integrations
```

5. Run this command with these arguments... `redocly lint openapi.yaml`
6. See error

**Expected behavior**

I want to be able to write a configurable rule that uses the new type.

**Redocly Version(s)**

1.0.0

**`Node.js` Version(s)**
v16.20.0

PR #1511, opened 2024-04-03
fix: not assert extended types when linting configuration

DmitryAnansky:
## What/Why/How?

This PR holds a small changes that will allow to ignore custom extended types during configuration file validation phase.

NOTE: supported files will be removed later

- [x] Ignore extended types
- [x] Test VSCode extention
- [ ] Delete supported content before merge

Alternative solution would include relatively extensive refactoring and influence all commands.
This will include changing lintConfig order and resolving plugins and types earlier, witch might have another complexity with extension located elsewhere. 

## Reference

Closes [#1223](https://github.com/Redocly/redocly-cli/issues/1223)

## Check yourself

## Security

### DIFF
--- a/bughunting/redocly.yaml
+++ b/bughunting/redocly.yaml
@@ -1,0 +1,13 @@
+extends: []
+
+plugins:
+  - plugins/type-extention.js
+
+rules:
+  spec: warn
+  rule/metadata-lifecycle:
+    subject:
+      type: XMetaData
+      property: 'lifecycle'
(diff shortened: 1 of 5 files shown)


bughunting/redocly.yaml
```diff
@@ -0,0 +1,13 @@
+extends: []
```
DmitryAnansky, line 1: TODO: support files will be removed before merge

PR #1527, opened 2024-04-15
fix: resolve config extensions using only one resolver

tatomyr:
## What/Why/How?

Currently, when resolving the config, the external refs' resolver gets re-declared inside the `resolveConfig` function. That makes caching some external resources (particularly those from the `extends` config section) impossible. Also, those disconnected resolvers make it impossible to substitute the fs provider (it's useful when working in a browser environment).

This PR eliminates the discrepancy and allows using one `externalRefResolver` throughout the entire `loadConfig` function.

## Check yourself

## Security

### DIFF
--- a/.changeset/many-steaks-hope.md
+++ b/.changeset/many-steaks-hope.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Changed the `loadConfig` function to avoid overwriting the `externalRefResolver` internally. That improves caching external config resources.
--- a/.changeset/many-steaks-hope.md
+++ b/.changeset/many-steaks-hope.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Changed the `loadConfig` function to avoid overwriting the `externalRefResolver` internally. That improves caching external config resources.
+Improved caching for external configuration resources.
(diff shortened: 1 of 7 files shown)


RomanHotsiy (approved):
LGTM

.changeset/many-steaks-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Changed the `loadConfig` function to avoid overwriting the `externalRefResolver` internally. That improves caching external config resources.
```
lornajane, line 5: Could you re-visit this explanation? What changed for the user?

tatomyr: @lornajane Nothing really changes from the user's prospective. This is more of an internal change which helps improving performance in other tools (like the VS Code extension, Realm &c.).

```suggestion
Fixed the use of `externalRefResolver` in the `loadConfig` function, which improves caching external config resources.
```

DmitryAnansky (approved):

.changeset/many-steaks-hope.md
lornajane: I don't think the user has much knowledge of our internal naming, so how about 
```suggestion
Improved caching for external configuration resources.
```

tatomyr: Much better, thanks!

lornajane (approved):

PR #1536, opened 2024-04-23
fix: improve error messages

tatomyr:
## What/Why/How?

When an assertion is supposed to run only on object keys and the assertion subject contains the `property` key, the error is `  - <the assertion name> can't be used on a single property. Please use 'property'..` which should be the other way around (thanks @adamaltman  for pointing on that). 
This PR fixes that, and also cleans up the error messages by removing needless periods and using better formatting. 

## Check yourself

## Security


__tests__/lint/assertions-property-key-not-allowed/snapshot.js
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E lint assertions-property-key-not-allowed 1`] = `
+
+validating /openapi.yaml...
+Something went wrong when processing /Users/andrewtatomyr/Tests/redocly-cli-backup/__tests__/lint/assertions-property-key-not-allowed/openapi.yaml:
+
+  - The 'disallowed' assertion can't be used on properties. Please remove the 'property' key.
```
tatomyr, line 8: Currently the error is:

`  - disallowed can't be used on a single property. Please use 'property'..`

__tests__/lint/assertions-property-key-required/snapshot.js
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E lint assertions-property-key-required 1`] = `
+
+validating /openapi.yaml...
+Something went wrong when processing /Users/andrewtatomyr/Tests/redocly-cli-backup/__tests__/lint/assertions-property-key-required/openapi.yaml:
+
+  - The 'nonEmpty' assertion can't be used on all keys. Please provide a single property.
```
tatomyr, line 8: The current error is:
` - nonEmpty can't be used on all keys. Please provide a single property.`

adamaltman (approved):

RomanHotsiy (approved):

PR #1538, opened 2024-04-24
docs: clarify the use of the `where` section in configurable rules

tatomyr:
## What/Why/How?

This PR clarifies how to use the `where` section in configurable rules.

## Check yourself

## Security


docs/rules/configurable-rules.md
```diff
@@ -65,6 +65,8 @@ A minimum of one assertion property is required to be defined.
 ## Where object
 
 The `where` object is part of a `where` list which must be defined in order from the root node.
+A node cannot be used in the `wrhere` list twice.
```
adamaltman, line 68:
```suggestion
A node cannot be used in the `where` list twice.
```

docs/rules/configurable-rules.md
```diff
@@ -65,6 +65,8 @@ A minimum of one assertion property is required to be defined.
...
+However, the same node could be used in the latest `where` object and in the root `subject` object.
```
adamaltman, line 69: I wonder if "latest" is clear? is that first? last?

tatomyr: I meant to write last, thanks.

docs/rules/configurable-rules.md
```diff
@@ -65,6 +65,8 @@ A minimum of one assertion property is required to be defined.
 ## Where object
 
 The `where` object is part of a `where` list which must be defined in order from the root node.
+A node cannot be used in the `where` list twice.
+However, the same node could be used in the latest `where` object and in the root `subject` object.
```
tatomyr, line 69:
```suggestion
However, the same node could be used in the last `where` object and in the root `subject` object.
```

docs/rules/configurable-rules.md
```diff
@@ -65,6 +65,8 @@ A minimum of one assertion property is required to be defined.
 ## Where object
 
 The `where` object is part of a `where` list which must be defined in order from the root node.
+A node cannot be used in the `where` list twice.
```
lornajane, line 68: Do you have a handy example to illustrate what you're trying to say here? I'm not sure I quite get it

tatomyr: Yes, I do. 

A rule like the following won't work because of the duplicated `Schema` subject in the list:

```yaml
    subject: 
      type: Schema
    assertions: 
      required: 
        - minLength
        - maxLength
    where: 
      - subject: 
          type: Schema # <-- 
          property: type
        assertions:  
          const: string
      - subject: 
          type: Schema # <--
          property: enum
        assertions: 
          defined: false        

```

Instead, we have to group the assertions the other way:

```yaml
    subject: 
      type: Schema
    assertions: 
      requireAny: 
        - minLength
        - maxLength
        - enum
      mutuallyRequired: 
        - minLength
        - maxLength
    where: 
      - subject: 
          type: Schema # <-- used once in the list
          property: type
        assertions:  
          const: string
```

lornajane: I think I'm getting there. How about this:

```suggestion
Each node can only be used in one `where` objects for each assertion.
Rules that use multiple `where` objects must target each one on a different node.
```

docs/rules/configurable-rules.md
```diff
@@ -65,19 +65,25 @@ A minimum of one assertion property is required to be defined.
 ## Where object
 
 The `where` object is part of a `where` list which must be defined in order from the root node.
+Each node can only be used in one `where` objects for each assertion.
```
tatomyr, line 68:
```suggestion
Each node can only be used in one `where` object for each assertion.
Each subsequent node must be a descendant of the previous one.
```

lornajane (approved):
Thanks for adding this

Issue #1569, opened 2024-05-27
Ability to exclude some operations from security-defined rule


RomanHotsiy:
**Is your feature request related to a problem? Please describe.**

Very often API define a health endpoint which is not protected by any kind of auth.

Our `security-defined` rule always reports this as a problem.

Surely, it can be ignored with ignore file but this is a problem for users who constantly onboard new APIs.
Configurable rule can't be used here as it doesn't cover security defined on different levels (operation vs root).

**Describe the solution you'd like**

Add support for `exceptions` config option to the rule:

```yaml
rules:
  security-defined:
    severity: error
    exceptions:
      - path: '/health'
        method: GET
```

**Describe alternatives you've considered**

I used `exceptions` as we use this term in `path-segment-plural` rule. Let me know if you have better suggestions.

Another alternative is to have a simpler list of exceptions (only path)

```yaml
rules:
  security-defined:
    severity: error
    exceptions:
      - '/health'
```

What do you think?

adamaltman:
FYI:

> If you define security on the operation or path item as `security: []` it means security is explicitly defined as no security, it won't trigger the rule.

I mention it in the docs here: https://redocly.com/docs/cli/rules/security-defined/#api-design-principles

tatomyr:
@RomanHotsiy do you think we still need this adjustment? Is there a case that couldn't be easily covered by the empty security as @adamaltman suggests?

RomanHotsiy:
Yes. Still makes sense.

PR #1570, opened 2024-05-27
feat: add ability to exclude some operations from security-defined rule

tatomyr:
## What/Why/How?

Adds the ability to exclude some paths or particular operations from the `security-defined` rule.

Here's an example:

**redocly.yaml**
```yaml
rules:  
  security-defined: 
    exceptions:
      - path: /partially-skipped
        methods:
          - get
      - path: /fully-skipped
```

**openapi.yaml**
```yaml
openapi: 3.1.0
paths:
  /partially-skipped:
    get:
      description: Skipped.
    post: 
      description: Has security.
      security: []
    delete: 
      description: Should have security defined!
  /fully-skipped:
    get:
      description: Skipped.
    post: 
      description: Skipped.
  /required:
    get: 
      summary: Should have security defined!
```

The ouput:

image

## Reference

Resolves: https://github.com/Redocly/redocly-cli/issues/1569

## Check yourself

## Security


RomanHotsiy (approved):
LGTM 👍

lornajane:
Looking at #1569 , it would be better to have the users declare the security as open than to omit it - the benefit to the user is that the good practice is part of their OpenAPI description so will apply throughout the lifecycle and all tools.

RomanHotsiy:
> Looking at https://github.com/Redocly/redocly-cli/issues/1569 , it would be better to have the users declare the security as open than to omit it - the benefit to the user is that the good practice is part of their OpenAPI description so will apply throughout the lifecycle and all tools.

Yes! Definitely it's better and that's why we have this rule. Unfortunately there are cases when it can't be done due to various reasons (like not controlling the openapi description directly multiplied by 100s of APIs). So it's good to have some way for exceptions.

JLekawa (approved):

PR #1571, opened 2024-05-28
chore: update error messages

tatomyr:
## What/Why/How?

Updated a couple of error messages to make the format more consistent.

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/commands/join.ts
+++ b/packages/cli/src/commands/join.ts
@@ -68,7 +68,7 @@ export async function handleJoin(argv: JoinOptions, config: Config, packageVersi
   const startedAt = performance.now();
 
   if (argv.apis.length < 2) {
-    return exitWithError(`At least 2 apis should be provided. \n\n`);
+    return exitWithError(`At least 2 apis should be provided.`);
   }
 
   const fileExtension = getAndValidateFileExtension(argv.output || argv.apis[0]);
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/join.ts
```diff
@@ -627,8 +625,8 @@ export async function handleJoin(argv: JoinOptions, config: Config, packageVersi
     const firstApi = documents[0];
     const openapi = firstApi.parsed;
     const componentsPrefix = getInfoPrefix(openapi.info, prefixComponentsWithInfoProp, COMPONENTS);
-    if (!openapi.openapi) exitWithError('Version of specification is not found in. \n');
-    if (!openapi.info) exitWithError('Info section is not found in specification. \n');
+    if (!openapi.openapi) exitWithError('Version of specification is not found in.');
```
lornajane, line 628: This isn't a sentence. I know we're mostly doing formatting in this PR but I can't unsee this now!

```suggestion
    if (!openapi.openapi) exitWithError('Version of specification is not found.');
```

lornajane (commented):
Minor thing, also I suggest we add a changelog here despite it being a chore. It wouldn't be too surprising if people were matching error messages in their own code, so we should probably let them know that they might see a change.

lornajane (approved):

PR #1585, opened 2024-06-10
docs: fix redocly cli quickstart

### LINKED ISSUE #1517: Review and update the quickstart guide
The [Redocly CLI Quickstart guide](https://redocly.com/docs/cli/quickstart) has not been updated in quite some time.

Could we please:
 - [ ] check the existing instructions still work
 - [ ] improve the copy on this page, including something about why each task is useful
 - [ ] add some additional common first tasks, such as defining a ruleset or bundling a file

techstacksavvy:
## What/Why/How?
Validated existing instructions, refined page copy with explanations of the utility of each task, and added a common initial task

## Reference
Resolves #1517 

## Check yourself

- [x] check the existing instructions still work
- [x] improve the copy on this page, including something about why each task is useful
- [x] add some additional common first tasks, such as defining a ruleset or bundling a file


lornajane:
Thanks @techstacksavvy, this is an improvement already!

The first two sections are good, I just removed the part about cloning the starter repo since people probably don't want to do that if they already have an API description to use - it's there in the admonition as an optional resource.

For the final section, could you please revisit the example? We already have a rule that requires a summary (https://redocly.com/docs/cli/rules/operation-summary/) so users shouldn't be creating a custom rule for that. A better example might be "craft a custom ruleset" - get the users to add a `redocly.yaml` config file using the minimal ruleset and adding a couple of other built-in rules on top.

I would also like to have a "more resources" section at the end of the quickstart to send people to more resources in case they are interested. If you can add this as well, that would be excellent. (I added a commit to your branch so make sure you update before doing more editing, and let us know if you have questions!)

lornajane (approved):
Thanks @techstacksavvy, I just tidied up a few ends so we can publish your upgrades.

PR #1595, opened 2024-06-19
chore: replace ➡️ with 🔀 to align the words

arkbriar:
## What/Why/How?

The ➡️ should be the same width as other emojis normally though it shows differently in GitHub's code viewer (`➡️`). Currently, the additional space makes the `stats` subcommand print such outputs, which is weird

image

The PR fixes it by removing the space.

## Check yourself

## Security

### DIFF
--- a/.changeset/shy-cherries-tie.md
+++ b/.changeset/shy-cherries-tie.md
@@ -1,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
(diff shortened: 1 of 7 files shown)


tatomyr:
Hi @arkbriar! Could you replace the ➡️ emoji with 🔀? I think it should eliminate the original issue with improper alignment: 
image

lornajane (approved):
Thanks @arkbriar ! This looks much better in all output formats

PR #1601, opened 2024-06-26
feat: lint arazzo test descriptions

DmitryAnansky:
## What/Why/How?

Add Arazzo validation to the `lint` cli command.
Now it is possible to lint Arazzo description file by running `redocly lint museum.yaml`

## Check yourself

## Security


packages/core/src/lint.ts
```diff
@@ -105,7 +105,10 @@ export async function lintDocument(opts: {
     resolvedRefMap,
     ctx,
   });
-  return ctx.problems.map((problem) => config.addProblemToIgnore(problem));
+  return {
```
DmitryAnansky, line 108: This change is needed in order to detect specVersion for proper validation message replacing `API` with `Test` as message subject.
But will affect dependent packages:
`const { problems: results } = await lintDocument({`
 instead of
`const result = await lintDocument({`


@RomanHotsiy / @tatomyr will it be ok to apply this ?
Alternative solution can be read document one more time to detect specVersion, or change the final message not to have specific subject information, like `API`

packages/core/src/types/arazzo.ts
```diff
@@ -0,0 +1,345 @@
...
+          additionalProperties: onFailureObject,
+        },
+      },
+    },
+  },
+  additionalProperties: false,
+  required: ['arazzo', 'info', 'sourceDescriptions', 'workflows'],
+} as const;
+
+export const ArazzoTypes: Record<string, NodeType> = getNodeTypesFromJSONSchema(
```
DmitryAnansky, line 342: @tatomyr / @RomanHotsiy There is a question where to store the Arazzo schema.
This is the second place I duplicate it, first was Spot-exp.
I would prefer to keep it centralised.
 Best candidates are Spot-ext (looks weird to import smth from it to cli), another one is maybe @redocly/config ?

We could import schema from @redocly/config in CLI and Spot-exp and have it as a single source of truce for Arazzo schema.
Please let me know what you think about it.

RomanHotsiy: why it can't be in this repo?

packages/cli/src/utils/miscellaneous.ts
```diff
@@ -314,11 +314,17 @@ export function handleError(e: Error, ref: string) {
...
+export function printLintTotals(
+  totals: Totals,
+  definitionsCount: number,
+  documentSpecVersion?: string
+) {
   const ignored = totals.ignored
     ? yellow(`${totals.ignored} ${pluralize('problem is', totals.ignored)} explicitly ignored.\n\n`)
     : '';
 
+  const lintSubject = documentSpecVersion === 'arazzo' ? 'Test' : 'API';
```
RomanHotsiy, line 326: Arrazo is not a "Test". It can be used as a test.

Let's change the message to:

> Woohoo! Your description documents are valid.

@lornajane do you have any better suggestions?

packages/core/src/types/arazzo.ts
DmitryAnansky: We will have duplicated schema. So there will be Arazzo JSON schema copies in Redocly/cli and Spot.

docs/commands/lint.md
```diff
@@ -24,7 +24,7 @@ redocly lint --version
 
 | Option                 | Type     | Description                                                                                                                                                                          |
 | ---------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| apis                   | [string] | Array of API description filenames that need to be linted. See [the Apis section](#apis) for more options.                                                                           |
+| apis                   | [string] | Array of API or Arazzo description filenames that need to be linted. See [the Apis section](#apis) for more options.                                                                 |
```
DmitryAnansky, line 27: @lornajane maybe we might need more renames here in the docs as we can pass Arazzo description file here.

JLekawa (approved):

packages/cli/src/utils/miscellaneous.ts
lornajane: I think we could literally say the Arazzo description is valid if we're specifically testing for an arazzo file here. I think these are a little different to more interchangeable OpenAPI/AsyncAPI formats.

What do we output for config linting?

DmitryAnansky: It used to be this message `Woohoo! Your API description is valid. 🎉`.
But if we want to provide custom and different message for API and Arazzo - we will introduce a breaking change or additional redundancy to read document two times. The reason for this is to pass SpecVersion.
So we ended up wit more generic version for validation`Woohoo! Your description document is valid. 🎉`

@lornajane , please check previous discussion https://github.com/Redocly/redocly-cli/pull/1601#discussion_r1654752154

docs/commands/lint.md
DmitryAnansky: More mentions updated, thx.

packages/core/src/oas-types.ts
```diff
@@ -100,6 +103,10 @@ export function detectSpec(root: any): SpecVersion {
     throw new Error(`Unsupported AsyncAPI version: ${root.asyncapi}`);
   }
 
+  if (root.arazzo) {
+    return SpecVersion.Arazzo;
```
tatomyr, line 107: Could we also check the version just in case? 
Like this: `root.asyncapi.startsWith('2.'))`

tatomyr (commented):
Left a few comments. Otherwise looks good.

museum.yaml
```diff
@@ -0,0 +1,37 @@
+arazzo: 1.0.0
```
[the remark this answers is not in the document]
DmitryAnansky: Removed

packages/core/src/config/__tests__/config.test.ts
```diff
@@ -249,6 +256,10 @@ describe('getMergedConfig', () => {
           },
           "recommendedFallback": false,
           "rules": {
+            "arazzo": {
+              "no-empty-servers": "error",
```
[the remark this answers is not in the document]
tatomyr: > Can they use plugins?

In theory they should be able to use plugins. @DmitryAnansky can you check that?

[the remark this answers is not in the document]
DmitryAnansky: The line you highlighted is based on the mock rules in the test

packages/cli/src/utils/miscellaneous.ts
lornajane: I recommend we keep the existing `Woohoo! Your API description is valid. 🎉` feedback for existing formats (OpenAPI, AsyncAPI). I don't mind if we do the same or something different for Arazzo, but let's not disrupt the existing functionality.

RomanHotsiy: That's the problem. If we want to keep the existing message for existing format but have a different message for Arazzo we will have to introduce a breaking change in the API or re-design some internals.
I was hoping to avoid it so my suggestion is to have some generic message.

packages/core/src/types/arazzo.ts
```diff
@@ -0,0 +1,345 @@
...
+  },
+  additionalProperties: false,
+  required: ['payload'],
+} as const;
+const criteriaObject = {
+  type: 'object',
+  properties: {
+    condition: { type: 'string' },
+    context: { type: 'string' },
+    type: { type: 'string', enum: ['regex', 'JSONPath'] },
```
lornajane, line 168: I'm seeing examples that have this set to `jsonpath`, do you think we should use a case-insensitive match here to accept those?

packages/core/src/config/config.ts
```diff
@@ -178,6 +181,8 @@ export class StyleguideConfig {
             if (!plugin.typeExtension.async2) continue;
             extendedTypes = plugin.typeExtension.async2(extendedTypes, version);
             break;
+          case SpecVersion.Arazzo:
+            break;
```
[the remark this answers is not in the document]
DmitryAnansky: Added

DmitryAnansky:
> It's a small thing, but we should also update the CLI help output to add arazzo to the lint command description.
> 
> Linting some of the Arazzo examples from the main project, I get some errors "The field `in` must be present on this level." but the entry looks like this:
> 
> ```
> workflowId: place-order
> parameters:
>   - name: pet_id
>     value: $steps.find-pet.outputs.my_pet_id
> ```
> 
> Looking at the [spec document for parameters](https://github.com/OAI/Arazzo-Specification/blob/main/versions/1.0.0.md#parameter-object), I think the "in" isn't required when it's a reference to something that's already in the scope of an Arazzo workflow.

Thanks Lorna,
We have a task for this issue => https://github.com/orgs/Redocly/projects/51/views/1?pane=issue&itemId=65948722

packages/core/src/types/arazzo.ts
DmitryAnansky: Its from Arazzo description, so I guess we should keep it as it is

```
    - context: $response.body
      condition: $[?count(@.pets) > 0]
      type: JSONPath
 ```

Comment (lornajane) [reply packages/core/src/types/arazzo.ts:168]: I will fix their example instead ;)

File: packages/core/src/types/arazzo.ts [hunk 10]
```diff
@@ -0,0 +1,345 @@
...
+} as const;
+const inherit = {
+  type: 'string',
+  enum: ['auto', 'none'],
+} as const;
+const onSuccessObject = {
+  type: 'object',
+  properties: {
+    name: { type: 'string' },
+    type: { type: 'string', enum: ['goto'] },
```
lornajane, line 185: I think `end` is also supported here, but maybe you can double check? I'm reading it here https://github.com/OAI/Arazzo-Specification/blob/main/versions/1.0.0.md#success-action-object

DmitryAnansky: We don't support it in our implementation

DmitryAnansky: @RomanHotsiy / @tatomyr this actually brings another topic to discuss - as here we validate our Arazzo implementation, including some extensions and things we don't support.
Should we lint with pure Arazzo schema ? 
In this case we can keep two Arazzo schemas in cli and in Spot - as for sure they will be different.

lornajane: Could we support `end` when linting when we don't have support for it in Spot? If it's a lot of work then we can publish a list of limitations, but I think if we want to say that we support Arazzo, we should try to get as much of it as we can. (sorry, I realise this probably makes more work!).

DmitryAnansky: As discussed offline =>

-  I will update the schema to have pure Arazzo spec.
- Spot difference in spec should be handled with additional rules in the future (out of this PR scope)

lornajane:
We don't have support for references in step parameters (or I'm doing something wrong! Testing [the pet coupons example](https://github.com/OAI/Arazzo-Specification/blob/main/examples/1.0.0/pet-coupons.arazzo.yaml#L73) ), will we list that as not supported, or should we add it? https://github.com/OAI/Arazzo-Specification/blob/main/versions/1.0.0.md#step-object

packages/cli/src/utils/miscellaneous.ts
tatomyr: I think we can keep the existing message for the sake of simplicity. Any Arazzo document could also be considered an API description to some extent. Later we can try customising the message if possible. 
WDYT?

RomanHotsiy: Yes! Good idea 😄

DmitryAnansky: Ok, I will roll back message and snapshots to prev state

packages/core/src/types/arazzo.ts
DmitryAnansky: I just discovered it was change at some point in time - so you are right. Updating schemas....

packages/core/src/types/arazzo.ts
DmitryAnansky: Arazzo schama is updated.

JSONSchema have some place for improvements - but good enough to digest Basic Arazzo examples, they are passing as a valid one now.

__tests__/lint/arazzo-type-extensions-with-plugin/snapshot.js
```diff
@@ -0,0 +1,8 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`E2E lint arazzo-type-extensions-with-plugin 1`] = `
+
+No APIs were provided.
```
tatomyr, line 5: Why there's no API? Is it expected?

packages/core/src/config/types.ts
```diff
@@ -51,23 +54,26 @@ export type StyleguideRawConfig<T = undefined> = {
   doNotResolveExamples?: boolean;
   recommendedFallback?: boolean;
 
-  rules?: RuleMap<BuiltInCommonRuleId | BuiltInCommonOASRuleId, RuleConfig, T>;
+  rules?: RuleMap<BuiltInCommonOASRuleId, RuleConfig, T>;
```
tatomyr, line 57: I believe the `rules` section should contain the AsyncAPI rules as well:

```suggestion
  rules?: RuleMap<BuiltInCommonOASRuleId | BuiltInAsync2RuleId, RuleConfig, T>;
```

packages/core/src/decorators/arazzo/index.ts
```diff
@@ -0,0 +1 @@
+export const decorators = {};
```
tatomyr, line 1: Where do you import it?

packages/core/src/rules/arazzo/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Spec } from '../common/spec';
+import type { ArazzoRuleSet } from '../../oas-types';
+import { Assertions } from '../common/assertions';
+
+export const rules: ArazzoRuleSet<'built-in'> = {
+  spec: Spec as ArazzoRule,
+  assertions: Assertions,
+};
+
+export const preprocessors = {};
```
tatomyr, line 11: Do you use it?

packages/core/src/types/arazzo.ts
```diff
@@ -0,0 +1,386 @@
...
+    },
+  },
+  additionalProperties: false,
+  required: ['workflowId', 'steps'],
+} as const;
+const workflows = {
+  type: 'array',
+  items: workflow,
+} as const;
+export const workflowSchema = {
```
tatomyr, line 336: Wouldn't it be better to rename it for consistency?

```suggestion
export const arazzoSchema = {
```

docs/commands/lint.md
```diff
@@ -22,19 +22,19 @@ redocly lint --version
...
+| Option                 | Type     | Description                                                                                                                                                                          |
+| ---------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| apis                   | [string] | Array of API or Arazzo description filenames that need to be linted. See [the API section](#specify-api) for more options.                                                           |
+| --config               | string   | Specify path to the [configuration file](#use-custom-configuration-file).                                                                                                            |
+| --extends              | [string] | [Extend a specific configuration](#extend-configuration) (defaults or config file settings).                                                                                         |
```
tatomyr, line 29: Don't we want to say something about the built-in ruelsets here?

tatomyr (commented):
Almost there. A few more comments.

packages/core/src/decorators/arazzo/index.ts
DmitryAnansky: No where - just template for the future

packages/core/src/rules/arazzo/index.ts
DmitryAnansky: No where - just template for the future

tatomyr: Then I don't think it makes sense to create that now (especially to create preprocessors in the rules folder -- I doubt anyone will be looking there).

DmitryAnansky: I keep it in sync with other folders - it just mean that there are no preprocessors | decorators for now - but at least we have a placeholder ready for this.
If you strictly against, I can remove it, have those settings to keep in sync with other types like oas3, async .....

packages/core/src/config/types.ts
DmitryAnansky: Why do you think so?
This change will affect minimal, all, recomended rules and change the previous rule sets.

DmitryAnansky: Discussed offline - out of this scope.

tatomyr: Oh, indeed. I'll create an issue to fix that separately.

docs/commands/lint.md
tatomyr:
```suggestion
| --extends              | [string] | [Extend a specific configuration](#extend-configuration) (defaults or config file settings).  **Built-in rulesets:** `minimal`, `recommended`, `recommended-strict`, `all`. Default value is `recommended`.                                                                                       |
```

__tests__/lint/arazzo-type-extensions-with-plugin/snapshot.js
DmitryAnansky: Test updated

__tests__/commands.test.ts
```diff
@@ -22,20 +22,61 @@ callSerializer();
 
 describe('E2E', () => {
   describe('lint', () => {
+    const excludeFolders = ['arazzo-type-extensions-with-plugin'];
```
tatomyr, line 25: Could you enlist all excluded folders (arazzo-valid-test-description, arazzo-not-valid-test-description)?

DmitryAnansky: They are skipped due to missing redocly.yaml, but I can add it here as well

tatomyr (approved):

lornajane (approved):
Very nice, works well with the samples in the main project (and I have some tiny fixes to send them that I found using this tool, so it's already proven its value!. Nice work!

PR #1608, opened 2024-07-03
Updated asyncapi v2 typings to conform to JSON Schema draft-07

jeremyfiel:
Refactors the v2 typings ONLY. 

Per the specification, AsyncAPI v2 supports a superset of JSON Schema draft-07. 

related https://github.com/Redocly/redoc/issues/1219

## What/Why/How?
The types defined in asyncapi.ts were copied from oas3_1 types but AsyncAPI does not support JSON Schema 2020-12, thus, many of the typings were invalid for proper validation and linting. 

## Reference
https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#schema-object

the list defined in their docs does not represent the terminology they have used. The list is copied from their earlier version specification documents and only minor updates were made.  They indicate any JSON Schema core or validation keywords are supported from draft-07, so those updates were integrated here. 

## Check yourself

## Security


tatomyr:
@jeremyfiel could you re-arrange the properties back just to see what's been changed?

packages/core/src/types/asyncapi.ts
```diff
@@ -417,23 +389,24 @@ const Schema: NodeType = {
...
     deprecated: { type: 'boolean' },
     const: null,
     $comment: { type: 'string' },
-    dependencies: { type: 'object' }, // TODO
+    additionalItems: (value: any) => {
+      return typeof value === 'boolean' ? { type: 'boolean' } : 'Schema';
+    },
+    dependencies: (value: any) => {
+      return Array.isArray(value) ? 'PropertyDependency' : 'Schema';
+    },
```
tatomyr, line 402: The `dependencies` property couldn't be an array. It must be an object with additionalProperties which could be either an array or a schema: 
https://json-schema.org/draft-07/json-schema-validation#rfc.section.6.5.7

Could you update that to something like the following?

```suggestion
    dependencies: {
      properties: {}
      additionalProperties: value => Array.isArray(value) ? { type: array, ... } : 'Schema'
    },
```

JLekawa (approved):

packages/core/src/types/asyncapi.ts
```diff
@@ -417,23 +389,22 @@ const Schema: NodeType = {
...
     examples: { type: 'array' },
     example: { isExample: true },
     deprecated: { type: 'boolean' },
     const: null,
     $comment: { type: 'string' },
-    dependencies: { type: 'object' }, // TODO
+    additionalItems: (value: any) => {
+      return typeof value === 'boolean' ? { type: 'boolean' } : 'Schema';
+    },
+    dependencies: 'PropertyDependency',
```
tatomyr, line 400: Don't you mind renaming it to just `Dependencies`?

```suggestion
    dependencies: 'Dependencies',
```

AlexVarchuk (approved):

tatomyr (approved):
Thank you for the contribution @jeremyfiel!

PR #1612, opened 2024-07-05
chore: change link checker config to internal links only

lornajane:
## What/Why/How?

We have a link checker, but if a 3rd party site is down then we see a build failure when in fact it's not a fault on our side. This pull request adjusts the link checking to be local-only in the CI pipeline that is used on every pull request.

This change brings the checks in line with the other docs sites.

## Check yourself

## Security

### DIFF
--- a/.mlc.toml
+++ b/.mlc.toml
@@ -5,3 +5,5 @@ ignore-links=["../../*", "/docs/*",
 ]
 # Path to the root folder used to resolve all relative paths
 root-dir="./docs"
+offline = true
+


.mlc.toml
```diff
@@ -5,3 +5,5 @@ ignore-links=["../../*", "/docs/*",
 ]
 # Path to the root folder used to resolve all relative paths
 root-dir="./docs"
+offline = true
```
tatomyr, line 8: But checking external links is also valuable. It can warn us if a resource we refer to is no longer present. Is it possible to retry on broken links instead of completely cutting them off?

lornajane:
What I'd like to do is add a separate CI task that runs the check every week and opens an issue if any link, including external links, is broken. The majority of external link check failures are other people's problems, not ours, and having that block every pull request isn't helping.

PR #1617, opened 2024-07-09
chore: re-enable the Reference Docs options in the theme.openapi config

tatomyr:
## What/Why/How?

- [x] Updated the config version to allow the Reference Docs options back into theme.openapi (waiting [this](https://github.com/Redocly/redocly/pull/9451) to be merged before + Redocly config released)
- [x] Added the notification about the unsupported theme.openapi options  to the `check-config` command
- [x] Refactored imports and cleaned up the codebase
- [x] Updated the `@redocly/config` package to v0.7.0

## Reference

https://github.com/orgs/Redocly/projects/12/views/5?pane=issue&itemId=70133042

## Screenshots (optional)

 
Screenshot 2024-07-17 at 16 20 18

## Check yourself

## Security


packages/cli/src/utils/miscellaneous.ts
```diff
@@ -662,3 +659,23 @@ export function checkForDeprecatedOptions<T>(argv: T, deprecatedOptions: Array<k
...
+    const intersection = propertiesSet.filter((prop) => deprecatedSet.includes(prop));
+    if (intersection.length > 0) {
+      process.stderr.write(
+        yellow(
+          `\n${pluralize('Property', intersection.length)} ${gray(
+            intersection.map((prop) => `'${prop}'`).join(', ')
+          )} ${pluralize(
+            'is',
+            intersection.length
+          )} specific to Redoc v1 or Reference Docs and not supported by Redoc v2.\n\n`
```
tatomyr, line 676: This is the message that users will see if they use the old Reference Docs options in theme.openapi.

packages/cli/src/commands/lint.ts
```diff
@@ -130,7 +131,9 @@ export function lintConfigCallback(
     return;
   }
 
-  return async ({ document, resolvedRefMap, config }) => {
+  return async ({ document, resolvedRefMap, config, parsed: { theme = {} } }) => {
+    notifyAboutIncompatibleConfigOptions(theme.openapi);
```
tatomyr, line 135: This is the only real change.

packages/cli/src/utils/miscellaneous.ts
lornajane: The current open source version of Redoc is 2.x, and I think that does still support some of those fields. How about:

" ... only used in API Reference Docs and Redoc version 2.x or earlier."

I suggested that with the major divergence (API Reference Docs forked years ago and was sort of a v3 under other branding), we make the realm version of Redoc a v4 release, but I'm not sure I've won my case. Either way, it'll have to be at least one major version up from the current 2.x series so I think this message will work whichever way we go on that.

tatomyr: Oh, indeed. Thanks for the correction. Will push an update.

tatomyr: @lornajane updated

JLekawa (approved):

RomanHotsiy (approved):

.changeset/clever-fans-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Allowed the theme.openapi to accept config options specific to Redoc 2.x and earlier.
```
lornajane, line 6:
```suggestion
Allowed the `theme.openapi` configuration option to accept settings specific to Redoc 2.x and earlier.
```

lornajane (approved):
Minor tweak to the changeset wording.

PR #1621, opened 2024-07-10
fix: arazzo parameter schema update

DmitryAnansky:
## What/Why/How?

Arazzo parameter schema update: added `body` to the allowed parameter `in` value.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/types/arazzo.ts
+++ b/packages/core/src/types/arazzo.ts
@@ -87,7 +87,7 @@ const parameter = {
     {
       type: 'object',
       properties: {
-        in: { type: 'string', enum: ['header', 'query', 'path', 'cookie'] },
+        in: { type: 'string', enum: ['header', 'query', 'path', 'cookie', 'body'] },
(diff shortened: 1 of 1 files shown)


packages/core/src/types/arazzo.ts
```diff
@@ -87,7 +87,7 @@ const parameter = {
     {
       type: 'object',
       properties: {
-        in: { type: 'string', enum: ['header', 'query', 'path', 'cookie'] },
+        in: { type: 'string', enum: ['header', 'query', 'path', 'cookie', 'body'] },
```
tatomyr, line 90: Can we add a Spot-specific rule that will error on this?

tatomyr (approved):

packages/core/src/types/arazzo.ts
DmitryAnansky: In future -yes.
There should be a list of similar rules

PR #1626, opened 2024-07-12
chore: use dequal instead of lodash.isequal

tatomyr:
## Check yourself

## Security


RomanHotsiy (approved):
LGTM. Just add a reference to the source of the dequal function.

JLekawa (approved):

Issue #1645, opened 2024-07-23
[OAS 3 spec] License object `identifier` and `url` keys are mutually-exclusive


bradjones1:
**Describe the bug**

According to [the OAS 3 spec](https://swagger.io/specification/#license-object), `identifier` and `url` are mutually-exclusive in a `license` object. Redocly CLI wrongly asserts that the `url` field is required.

**To Reproduce**

Specify a `name` and `identifier` in a `license` object. Redocly reports

```
License object should contain `url` field.
```

**Expected behavior**

The license object rule actually enforces _either_ `identifier` or `url` (or neither) are present, but not both. It should also assert that `name` is specified, as it is required.

**Logs**

```
License object should contain `url` field.

4 | title: 'Meet Kinksters - JSON API'
5 | version: 'Versioning not (yet) implemented.'
6 | license:
  | ^^^^^^^
7 |   name: 'GNU General Public License v2.0 or later'
8 |   identifier: ' GPL-2.0-or-later'

Warning was generated by the info-license-url rule.
```

**Redocly Version(s)**

`1.18.1`

tatomyr:
Hi @bradjones1, I recommend you switch off the `info-license-url` rule in your **redocly.yaml**. Instead, you can write a [configurable rule](https://redocly.com/docs/cli/rules/configurable-rules) like the one below to ensure that either `url` or `identifier` are present in your `license`:


```yaml
rules:
  info-license-url: off
  rule/info-licence-required-fields:
    subject:
      type: License
    assertions:
      requireAny:
        - url
        - identifier
      mutuallyExclusive:
        - url
        - identifier
```

I hope that helps.

bradjones1:
I certainly could, but I was thinking more of changing the rule that ships with redocly core to reflect the spec. Is there a reason why that change shoudln't be made?

tatomyr:
> Is there a reason why that change shoudln't be made?

It's a good question.
First, this particular rule requires a URL in the `license` field, so I'd imagine it does what it's supposed to do. Both the `url` and `identifier` fields are optional, so the rule isn't about compliance with the spec. It's up to users to configure the setup (including turning the rule off).
Second, I noticed that Redocly CLI lacks support for the `identifier` field in the OAS 3 node tree, and the spec rule will error if the field is present (regardless of the `url` field). This actually violates the spec and should be fixed.

[...]

PR #1653, opened 2024-07-31
feat: add info-license-strict rule

tatomyr:
## What/Why/How?

Added `info-license-strict` rule. Now it can be used instead of the `info-license-url` which is deprecated.

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/1645

## Check yourself

## Security


packages/core/src/rules/common/info-license-strict.ts
```diff
@@ -0,0 +1,23 @@
...
+  let specVersion: string | undefined;
+  return {
+    Root: {
+      enter(root: any) {
+        specVersion = detectSpec(root);
+      },
+      License: {
+        leave(license, ctx) {
+          if (specVersion === 'oas3_1') {
+            validateOneOfDefinedAndNonEmpty(['url', 'license'], license, ctx);
```
lornajane, line 15: I had to adjust like this to make the rule do what I expected.

```suggestion
            validateOneOfDefinedAndNonEmpty(['url', 'identifier'], license, ctx);
```

tatomyr: Thanks! That was what I actually meant 😅

tatomyr: Also fixed tests not failing for that.

packages/core/src/config/recommended.ts
```diff
@@ -4,7 +4,8 @@ const recommended: PluginStyleguideConfig<'built-in'> = {
   rules: {
     'info-contact': 'off',
     'info-license': 'warn',
-    'info-license-url': 'warn',
+    'info-license-url': 'off',
+    'info-license-strict': 'warn',
```
tatomyr, line 8: Turned the old rule off and added the new one as a warn.

.changeset/shy-crews-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added `info-license-strict` rule as a replacement of the `info-license-url`.
```
lornajane, line 6:
```suggestion
Added `info-license-strict` rule as a replacement of the `info-license-url` to support the OpenAPI 3.1 changes to allow identifier or URL license details.
```

packages/core/src/config/recommended.ts
```diff
@@ -90,6 +91,7 @@ const recommended: PluginStyleguideConfig<'built-in'> = {
   async2Rules: {
     spec: 'error',
     'info-contact': 'off',
+    'info-license-strict': 'warn',
```
lornajane, line 94: I'm not sure I would make this change to the recommended ruleset if the old rule wasn't already here (also for AsyncAPI 3)

lornajane (commented):
Minor changelog suggestion and I don't think we should add the rule where it wasn't included before (fair enough in "strict", but "recommended" is our default so surprise behaviour changes may not be ideal).  Also: AsyncAPI does not support an identifier for license, so I think it might be better to leave the old rule in place for that format. Otherwise if someone tries to use OpenAPI format, we will mark it as valid and it won't be.

Otherwise - works perfectly and I'm ready to merge.

packages/core/src/config/recommended.ts
tatomyr: It just happens to cover the same structure requirements as in OAS3/3.1, so I decided to include it to align more closely with our OAS ruleset. I don't see how a recommended rule for the OAS info section could be harmful for AsyncAPI, especially considering that we only recently added it and most likely just forgot to include the rule in AsyncAPI's ruleset. It's still a warning anyway.

tatomyr:
@lornajane 
> AsyncAPI does not support an identifier for license, so I think it might be better to leave the old rule in place for that format. Otherwise if someone tries to use OpenAPI format, we will mark it as valid and it won't be.

No, the rule itself covers the difference in specs (I even have a [test](https://github.com/Redocly/redocly-cli/pull/1653/files#diff-ff654a38a91e7c5152aa3497451a66c97e0e35272be686f067b142366bfe4a12R112) for that). I think we should abandon the old rule in favour of the new one as the latter is more universal.

lornajane (approved):

PR #1656, opened 2024-08-02
chore: remap arazzo schema to our internal type

DmitryAnansky:
## What/Why/How?

- Remap Arazzo JSON schema to our internal type 
- Add Spot rule sample
- Change makeConfig test util

## Check yourself

## Security


DmitryAnansky:
@RomanHotsiy / @tatomyr 
There is an open question if we need to keep `export const arazzoSchema` in cli or move it back to Spot?

Keeping it in cli will give us single place to manage schemas and not switching between projects.
Also there is some ongoing activities on Arazzo side - they are going to have official schema and we might use it as source of truth in future.

DmitryAnansky:
> @RomanHotsiy / @tatomyr There is an open question if we need to keep `export const arazzoSchema` in cli or move it back to Spot?
> 
> Keeping it in cli will give us single place to manage schemas and not switching between projects. Also there is some ongoing activities on Arazzo side - they are going to have official schema and we might use it as the source of truth in future.

Discussed offline - I will move it back to Spot codebase.

packages/core/__tests__/utils.ts
```diff
@@ -69,17 +69,24 @@ export function makeConfigForRuleset(
   });
 }
 
-export async function makeConfig(
-  rules: Record<string, RuleConfig>,
-  decorators?: Record<string, DecoratorConfig>,
-  configPath?: string
-) {
+export async function makeConfig({
```
DmitryAnansky, line 72: @tatomyr I changed this helper function arguments to make it flexible and not pass params in special order and also to handle Arazzo rules case - as it does not merge `rules` with `ArazzoRules` as other specs.

packages/core/src/visitors.ts
```diff
@@ -220,6 +237,23 @@ type Async3FlatVisitor = {
...
+  ExpectSchema?: VisitFunctionOrObject<ExpectSchema>;
+  Replacement?: VisitFunctionOrObject<Replacement>;
+  RequestBody?: VisitFunctionOrObject<RequestBody>;
+  CriteriaObject?: VisitFunctionOrObject<CriteriaObject>;
+  OnSuccessObject?: VisitFunctionOrObject<OnSuccessObject>;
+  OnFailureObject?: VisitFunctionOrObject<OnFailureObject>;
+  Step?: VisitFunctionOrObject<Step>;
+  Steps?: VisitFunctionOrObject<Record<string, Step>>;
+  Workflow?: VisitFunctionOrObject<Workflow>;
+  Workflows?: VisitFunctionOrObject<Record<string, Workflow>>;
```
tatomyr, line 256: Isn't it an array instead?

```suggestion
  Workflows?: VisitFunctionOrObject<Workflow[]>;
```

packages/core/src/types/arazzo.ts
```diff
@@ -1,391 +1,308 @@
...
+  required: ['arazzo', 'info', 'sourceDescriptions', 'workflows'],
+  extensionsPrefix: 'x-',
+};
+const Components: NodeType = {
+  properties: {
+    inputs: 'NamedInputs',
+    parameters: {
+      type: 'object',
+      additionalProperties: 'ParameterObject',
+    },
```
tatomyr, line 23: Let's have separate names for components (and others too):

```suggestion
    parameters: 'NamedParameters',
    },
```

packages/core/src/types/arazzo.ts
```diff
@@ -1,391 +1,308 @@
...
-  type: 'array',
-  items: parameter,
-} as const;
-export const infoObject = {
-  type: 'object',
+  },
+  required: ['reference'],
+  extensionsPrefix: 'x-',
+};
+const ParameterObject: NodeType = {
```
tatomyr, line 96: Why not just this?

```suggestion
const Parameter: NodeType = {
```

packages/core/src/types/arazzo.ts
```diff
@@ -1,391 +1,308 @@
...
-export const criteriaObject = {
-  type: 'object',
+  extensionsPrefix: 'x-',
+};
+const Replacement: NodeType = {
+  properties: {
+    target: { type: 'string' },
+    value: {},
+  },
+  required: ['target', 'replacement'],
```
tatomyr, line 175: What is the `replacement` property? Did you mean `value`?

packages/core/src/types/oas3.ts
```diff
@@ -44,7 +44,7 @@ const TagGroup: NodeType = {
   extensionsPrefix: 'x-',
 };
 
-const ExternalDocs: NodeType = {
+export const ExternalDocs: NodeType = {
```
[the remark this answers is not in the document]
tatomyr: You mean the schema types? I think it makes sense to import some parts from Oas3 into Oas3.1 as it's the successor. You can try declaring something like `Oas3Schema` and extend it into `Oas3_1Schema`, and then reuse the latter for Arazzo.

tatomyr (commented):
Left a few comments.

JLekawa (approved):

packages/core/src/types/oas3.ts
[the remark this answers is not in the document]
DmitryAnansky: Lets handle this in a separate PR

packages/core/src/rules/spot/parameters-no-body-inside-in.ts
```diff
@@ -0,0 +1,17 @@
+import type { ArazzoRule } from '../../visitors';
+import type { UserContext } from '../../walk';
+
+export const ParametersNoBodyInsideIn: ArazzoRule = () => {
+  return {
+    Parameter: {
+      enter(parameter, { report, location }: UserContext) {
+        if (parameter.in === 'body') {
+          report({
+            message: 'Parameters `in` property not allowed to have `body` value.',
```
tatomyr, line 10: Technically it is allowed, but we don't support that. What about this:

```suggestion
            message: 'The `body` value of the `in` property is not supported.',
```

tatomyr (commented):
Do we have any docs on Arazzo linting?

tatomyr (approved):

packages/core/src/rules/spot/parameters-no-body-inside-in.ts
DmitryAnansky: 'The `body` value of the `in` property is not supported by Spot'

Issue #1629, opened 2024-07-15
Translations and eject command wrapper


adamaltman:
**Is your feature request related to a problem? Please describe.**

More info here in a private repo: https://github.com/Redocly/redocly/issues/9612 (discussions in that repo), this ticket is added here for visibility and prioritization.

PR #1668, opened 2024-08-16
feat: add the eject and translate commands

tatomyr:
## What/Why/How?

Added the `translate` and `eject` commands. 

Also, refactored types.

~Open questions: https://github.com/Redocly/redocly/issues/9612#issuecomment-2298159934 .~

~Another question: do we want to use options in camel case like `--configDir`? We mostly use kebab-case for options (except for `--outDir` in the `split` command and some options in the `build-docs` command which is legacy of redoc-cli I believe).~

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/1629

The eject docs PR: https://github.com/Redocly/redocly/pull/10344

## Check yourself

## Security


RomanHotsiy (approved):

JLekawa (approved):

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
       commandWrapper(handlerBuildCommand)(argv as Arguments<BuildDocsArgv>);
     }
   )
+  .command(
+    'translations',
+    'Creates or updates translations.yaml files and fills them with missing built-in translations and translations from the redocly.yaml and sidebars.yaml files.',
+    (yargs) =>
+      yargs.options({
+        contentDir: {
```
lornajane, line 785: Could we use `source-dir` here, to be consistent with other commands such as `preview`

lornajane: Ah, I thought it was the directory where the project itself lived, so that the translations could be added to that structure as appropriate. Now I'm not clear if this argument is input or output! Can we get some clarification?

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+          alias: 'd',
+          type: 'string',
+          description: 'Project content directory.',
+          default: '.',
+        },
+        locale: {
```
lornajane, line 791: If the locale is required, then let's make it a positional argument. This approach isn't completely consistent throughout the CLI but we get complaints about the main other place where we didn't get this correct (the outDir on split)

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+          alias: 'l',
+          type: 'string',
+          array: true,
+          description: 'Locale or a list of locales to generate translations for.',
```
lornajane, line 795:
```suggestion
          description: 'Locale code to generate translations for, or `all` for all current project locales.',
```

packages/cli/src/index.ts
lornajane: I realise this means we can only translate one locale or all of them, but I think people will manage to repeat the command if they need to.

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+  )
+  .command(
+    'eject <type> <name>',
+    'Ejects components.',
+    (yargs) =>
+      yargs
+        .positional('type', { required: true, choices: ['component'] })
+        .positional('name', { type: 'string', required: true })
+        .options({
+          contentDir: {
```
lornajane, line 817: I suggest something like `destination` or `destination-dir` as the name here.

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+        },
+      }),
+    (argv) => {
+      process.env.REDOCLY_CLI_COMMAND = 'translations';
+      commandWrapper(handleTranslations)(argv);
+    }
+  )
+  .command(
+    'eject <type> <name>',
+    'Ejects components.',
```
lornajane, line 811:
```suggestion
    'Helper function to eject project elements for customization.',
```

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+    (yargs) =>
+      yargs
+        .positional('type', { required: true, choices: ['component'] })
+        .positional('name', { type: 'string', required: true })
```
lornajane, line 815: How do users find out what things they can name?

packages/cli/src/index.ts
```diff
@@ -764,6 +777,66 @@ yargs
...
+      process.env.REDOCLY_CLI_COMMAND = 'translations';
+      commandWrapper(handleTranslations)(argv);
+    }
+  )
+  .command(
+    'eject <type> <name>',
+    'Ejects components.',
+    (yargs) =>
+      yargs
+        .positional('type', { required: true, choices: ['component'] })
```
lornajane, line 814: Both these positional arguments need description fields (type and name)

lornajane:
If we are shipping this CLI update without docs (I assume we are, since time is pressing), please open one issue per command for documentation, and link those issues to this pull request. I'll pick them up (we sort of have some old translations docs, but they're not ready to publish as they are)

.changeset/short-panthers-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Added the `eject` and `translations` commands.
```
lornajane, line 5:
```suggestion
Added the `eject` and `translations` commands for use with the new Reunite-hosted product family.
```

lornajane (changes requested):
Added a few comments of things to tidy up before we can publish, trying to make things clear and consistent with existing CLI patterns.

packages/cli/src/index.ts
tatomyr: I'd also vote for adopting kebab-case version, but I'm not sure about the naming itself. 
@RomanHotsiy could you help us to decide on naming?

packages/cli/src/index.ts
tatomyr: I don't see any benefits of making it positional in Redocly CLI while it's an option in Realm. Wouldn't it be better to keep it as much consistent as we can?

lornajane: Redocly CLI should be the only interface to all these commands, so the consistency should be with the rest of this tool.

packages/cli/src/index.ts
lornajane: it should still be kebab-case, whatever we call it! "Content" sounds like the content to translate, but I think this is actually the directory where the translations should live, is that right? Maybe `translations-dir` if that's the case

packages/cli/src/index.ts
```diff
@@ -764,6 +777,76 @@ yargs
...
   )
+  .command(
+    'translations',
+    'Creates or updates translations.yaml files and fills them with missing built-in translations and translations from the redocly.yaml and sidebars.yaml files.',
+    (yargs) =>
+      yargs.options({
+        contentDir: {
+          alias: 'd',
+          type: 'string',
+          description: 'A directory where to put the translations content.',
```
lornajane, line 788:
```suggestion
          description: 'The directory to store the translations content.',
```

packages/cli/src/index.ts
RomanHotsiy: I don't like naming this argument (contentDir) different in different commands.

I think we should name it same for develop/eject/translate etc.

contentDir is a path to the folder with the content of the portal. By default it uses current working dir but someone can change it like:

```
-d src/content
```

It's not widely used.

So the description here is invalid. It's not a folder to eject components into: if you provide `-d src/content` it won't eject into `src/content`, it will eject into `src/content/@theme/components/...`.

I am fine with kebab-case. We can also rename it to something else but I want to keep it consistent for all the commands.

We need to open a ticket to sync argument names with CLI and realm ASAP too.

packages/cli/src/index.ts
```diff
@@ -652,6 +660,11 @@ yargs
           description: 'Project directory.',
           default: '.',
         },
+        'lint-config': {
```
RomanHotsiy, line 663: I noticed that `-p` is not working for this command, can we add it?

packages/cli/src/index.ts
lornajane: For `redocly preview` (what was the `develop` command in portal/realm package), we use `--source-dir` or `-d`. That command is already live, in case that influences anyone.

tatomyr: Created the issue: https://github.com/Redocly/redocly/issues/10374
@RomanHotsiy @lornajane please help with finalising the names.

packages/cli/src/index.ts
lornajane: What would you expect `-p` to do here? We have both `--plan` and `--port`, so we didn't add shorteners. Or were you thinking of another option that should be controlled with `-p` ?

packages/cli/src/index.ts
BohdanPetryshyn: @tatomyr, in the issue you created, you suggest renaming the parameter to `content-dir`. As @lornajane pointed out, we already use the name `--source-dir` in `redocly preview` and I agree that all the commands should use the same name for this property.

As for renaming the properties in Realm, we will do this soon but we won't be able to do it today. Let's use the current Realm interface for now.

packages/cli/src/index.ts
RomanHotsiy: `-p` is a common alias for `--port`. This is how it works with realm CLI.

packages/cli/src/index.ts
RomanHotsiy: > For redocly preview (what was the develop command in portal/realm package), we use --source-dir or -d. That command is already live, in case that influences anyone.

So let's use `--source-dir` everywhere.
We can implement the change for realm in the next PR tomorrow. I'm pretty sure very few people if any use --source-dir at the moment. We can change it easily now and we won't be able to do it soon.

@BohdanPetryshyn how about renaming `develop` to `preview` in realm too? (we can support two until 1.0)

RomanHotsiy: `preview` vs `develop` is not as easy to change though as, in contrast to `--source-dir`, it's used by everyone.
I would suggest adding `preview` to the realm but keep using `develop` in CLI until the 1.0 of `realm` family.

packages/cli/src/index.ts
tatomyr: Added the alias.

packages/cli/src/index.ts
BohdanPetryshyn: > We can change it [Realm] easily now and we won't be able to do it soon.

Our latest plan was to completely wrap all the realm/revel/reef/redoc CLI interactions with Redocly CLI and make the interface internal implementation detail. 

We already have an issue to remove any mentions of realm CLI from docs: https://github.com/Redocly/redocly/issues/10068. We should also ask the existing customers to use `redocly preview` in their scripts in package.json

@RomanHotsiy, what do you think? I'm trying to decrease the priority of this change in Realm as we have limited capacity at the moment.

RomanHotsiy: I understand that this is an implementation detail but why keep it unsynchronized if we can change it now? We won't be able to change it easily in the future.

I can pick it up if you are busy. Let me know.

tatomyr: Prepared PR for that: https://github.com/Redocly/redocly/pull/10415.

docs/commands/preview.md
```diff
@@ -14,19 +14,19 @@ This command is for our pre-release products, currently open for early access to
 redocly preview
 redocly preview --product=revel
 redocly preview --product=reef --plan=pro
-redocly preview --product=reef --plan=pro --source-dir=./my-docs-project --port=4001
+redocly preview --product=reef --plan=pro --project-dir=./my-docs-project --port=4001
```
lornajane, line 17: Is this a breaking change? Will the old alias still work?

docs/commands/preview.md
```diff
@@ -14,19 +14,19 @@ This command is for our pre-release products, currently open for early access to
...
+| --port, -p        | number  | The port to run the preview server on. The default value is `4000`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
lornajane, line 28: This may not be a helpful change for a command that has port, plan and product parameters. Not a blocking request for change, but a general note that we did discuss this at the time that the command was implemented.

packages/cli/src/index.ts
```diff
@@ -764,6 +779,78 @@ yargs
       commandWrapper(handlerBuildCommand)(argv as Arguments<BuildDocsArgv>);
     }
   )
+  .command(
+    'translations <locales...>',
+    'Creates or updates translations.yaml files and fills them with missing built-in translations and translations from the redocly.yaml and sidebars.yaml files.',
+    (yargs) =>
+      yargs
+        .positional('locales', {
```
lornajane, line 787: This should be singular, the description is a singular code and that is the expected behaviour. We already had to amend some other command where we tried to allow multiple values to a positional argument and it caused confusion!

docs/commands/preview.md
tatomyr: Added `--source-dir` as an alias.

RomanHotsiy: Add a comment to remove it in a few month then.

docs/commands/preview.md
RomanHotsiy: I don't see the problem with that at all.

Here is an example of a `psql` (Postgress client) CLI docs: 
image

They have multiple args that start with letter `p` but still use `-p` as `--port`.

I also see this very common to in linux CLI tools, e.g. `curl -i` where -i is an alias for `--show-headers` while curl has many other arguments starting with letter `i`: `--include`, `--insecure`, `--interface`.

.changeset/short-panthers-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Added the `eject` and `translations` commands for use with the new Reunite-hosted product family.
```
adamaltman, line 5: I don't want to be annoying in the final hour. I think we should rename the `translations` command to `translate`. Most of our commands are in verb format:

- lint
- bundle
- eject
- preview

Why not `translate`?

RomanHotsiy: good idea!

packages/cli/src/commands/translations.ts
```diff
@@ -0,0 +1,19 @@
...
+  locale: string;
+  'project-dir'?: string;
+} & VerifyConfigOptions;
+
+export const handleTranslations = async ({ argv }: CommandArgs<TranslationsOptions>) => {
+  process.stdout.write(`\nLaunching translations using NPX.\n\n`);
+  const npxExecutableName = process.platform === 'win32' ? 'npx.cmd' : 'npx';
+  spawn(
+    npxExecutableName,
+    ['-y', '@redocly/realm', 'translations', argv.locale, `--project-dir=${argv['project-dir']}`],
```
adamaltman, line 16: I would also recommend we rename it in the original source to `translate` too.

.changeset/short-panthers-smoke.md
tatomyr: I thought the reason it originally wasn't `translate` is that the command doesn't translate, just generate translation stubs.

tatomyr: Here's the corresponding PR for the original source: https://github.com/Redocly/redocly/pull/10425

RomanHotsiy (approved):

PR #1677, opened 2024-08-22
feat: report endpoint sunset

roman-sainchuk:
## What/Why/How?
Warn user about planned endpoint sunset.

Currently there is no way to let user know about breaking changes in the endpoints or endpoint deprecation used for specific command.

Lookup for `Sunset` header in responses received from Reunite to notify users if there are resources that is going to deprecate or is already deprecated (but is still functioning).   

## Screenshots (optional)

*Endpoint is about to be deprecated*

365737829-3ab90010-0739-4d3b-8ce2-c0a0505b3be9

*Endpoint is already deprecated*

image

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/cms/api/api-client.ts
+++ b/packages/cli/src/cms/api/api-client.ts
@@ -335,16 +335,15 @@ export class ReuniteApi {
     if (sunsetWarnings.length) {
       const [{ isSunsetExpired, sunsetDate }] = sunsetWarnings.sort(
         (a: SunsetWarning, b: SunsetWarning) => {
-        // First, prioritize by expiration status
-        if (a.isSunsetExpired !== b.isSunsetExpired) {
-          return a.isSunsetExpired ? -1 : 1;
+          // First, prioritize by expiration status
+          if (a.isSunsetExpired !== b.isSunsetExpired) {
+            return a.isSunsetExpired ? -1 : 1;
+          }
+
+          // If both are either expired or not, sort by sunset date
+          return a.sunsetDate > b.sunsetDate ? 1 : -1;
         }
-
-        // If both are either expired or not, sort by sunset date
-        return a.sunsetDate > b.sunsetDate ? 1 : -1;
-      }
-    );
-
+      );
 
       const updateVersionMessage = `Update to the latest version by running "npm install @redocly/cli@latest".`;
 
--- a/packages/cli/src/cms/api/api-client.ts
+++ b/packages/cli/src/cms/api/api-client.ts
@@ -51,29 +51,26 @@ class ReuniteApiClient implements BaseApiClient {
   private collectSunsetWarning(response: Response) {
     const sunsetTime = this.getSunsetDate(response);
 
-    if (sunsetTime) {
-      const sunsetDate = new Date(sunsetTime);
-      let sunsetWarning: SunsetWarning;
-
-      if (sunsetTime > Date.now()) {
-        sunsetWarning = {
+    if (!sunsetTime) return
+    
+    const sunsetDate = new Date(sunsetTime);
+    if (sunsetTime > Date.now()) {
+        this.sunsetWarnings.push({
           sunsetDate,
           isSunsetExpired: false,
           warning: `Endpoint ${
             response.url
           } is marked for removal after ${sunsetDate.toISOString()}`,
-        };
+        });
       } else {
-        sunsetWarning = {
+        this.sunsetWarnings.push({
           sunsetDate,
           isSunsetExpired: true,
           warning: `Endpoint ${
             response.url
           } support expired on ${sunsetDate.toISOString()} and will be removed.`,
         };
-      }
-
-      this.sunsetWarnings.push(sunsetWarning);
+      })
     }
   }
 
--- a/packages/cli/src/cms/api/api-client.ts
+++ b/packages/cli/src/cms/api/api-client.ts
@@ -84,19 +84,8 @@ class ReuniteApiClient implements BaseApiClient {
       return;
     }
 
-    let sunsetDate = headers.get('sunset');
-
-    if (sunsetDate) {
-      return Date.parse(sunsetDate);
-    }
-
-    sunsetDate = headers.get('Sunset');
-
-    if (sunsetDate) {
-      return Date.parse(sunsetDate);
-    }
-
-    return;
+    const sunsetDate = headers.get('sunset') || headers.get('Sunset');
+    return sunsetDate && Date.parse(sunsetDate)
   }
 }
 
--- a/packages/cli/src/cms/api/api-client.ts
+++ b/packages/cli/src/cms/api/api-client.ts
@@ -61,7 +61,7 @@ class ReuniteApiClient implements BaseApiClient {
             isSunsetExpired: false,
             warning: `Endpoint ${
               response.url
-            } is deprecated for removal on ${sunsetDate.toISOString()}`,
+            } is marked for removal after ${sunsetDate.toISOString()}`,
           };
         } else {
           sunsetWarning = {
@@ -69,7 +69,7 @@ class ReuniteApiClient implements BaseApiClient {
             isSunsetExpired: true,
             warning: `Endpoint ${
               response.url
-            } was deprecated for removal on ${sunsetDate.toISOString()} and could be removed AT ANY TIME`,
+            } support expired on ${sunsetDate.toISOString()} and will be removed.`,
           };
         }
 
@@ -361,16 +361,15 @@ export class ReuniteApi {
         }
       });
 
-      const sunsetMessagePrevix = `Endpoint required for proper work of "${this.command}" command is`;
       const updateVersionMessage = `Please update the version of Redocly CLI.`;
 
       if (isSunsetExpired) {
         process.stdout.write(
-          red(`${sunsetMessagePrevix} DEPRECATED on ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`)
+          red(`Support for "${this.command}" command is deprecated after ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`)
         );
       } else {
         process.stdout.write(
-          yellow(`${sunsetMessagePrevix} going to deprecate on ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`)
+          yellow(`Support for "${this.command}" command is marked for deprecation after ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`)
         );
       }
     }
--- a/packages/cli/src/cms/api/api-client.ts
+++ b/packages/cli/src/cms/api/api-client.ts
@@ -1,3 +1,4 @@
+import { yellow, red } from 'colorette';
 import * as FormData from 'form-data';
 import fetchWithTimeout, {
   type FetchWithTimeoutOptions,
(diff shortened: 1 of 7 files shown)


packages/cli/src/cms/api/middlewares/sunset.middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+
+      if (sunsetTime) {
+        const sunsetDate = new Date(sunsetTime);
+        let sunsetWarning: SunsetWarning;
+
+        if (sunsetTime > Date.now()) {
+          sunsetWarning = {
+            sunsetDate,
+            isSunsetExpired: false,
+            warning: `Endpoint ${
```
lornajane, line 45: Where does this content go to? We usually don't use "endpoint" in customer facing language, or expose the underlying URL. If you can put together some examples of the expected output, I'm happy to help with re-wording.

packages/cli/src/cms/api/middlewares/sunset.middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+              response.url
+            } is deprecated for removal on ${sunsetDate.toISOString()}`,
```
lornajane, line 47:
```suggestion
            } is marked for removal after ${sunsetDate.toISOString()}`,
```

packages/cli/src/cms/api/middlewares/sunset.middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+          };
+        } else {
+          sunsetWarning = {
+            sunsetDate,
+            isSunsetExpired: true,
+            warning: `Endpoint ${
+              response.url
+            } was deprecated for removal on ${sunsetDate.toISOString()} and could be removed AT ANY TIME`,
```
lornajane, line 55: Please, never use upper case in user-facing content unless there is danger to life (or data).

```suggestion
            } support expired on ${sunsetDate.toISOString()} and will be removed.`,
```

packages/cli/src/cms/commands/utils.ts
```diff
@@ -64,3 +67,62 @@ export function handleReuniteError(
...
+    });
+  }
+
+  return {
+    middleware,
+    report: () => {
+      if (warningBuffer.length) {
+        const [{ isSunsetExpired, sunsetDate }] = sortWarningsBySunsetDate(warningBuffer);
+
+        const sunsetMessagePrevix = `Endpoint required for proper work of "${command}" command is`;
```
lornajane, line 113: Check variable spelling: probably `... Prefix`

This sentence doesn't make sense (I realise it's not a whole sentence). Something like:

```suggestion
        const sunsetMessagePrevix = `Support for "${command}" command is`;
```

packages/cli/src/cms/commands/utils.ts
```diff
@@ -64,3 +67,62 @@ export function handleReuniteError(
...
+        const updateVersionMessage = `Please update the version of Redocly CLI`;
+
+        if (isSunsetExpired) {
+          logger.warn(
+            `${sunsetMessagePrevix} DEPRECATED on ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`
```
lornajane, line 118:
```suggestion
            `${sunsetMessagePrevix} marked for deprecation after ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`
```

packages/cli/src/cms/commands/utils.ts
```diff
@@ -64,3 +67,62 @@ export function handleReuniteError(
...
+          );
+        } else {
+          logger.warn(
+            `${sunsetMessagePrevix} going to deprecate on ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`
```
lornajane, line 122:
```suggestion
            `${sunsetMessagePrevix} deprecated after ${sunsetDate.toLocaleString()}. ${updateVersionMessage}\n\n`
```

lornajane (commented):
I'm not sure which ticket this relates to or what problem this solves, but I added some wording recommendations which hopefully make sense in the wider context.

packages/cli/src/cms/api/middlewares/sunset.middleware.ts
roman-sainchuk: This content will be shown to user at the end of `push` or `push-status` command logs. I've updated the description and added screenshots to explain intentions of this functionality.

> We usually don't use "endpoint" in customer facing language, or expose the underlying URL.

Would be better to use "resource" in this case?

lornajane: I'm not sure! What are we actually trying to say?

roman-sainchuk: We're going to notify user that some endpoint used by `push` or `push-status` command will not be/is not compatible with current CLI version due to breaking change or endpoint shutdown.

packages/cli/src/cms/api/api-client.ts
```diff
@@ -13,54 +14,115 @@ import type {
...
+            response.url
+          } is marked for removal after ${sunsetDate.toISOString()}`,
+        };
+      } else {
+        sunsetWarning = {
+          sunsetDate,
+          isSunsetExpired: true,
+          warning: `Endpoint ${
+            response.url
+          } support expired on ${sunsetDate.toISOString()} and will be removed.`,
```
adamaltman, line 72: This doesn't make sense. When the sunset date is reached, the API shouldn't be usable. Also, it's not that the person can't push, it's that they can't push from this version of Redocly CLI. Also the response.url -- what does that have to do with it? Are you migrating? or changing the version and keeping the same URL?

https://datatracker.ietf.org/doc/html/rfc8594

roman-sainchuk: >When the sunset date is reached, the API shouldn't be usable.

It's not always the case, expected sunset date might be expired but the API may still be functioning.
From the `Sunset` [header docs](https://datatracker.ietf.org/doc/html/rfc8594#section-3):

> It is safest to consider timestamps in the past mean the present
   time, meaning that the resource is expected to become unavailable at
   any time.

roman-sainchuk: >Also the response.url -- what does that have to do with it? Are you migrating? or changing the version and keeping the same URL?

`response.url` is not the `url` field from the response body, its a URL we made request to. So we just outline the URL that sent us this header.

The message mentioned above is the default (fallback) message in case we get the `Sunset` header from any endpoint, but is not the message we currently show to user. The actual message is:

https://github.com/Redocly/redocly-cli/blob/45612f32d7f7848b1e3c58dbae0c9b442276efbc/packages/cli/src/cms/api/__tests__/api.client.test.ts#L352

lornajane:
Can we go back a step, I don't think I understand what's happening here - are we adding a user-friendly outcome in the CLI for what happens when the CLI tries to do something that we've stopped supporting in our API? Is it generic, or specifically for a planned change to the push command? What's the change?

In terms of communicating to users, the user thinks of the CLI has having "commands", they don't care about the implementation details such as endpoints. Can we say "this command is unavailable, updating the CLI tool might help " ?

packages/cli/src/cms/api/api-client.ts
RomanHotsiy: I think this message should be updated. The push command is not deprecated. It's just this version of CLI can no longer work with push.

roman-sainchuk: What about 
`"push" command will be incompatible with current version of Redocly CLI after <sunsetDate>. Please run "npm install @redocly/cli@latest" to update.` 

and 

`"push" command is not supported anymore by current version of Redocly CLI. Please run "npm install @redocly/cli@latest" to update.`

?

roman-sainchuk: @RomanHotsiy @adamaltman what do you think? 👆

RomanHotsiy: The first one is okay. The second one is misleading ("not supported"). Should say:

> The "push" command is not compatible with your version of Redocly CLI. Update to the latest version by running "npm install @redocly/cli@latest"

Update the first one to:

> The "push" command will be incompatible with your version of Redocly CLI after <sunsetDate>. Update to the latest version by running "npm install @redocly/cli@latest".

roman-sainchuk: Thanks, updated!

https://github.com/Redocly/redocly-cli/blob/164c5794a27ab2d37d759bd3c9984c40c89f4971/packages/cli/src/cms/api/__tests__/api.client.test.ts#L376

https://github.com/Redocly/redocly-cli/blob/164c5794a27ab2d37d759bd3c9984c40c89f4971/packages/cli/src/cms/api/__tests__/api.client.test.ts#L401

packages/cli/src/cms/api/api-client.ts
```diff
@@ -254,9 +320,67 @@ export class ReuniteApiClient {
...
+          }
+
+          if (a.sunsetDate > b.sunsetDate) {
+            // b will shutdown earlier
+            return 1;
+          } else {
+            return -1;
+          }
+        }
+      );
```
tatomyr, line 364: Looks like it could be simplified: 

```suggestion
      const [{ isSunsetExpired, sunsetDate }] = sunsetWarnings.sort(
        (a: SunsetWarning, b: SunsetWarning) => {
        // First, prioritize by expiration status
        if (a.isSunsetExpired !== b.isSunsetExpired) {
          return a.isSunsetExpired ? -1 : 1;
        }

        // If both are either expired or not, sort by sunset date
        return a.sunsetDate > b.sunsetDate ? 1 : -1;
      }
    );

```

roman-sainchuk: Much better, thanks!

packages/cli/src/cms/api/api-client.ts
```diff
@@ -13,54 +14,115 @@ import type {
...
+
+    if (sunsetTime) {
+      const sunsetDate = new Date(sunsetTime);
+      let sunsetWarning: SunsetWarning;
+
+      if (sunsetTime > Date.now()) {
+        sunsetWarning = {
+          sunsetDate,
+          isSunsetExpired: false,
+          warning: `Endpoint ${
```
tatomyr, line 62: Do we actually use the warnings somewhere? Looks like it duplicates this one: https://github.com/Redocly/redocly-cli/pull/1677/files#diff-6c6d034426b6f96932bea21f0920f55c750b51dab2f7544c4154880f3cae2182R361

packages/cli/src/cms/api/api-client.ts
```diff
@@ -13,54 +14,115 @@ import type {
...
+        sunsetWarning = {
+          sunsetDate,
+          isSunsetExpired: true,
+          warning: `Endpoint ${
+            response.url
+          } support expired on ${sunsetDate.toISOString()} and will be removed.`,
+        };
+      }
+
+      this.sunsetWarnings.push(sunsetWarning);
```
tatomyr, line 76:
```suggestion
    if (!sunsetTime) return
    
    const sunsetDate = new Date(sunsetTime);
    if (sunsetTime > Date.now()) {
        this.sunsetWarnings.push({
          sunsetDate,
          isSunsetExpired: false,
          warning: `Endpoint ${
            response.url
          } is marked for removal after ${sunsetDate.toISOString()}`,
        });
      } else {
        this.sunsetWarnings.push({
          sunsetDate,
          isSunsetExpired: true,
          warning: `Endpoint ${
            response.url
          } support expired on ${sunsetDate.toISOString()} and will be removed.`,
        };
      })
```

packages/cli/src/cms/api/api-client.ts
```diff
@@ -13,54 +14,115 @@ import type {
...
+      return Date.parse(sunsetDate);
+    }
+
+    sunsetDate = headers.get('Sunset');
+
+    if (sunsetDate) {
+      return Date.parse(sunsetDate);
+    }
+
+    return;
```
tatomyr, line 99: What about this?

```suggestion
    const sunsetDate = headers.get('sunset') || headers.get('Sunset');
    return sunsetDate && Date.parse(sunsetDate)
```

tatomyr (commented):
Left a couple more suggestions. Please have a look.

packages/cli/src/cms/api/api-client.ts
roman-sainchuk: Makes sense, thanks!

tatomyr (approved):
LGTM

lornajane (approved):

PR #1688, opened 2024-08-28
fix: fix validation of redocly.yaml for product folder

RomanHotsiy:
## What/Why/How?

The product config was missing root and output types so validation failed.

## Reference
Fixes https://github.com/Redocly/redocly/issues/10504

## Testing
I have added unit tests.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/types/redocly-yaml.ts
+++ b/packages/core/src/types/redocly-yaml.ts
@@ -7,6 +7,7 @@ import { SpecVersion, getTypes } from '../oas-types';
 import type { JSONSchema } from 'json-schema-to-ts';
 import type { NodeType } from '.';
 import type { Config } from '../config';
+import { output } from '../output';
 
 const builtInCommonOASRules = [
   'spec',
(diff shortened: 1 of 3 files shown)


packages/core/src/types/redocly-yaml.ts
```diff
@@ -313,6 +314,8 @@ const createConfigApisProperties = (nodeTypes: Record<string, NodeType>): NodeTy
   ...nodeTypes['rootRedoclyConfigSchema.apis_additionalProperties'],
   properties: {
     ...nodeTypes['rootRedoclyConfigSchema.apis_additionalProperties']?.properties,
+    root: { type: 'string' },
+    output: { type: 'string' },
```
tatomyr, line 318: Those two should have come from the [config](https://github.com/Redocly/redocly/blob/main/packages/redocly-config/src/root-config-schema.ts#L138) itself. If they are missing, then something is wrong on Realm's side.

RomanHotsiy: You are right. it was a bug in config.

Issue #1649, opened 2024-07-24
Remove all usages of component marked x-internal


rubypersona:
**Is your feature request related to a problem? Please describe.**

We're using the [remove-x-internal](https://redocly.com/docs/cli/decorators/remove-x-internal) decorator:

```
decorators:
      remove-x-internal:
        internalFlagProperty: 'x-remove-from-github'
```

I have a component that looks like this:

```
x-remove-from-github: true
title: candy-cane
type: object
```

And another component (let's call it `christmas-tree`) that references this component:

```
type: array
items:
  discriminator:
    propertyName: type
    mapping:
      ...
      candy-cane: ../candy-cane.yaml
  anyOf:
    - $ref: ../candy-cane.yaml
    - ...
```

When we bundle with `x-remove-from-github`, the resulting spec has `candy-cane` removed from `components` (as expected), but it's still referenced in `components/schema/christmas-tree`, which results in the spec being invalid.

**Describe the solution you'd like**

Is there a way for all references to the `candy-cane` component to be removed when `x-remove-from-github` is applied?

**Describe alternatives you've considered**

I looked into using decorators, but couldn't find a way to determine whether a given node has a given property, and then delete it from the spec. If there is a way to do this with decorators, some guidance would be appreciated. Thanks!

jeremyfiel:
How would your discriminator work if you remove the reference to the schema being discriminated?

rubypersona:
@jeremyfiel is your concern what happens if an API response does contain an object of type `candy-cane`? That's a valid point to consider. Are there still concerns if we know the API hasn't started returning this type of object yet (let's say that feature is in development and we proactively updated our OpenAPI schema but we don't want to expose this object type publicly yet)?

jeremyfiel:
If the schema with the discriminator is encountered, no discrimination would be available if the candy cane is removed 

What behavior would you expect from that schema without the discriminator working?

rubypersona:
Oh the schema actually looks like this:

```
type: array
items:
  discriminator:
    propertyName: type
    mapping:
      candy-cane: ../candy-cane.yaml
      popcorn: ../popcorn.yaml
      cranberry: ../cranberry.yaml
  anyOf:
    - $ref: ../candy-cane.yaml
    - $ref: ../popcorn.yaml
    - $ref: ../cranberry.yaml
 ```

jeremyfiel:
This needs more investigation.  Someone will come back to you soon.  Thanks for your patience

[...]

PR #1694, opened 2024-08-29
fix: clean up the discriminator mapping in the remove-x-internal decorator

tatomyr:
## What/Why/How?

Fixed the `remove-x-internal` decorator, which was not removing the reference in the corresponding discriminator mapping while removing the original $ref.

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/1649

## Check yourself

## Security

### DIFF
--- a/packages/core/src/decorators/common/remove-x-internal.ts
+++ b/packages/core/src/decorators/common/remove-x-internal.ts
@@ -22,7 +22,7 @@ export const RemoveXInternal: Oas3Decorator | Oas2Decorator = ({
           const resolved = ctx.resolve(node[i]);
           if (resolved.node?.[internalFlagProperty]) {
             // First, remove the reference in the discriminator mapping, if it exists:
-            if (parent.discriminator?.mapping) {
+            if (isPlainObject(parent.discriminator?.mapping)) {
               for (const mapping in parent.discriminator.mapping) {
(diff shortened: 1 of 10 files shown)


packages/core/src/decorators/common/remove-x-internal.ts
```diff
@@ -6,39 +6,50 @@ import type { UserContext } from '../../walk';
...
     const { parent, key } = ctx;
     let didDelete = false;
     if (Array.isArray(node)) {
       for (let i = 0; i < node.length; i++) {
         if (isRef(node[i])) {
           const resolved = ctx.resolve(node[i]);
-          if (resolved.node?.[hiddenTag]) {
+          if (resolved.node?.[internalFlagProperty]) {
+            // First, remove the reference in the discriminator mapping, if it exists:
+            if (parent.discriminator?.mapping) {
```
RomanHotsiy, line 25: add extra check for typeof object

RomanHotsiy (approved):

JLekawa (approved):

packages/core/src/decorators/common/remove-x-internal.ts
tatomyr: Added [here](https://github.com/Redocly/redocly-cli/pull/1694/commits/03f55f8f73bb4509a06a0f648460fa61572fd43e).

Issue #1673, opened 2024-08-21
After upgrading v1.19.0->1.20.0: failed to load plugin...please provide a valid path


el2iot2:
After upgrading to v1.20.0, the JavaScript plugin that I wrote was no longer found and prevented commands from working:

## Works in v1.19.0 #
```Powershell
> npx @redocly/cli@1.19.0 build-docs myapi -o yada.html

    ╔═══════════════════════════════════════════════════════╗
    ║                                                       ║
    ║  A new version of Redocly CLI (1.20.0) is available.  ║
    ║  Update now: `npm i -g @redocly/cli@latest`.          ║
    ║  Changelog: https://redocly.com/docs/cli/changelog/   ║
    ║                                                       ║
    ╚═══════════════════════════════════════════════════════╝

Found .redocly.yaml and using theme.openapi options
Prerendering docs
Non-existing tag "tenant-provisioning" is added to the group "Administration / Operation"
Non-existing tag "trait-async-event" is added to the group "Traits"
Non-existing tag "trait-relo-distribution" is added to the group "Traits"

🎉 bundled successfully in: yada.html (1787 KiB) [⏱ 3ms].

> npx @redocly/cli@1.19.0 check-config

    ╔═══════════════════════════════════════════════════════╗
    ║                                                       ║
    ║  A new version of Redocly CLI (1.20.0) is available.  ║
    ║  Update now: `npm i -g @redocly/cli@latest`.          ║
    ║  Changelog: https://redocly.com/docs/cli/changelog/   ║
    ║                                                       ║
    ╚═══════════════════════════════════════════════════════╝


Properties 'enumSkipQuotes', 'sortOperationsAlphabetically', 'sortTagsAlphabetically', 'htmlTemplate', 'theme' are only used in API Reference Docs and Redoc version 2.x or earlier.

✅  Your config is valid.
```

## Fails in v1.20.0
```
> npx @redocly/cli@1.20.0 build-docs myapi -o yada.html
Something went wrong when processing :

  - Failed to load plugin "plugins/myplugin.js". Please provide a valid path

> npx @redocly/cli@1.20.0 check-config
Something went wrong when processing :

  - Failed to load plugin "plugins/myplugin.js". Please provide a valid path
```
## OpenAPI description

OpenApi: 3.1

## .redocly.yaml

```yaml
# See https://redoc.ly/docs/cli/configuration/ for more information.
apis:
  myapi:
    root: yada/openapi/openapi_myapi.yaml
[...]

PR #1705, opened 2024-09-01
fix resolve path for import plugin, after upgrade 1.20.0

Artemeey:
Error with use plugin:

```node
Something went wrong when processing :                                                                                  
  - Failed to load plugin "plugins/myapi.js": Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 's:'
```

More info: https://github.com/Redocly/redocly-cli/issues/1673

Tested in Windows

No tested in Linux.
Please, test in Linux.

### DIFF
--- a/.changeset/strange-gifts-clap.md
+++ b/.changeset/strange-gifts-clap.md
@@ -1,0 +1,5 @@
+---
+"@redocly/cli": major
+---
+
+Fix absolute path for import plugin in Windows
--- a/.changeset/strange-gifts-clap.md
+++ b/.changeset/strange-gifts-clap.md
@@ -2,4 +2,4 @@
 "@redocly/cli": major
 ---
 
-Fix absolute path for import plugin in Windows
+Fixed the absolute path for importing plugins in Windows.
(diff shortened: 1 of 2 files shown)


.changeset/strange-gifts-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": major
+---
+
+Fix absolute path for import plugin in Windows
```
JLekawa, line 5:
```suggestion
Fixed the absolute path for importing plugins in Windows.
```

.changeset/strange-gifts-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": major
```
tatomyr, line 2:
```suggestion
"@redocly/cli": patch
```

JLekawa (approved):

tatomyr:
Note: I've tested this in Workflows separately. 

@Artemeey thank you for the fix! It will be available in the next release.

tatomyr (approved):

Issue #1579, opened 2024-06-04
Add `output` as a per-API configuration option


lornajane:
**Is your feature request related to a problem? Please describe.**

Our universal config supports `output` for where to put a bundled API, but the CLI doesn't use the same thing yet. We should add it.

**Describe alternatives you've considered**

Do nothing, but then the universal config wouldn't be universal.

Related: #1294 which needs `output-docs` as a filename too.

PR #1708, opened 2024-09-03
feat: add output as a per-API configuration option

tatomyr:
## What/Why/How?

Added support for the `output` option in the per-API configuration. This adds the possibility to store bundled files for all `apis` that have the option specified.

For example, for the following `redocly.yaml`:

```yaml
apis:
   v1: 
     root: openapi1.yaml
     output: dist/v1.yaml
   v2: 
     root: openapi2.yaml
     output: dist/v2.yaml
```

Executing `redocly bundle` will store bundled files under the `dist/` directory. 
Executing `redocly bundle openapi1.yaml` will result in printing only the openapi1 bundle to stdout.

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/1579

## Check yourself

## Security


lornajane (changes requested):
The feature behaves differently depending if we're bundling one API or many, which I did not expect. I can tell from the documentation that this is an intentional exception - but I don't know why we did this. In every other context, the API is linted/bundled the same way whether it's done alone by alias, or others are done at the same time. With this configuration, the bundle command will apply per-api decorators, but not per-api output, which seems wrong.

The command line parameter should still take priority if it's supplied, but the configuration should apply if it exists, regardless of whether there are other APIs being bundled at the same time or not.

tatomyr:
@lornajane do you mean that having  this `redocy.yaml`:
```yaml
apis:
  main@v1:
    root: openapi.yaml
    output: bunlded.yaml
```
and running `redocly bundle main@v1` it should store it to `bundled.yaml`?

lornajane:
Yes, running `redocly bundle [alias]` and running `redocly bundle` should do the same thing for the API named by `[alias]`.

tatomyr:
@lornajane but  how would you print it out to the stdout for the alias if needed?

lornajane:
Oh, of course. You can't use the filename without silently picking up the alias :( I guess another alias, or a special value for `--output`

tatomyr:
@lornajane another alias means you have to maintain consistency between those two. And a special output value could collide with a filename (users are quite creative when it comes to naming things!).

tatomyr:
I believe this option is mainly aimed for bundling and storing large amount of apis which is quite tedious to do via the command line. So that’s where the ‘output’  really comes in handy. If a user wants to bundle one alias, it’s not hard to specify the output destination as well.

lornajane:
The current behaviour is unexpected - test it alone and you get one behaviour. Run it in a batch and it does something different. I'm not comfortable with us shipping it as it is.

tatomyr:
@lornajane changed the behaviour. Now if you bundle an alias, it will take the output from the config if it exists.

lornajane (approved):

lornajane:
Everything here looks great to me, but I was surprised that I could bundle a file by name without it picking up the alias magically (since the file is mentioned in the config file). So I like everything here, but I wanted to mention it in case the change was not intentional.

packages/cli/src/utils/miscellaneous.ts
```diff
@@ -356,33 +364,20 @@ export function printConfigLintTotals(totals: Totals, command?: string | number)
...
-    }
-    outputFile = join(dirname(outputFile), basename(outputFile, extname(outputFile))) + '.' + ext;
+  if (output) {
+    ext = ext || (extname(output).substring(1) as BundleOutputFormat);
+  }
+  ext = ext || (extname(entrypoint).substring(1) as BundleOutputFormat);
+  if (!outputExtensions.includes(ext)) {
+    throw new Error(`Invalid file extension: ${ext}.`);
   }
+  outputFile = join(dirname(outputFile), basename(outputFile, extname(outputFile))) + '.' + ext;
```
RomanHotsiy, line 380: I think there may be aproblem here. The `output` in `redocly.yaml` should be relative to the `reodcly.yaml`'s folder, not to the current working dir.

RomanHotsiy (commented):
Left one comment, otherwise looks good.

packages/cli/src/utils/miscellaneous.ts
tatomyr: You're right, good catch!

RomanHotsiy (approved):

Issue #1718, opened 2024-09-06
Configurable rules for non-OpenAPI formats


lornajane:
**Describe the bug**

I expected to be able to write configurable rules to check my Arazzo descriptions, but the rules don't seem to apply as I expected.

**To Reproduce**

1. Given this `redocly.yaml` file

```
extends: 
  - recommended

arazzoRules:
  sourceDescription-name-unique: warn
  workflowId-unique: warn

rules:
  rule/flow-prefix:
    subject:
      type: Workflow
      property: workflowId
    assertions:
      pattern: /test/

  rule/info-description:
    subject:
      type: Info
      property: description
    assertions:
      defined: true
      minLength: 300

```

3. And this ~~OpenAPI file(s)~~ Arazzo file: https://github.com/Redocly/museum-openapi-example/blob/main/arazzo/museum-api.arazzo.yaml
4. Run this command with these arguments... `redocly lint museum-api.arazzo.yaml`
5. No configurable rules are applied

**Expected behavior**

Lint errors because the API description doesn't meet the rules.

**Redocly Version(s)**

1.23.0

**`Node.js` Version(s)**

20.4.0

PR #1721, opened 2024-09-06
fix: apply configurable rules for arazzo

DmitryAnansky:
## What/Why/How?

 Apply configurable rules for arazzo specification.

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/1718

## Check yourself

## Security

### DIFF
--- a/.changeset/tough-shirts-invite.md
+++ b/.changeset/tough-shirts-invite.md
@@ -3,4 +3,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed an issue where custom rules were not applied to Arazzo configuration.
+Fixed an issue where custom rules were not applied to Arazzo descriptions.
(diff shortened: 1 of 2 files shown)


tatomyr (approved):
This is more like a workaround rather than fix. We need to refactor the internal representation of the rules to be able to handle such cases better.

JLekawa (approved):

.changeset/tough-shirts-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed an issue where custom rules were not applied to Arazzo configuration.
```
lornajane, line 6:
```suggestion
Fixed an issue where custom rules were not applied to Arazzo descriptions.
```

lornajane (approved):
Thanks! I agree that we need to think/talk more about all these features, but it meets my current needs (and I installed from local source so I can use it like this for now!)

PR #1749, opened 2024-10-03
fix: resolve plugins from npm packages correctly

volodymyr-rutskyi:
## What/Why/How?

When the CLI is installed globally and tries to resolve a plugin from NPM package, it can't find it because by default the `require.resolve` tries to look for the package from the CLI installation location. This would cause errors like below even though the plugin package is installed in the project's `node_modules` directory:

```
Something went wrong when processing :

  - Failed to load plugin "@redocly/portal-plugin-async-api/plugin.js": Cannot find module '@redocly/portal-plugin-async-api/plugin.js'
```

Added the config directory as an additional path to look for NPM packages when resolving plugins.

## Reference

Reported in internal conversation

## Check yourself

## Security

### DIFF
--- a/.changeset/strange-starfishes-press.md
+++ b/.changeset/strange-starfishes-press.md
@@ -3,4 +3,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed a bug where the plugins in external NPM packages could not be resolved if the CLI package was installed globally.
+Fixed an issue where the plugins in external NPM packages could not be resolved if the CLI package was installed globally.
--- a/.changeset/strange-starfishes-press.md
+++ b/.changeset/strange-starfishes-press.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed a bug where the plugins in external NPM packages could not be resolved if the CLI package was installed globally.
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 20
-      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz" "npm run"
+      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz -g" "npm run"
 
   run-smoke--npm--node-20--redoc:
     needs: prepare-smoke
@@ -63,7 +63,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz" "npm run"
+      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz -g" "npm run"
 
   run-smoke--npm--node-18--redoc:
     needs: prepare-smoke
@@ -89,7 +89,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 16
-      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz" "npm run"
+      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz -g" "npm run"
 
   run-smoke--npm--node-16--redoc:
     needs: prepare-smoke
@@ -115,7 +115,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 14
-      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz" "npm run"
+      - run: bash ./__tests__/smoke/run-smoke.sh "npm i redocly-cli.tgz -g" "npm run"
 
   run-smoke--npm--node-14--redoc:
     needs: prepare-smoke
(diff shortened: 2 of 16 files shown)


packages/core/src/config/config-resolvers.ts
```diff
@@ -139,7 +139,7 @@ export async function resolvePlugins(
         const absolutePluginPath = existsSync(maybeAbsolutePluginPath)
           ? maybeAbsolutePluginPath
           : // For plugins imported from packages specifically
-            require.resolve(plugin);
+            require.resolve(plugin, { paths: [configDir] });
```
adamaltman, line 142: Is there one and only one config dir? What happens if there are multiple nested configs?

.changeset/strange-starfishes-press.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed a bug where the plugins in external NPM packages could not be resolved if the CLI package was installed globally.
```
HCloward, line 6:
```suggestion
Fixed an issue where the plugins in external NPM packages could not be resolved if the CLI package was installed globally.
```

packages/core/src/config/config-resolvers.ts
volodymyr-rutskyi: The `node_modules` directory will always be in the root of the project. If the `configDir` is a nested directory, the plugin will still get resolved.

__tests__/smoke-plugins/run-smoke.sh
```diff
@@ -0,0 +1,19 @@
...
+echo
+echo "Directory content:"
+ls -a
+echo
+
+# Install plugin package
+npm i
+
+# Install CLI globally
+npm i redocly-cli.tgz -g
```
tatomyr, line 15: Let's move it under package.json's dependencies.

.github/workflows/smoke.yaml
```diff
@@ -247,3 +247,55 @@ jobs:
...
+    needs: prepare-smoke
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/cache@v3
+        with:
+          path: __tests__/smoke-plugins/
+          key: cache-${{ github.run_id }}-${{ github.run_attempt }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
tatomyr, line 300:
```suggestion
  run-smoke-plugins--npm--node-22:
    needs: prepare-smoke
    runs-on: ubuntu-latest
    steps:
      - uses: actions/cache@v3
        with:
          path: __tests__/smoke-plugins/
          key: cache-${{ github.run_id }}-${{ github.run_attempt }}
      - uses: actions/setup-node@v3
        with:
          node-version: 22
```

I think it makes more sense to test against the latest Node 😄.

__tests__/smoke/run-smoke.sh
```diff
@@ -8,6 +8,9 @@ echo "Directory content:"
 ls -a
 echo
 
+# Install npm plugin
+npm i
```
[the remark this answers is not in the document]
volodymyr-rutskyi: Reverted this change

__tests__/smoke-plugins/run-smoke.sh
volodymyr-rutskyi: To reproduce the original issue this PR fixes, the CLI needs to be installed globally.

.changeset/strange-starfishes-press.md
volodymyr-rutskyi: Applied the suggestion

.github/workflows/smoke.yaml
volodymyr-rutskyi: Sure. Added jobs for node 22 for all smoke tests

.github/workflows/smoke.yaml
```diff
@@ -26,6 +26,49 @@ jobs:
           path: __tests__/smoke/
           key: cache-${{ github.run_id }}-${{ github.run_attempt }}
 
+  prepare-smoke-plugins:
```
tatomyr, line 29: Can we split smokes and plugin-smokes into different workflows? I feel like they should be updated separately.

scripts/prepare-smoke.sh
```diff
@@ -7,6 +7,7 @@ npm run webpack-bundle
 npm run pack:prepare
 
 cp ./redocly-cli.tgz ./openapi-core.tgz ./dist/bundle.js ./__tests__/smoke/
+cp ./redocly-cli.tgz ./openapi-core.tgz ./__tests__/smoke-plugins
```
tatomyr, line 10: I'd move that to a separate script. It's being called twice anyway: https://github.com/Redocly/redocly-cli/pull/1749/files#diff-e5a87d0236e416c1fa636174d597bd7628a7434d6fe5bacbe4badd042b94804bR40

tatomyr (commented):
Left an optional suggestion. Otherwise looks good.

.github/workflows/smoke.yaml
volodymyr-rutskyi: Done

scripts/prepare-smoke.sh
volodymyr-rutskyi: Created a separate script for plugins smoke preparation

tatomyr (approved):

HCloward (approved):

PR #1755, opened 2024-10-08
Chore/update dependencies to remove punycode depricated package

DmitryAnansky:
## Check yourself

## Security


packages/core/src/utils.ts
```diff
@@ -1,16 +1,31 @@
...
 import type { HttpResolveConfig } from './config';
 import type { UserContext } from './walk';
 
+let fetch: ((input: RequestInfo | URL, init?: RequestInit) => Promise<Response>) & {
+    (input: RequestInfo | URL, init?: RequestInit): Promise<Response>;
+    (input: string | URL | globalThis.Request, init?: RequestInit): Promise<Response>;
+  },
+  ProxyAgent: (new (arg0: string) => any) | null;
+
+if (typeof window === 'undefined') {
```
RomanHotsiy, line 18: this is has to be fixed very carefully

DmitryAnansky: Its for browser support, I found some logger tests for that.
Going to test in vscode / monorepo with some verdaccio package.

DmitryAnansky: I changed it to `isBrowser` from `.env`. Or should we use some other workaround?

packages/core/package-lock.json
```diff
@@ -1,555 +0,0 @@
-{
```
DmitryAnansky, line 1: @tatomyr 
This package-lock looks obsolete in the mono-repo setup, but please tell me if there is some idea behind it.

DmitryAnansky:
Going to test another approach with native fetch.

packages/core/package-lock.json
tatomyr: You can try removing it and see what happens.

PR #1759, opened 2024-10-10
fix: component-name-unique rule should report location

pboos:
## What/Why/How?
What: This sets the location on the reported problem & makes the message one line instead of multiple lines.

Why: The `component-name-unique` didn't report any location. That location is specifically useful when using the ignore file. The message is more aligned with the other reports.

How: By using passing the `location` and reporting it instead of just passing the `absolutePointer`.

## Testing
See updated test code.

## Check yourself

## Security

### DIFF
--- a/.changeset/real-chicken-cheat.md
+++ b/.changeset/real-chicken-cheat.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": minor
 ---
 
-Fixed `component-name-unique` problems to include correct location
+Fixed `component-name-unique` problems to include correct location.
--- a/.changeset/real-chicken-cheat.md
+++ b/.changeset/real-chicken-cheat.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+Fixed `component-name-unique` problems to include correct location
--- a/packages/core/src/rules/oas3/component-name-unique.ts
+++ b/packages/core/src/rules/oas3/component-name-unique.ts
@@ -59,9 +59,11 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
           if (value.absolutePointers.size > 1) {
             const component = getComponentFromKey(key);
             const optionComponentName = getOptionComponentNameForTypeName(component.typeName);
-            const definitions = Array.from(value.absolutePointers).join(', ');
+            const definitions = Array.from(value.absolutePointers)
+              .map((v) => `- ${v}`)
+              .join('\n');
             const problem: Problem = {
-              message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at: ${definitions}`,
+              message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at:\n${definitions}`,
             };
             problem.location = {
               ...value.locations.sort((a, b) =>
--- a/packages/core/src/rules/oas3/component-name-unique.ts
+++ b/packages/core/src/rules/oas3/component-name-unique.ts
@@ -59,12 +59,10 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
           if (value.absolutePointers.size > 1) {
             const component = getComponentFromKey(key);
             const optionComponentName = getOptionComponentNameForTypeName(component.typeName);
-            const definitions = Array.from(value)
-              .map((v) => `- ${v}`)
-              .join('\n');
+            const definitions = Array.from(value.absolutePointers).join(', ');
 
             const problem: Problem = {
-              message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at:\n${definitions}`,
+              message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at: ${definitions}`,
             };
             problem.location = {
             ...value.locations[0],
--- a/packages/core/src/rules/oas3/component-name-unique.ts
+++ b/packages/core/src/rules/oas3/component-name-unique.ts
@@ -60,7 +60,7 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
             const component = getComponentFromKey(key);
             const optionComponentName = getOptionComponentNameForTypeName(component.typeName);
             const componentSeverity = optionComponentName ? options[optionComponentName] : null;
-            value.locations.forEach((location) => {
+            for (const location of value.locations) {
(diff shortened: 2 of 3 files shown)


tatomyr:
Thank you for the contribution @pboos! I'll take a look into this once I have time.

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -46,32 +49,47 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
...
+                a.absolutePointer.localeCompare(b.absolutePointer)
+              )[0],
             };
 
             const componentSeverity = optionComponentName ? options[optionComponentName] : null;
             if (componentSeverity) {
               problem.forceSeverity = componentSeverity;
             }
             ctx.report(problem);
+            // value.locations.forEach((location) => {
```
lornajane, line 77: Please remove these commented lines if they are not needed

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -46,32 +49,47 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
...
         components.forEach((value, key, _) => {
-          if (value.size > 1) {
+          if (value.absolutePointers.size > 1) {
             const component = getComponentFromKey(key);
             const optionComponentName = getOptionComponentNameForTypeName(component.typeName);
-            const definitions = Array.from(value)
-              .map((v) => `- ${v}`)
-              .join('\n');
-
+            const definitions = Array.from(value.absolutePointers).join(', ');
```
lornajane, line 62: I'm not sure about this change, because it reformats the layout to be more difficult to read.

Previously we had

```
Component 'schemas/SomeSchema' is not unique. It is defined at:
- .../foobar.yaml#/components/schemas/SomeSchema
- .../test.yaml#/components/schemas/SomeSchema
```

Whereas now those all flow on one line, making it harder to read and compare the paths. The additional location information is great, but can we retain the original formatting on the output message please?

lornajane (commented):
This a nice improvement to have the accurate location of one of the duplicates. I added a couple of comments about keeping the output formatting, and removing some commented code, could you take a look?

packages/core/src/rules/oas3/component-name-unique.ts
pboos: Reverted it now to how it was before. You are right that it is more readable in the old format.

packages/core/src/rules/oas3/component-name-unique.ts
pboos: Oh my, totally forgot to remove them 🙈 . Removed them now.

lornajane (approved):
Thanks @pboos , this is definitely an improvement :)

lornajane:
Also tagging @tatomyr - do we need to update some test snapshots to go with this change?

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -46,26 +49,27 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
...
+            const definitions = Array.from(value.absolutePointers)
               .map((v) => `- ${v}`)
               .join('\n');
-
             const problem: Problem = {
               message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at:\n${definitions}`,
             };
+            problem.location = {
+              ...value.locations.sort((a, b) =>
+                a.absolutePointer.localeCompare(b.absolutePointer)
```
tatomyr, line 70: What's the reason of using `localeCompare`?

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -46,26 +49,27 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
...
+              )[0],
```
tatomyr, line 71: Could we simply report each duplicated component? To me, it seems that would make spotting duplication easier.

.changeset/real-chicken-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+Fixed `component-name-unique` problems to include correct location
```
tatomyr, line 5:
```suggestion
Fixed `component-name-unique` problems to include correct location.
```

packages/core/src/rules/oas3/component-name-unique.ts
pboos: I wanted to keep the code simple. I am not aware of a `compare` function that is without the `locale` part. Even though the locale part is not needed. But a simple `===` won't work as we need -1, 0, 1 return values for it to sort corectly.

packages/core/src/rules/oas3/component-name-unique.ts
tatomyr: >  But I guess having only one on output but multiple in ignore is not possible, or is it?

I don't think it's possible. You can only ignore items that are in the output.

> Then those two reports basically are duplicates just with a different location

Perhaps you could filter out the current location and list only the duplicates to reduce noise. In most cases, there will likely be only one duplicate.

tatomyr: Frankly, I'm not against having a single output location. I'm just curious if there’s a better option. If your approach works for you, let’s go with it 👍

pboos: Change pushed.

tatomyr (commented):
Looks great! Left one small style comment.

packages/core/src/rules/oas3/component-name-unique.ts
```diff
@@ -46,32 +49,32 @@ export const ComponentNameUnique: Oas3Rule | Oas2Rule = (options) => {
...
-            const problem: Problem = {
-              message: `Component '${optionComponentName}/${component.componentName}' is not unique. It is defined at:\n${definitions}`,
-            };
-
             const componentSeverity = optionComponentName ? options[optionComponentName] : null;
-            if (componentSeverity) {
-              problem.forceSeverity = componentSeverity;
-            }
-            ctx.report(problem);
+            value.locations.forEach((location) => {
```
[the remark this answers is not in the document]
pboos: done

tatomyr (approved):
LGTM. I'm going to merge this PR soon. 
Thank you very much @pboos for making Redocly CLI better!

PR #1697, opened 2024-08-29
docs: add arazzo rules documentation

DmitryAnansky:
## What/Why/How?

Add documentation for Arazzo and AsyncAPI rules

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/1691

## Check yourself

- [ ] Code changed? - Tested with redoc/reference-docs/workflows (internal)
- [ ] All new/updated code is covered with tests
- [ ] New package installed? - Tested in different environments (browser/node)

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #1773, opened 2024-10-17
feat: removed the `type:none` sourceDescriptions Arazzo extension

DmitryAnansky:
## What/Why/How?

- Removed the `type:none` sourceDescriptions Arazzo extension.
- Add additional lint rule for sourceDescriptions:

sourceDescriptions | [Source Description Object] | REQUIRED. A list of source descriptions (such as an OpenAPI description) this Arazzo Description SHALL apply to. The list MUST have at least one entry.
-- | -- | --

## Reference

https://github.com/Redocly/redocly/issues/11343

## Check yourself

## Security


docs/rules/arazzo/sourceDescriptions-empty.md
```diff
@@ -0,0 +1,119 @@
+# sourceDescriptions-empty
+
+The Source Description object must be defined when `operationId` or `operationPath` is used inside Step to referense description operation.
```
adamaltman, line 3:
```suggestion
The Source Description object must be defined when `operationId` or `operationPath` is used inside Step to reference the description operation.
```

docs/rules/arazzo/sourceDescriptions-empty.md
```diff
@@ -0,0 +1,119 @@
...
+`sourceDescriptions` can be undefined in the Arazzo description in case there are no OpenAPI description exist and `x-operation` extension is used inside Step.
+
+| Arazzo | Compatibility |
+| ------ | ------------- |
+| 1.0.0  | ✅            |
+
+## API design principles
```
adamaltman, line 10: This heading isn't good.

docs/rules/arazzo/sourceDescriptions-empty.md
```diff
@@ -0,0 +1,119 @@
+# sourceDescriptions-empty
```
adamaltman, line 1: This rule doesn't really make sense. 

https://spec.openapis.org/arazzo/latest.html#arazzo-specification-object

`sourceDescriptions` is required as part of the spec.

DmitryAnansky: Yes, it is required, but in that case you can`t use only `x-operation` without adding some dummy sourceDescription.
I think we need to have this rule and it will allow you to omit sourceDescription if you don't have any, or will tell you to add one if `opearationId` | `oprationPath` is used.

docs/rules/arazzo/sourceDescriptions-empty.md
DmitryAnansky: any suggestions ?
I took it from some existing examples here https://github.com/Redocly/redocly-cli/pull/1697
But I saw there are no final decision on the format.

DmitryAnansky: cc: @lornajane

docs/rules/arazzo/sourceDescriptions-empty.md
```diff
@@ -0,0 +1,119 @@
...
+  title: Cool API
+  version: 1.0.0
+  description: A cool API
+sourceDescriptions:
+  - name: museum-api
+    type: openapi
+    url: openapi.yaml
+  - name: museum-api
+    type: openapi
+    url: openapi.yaml
```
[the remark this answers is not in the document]
DmitryAnansky: updated, thx

docs/rules/arazzo/sourceDescriptions-empty.md
DmitryAnansky: updated, thx

docs/rules/arazzo/sourceDescriptions-empty.md
DmitryAnansky: Alternative solution might be allow `sourceDescriptions` to be empty list.
But this will be against Arazzo spec as it should have at least one described item.

DmitryAnansky: Discussed offline  with @adamaltman ->  I will transform this rule to check that there is at least one sourceDescription as per Arazzo Spec.
There will be no way to use Spot without description added to sourceDescriptions.

docs/rules/arazzo/sourceDescriptions-empty.md
DmitryAnansky: If no suggestion - I can leave it as it is - so we can update heading later.

lornajane:
I can fix the docs header but I thought we were holding off on adding docs until we were happy with how they were all shaping up in #1697 . Maybe we can make the code changes here, and put the docs over there? (where I will also revisit the section headers on the Arazzo rules docs)

DmitryAnansky:
> I can fix the docs header but I thought we were holding off on adding docs until we were happy with how they were all shaping up in #1697 . Maybe we can make the code changes here, and put the docs over there? (where I will also revisit the section headers on the Arazzo rules docs)

Also I can merge changes just to have rule described - so we can update docs later.

lornajane (commented):
Rule works as expected and I agree it should be in our recommended set. I think we should remove the docs though since we haven't added other docs yet - could you transfer the new rule page into #1697?

DmitryAnansky:
> Rule works as expected and I agree it should be in our recommended set. I think we should remove the docs though since we haven't added other docs yet - could you transfer the new rule page into #1697?

Moved docs update to this [PR](https://github.com/Redocly/redocly-cli/pull/1697)

.changeset/dull-pigs-jump.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/cli": patch
+"@redocly/openapi-core": patch
+---
+
+Removed the `type:none` sourceDescriptions Arazzo extension.
```
lornajane, line 6:
```suggestion
Updated sourceDescriptions to require a valid type, to align with the Arazzo specification.
```

lornajane (approved):
Please apply the changelog suggestion!

DmitryAnansky:
> I don't like this. It seems like it could be part of the spec rule. What am I missing?

It is by the Spec rule:
We have `sourceDescriptions` required in type and checking that it has at least one sourceDescription with the rule.

packages/core/src/rules/arazzo/__tests__/sourceDescriptions-not-empty.test.ts
```diff
@@ -0,0 +1,106 @@
...
+              description: >-
+                Get museum hours by resolving request details with getMuseumHours operationId from openapi.yaml description.
+              operationId: museum-api.getMuseumHours
+              successCriteria:
+                - condition: $statusCode == 200
+    `,
+    'arazzo.yaml'
+  );
+
+  it('should not report an error when sourceDescriptions have at leasr one entry.', async () => {
```
DmitryAnansky, line 65:
```suggestion
  it('should not report an error when sourceDescriptions have at least one entry.', async () => {
```

DmitryAnansky:
> I don't like this. It seems like it could be part of the spec rule. What am I missing?

This can be also done as part of type function check

```

const Root: NodeType = {
  properties: {
    arazzo: { type: 'string' },
    info: 'Info',
    sourceDescriptions: (items) => {
      if (items.length === 0) {
        throw new Error('SourceDescriptions must not be empty');
      }
      return 'SourceDescriptions'
    },
    workflows: 'Workflows',
    components: 'Components',
  },
  required: ['arazzo', 'info', 'sourceDescriptions', 'workflows'],
  extensionsPrefix: 'x-',
};
```
But error report is worse than having a rule: 

Screenshot 2024-10-21 at 11 43 04


With the rule:

Screenshot 2024-10-21 at 11 45 36


@adamaltman , please clarify what variant would you prefer.

DmitryAnansky:
> @DmitryAnansky I think you're missing my point. It must be part of the spec rule. When the spec rule covers less and less of the spec we shouldn't even call it a spec rule then. Get my point?

In current cli implementation of the rules - `spec` rule is an abstraction on checking defined custom type.
It is not extendable to apply similar logic as in the rule I introduced.
And our custom types does not have minItems.

Please correct me if I am wrong @tatomyr , @RomanHotsiy

DmitryAnansky:
> @DmitryAnansky I think you're missing my point. It must be part of the spec rule. When the spec rule covers less and less of the spec we shouldn't even call it a spec rule then. Get my point?

@adamaltman moreover some of existing unique rules can't be defined in our `spec` rule.

tatomyr:
> And our custom types does not have minItems.

@DmitryAnansky, yes, this is a limitation of the existing `spec` rule. It doesn't include this type of check out of the box. It can be extended if needed, but I noticed we have a few exceptions in the OAS `spec` rule as well. For instance, the [no-example-value-and-externalValue](https://redocly.com/docs/cli/rules/no-example-value-and-externalValue) rule which covers the requirement that the fields are mutually exclusive by [the specification](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-15), or [operation-description](https://redocly.com/docs/cli/rules/operation-description) which must be [unique](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-7).

adamaltman:
@tatomyr those were honest omissions. Mostly added as separate rules in the future for backwards compatibility reasons.

We are not in that situation with Arazzo. @DmitryAnansky we should not start the day with technical debt.

tatomyr:
@adamaltman thanks for the clarification!

adamaltman:
Given this information, do we have enough information to move forward?

RomanHotsiy:
@adamaltman I would like to discuss it with you quickly before we proceed.
If we want to make changes in this area we should consider bigger changes or we may end up with not the most optimal design.

The problem is that the `spec` rule is named incorrectly in my opinion. It should be named `struct` which means "Structural validation". It was never designed to validate the "spec" conformance.

Sure, we can wrap up all the various spec-related checks under the `spec` rule but it has some disadvantages:

- all the spec related errors will end up with `spec` ruleId in the output which may limit tooling in the future (or tooling will have to rely on error message)
- we loose some extra flexibility to be able to disable some aspects of spec validation (it doesn't make sense in most of cases but `openapi-core` can be used for various use-cases so you never know).
- (minor) `spec` rule code will get much harder to maintain if we want to keep same performance

I think the better solution would be to:

- rename the `spec` rule into `struct` rule (or something like that)
- prefix all of the other spec-related rules with `spec-` prefix (we already have a few like that)
- create a `spec` ruleset which includes all of the spec-related rules for each type of spec (openapi, asyncapi, arazzo)

This solution is a breaking change so I'm not suggesting doing it right now (or maybe?). I want to bring it up though so whatever we do right now is not blocking us in the future.

Let's discuss it offline and decide?

adamaltman:
Yes, this is exactly what I was thinking (I like `struct` as I had a similar name in mind). I think when you say `spec-{specname}` you mean it is a config that someone would use like:

```
extends:
  - spec
```

And other configs like `recommended` could extend that itself too. Right?

RomanHotsiy:
> And other configs like recommended could extend that itself too. Right?

yes!

adamaltman:
@DmitryAnansky we discussed not to rename the spec rule but rather to start a similar `struct` rule (for Arazzo). We would need to make any breaking changes for OpenAPI and AsyncAPI in version 2 with the `struct` rule.

The idea is that the struct rule is part of a ruleset that includes the `struct` rule and other rules such as `sourceDescriptions` has at least one value, IDs are unique, etc...  It is misleading as a `spec` rule because it is only the basic structure of the spec, not actual things like there must be at least one sourceDescription and workflow IDs are unique.

DmitryAnansky:
Mobing code changes to https://github.com/Redocly/redocly-cli/pull/1800

PR #1795, opened 2024-11-05
feat(typings): update OpenAPI 3.0 and 3.1 typing declarations

### LINKED ISSUE #1547: Missing const typing support
**Describe the bug**

Are we missing `const` from here: https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/typings/openapi.ts 

- https://json-schema.org/understanding-json-schema/reference/const

**To Reproduce**
n/a

**Expected behavior**

To be able to use the type tree to traverse the values including enum (is already in the typings) and const (missing) values.

**Logs**

n/a

**OpenAPI description**

3.1

**Redocly Version(s)**

latest

**`Node.js` Version(s)**

n/a

**Additional context**

Relates to writing some advanced plugins/extensions.

jeremyfiel:
fixes   #1547

## What/Why/How?
The typings for OAS 3 and OAS 3_1 schemas were incorrectly defined per the OpenAPI Specification.

OAS 3.0 has a specific requirement which defines an OpenAPI Schema rather than a JSON Schema schema. There are subtleties to the OpenAPI 3.0 Schema. 

OpenAPI 3.1 Schemas are equivalent to a JSON Schema 2020-12 schema. 
 
## Reference

Per the OpenAPI 3.0 Specification

[4.7.24.1 JSON Schema Keywords](https://spec.openapis.org/oas/v3.0.4#json-schema-keywords)
The following keywords are taken directly from the JSON Schema definition and follow the same specifications:

* title
* multipleOf
* maximum
* exclusiveMaximum
* minimum
* exclusiveMinimum
* maxLength
* minLength
* pattern (This string SHOULD be a valid regular expression, according to the [Ecma-262 Edition 5.1 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1) dialect)
* maxItems
* minItems
* uniqueItems
* maxProperties
 * minProperties
* required
* enum

The following keywords are taken from the JSON Schema definition **_but their definitions were adjusted to the OpenAPI Specification._**

* type - Value MUST be a string. Multiple types via an array are not supported.
* allOf - Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema.
* oneOf - Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema.
* anyOf - Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema.
* not - Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema.
* items - Value MUST be an object and not an array. Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema. items MUST be present if type is "array".
* properties - Property definitions MUST be a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema (inline or referenced).
* additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a [Schema Object](https://spec.openapis.org/oas/v3.0.4#schema-object) and not a standard JSON Schema. Consistent with JSON Schema, additionalProperties defaults to true.
* description - [[CommonMark](https://spec.openapis.org/oas/v3.0.4#bib-commonmark)] syntax MAY be used for rich text representation.
* format - See [Data Type Formats](https://spec.openapis.org/oas/v3.0.4#data-type-format) for further details. While relying on JSON Schema’s defined formats, the OAS offers a few additional predefined formats.
* default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if type is "string", then default can be "foo" but cannot be 1.

Alternatively, any time a Schema Object can be used, a [Reference Object](https://spec.openapis.org/oas/v3.0.4#reference-object) can be used in its place. This allows referencing definitions instead of defining them inline.

Additional keywords defined by the JSON Schema specification that are not mentioned here are strictly unsupported.

Other than the JSON Schema subset fields, the following fields MAY be used for further schema documentation:

4.7.24.2 Fixed Fields
Field Name | Type | Description
-- | -- | --
nullable | boolean | This keyword only takes effect if type is explicitly defined within the same Schema Object. A true value indicates that both null values and values of the type specified by type are allowed. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of null as a value. A false value leaves the specified or default type unmodified. The default value is false.
discriminator | Discriminator Object | Adds support for polymorphism. The discriminator is used to determine which of a set of schemas a payload is expected to satisfy. See Composition and Inheritance for more details.
readOnly | boolean | Relevant only for Schema Object properties definitions. Declares the property as “read only”. This means that it MAY be sent as part of a response but SHOULD NOT be sent as part of the request. If the property is marked as readOnly being true and is in the required list, the required will take effect on the response only. A property MUST NOT be marked as both readOnly and writeOnly being true. Default value is false.
writeOnly | boolean | Relevant only for Schema Object properties definitions. Declares the property as “write only”. Therefore, it MAY be sent as part of a request but SHOULD NOT be sent as part of the response. If the property is marked as writeOnly being true and is in the required list, the required will take effect on the request only. A property MUST NOT be marked as both readOnly and writeOnly being true. Default value is false.
xml | XML Object | This MAY be used only on property schemas. It has no effect on root schemas. Adds additional metadata to describe the XML representation of this property.
externalDocs | External Documentation Object | Additional external documentation for this schema.
example | Any | A free-form field to include an example of an instance for this schema. To represent examples that cannot be naturally represented in JSON or YAML, a string value can be used to contain the example with escaping where necessary.
deprecated | boolean | Specifies that a schema is deprecated and SHOULD be transitioned out of usage. Default value is false.

---

Per the OpenAPI 3.1 Specification

[4.8.24 Schema Object](https://spec.openapis.org/oas/v3.1.1#schema-object)
The Schema Object allows the definition of input and output data types. These types can be objects, but also primitives and arrays. This object is a superset of the [JSON Schema Specification Draft 2020-12](https://spec.openapis.org/oas/v3.1.1#bib-json-schema-2020-12). The empty schema (which allows any instance to validate) MAY be represented by the boolean value true and a schema which allows no instance to validate MAY be represented by the boolean value false.

For more information about the keywords, see [JSON Schema Core](https://spec.openapis.org/oas/v3.1.1#bib-json-schema-2020-12) and [JSON Schema Validation](https://spec.openapis.org/oas/v3.1.1#bib-json-schema-validation-2020-12).

Unless stated otherwise, the keyword definitions follow those of JSON Schema and do not add any additional semantics; this includes keywords such as $schema, $id, $ref, and $dynamicRef being URIs rather than URLs. Where JSON Schema indicates that behavior is defined by the application (e.g. for annotations), OAS also defers the definition of semantics to the application consuming the OpenAPI document.

## Check yourself

## Security


jeremyfiel:
ok, i think i'm done with this. missed a few keywords for Oas3_1 schemas. 

ready for review.  thanks

.changeset/selfish-pandas-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Update typings for OAS 3.0 and OAS 3.1 Schemas
```
tatomyr, line 6:
```suggestion
"@redocly/openapi-core": patch
---

Updated typings for OAS 3.0 and OAS 3.1 Schemas.
```

packages/core/src/typings/openapi.ts
```diff
@@ -112,13 +112,10 @@ export interface Oas3Xml {
 }
 
 // common fields for OpenAPI Schema v3.x
-interface Oas3XSchemaBase<T> {
+interface Oas3XSchemaBase {
```
tatomyr, line 115: I prefer using the generic interface because it's less likely for the common properties to diverge. Not sure which approach would be cleaner though. Why did you decide to change the approach?

jeremyfiel: the keywords which utilized the generic interface were moved into the `Oas3Schema` and `Oas3_1Schema`.  So it was unused.

tatomyr: Yes, but what was the reason for moving the keywords from `Oas3XSchemaBase`? Was there an issue with leaving them as they were?

jeremyfiel: Yes, they have different behavior between 30 asd  31.  So the extension of Oas3XSchemaBase  was incorrect and the keywords should be assigned to each version

jeremyfiel:
I added a few extra JSON Schema 2020-12 keywords that were missed

packages/core/src/typings/openapi.ts
tatomyr: Ok. I'll test that once I have time. Thank you for contributing!

tatomyr (commented):
Let's make it reuse the common props slightly more. 
Otherwise looks good!

jeremyfiel:
i'm stuck on the visitors.ts and custom rule files.  Any help appreciated.

packages/core/src/typings/openapi.ts
```diff
@@ -202,46 +230,59 @@ export type Oas3ParameterStyle =
...
+  headers?: { [name: string]: Referenced<Oas3Header<T>> };
   securitySchemes?: { [name: string]: Referenced<Oas3SecurityScheme> };
   links?: { [name: string]: Referenced<Oas3Link> };
-  callbacks?: { [name: string]: Referenced<Oas3Callback> };
+  callbacks?: { [name: string]: Referenced<Oas3Callback<T>> };
+}
+
+export interface Oas3_1Components extends Oas3ComponentsBase<Oas3Schema> {
+  pathItems?: { [name: string]: Referenced<Oas3PathItem<Oas3Schema | Oas3_1Schema>> };
 }
```
tatomyr, line 283: I'd add a separate `Oas3Components` and slightly improve the `Oas3_1Components` type to use `Oas3_1Schema` only:

```suggestion

export interface Oas3_1Components extends Oas3ComponentsBase<Oas3_1Schema> {
  pathItems?: { [name: string]: Referenced<Oas3PathItem<Oas3_1Schema>> };
}

export interface Oas3Components extends  Oas3ComponentsBase<Oas3Schema>
```

packages/core/src/visitors.ts
```diff
@@ -159,36 +162,40 @@ type Oas3FlatVisitor = {
...
   Link?: VisitFunctionOrObject<Oas3Link>;
-  Schema?: VisitFunctionOrObject<Oas3Schema>;
+  Schema?: VisitFunctionOrObject<Oas3Schema | Oas3_1Schema>;
   Xml?: VisitFunctionOrObject<Oas3Xml>;
   SchemaProperties?: VisitFunctionOrObject<Record<string, Oas3Schema>>;
   DiscriminatorMapping?: VisitFunctionOrObject<Record<string, string>>;
   Discriminator?: VisitFunctionOrObject<Oas3Discriminator>;
-  Components?: VisitFunctionOrObject<Oas3Components>;
+  Components?: VisitFunctionOrObject<
+    Oas3ComponentsBase<Oas3Schema | Oas3_1Schema> | Oas3_1Components
```
tatomyr, line 186:
```suggestion
    Oas3Components | Oas3_1Components
```

packages/cli/src/commands/split/index.ts
```diff
@@ -264,15 +262,17 @@ function getFileNamePath(componentDirPath: string, componentName: string, ext: s
 }
 
 function gatherComponentsFiles(
-  components: Oas3Components,
+  components: Oas3ComponentsBase<Oas3Schema | Oas3_1Schema> | Oas3_1Components,
```
tatomyr, line 265: Related to https://github.com/Redocly/redocly-cli/pull/1795/files?diff=split&w=1#r1898410733 :

```suggestion
  components: Oas3Components | Oas3_1Components,
```

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -2,17 +2,28 @@ import { isEmptyObject } from '../../utils';
...
+  Oas3_1Schema,
+} from '../../typings/openapi';
 
 export const RemoveUnusedComponents: Oas3Decorator = () => {
   const components = new Map<
     string,
-    { usedIn: Location[]; componentType?: keyof Oas3Components; name: string }
+    {
+      usedIn: Location[];
+      componentType?: keyof (Oas3ComponentsBase<Oas3Schema | Oas3_1Schema> | Oas3_1Components);
```
tatomyr, line 19: Related to https://github.com/Redocly/redocly-cli/pull/1795/files?diff=split&w=1#r1898410733 :
```suggestion
      componentType?: keyof (Oas3Components | Oas3_1Components);
```

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -2,17 +2,28 @@ import { isEmptyObject } from '../../utils';
...
+      name: string;
+    }
   >();
 
   function registerComponent(
     location: Location,
-    componentType: keyof Oas3Components,
+    componentType: keyof (Oas3ComponentsBase<Oas3Schema | Oas3_1Schema> | Oas3_1Components),
```
tatomyr, line 26:
```suggestion
    componentType: keyof (Oas3Components | Oas3_1Components),
```

packages/core/src/typings/openapi.ts
```diff
@@ -1,19 +1,26 @@
-export interface Oas3Definition {
+// common fields for OAS descriptions v3.x
+interface Oas3DefinitionBase<T extends Oas3Schema | Oas3_1Schema> {
   openapi: string;
   info?: Oas3Info;
   servers?: Oas3Server[];
-  paths?: Oas3Paths;
-  components?: Oas3Components;
+  paths?: Oas3Paths<T>;
+  components?: T extends Oas3_1Schema ? Oas3_1Components : Oas3ComponentsBase<T>;
```
tatomyr, line 7: Related to https://github.com/Redocly/redocly-cli/pull/1795/files?diff=split&w=1#r1898410733 :

```suggestion
  components?: T extends Oas3_1Schema ? Oas3_1Components : Oas3Components;
```

HCloward (approved):

packages/core/src/decorators/oas3/remove-unused-components.ts
tatomyr: Did you actually update it?

packages/core/src/typings/openapi.ts
tatomyr: Do you see any issues with splitting 3 and 3.1 components?

packages/core/src/typings/openapi.ts
```diff
@@ -202,46 +230,61 @@ export type Oas3ParameterStyle =
...
+export interface Oas3RequestBody<T extends Oas3Schema | Oas3_1Schema = Oas3Schema | Oas3_1Schema> {
   description?: string;
   required?: boolean;
-  content: { [mime: string]: Oas3MediaType };
+  content: { [mime: string]: Oas3MediaType<T> };
 }
 
-export interface Oas3Responses {
-  [code: string]: Oas3Response;
+export interface Oas3Responses<T extends Oas3Schema | Oas3_1Schema> {
```
tatomyr, line 239: Let's add a default for backward compatibility:

```suggestion
export interface Oas3Responses<T extends Oas3Schema | Oas3_1Schema = Oas3Schema | Oas3_1Schema> {
```

packages/core/src/typings/openapi.ts
```diff
@@ -202,46 +230,61 @@ export type Oas3ParameterStyle =
...
 }
 
-export type Oas3ComponentName = keyof Oas3Components;
+export interface Oas3_1Components extends Oas3ComponentsBase<Oas3_1Schema> {
+  pathItems?: { [name: string]: Referenced<Oas3PathItem<Oas3_1Schema>> };
+}
+
+export interface Oas3Components extends Oas3ComponentsBase<Oas3Schema> {}
+
+export type Oas3ComponentName<T extends Oas3Schema | Oas3_1Schema> = keyof Oas3ComponentsBase<T>;
```
tatomyr, line 287: Let's add a default for backward compatibility:

```suggestion
export type Oas3ComponentName<T extends Oas3Schema | Oas3_1Schema = Oas3Schema | Oas3_1Schema> = keyof Oas3ComponentsBase<T>;
```

tatomyr (changes requested):
Added a couple suggestions. Otherwise all good.

packages/core/src/typings/openapi.ts
```diff
@@ -112,63 +119,84 @@ export interface Oas3Xml {
 }
 
 // common fields for OpenAPI Schema v3.x
-interface Oas3XSchemaBase<T> {
+interface Oas3XSchemaBase<T extends Oas3Schema | Oas3_1Schema> {
   $ref?: string;
-  properties?: { [name: string]: T };
-  additionalProperties?: boolean | T;
+  properties?: { [name: string]: Referenced<T> };
+  additionalProperties?: boolean | { [name: string]: Referenced<T> };
```
tatomyr, line 125: Oh, one more thing. I guess it should be just `T` as `additionalProperties` accepts a schema:

```suggestion
  additionalProperties?: boolean | T;
```

tatomyr (changes requested):

PR #1804, opened 2024-11-19
fix: use shell when running child_process.spawn

mallachari:
## What/Why/How?
CLI preview command was failing on Windows with specific Node versions.  
Turns out there is a [breaking change](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2) in Node API that prevents passing `.cmd` files to `spawn` function without using `shell: true` option. Since we're using `npx.cmd` `shell` option is now necessary.

Using `shell: true` brings an actual shell which introduces a risk of shell injection. This means that the arguments should be sanitized before passing. In our case those that require to be sanitized are strings not validated by `yargs` itself: paths and locale. For these values there are sanitize functions restricting input to the set of allowed characters, which significantly limits the possible risk (but also reduces characters that can be used in regular paths).

To limit the risk exposure `shell: true` is used only on windows platform.

Other options considered:
- using `execFile` instead of spawn - `execFile` doesn't spawn a shell which makes it safer, however it behaves differently and doesn't handle long running processes and user interaction like `spawn` does. `eject` use case is fine with `execFile`, but `preview` and `translate` won't work in this case
- escaping arguments - it would be fine if we were using single platform/shell, but in this case we need to support both unix and windows. Within windows we also have multiple shells (cmd, powershell, git-bash) each having different way of escaping commands. I didn't find any library that could cover all these cases and writing it ourselves would require shell detection on windows which might be hard to do (and not always work)
- checking if path exists - checking a path with `fs.existsSync` wouldn't prevent injections since a malformed path might be created (there are no limitations in characters)
- validating path format with regex - that still would need to include the whitelisted character set so it wouldn't bring additional protection

## Reference
https://github.com/Redocly/redocly/issues/11284

## Testing
Tested on Windows and MacOS with Node versions from before and after the change.

## Check yourself

## Security


.changeset/lemon-cougars-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
```
mallachari, line 2: Should it include `@redocly/openapi-core` bump too?

mallachari:
Converting back to draft as I'd like to add some input sanitization

.changeset/lemon-cougars-cheat.md
tatomyr: No as all the changes are scoped to the `cli` package.

packages/cli/src/commands/eject.ts
```diff
@@ -12,18 +13,34 @@ export type EjectOptions = {
...
-  const npxExecutableName = process.platform === 'win32' ? 'npx.cmd' : 'npx';
-  spawn(
+  const isWindowsPlatform = process.platform === 'win32';
+
+  const npxExecutableName = isWindowsPlatform ? 'npx.cmd' : 'npx';
+  const path = isWindowsPlatform ? sanitizePath(argv.path ?? '') : argv.path ?? '';
+  const projectDir =
+    isWindowsPlatform && argv['project-dir']
+      ? sanitizePath(argv['project-dir'])
+      : argv['project-dir'];
```
tatomyr, line 23: I find it hard to read. I wonder if it would be actually simpler to split this command execution for different platforms using `if`:

```ts
if (isWindowsPlatform) {
  ...
} else {
  ...
}
```

Also, please check if we can abstract the code for all the Realm wrapper commands (perhaps per platform), as I believe they might share a lot in common (just to avoid duplication/divergence).

tatomyr (commented):
Left one comment. 
Regarding the test failure -- the coverage just doesn't meet the threshold. Either add more tests or lower the threshold.

packages/cli/src/utils/platform.ts
```diff
@@ -0,0 +1,26 @@
+export const sanitizePath = (input: string): string => {
+  return input.replace(/[^a-zA-Z0-9 ._\-:\\/@]/g, '');
+};
+
+export const sanitizeLocale = (input: string): string => {
+  return input.replace(/[^a-zA-Z0-9@._-]/g, '');
+};
+
+export function getPlatformArgs(argv: { path?: string; locale?: string; 'project-dir'?: string }) {
```
tatomyr, line 9: These properties are specific to individual commands, whilst utils are meant to be context-agnostic. Could you move them out to make the getter more general? I’d recommend returning sanitizers instead, so they can be used directly in the command handlers.

packages/cli/src/commands/translations.ts
```diff
@@ -10,10 +11,19 @@ export type TranslationsOptions = {
...
   process.stdout.write(`\nLaunching translate using NPX.\n\n`);
-  const npxExecutableName = process.platform === 'win32' ? 'npx.cmd' : 'npx';
-  spawn(
+  const { npxExecutableName, projectDir, locale, shell } = getPlatformArgs(argv);
+
+  const child = spawn(
     npxExecutableName,
-    ['-y', '@redocly/realm', 'translate', argv.locale, `-d=${argv['project-dir']}`],
-    { stdio: 'inherit' }
+    ['-y', '@redocly/realm', 'translate', locale ?? '', `-d=${projectDir}`],
```
tatomyr, line 18: Could you move the `?? ''` into the sanitizer itself? For me it looks like it makes sense to transform all empty arguments into empty spaces.

packages/cli/src/utils/platform.ts
```diff
@@ -0,0 +1,26 @@
+export const sanitizePath = (input: string): string => {
+  return input.replace(/[^a-zA-Z0-9 ._\-:\\/@]/g, '');
+};
+
+export const sanitizeLocale = (input: string): string => {
```
tatomyr, line 5: Could you leave a comment with an explanation regarding what is being sanitized? Later, it could be hard to figure out the difference between the sanitizers 😄
Or perhaps you could come up with more descriptive names?

packages/cli/src/utils/platform.ts
mallachari: I made it more agnostic by returning generic `sanitize` function.

tatomyr (approved):
LGTM, thanks for your input!

HCloward (approved):

PR #1834, opened 2024-12-19
feat: support custom message for built-in rules

tatomyr:
## What/Why/How?

Added the ability to override default problem messages for built-in rules.

Configuration example (the original problem will replace `{{message}}`):

```yaml
struct:
  severity: error
  message: CODE-123. Invalid structure. {{message}}
```

## Reference

The PR is based on this one: https://github.com/Redocly/redocly-cli/pull/1821

## Check yourself

## Security


docs/configuration/reference/rules.md
```diff
@@ -38,6 +38,14 @@ The `rules` block can be used at the root of a configuration file, or inside an
 
 ---
 
+- message
+- string
+- Optional custom message for this rule.
+  Example: `My Error Derscription. {{message}}`.
+  The {{message}} placeholder (if provided) gets replaced with the actual message of the rule.
```
RomanHotsiy, line 45: I think this has to be rephrased. cc @Redocly/technical-writers

RomanHotsiy (approved):
LGTM, left a comment on docs.

docs/configuration/reference/rules.md
```diff
@@ -38,6 +38,14 @@ The `rules` block can be used at the root of a configuration file, or inside an
 
 ---
 
+- message
+- string
+- Optional custom message for this rule.
+  Example: `My Error Derscription. {{message}}`.
```
HCloward, line 44:
```suggestion
  Example: `My Error Description. {{message}}`.
```

HCloward: I'm not sure I understand. Is the custom message what goes before {{message}} ? Or do users put the message in the brackets in place of "message"?

docs/configuration/reference/rules.md
HCloward: What is the "actual" message of the rule?

tatomyr: Each rule, when violated, produces an error message. That is what I called the "actual" one. The `message` field allow users to replace the "actual" message with a custom one.

docs/configuration/reference/rules.md
tatomyr: >  Is the custom message what goes before {{message}} ?

Yes. The `{{message}}` placeholder is to be replaced by the linter. BTW, users could just put something like `My Custom Message` there without using the `{{message}}` at all. Then, they won't be able to see the default (or the "actual") linter message.

docs/configuration/reference/rules.md
```diff
@@ -38,6 +38,14 @@ The `rules` block can be used at the root of a configuration file, or inside an
 
 ---
 
+- message
+- string
+- Optional custom message for this rule.
+  Example: `My Error Description. {{message}}`.
+  The {{message}} placeholder (if provided) gets replaced with the actual message of the rule.
```
HCloward, line 45:
```suggestion
The {{message}} placeholder renders with the default error message for the rule.  Include the {{message}} placeholder if you want to provide the user with your custom message as well as the default error message for the rule.
```

PR #1835, opened 2024-12-19
feat: add redocBundleURL option

dmosberger:
## What/Why/How?

It is helpful to be able render documentation with a non-standard redoc bundle, e.g., for testing.  In theory, one could use a custom HTML template for this purpose but in practice that's not straight-forward because the template would have to be different depending on whether or not `preview-docs` or `build-docs` is run. The template would also have to include all the CSS code.

## Testing

unit tests and e2e tests all passed.

### DIFF
--- a/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
+++ b/packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
@@ -10,7 +10,9 @@ import type { IncomingMessage } from 'http';
 
 function getPageHTML(
   htmlTemplate: string,
-  redocOptions: object = {},
+  redocOptions: {
+    redocBundleURL?: string;
+  } = {},
   useRedocPro: boolean,
   wsPort: number,
   host: string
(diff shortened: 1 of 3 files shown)


tatomyr:
Thank you @dmosberger! We'll take a look at this soon.

packages/cli/src/commands/preview-docs/preview-server/preview-server.ts
```diff
@@ -10,7 +10,9 @@ import type { IncomingMessage } from 'http';
 
 function getPageHTML(
   htmlTemplate: string,
-  redocOptions: object = {},
+  redocOptions: {
+    redocBundleURL?: string;
```
AlexVarchuk, line 14: We don't want to have inside config

AlexVarchuk (changes requested):
@dmosberger Thank you for your contribution. I left few comments.

Some time ago we deprecated this because we had too many problems where Redoc CE of version x didn't work in Redocly/cli version y. Not all versions of cli could support all versions of Redoc CE. If we want to add this option back, we want to mention that we guarantee that only the default version (installed in cli) will work as expected. 

I think we should write this in the docs or mention it somewhere else, otherwise it will be very hard to maintain.

tatomyr:

PR #1838, opened 2025-01-07
fix: consider the cli install directory when resolving plugins

volodymyr-rutskyi:
## What/Why/How?

There is an issue when running the `npx @redocly/cli preview` command when the `@redocly/realm` package is not installed in the project's `node_modules`: it's dependent package cannot be resolved.

This happens because the npx installs both the `@redocly/realm` and it's dependency `@redocly/openapi-core` into a cache directory (`~/.npm/_npx` for mac and linux) but when resolving plugins, only the project directory is considered.

This PR adds the package location to the list of directories to resolve the Realm plugins from.

## Check yourself

## Security

### DIFF
--- a/.changeset/tender-forks-argue.md
+++ b/.changeset/tender-forks-argue.md
@@ -3,4 +3,5 @@
 "@redocly/cli": patch
 ---
 
-Fixed a bug where running a preview command when Realm is not in the `node_modules` of the project failed because one of it's dependencies could not be resolved.
+Fixed an issue where running the `preview` command failed because one of its dependencies could not be resolved.
+The issue occurred when Realm was not installed in the `node_modules` of the project.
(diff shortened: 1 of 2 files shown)


tatomyr (approved):
Please leave a comment to explain why it was added.

.changeset/tender-forks-argue.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed a bug where running a preview command when Realm is not in the `node_modules` of the project failed because one of it's dependencies could not be resolved.
```
JLekawa, line 6:
```suggestion
Fixed an issue where running the `preview` command failed because one of its dependencies could not be resolved.
The issue occurred when Realm was not installed in the `node_modules` of the project.
```

PR #1852, opened 2025-01-21
chore(dockerfile): optimize image size

weisdd:
## What/Why/How?

The current structure of the Dockerfile results in a bloated image, because files get cleaned up not in the same layers where they were created (=> they do not disappear from the end image, just get "hidden" by OverlayFS).
By slightly rearranging and merging RUN instructions + removing apk cache, we can reduce the image size from ~523MB to ~248MB.


weisdd:
@tatomyr not sure whom I need to ping to have the PR reviewed, so would be great if you could point at the right person.
Thanks!

tatomyr:
Closing this as I believe it's outdated after the last release.

PR #1876, opened 2025-01-31
feat(rules): update `operation-tag-defined` rule

jeremyfiel:
adds a check for `tags` existence in `Operation` node

## Check yourself

## Security


.changeset/tasty-cougars-dress.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Updated `operation-tag-defined` configurable rule.
```
HCloward, line 6: what was was updated about the rule and why?

jeremyfiel: The rule didn't check for a `tags` definition at the operation level.  I also added the tests for it.

packages/core/src/rules/common/operation-tag-defined.ts
```diff
@@ -11,6 +11,13 @@ export const OperationTagDefined: Oas3Rule | Oas2Rule = () => {
       definedTags = new Set((root.tags ?? []).map((t) => t.name));
     },
     Operation(operation: Oas2Operation | Oas3Operation, { report, location }: UserContext) {
+      if (!operation?.hasOwnProperty('tags')) {
+        report({
+          message: `Operation tags should be defined`,
+          location: location.child('tags').key(),
```
tatomyr, line 17: If there's no `operation.tags`, then the location couldn't be `tags`, I believe. Would the following work?

```suggestion
          location: location.key(),
```

packages/core/src/rules/common/operation-tag-defined.ts
```diff
@@ -11,6 +11,13 @@ export const OperationTagDefined: Oas3Rule | Oas2Rule = () => {
...
+        });
+        return;
+      }
       if (operation.tags) {
```
tatomyr, line 21: Wouldn't it be better to add the `else` clause to this `if` and put the changes there? 

```suggestion
      if (operation?.tags) {
```

.changeset/tasty-cougars-dress.md
tatomyr: What was the issue you're trying to solve @jeremyfiel? Was the linter failing? Or just not reporting the problem?

packages/core/src/rules/common/operation-tag-defined.ts
jeremyfiel: let's try it. .the snapshot seemed to work ok, but i'll try updating it. your comment makes sense

packages/core/src/rules/common/operation-tag-defined.ts
jeremyfiel: i put the check first to fail fast.  the entire rule can be refactored to check for tags, then check for global tags without having two if statements.  Let me know if you want me to do that.

jeremyfiel: updated per your suggestion

.changeset/tasty-cougars-dress.md
HCloward, line 6:
```suggestion
Updated the `operation-tag-defined` built-in rule to check that the globally defined tags are used at the operation level.
```

HCloward: I was only able to find the built-in rule `operation-tag-defined`. I also think it would be helpful to add some information to the rule's documentation with this update - https://redocly.com/docs/cli/rules/oas/operation-tag-defined

HCloward (approved):

tatomyr (approved):
Well done, thank you @jeremyfiel!

PR #1878, opened 2025-02-03
feat: add login via Reunite API

tatomyr:
## What/Why/How?

- Added the ability to log in via Reunite API with `--next` flag
- Renamed folder `cms/` to `reunite/` for clarity
- Renamed the `--region` option to `--residency` as it may take a full domain alongside `us` and `eu` (but retained `--region` alias for backward compatibility)

## Reference

Internal request: https://github.com/Redocly/openapi-vs-code/issues/231#issuecomment-2598584274

## Screenshots (optional)

Old login:

image

New login:
image

## Check yourself

## Security


DmitryAnansky:
Should we update some documentation?

packages/cli/src/commands/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+import { RedoclyOAuthClient } from '../auth/oauth-client';
+import { getReuniteUrl } from '../reunite/api';
+
+import type { CommandArgs } from '../wrapper';
+import type { Region } from '@redocly/openapi-core';
+
+export function promptClientToken(domain: string) {
+  return promptUser(
+    green(
+      `\n  🔑 Copy your API key from ${blue(`https://app.${domain}/profile`)} and paste it below`
```
DmitryAnansky, line 13: It looks to me that for now this message is misleading.
We don't apply new `isAuthorized()` function anywhere yet, and this will not have any effect on previously protected code.
Please correct me if I am wrong.

tatomyr: We still support the old login which is needed to render Reference Docs with the `preview-docs` command. At the same time we need the `isAuthorized` method to use it in places that will require auth (probably in the VS Code extension).

DmitryAnansky: Yes, but will `preview-docs` work if you run new login ?

DmitryAnansky: Maybe we can change the message text.

tatomyr: > will preview-docs work if you run new login ?

No, I doubt it makes sense since Reference Docs are only related to our previous product. And also, it won't allow to `push` to the Redocly registry. 

@RomanHotsiy could you check if the message is OK or could be improved?

docs/commands/login.md
```diff
@@ -22,15 +22,24 @@ redocly login [--help] [--verbose] [--version]
 redocly login --verbose
 ```
 
+To authenticate using **Reunite** API, use `--next` option.
```
JLekawa, line 25:
```suggestion
To authenticate using **Reunite** API, use the `--next` option.
```

docs/commands/login.md
```diff
@@ -22,15 +22,24 @@ redocly login [--help] [--verbose] [--version]
...
+
+```bash
+redocly login --next
+```
+
+Please note that login with **Reunite** API does not allow you to use the `push` command.
```
JLekawa, line 31:
```suggestion
Note that logging in with **Reunite** API does not allow you to use the `push` command.
```

packages/cli/src/commands/auth.ts
```diff
@@ -0,0 +1,66 @@
...
+  verbose?: boolean;
+  region?: string;
+  config?: string;
+  next?: boolean;
+};
+
+export async function handleLogin({ argv, config, version }: CommandArgs<LoginOptions>) {
+  if (argv.next) {
+    try {
+      const reuniteUrl = getReuniteUrl(argv.region);
```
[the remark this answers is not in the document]
tatomyr: Users can provide any string as an input. We just transform known regions into domains. I think it's just a wrong name. I'll rename it to `residency` instead (and retain `region` as an alias).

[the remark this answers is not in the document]
tatomyr: Updated.

packages/cli/src/auth/device-flow.ts
```diff
@@ -0,0 +1,175 @@
...
+      const cmd =
+        process.platform === 'win32'
+          ? `start ${url}`
+          : process.platform === 'darwin'
+          ? `open ${url}`
+          : `xdg-open ${url}`;
+
+      childProcess.execSync(cmd);
+    } catch {
+      // silently fail if browser cannot be opened
```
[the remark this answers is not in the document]
roman-sainchuk: Ah, missed that, thanks! No need for additional message

roman-sainchuk (approved):

packages/cli/src/reunite/api/__tests__/domains.test.ts
```diff
@@ -0,0 +1,37 @@
+import { getDomain } from '../domains';
+import { getReuniteUrl } from '../domains';
+
+import type { Region } from '@redocly/openapi-core';
+
+describe('getDomain()', () => {
+  it('should return the domain from environment variable', () => {
+    process.env.REDOCLY_DOMAIN = 'test-domain';
```
DmitryAnansky, line 8: Do you need to clean it/ delete after test finished.

tatomyr: Good point. Updating it.

packages/cli/src/commands/auth.ts
RomanHotsiy: This is fine for now.

RomanHotsiy (approved):

Issue #1871, opened 2025-01-30
Merge-resolve per-API with root configuration not fully supported


mscheerschmidt-mhp:
**Describe the bug**

Parameters of custom plugins (decorators) are not being merged with parameters from root configuration.

**To Reproduce**

1. Given a `redocly.yaml` file
2. defining a custom decorator
3. which has two options _tag_ and _version_
4. when the root decorator config defines values for both
5. when configuring per-API with version=v2, leaving tag undefined
6. then param **tag remains undefined** in the decorator execution

```
extends:
    - recommended
plugins:
    - plugins/mine.js
decorators:
    mine/add-api-id:
        tag: x-mine-api-id
        version: v1
theme:
    openapi:
        showExtensions: [ x-mine-api-id ]
apis:
    mine@v2:
        root: ./mine_v2.yaml
        decorators:
            mine/add-api-id:
                version: v2
```
and
```
function OperationApiId({tag, version}) {
    console.log(`Decorating with tag ${tag} and version ${version}...`);
}
```
yields: `Decorating with tag undefined and version v2...`

Defining another api element which doesn't mention the plugin makes use of the root configuration.

**Expected behavior**

It should yield: `Decorating with tag x-mine-api-id and version v2...`.

Configured parameters should be merge-resolved just as the docs say:

> If per-API and root settings modify the same properties, per-API settings overrides root settings.

See https://redocly.com/docs/cli/configuration/apis

**Redocly Version(s)**

CLI 1.28.0

tatomyr:
@mscheerschmidt-mhp it's not a bug, rather some misunderstanding. By 'settings' we mean the decorators or rules themselves, not their parameters. For example, if you define a decorator with the same name (but different parameters) on the root and the api levels, the api-level one will override the root one (so only the api-level parameters will be used). I'm not sure how to better clarify that in the docs though. @Redocly/technical-writers could you suggest better wording?

mscheerschmidt-mhp:
Thanks for your quick reply. Would you maybe consider adding this as an enhancement?
The benefit would be that one can define defaults on the root level for decorators etc.
Within per-API definitions one only has to set what needs to be overridden.

tatomyr:
@mscheerschmidt-mhp I'm reluctant to any kind of implicit inheritance like that. It's easy to get it wrong with that and frankly speaking brings no benefits. Consider an opposite situation when you have two decorators on the root and apis level, and you actually want to completely replace the parameters:

```yaml

[...]

PR #1882, opened 2025-02-04
docs(all): clarifies what is overridden on a per-api basis

HCloward:
## What/Why/How?
There is some confusion with users about how per-API rules, decorators, and preprocessors merge with root settings. When there are per-API rules, decorators, and preprocessors, we merge with the top-level root properties. This merging does not include parameters specific to the rules, decorators, or preprocessors.

## Reference
Relates to #1871 

## Security


JLekawa (approved):

docs/configuration/apis.md
```diff
@@ -8,8 +8,8 @@ Every `name@version` combination listed in the object must be unique.
 For every API listed in the object, you must provide the path to the OpenAPI description using the `root` property.
 
 If `rules`, `decorators`, or `preprocessors` aren't defined for an API, root settings are used.
-If `rules`, `decorators`, or `preprocessors` are defined for an API, its settings apply together with the root configuration.
-If per-API and root settings modify the same properties, per-API settings overrides root settings.
+If `rules`, `decorators`, or `preprocessors` are defined for an API, its first-level properties apply together with the root configuration.
```
tatomyr, line 11: It's not first-level properties of rules &c., but rather rules, decorators, and preprocessors are the first-level properties of a per-api configuration. 

For instance, let's define this in **redocly.yaml**:

```yaml
decorators:
  decorator-one: # <-- first level `decorators` property is a decorator (they get merged)
     prop1: 1 # <-- decorator's properties are something we don't merge but replace
     prop2: 2
  decorator-two:
     prop3: 3
     prop4: 4

apis:
   override@v1:
      decorators:
         decorator-one:
             prop: 5
```

When applied to `override@v1`, that results in the following config:

```yaml
decorators:
  decorator-one: # <-- got replaced entirely (including its props)
             prop: 5
  decorator-two: # <-- remains unchanged as it gets merged with the per-api decorators
     prop3: 3
     prop4: 4
```

The same goes for `theme` first-level properties (which is deprecated already, the they should become just root-level properties).

So it's a bit trickier than that 😅.

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
 For every API listed in the object, you must provide the path to the OpenAPI description using the `root` property.
 
 If `rules`, `decorators`, or `preprocessors` aren't defined for an API, root settings are used.
-If `rules`, `decorators`, or `preprocessors` are defined for an API, its first-level properties apply together with the root configuration.
-If per-API `rules`, `decorators`, or `preprocessors` and root settings modify the same first-level properties, per-API `rules`, `decorators`, and `preprocessors` override root settings.
+If `rules`, `decorators`, or `preprocessors` are defined for an API, they apply together with the root configuration.
+If per-API `rules`, `decorators`, or `preprocessors` and root settings modify the same properties, per-API `rules`, `decorators`, and `preprocessors` override root settings.
+
+For example, if you include the same `decorator` at the root level and for a specific API, but with different properties, only the properties applied to the specific API will be applied to that API.
```
tatomyr, line 14: > only the properties applied to the specific API will be applied to that API.

Seems like a tautology to me :). 

```suggestion
For example, if you include the same `decorator` at the root level and for a specific API, but with different properties, the API-level settings replace the root ones.
```

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
...
+So if you have the following `redocly.yaml` configuration, adding `decorator-one` and `decorator-two` at the root level and applying `decorator-one` to the `override@v1` API:
+
+```yaml
+decorators:
+  decorator-one:
+    property-one: 1
+    property-two: 2
+  decorator-two:
+    property-three: 3
+    property-four: 4
```
tatomyr, line 25: I don't really like that `decorator-one` and `property-one` namings. Can we use something closer to the real-world scenario? Probably we can borrow something [here](https://github.com/Rebilly/api-definitions/blob/main/redocly.yaml).

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
...
+
+apis:
+  override@v1:
+    decorators:
+      decorator-one:
+        property-five: 5
+```
+
+Only `property-five` is applied to `decorator-one` for the `override@v2` API, and both `property-three` and `property-four` are applied for `decorator-two`.
```
tatomyr, line 34: Did you mean `@v1`?

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
 For every API listed in the object, you must provide the path to the OpenAPI description using the `root` property.
 
 If `rules`, `decorators`, or `preprocessors` aren't defined for an API, root settings are used.
-If `rules`, `decorators`, or `preprocessors` are defined for an API, its settings apply together with the root configuration.
-If per-API and root settings modify the same properties, per-API settings overrides root settings.
+If `rules`, `decorators`, or `preprocessors` are defined for an API, they apply together with the root configuration.
+If per-API `rules`, `decorators`, or `preprocessors` and root settings modify the same properties, per-API `rules`, `decorators`, and `preprocessors` override root settings.
```
tatomyr, line 12: It's not too obvious what properties/settings are in that context. Could it be rephrased like this?

```suggestion
If the same `rules`, `decorators`, or `preprocessors` are defined on `apis` and root level, per-API `rules`, `decorators`, and `preprocessors` override root ones.
```

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
...
+So if you have the following `redocly.yaml` configuration, adding `filter-in` and `plugin/change-title` at the root level and applying `plugin/change-title` with a different `title` property to the `storefront@latest` API:
+
+```yaml
+decorators:
+  filter-in:
+    property: x-products
+    value:
+      - Core
+  plugin/change-title:
+    title: All APIs
```
tatomyr, line 25: I'd add some other parameter here, so it's more clear that it will be overriden:

```suggestion
    title: All APIs
    extraProperty: will be ignored on per-API level
```

docs/configuration/apis.md
```diff
@@ -8,8 +8,33 @@ Every `name@version` combination listed in the object must be unique.
...
+apis:
+  storefront@latest:
+    decorators:
+      plugin/change-title:
+        title: Storefront APIs
+```
+
+The `plugin/change-title` decorator with the "Storefront APIs" `title` property is applied to the `storefront@latest` API, and the `filter-in` decorator is also applied to the `storefront@latest` API.
+
+For all other APIs, not including the `storefront@latest` API, `filter-in` and `plugin/change-title` with the "Core" `title` property are applied.
```
tatomyr, line 37:
```suggestion
The `plugin/change-title` decorator with the "Storefront APIs" `title` property is applied to the `storefront@latest` API with value `Storefront APIs`, and the `filter-in` decorator is also applied to the `storefront@latest` API.

For all other APIs, not including the `storefront@latest` API, `filter-in` and `plugin/change-title` with the "Core" `title` and `extraProperty` properties are applied.
```

tatomyr (commented):
Almost there. Left a couple of comments.

tatomyr (approved):
Thank you!

PR #1894, opened 2025-02-07
chore: bump styled-components version to 6.1.15

jeylost:
## What/Why/How?

This PR aims to address [CVE-2024-55565](https://nvd.nist.gov/vuln/detail/CVE-2024-55565).  
`styled-components` has released version [6.1.15](https://github.com/styled-components/styled-components/releases/tag/styled-components%406.1.15) with the fix.

## Check yourself

## Security


jeylost:
Should the bumping of a package be included in the package changeset?

tatomyr:
@jeylost no, it's not needed I think. I'll ignore that. 
However, the tests are failing due to snapshot size changes (which is expected, but we've just released a broken version because of assuming it's expected, so it's better to test it separately). Let's check that case offline.

PR #1943, opened 2025-02-21
feat: changes to success criteria checks display

DmitryAnansky:
## What/Why/How?

Add `condition` information to each SUCCESS_CRITERIA_CHECK display output.

## Screenshots (optional)

Screenshot 2025-02-21 at 18 59 27

## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/modules/cli-output/display-checks.ts
+++ b/packages/respect-core/src/modules/cli-output/display-checks.ts
@@ -9,6 +9,7 @@ import type { RuleSeverity } from '@redocly/openapi-core/lib/config/types';
 import type { Check, VerboseLog } from '../../types';
 
 const logger = DefaultLogger.getInstance();
+const MAX_CRITERIA_CONDITION_DISPLAY_LENGTH = 50;
 
 export function displayChecks(
   testNameToDisplay: string,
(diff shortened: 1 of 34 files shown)


packages/respect-core/src/types.ts
```diff
@@ -271,6 +271,7 @@ export type Check = {
   name: string;
   message?: string;
   additionalMessage?: string;
```
DmitryAnansky, line 273: We might also reuse `additionalMessage`, but imho having `criteriaCondition` is better describes the concept and will be more useful in json logs.

packages/respect-core/src/modules/cli-output/display-checks.ts
```diff
@@ -35,13 +36,25 @@ export function displayChecks(
...
 
     const failTestMessage = (checkName: string, severity?: RuleSeverity) =>
       `${severity === 'warn' ? yellow('⚠') : red('✗')} ${gray(checkName.toLowerCase())}${
-        check?.additionalMessage ? ' (' + check.additionalMessage + ')' : ''
-      }`;
+        check?.criteriaCondition
+          ? ` - ${red(
+              check.criteriaCondition.length > MAX_CRITERIA_CONDITION_DISPLAY_LENGTH
+                ? check.criteriaCondition.slice(0, MAX_CRITERIA_CONDITION_DISPLAY_LENGTH) + '...'
+                : check.criteriaCondition
```
tatomyr, line 54: Could you move the common logic to a utility function?

packages/respect-core/src/types.ts
tatomyr: For me it looks like `additionalMessage` is a better option since it serves the same exact purpose for checks derived from OpenAPI description in `checkStatusCodeFromDescription`.

DmitryAnansky: Not really, schema check or status code check or content type check does not have `criteriaCondition`.
And display format is different.

tatomyr: But they both are for explaining the additional details, just in different cases (`additionalMessage` for the cases you've mentioned and `criteriaCodition` for success criterias. The format shouldn't be different either, I believe.

PR #1945, opened 2025-02-24
chore: implement respect_max_steps

RomanHotsiy:
## What/Why/How?

Introduce MAX_STEPS env variable.

## Security


packages/respect-core/src/modules/__tests__/flow-runner/run-step.test.ts
```diff
@@ -1790,7 +1790,7 @@ describe('runStep', () => {
     });
 
     expect(displayChecks).toHaveBeenCalled();
-    expect(checkCriteria).toHaveBeenCalledTimes(5);
+    expect(checkCriteria).toHaveBeenCalledTimes(3);
```
DmitryAnansky, line 1793: Lets investigate this call times.

DmitryAnansky: It looks like it was actually broken in some prev commit in source branch

PR #1969, opened 2025-03-01
feat: add reference property to context.report() for external documentation

### LINKED ISSUE #1052: context.report(): add new property for external documentation
As an API governance architect, we have a number of design decisions enforced with Redocly rules. While we try to make the rules self-documenting, there are definitely some instances which require a deeper context on the decision.

I would like a property to include an external link to documentation supporting the design decision of a particular rule so a user of the linter can find reference material for the rule.

```
{
    "message": <string>,
    "location": Location{},
    "suggest": <string>[],
    "from": Location{},
    "reference": <string><uri-reference>
}

```

I'm currently trying out `\n\t` formatting in the message body but I would like something more formal. 

```
report({
      message: `this is a message with a reference to external docs\n\treference: https://www.redocly.com`
      })
```

```sh
  1:1  warning  custom-id/some-rule  this is a message with a reference to external docs
        reference: https://www.redocly.com
```

ImBIOS:
## What/Why/How?

Add the `reference` property to the `context.report()` method for external documentation links.

* **Documentation**: Update `docs/custom-plugins/custom-rules.md` to include the `reference` property in the `context.report()` method documentation.
* **New Rule**: Add `packages/core/src/rules/arazzo/reference-property.ts` to implement a new rule for validating the `reference` property in the report object.
* **Rule Integration**: Modify `packages/core/src/rules/arazzo/index.ts` to import and add the new `reference-property` rule to the existing rules.
* **Unit Tests**: Add `packages/core/src/rules/arazzo/__tests__/reference-property.test.ts` to test the presence and format validation of the `reference` property and error reporting for missing or incorrect `reference` property.

## Reference

Fix #1052

## Testing

ran npm run test

## Check yourself

## Security


packages/core/src/rules/arazzo/__tests__/reference-property.test.ts
```diff
@@ -0,0 +1,106 @@
...
+              name: Authorization
+              value: Basic Og==
+          steps:
+            - stepId: get-museum-hours
+              description: >-
+                Get museum hours by resolving request details with getMuseumHours operationId from openapi.yaml description.
+              operationId: museum-api.getMuseumHours
+              successCriteria:
+                - condition: $statusCode == 200
+                  reference: invalid-url
```
DmitryAnansky, line 32: This is not according to the Arazzo specification and I would prefer not to add additional properties to Criterion Object. In case we need one - this can be extension, before accepted and added to the Spec itself.

DmitryAnansky: In fact `reference` field already exist in Arazzo, but used for different type `Reusable Object`.

packages/core/src/__tests__/utils.test.ts
```diff
@@ -101,44 +102,133 @@ describe('utils', () => {
...
-          if (path.endsWith('.yaml')) {
-            return '.yaml';
-          } else if (path.endsWith('.yml')) {
-            return '.yml';
-          } else {
-            return '';
-          }
-        });
-      });
+  describe('isValidURL', () => {
```
RomanHotsiy, line 107: I don't think we need so many tests for this. I would also not validate this at all. I would add a very basic URL validation.

packages/core/src/rules/arazzo/__tests__/reference-property.test.ts
RomanHotsiy: This test is completely wrong.

RomanHotsiy (changes requested):
This PR is an AI hallucination. It doesn't make any sense.
The markdown file change looks valid but the code changes are completely irrelevant and wrong.

packages/core/src/__tests__/utils.test.ts
ImBIOS: you're right, URL constructor is also more performant and reliable.

ImBIOS:
I'm sorry. I misunderstood the issue. It should not be a new rule; it should just be an additional string in the reporter.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -156,6 +156,7 @@ const builtInArazzo1Rules = [
   'requestBody-replacements-unique',
   'no-criteria-xpath',
   'criteria-unique',
+  'reference-property',
```
RomanHotsiy, line 159: what is this?

RomanHotsiy (commented):
I think we may want to rename the `reference` to something else. Reference may be too generic.

I am thinking about:

- `referenceUrl`
- `externalDocsUrl`
- 
@jeremyfiel do you have some other options?

jeremyfiel:
Consistent with OpenAPI and Asyncapi would be `externalDocs`

They have an object definition of:  `description` and `url`

  I don't mind only having the url. 

Either of these works for me. 

`externalDocs` 
`externalDocsUrl`

jeremyfiel:
@ImBIOS   if you want to update this PR, the change should be on the `UserContext > Problem` type with the field name agreed by Redocly team.

https://github.com/Redocly/redocly-cli/blob/59c7b8e5e565fc5ae4e2284c821563ca261a9fae/packages/core/src/walk.ts#L74

tatomyr:

PR #1980, opened 2025-03-04
fix: validate nullable oas3 example correctly

tatomyr:
## What/Why/How?

Fixed the `no-invalid-schema-examples` rule validating nullable OpenAPI 3.0 schemas incorrectly.
Added the check for `null` example in nullable schemas (only valid for OAS3).

## Reference

Related issue: https://github.com/Redocly/redocly-vs-code/issues/42

## Testing

Tested in VS Code extension as well.

## Check yourself

## Security

### DIFF
--- a/.changeset/rich-falcons-act.md
+++ b/.changeset/rich-falcons-act.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Fixed the `no-invalid-schema-examples` rule validating nullable OpenAPI 3.0 schemas incorrectly.
+Fixed the `no-invalid-schema-examples` rule that incorrectly validated nullable OpenAPI 3.0 schemas.
--- a/.changeset/rich-falcons-act.md
+++ b/.changeset/rich-falcons-act.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed the `no-invalid-schema-examples` rule validating nullable OpenAPI 3.0 schemas incorrectly.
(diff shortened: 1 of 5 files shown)


packages/core/src/rules/common/no-invalid-schema-examples.ts
```diff
@@ -1,25 +1,37 @@
 import { getAdditionalPropertiesOption, validateExample } from '../utils';
 
 import type { UserContext } from '../../walk';
-import type { Oas3_1Schema } from '../../typings/openapi';
+import type { Oas3_1Schema, Oas3Schema } from '../../typings/openapi';
+import type { Oas2Rule, Oas3Rule } from '../../visitors';
 
-export const NoInvalidSchemaExamples: any = (opts: any) => {
+export const NoInvalidSchemaExamples: Oas3Rule | Oas2Rule = (opts: any) => {
```
DmitryAnansky, line 7: Should we rename common to common OAS rules?

DmitryAnansky (approved):

packages/core/src/rules/common/no-invalid-schema-examples.ts
tatomyr: Probably yes. But there are some rules that are common for all specs (e.g., related to the info section). I have to think through that.

.changeset/rich-falcons-act.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed the `no-invalid-schema-examples` rule validating nullable OpenAPI 3.0 schemas incorrectly.
```
JLekawa, line 5:
```suggestion
Fixed the `no-invalid-schema-examples` rule that incorrectly validated nullable OpenAPI 3.0 schemas.
```

JLekawa (approved):

RomanHotsiy (approved):

PR #2001, opened 2025-03-11
docs(respect): add respect commands to cli commands sidebar

DmitryAnansky:
## What/Why/How?

Add respect commands to cli commands sidebar

## Check yourself

## Security

### DIFF
--- a/docs/sidebars.yaml
+++ b/docs/sidebars.yaml
@@ -42,7 +42,6 @@
           page: commands/stats.md
         - label: translate
           page: commands/translate.md
-
     - group: Guides
       page: guides/index.md
       items:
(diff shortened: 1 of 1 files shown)


docs/sidebars.yaml
```diff
@@ -38,6 +38,11 @@
           page: commands/stats.md
         - label: translate
           page: commands/translate.md
+        - label: generate-arazzo
```
DmitryAnansky, line 41: HI @adamaltman , we noticed that respect command are not visible in the same section as other CLI commands.
Maybe this was intentional, could you please take a look?

1AnnaWi (approved):

docs/sidebars.yaml
adamaltman: Hi this was not intentional (it was accidentally made in the monorepo), but this PR isn't good because the commands should be listed in alphabetical order -- so this PR needs to be modified to fix that.

DmitryAnansky: Thx, commands are sorted alphabetically by label in sidebar now.

PR #2003, opened 2025-03-11
feat(respect): introduce global execution timeout-timer

DmitryAnansky:
## What/Why/How?

I would like to introduce global execution timeout timer, that will help to prevent infinity execution of some Arazzo files that can be looped with allowed `goto` and `workflowId` syntax.

### Changes in the Flow

- We will start Timer before we walk through the list of all Arazzo files we need to execute within this Respect run session
- Calculation decision is made on difference of RESPECT_TIMEOUT and elapsedTime

```
    const elapsedTime = Date.now() - this.startTime;
    return Math.max(0, Number(RESPECT_TIMEOUT) - elapsedTime);
```  
- There is one actual place where timer check done, all other places in code operate on its result: 
  -- on `Step` level. In this case we stop other steps execution and report error
  ```
    if (Timer.getInstance().isTimedOut()) {
    step.checks.push({
      name: CHECKS.GLOBAL_TIMEOUT_ERROR,
      message: `Global Respect timer reached`,
      passed: false,
      severity: ctx.severity['GLOBAL_TIMEOUT_ERROR'],
    });
    return { shouldEnd: true };
  }
  ```

### Command example 

```
 RESPECT_TIMEOUT=2500 redocly respect resources/museum-api.arazzo.yaml resources/museum-tickets.arazzo.yaml --json-output="out.json" --verbose
 ```

### Json Structure
This will affect `json-output` structure by adding additional property `globalTimeoutError` [boolean] on two levels:
- per file log level, that indicate that some timeout was hit or not during  that file execution;
- per workflow log level, that indicate that timeout was hit or not in workflow during its steps run.

Here is a simplified of the json-output:
```
{
  "files": {
    "resources/museum-api.arazzo.yaml": {
      "totalRequests": 7,
      "executedWorkflows": [
          { 
           "globalTimeoutError": false 
           ...
          }
       ],
      "totalTimeMs": 2524.251959,
      "globalTimeoutError": false
    },
    "resources/museum-tickets.arazzo.yaml": {
      "totalRequests": 1,
      "executedWorkflows": [
          { 
           "globalTimeoutError": false 
           ...
          }
      ],
      "totalTimeMs": 171.85491699999966,
      "globalTimeoutError": false
    }
  },
  "status": "success",
  "totalTime": 2697.196,
}
```

### Default timeout-timer value

Default timeout-timer value will be se to 1 hour, and can be changed by .env `RESPECT_TIMEOUT`

### New Checks types

New check error types were introduced to differentiate from `UNEXPECTED_ERROR`.
Those severity can't be changed and will be `error` by default.

```
  GLOBAL_TIMEOUT_ERROR: 'global timeout error',
  MAX_STEPS_REACHED_ERROR: 'maximum steps reached',
``` 

### Effect on the output

- As we can run in the condition when timeout timer hit on the start of workflow execution - the final table can contain failed file run record with 0 executed workflows.
- New error messages for failed steps

## Check yourself

## Security

### DIFF
--- a/.changeset/silent-suits-kiss.md
+++ b/.changeset/silent-suits-kiss.md
@@ -3,4 +3,4 @@
 "@redocly/cli": minor
 ---
 
-Add global execution timeout timer to `respect` command execution in order to prevent infinite test runs. The timer can be configured using the `RESPECT_TIMEOUT` environment variable (defaults to 1 hour).
+Added global execution timeout timer to `respect` command execution to prevent infinite test runs. You can configure this timer using the `RESPECT_TIMEOUT` environment variable (defaults to 1 hour).
--- a/.changeset/silent-suits-kiss.md
+++ b/.changeset/silent-suits-kiss.md
@@ -1,0 +1,6 @@
+---
+"@redocly/respect-core": minor
+"@redocly/cli": minor
+---
+
+Add global execution timeout timer to `respect` command execution in order to prevent infinite test runs. The timer can be configured using the `RESPECT_TIMEOUT` environment variable (defaults to 1 hour).
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,2 +1,2 @@
-export const RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
-export const MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
+export const DEFAULT_RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
+export const DEFAULT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,3 +1,3 @@
 export const DEFAULT_RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
-export const DEFAULT_RESPECT_MAX_FETCH_TIMEOUT = 60000; // 20 seconds in milliseconds
+export const DEFAULT_RESPECT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
 export const DEFAULT_RESPECT_MAX_STEPS = 2000;
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,2 +1,3 @@
 export const DEFAULT_RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
-export const DEFAULT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
+export const DEFAULT_RESPECT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
+export const DEFAULT_RESPECT_MAX_STEPS = 2000;
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,3 +1,3 @@
 export const DEFAULT_RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
-export const DEFAULT_RESPECT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
+export const DEFAULT_RESPECT_MAX_FETCH_TIMEOUT = 60000; // 20 seconds in milliseconds
 export const DEFAULT_RESPECT_MAX_STEPS = 2000;
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,2 +1,3 @@
-export const RESPECT_TIMEOUT = parseInt(process.env.RESPECT_TIMEOUT || '60 * 60 * 1000', 10); // 1 hour in milliseconds
+export const RESPECT_TIMEOUT = parseInt(process.env.RESPECT_TIMEOUT || '3600000', 10); // 1 hour in milliseconds
 export const MAX_STEPS = parseInt(process.env.RESPECT_MAX_STEPS || '2000', 10);
+export const MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,3 +1,2 @@
-export const RESPECT_TIMEOUT = parseInt(process.env.RESPECT_TIMEOUT || '3600000', 10); // 1 hour in milliseconds
-export const MAX_STEPS = parseInt(process.env.RESPECT_MAX_STEPS || '2000', 10);
+export const RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
 export const MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
--- a/packages/respect-core/src/consts.ts
+++ b/packages/respect-core/src/consts.ts
@@ -1,0 +1,2 @@
+export const RESPECT_TIMEOUT = parseInt(process.env.RESPECT_TIMEOUT || '60 * 60 * 1000', 10); // 1 hour in milliseconds
+export const MAX_STEPS = parseInt(process.env.RESPECT_MAX_STEPS || '2000', 10);
--- a/packages/respect-core/src/handlers/run.ts
+++ b/packages/respect-core/src/handlers/run.ts
@@ -12,8 +12,9 @@ import { DefaultLogger } from '../utils/logger/logger';
 import { exitWithError } from '../utils/exit-with-error';
 import { writeFileSync } from 'node:fs';
 import { indent } from '../utils/cli-outputs';
-import { Timer } from '../modules/timeout-timer';
-import { type JsonLogs, type CommandArgs, type RunArgv } from '../types';
+import { Timer } from '../modules/timeout-timer/timer';
+
+import type { JsonLogs, CommandArgs, RunArgv } from '../types';
 
 export type RespectOptions = {
   files: string[];
--- a/packages/respect-core/src/handlers/run.ts
+++ b/packages/respect-core/src/handlers/run.ts
@@ -12,8 +12,8 @@ import { DefaultLogger } from '../utils/logger/logger';
 import { exitWithError } from '../utils/exit-with-error';
 import { writeFileSync } from 'node:fs';
 import { indent } from '../utils/cli-outputs';
-
-import type { JsonLogs, CommandArgs, RunArgv } from '../types';
+import { Timer } from '../modules/timeout-timer';
+import { type JsonLogs, type CommandArgs, type RunArgv } from '../types';
 
 export type RespectOptions = {
   files: string[];
@@ -53,6 +53,7 @@ export async function handleRun({ argv, collectSpecData }: CommandArgs<RespectOp
   }
 
   try {
+    Timer.getInstance();
     const startedAt = performance.now();
     const testsRunProblemsStatus: boolean[] = [];
     const { files } = argv;
@@ -93,6 +94,7 @@ export async function handleRun({ argv, collectSpecData }: CommandArgs<RespectOp
             files: composeJsonLogsFiles(runAllFilesResult),
             status: hasProblems ? 'error' : hasWarnings ? 'warn' : 'success',
             totalTime: performance.now() - startedAt,
+            globalTimeoutError: Timer.getInstance().isTimedOut(),
           } as JsonLogs,
           null,
           2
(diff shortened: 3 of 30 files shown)

### CODE REVIEW

File: packages/respect-core/src/modules/checks/severity.ts [hunk 1]
```diff
@@ -12,6 +12,7 @@ export const DEFAULT_SEVERITY_CONFIGURATION: {
   CONTENT_TYPE_CHECK: 'error',
   UNEXPECTED_ERROR: 'error',
   NETWORK_ERROR: 'error',
+  GLOBAL_TIMEOUT_ERROR: 'error',
```
tatomyr, line 15: Do we maybe want to have a separate check for `MAX_STEPS` too? Or it's ok to have it under `UNEXPECTED_ERROR`? If so, maybe we don't have to introduce the separate `GLOBAL_TIMEOUT_ERROR`?

tatomyr (changes requested):
Left a couple of suggestions. Let's see if we can achieve the same goal in a simpler manner.

packages/respect-core/src/modules/checks/severity.ts
DmitryAnansky: I was thinking to introduce MAX_STEPS as well.
Its a draft version, I will do more changes here.

packages/respect-core/src/handlers/run.ts
```diff
@@ -93,6 +94,7 @@ export async function handleRun({ argv, collectSpecData }: CommandArgs<RespectOp
             files: composeJsonLogsFiles(runAllFilesResult),
             status: hasProblems ? 'error' : hasWarnings ? 'warn' : 'success',
             totalTime: performance.now() - startedAt,
+            globalTimeoutError: Timer.getInstance().isTimedOut(),
```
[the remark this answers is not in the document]
DmitryAnansky: Its better not to mix this two concepts.

packages/respect-core/src/handlers/run.ts
```diff
@@ -139,5 +141,6 @@ async function runFile(
     ctx,
     totalTimeMs: performance.now() - startedAt,
     totalRequests: totals.totalRequests,
+    globalTimeoutError: Timer.getInstance().isTimedOut(),
```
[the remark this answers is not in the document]
DmitryAnansky: I would prefer to isolate logic and not mix it, but I will check if it can be reused.

packages/respect-core/src/modules/checks/severity.ts
tatomyr: Please do as few changes as possible 😅.

packages/respect-core/src/handlers/run.ts
```diff
@@ -12,8 +12,8 @@ import { DefaultLogger } from '../utils/logger/logger';
 import { exitWithError } from '../utils/exit-with-error';
 import { writeFileSync } from 'node:fs';
 import { indent } from '../utils/cli-outputs';
-
-import type { JsonLogs, CommandArgs, RunArgv } from '../types';
+import { Timer } from '../modules/timeout-timer';
+import { type JsonLogs, type CommandArgs, type RunArgv } from '../types';
```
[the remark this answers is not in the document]
DmitryAnansky: Its not a big change.

packages/respect-core/src/modules/checks/severity.ts
DmitryAnansky: I still need to do those changes,
I will help you to resolve conflicts if there will be some.

packages/respect-core/src/consts.ts
```diff
@@ -0,0 +1,3 @@
+export const RESPECT_TIMEOUT = parseInt(process.env.RESPECT_TIMEOUT || '3600000', 10); // 1 hour in milliseconds
```
DmitryAnansky, line 1: I am wondering if we should count this in seconds and transform to ms already here.
This is probably tangled with usage in ci/cd.
cc: @RomanHotsiy , @malis42 , @sobanieca-redocly

packages/respect-core/src/modules/flow-runner/runner.ts
```diff
@@ -102,6 +103,9 @@ async function runWorkflows(testDescription: TestDescription, options: AppOption
       await handleDependsOn({ workflow, ctx });
     }
 
+    if (Timer.getInstance().isTimedOut()) {
```
DmitryAnansky, line 106: I would like to discuss this behavior more.
- Basically it enough to  check timeout timer on Step level only - but in this case we will get as an output a bunch off workflow records with single failed step (timeout error).
- With this additional check on the workflow level, we have output without other workflow that should be executed if not timeput, so in rare condition it might look like first workflow was successful and then we don't have any records and need to check for `globalTimeoutError` in logs.

Would like to here your ideas about the desired behavior
cc: @sobanieca-redocly , @malis42 , @RomanHotsiy , @tatomyr

sobanieca-redocly: Shouldn't we throw 'general' error in this case? I mean, do not even include steps/workflows inside (no json output)?

DmitryAnansky: The idea is different - we should stop and return the output on this moment, but report that further execution results are beyond our time frame.

DmitryAnansky: My personal preference would be to check only on Step level - this would be better for the output as there will be failed check in each workflow step.

packages/respect-core/src/handlers/run.ts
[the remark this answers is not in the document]
DmitryAnansky: @tatomyr , I have created quick [draft pr](https://github.com/Redocly/redocly-cli/pull/2006)  with alternative implementation avoiding singleton, passing parameters and using performance.now.
This alternative variant can be optimized additionally, but I think its enough to show the scale of changes the developer should follow and not messed up startedAt with global sessionStarted in some places.

I am still in favor of this first variant of implementation due to its compact location and not polluting functions with additional parameter. But we can discuss more.

packages/respect-core/src/consts.ts
malis42: Currently in Respect Monitoring config, the lowest value of `interval` we allow is `1m` - therefore I believe seconds would be a better unit for this specific use case. 

However I'm not sure if for general Respect purposes milliseconds wouldn't be better to allow more precise values to be set.

.changeset/silent-suits-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/respect-core": minor
+"@redocly/cli": minor
+---
+
+Add global execution timeout timer to `respect` command execution in order to prevent infinite test runs. The timer can be configured using the `RESPECT_TIMEOUT` environment variable (defaults to 1 hour).
```
JLekawa, line 6:
```suggestion
Added global execution timeout timer to `respect` command execution to prevent infinite test runs. You can configure this timer using the `RESPECT_TIMEOUT` environment variable (defaults to 1 hour).
```

JLekawa (approved):

sobanieca-redocly:
Let's remove `"globalTimeoutError": false` from root in json output? It should be sufficient for us to check at `globalTimeoutError` only for specific file.

packages/respect-core/src/handlers/run.ts
DmitryAnansky, line 97: As discussed I will move `globalTimeoutError` to workflow and file level and remove from root.

DmitryAnansky:
As discussed with @sobanieca-redocly and @malis42 

- `globalTimeoutError` moved on Workflow and File levels and removed from root json-logs level.
- there is only one place to check timeout timer and it is Step.
- timeout will be in ms.
- fixed issue with outputs not resolved when timeout happened.

packages/respect-core/src/handlers/run.ts
DmitryAnansky: updated

packages/respect-core/src/modules/flow-runner/runner.ts
DmitryAnansky: As discussed we will perform check timeout timer on Step level only.

packages/respect-core/src/consts.ts
DmitryAnansky: It was discussed and decided to leave it n ms for now. Can be changed in future.

packages/respect-core/src/modules/cli-output/display-files-summary-table.ts
```diff
@@ -69,9 +70,10 @@ export function displayFilesSummaryTable(
         : gray('-'.padEnd(9));
 
     // First pad the content, then add colors
-    const statusSymbol = testWorkflows.failed > 0 ? 'x' : '✓';
+    const statusSymbol = testWorkflows.failed > 0 || globalTimeoutError ? 'x' : '✓';
```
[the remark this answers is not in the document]
DmitryAnansky: Good catch, its a leftover on prev logic with 2 place timer check, I will revert desplay-table logic.

malis42 (approved):
Awesome 👍

packages/respect-core/src/modules/timeout-timer/index.ts
```diff
@@ -0,0 +1 @@
+export * from './timer';
```
tatomyr, line 1: Please do not use index files as we're going to remove them anyway in 2.0.

packages/respect-core/src/consts.ts
```diff
@@ -0,0 +1,2 @@
+export const RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
+export const MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
```
tatomyr, line 2:
```suggestion
export const DEFAULT_RESPECT_TIMEOUT = 3600000; // 1 hour in milliseconds
export const DEFAULT_MAX_FETCH_TIMEOUT = 20000; // 20 seconds in milliseconds
```

packages/respect-core/src/handlers/run.ts
```diff
@@ -123,6 +124,7 @@ async function runFile(
   const totals = calculateTotals(executedWorkflows);
   const hasProblems = totals.workflows.failed > 0;
   const hasWarnings = totals.workflows.warnings > 0;
+  const hasGlobalTimeoutError = executedWorkflows.some((workflow) => workflow.globalTimeoutError);
```
tatomyr, line 127: Why not just increase `totals.workflows.failed`? That way we could avoid passing `globalTimeoutError` down.

packages/respect-core/src/modules/timeout-timer/timer.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+    return Timer.instance;
+  }
+
+  public isTimedOut(): boolean {
+    const elapsedTime = Date.now() - this.startTime;
+    const timeout = parseInt(process.env.RESPECT_TIMEOUT || RESPECT_TIMEOUT.toString(), 10);
+    const remainingTime = Math.max(0, timeout - elapsedTime);
+
+    return remainingTime <= 0;
```
tatomyr, line 23:
```suggestion
    const timeout = isNaN(+process.env.RESPECT_TIMEOUT)
      ? DEFAULT_RESPECT_TIMEOUT 
      : +process.env.RESPECT_TIMEOUT;
    return  Date.now() - this.startTime > timeout;
```

packages/respect-core/src/modules/cli-output/display-errors.ts
```diff
@@ -48,7 +48,11 @@ export function displayErrors(workflows: WorkflowExecutionResult[]) {
 
       for (const failedCheckIndex in failedStepChecks) {
         const { name, message, severity } = failedStepChecks[failedCheckIndex];
-        const showErrorMessage = name !== CHECKS.UNEXPECTED_ERROR;
+        const showErrorMessage = ![
+          CHECKS.UNEXPECTED_ERROR,
+          CHECKS.GLOBAL_TIMEOUT_ERROR,
+          CHECKS.MAX_STEPS_REACHED_ERROR,
+        ].includes(name);
```
tatomyr, line 55:
```suggestion
        const showDetailedErrorMessage = ![
          CHECKS.UNEXPECTED_ERROR,
          CHECKS.GLOBAL_TIMEOUT_ERROR,
          CHECKS.MAX_STEPS_REACHED_ERROR,
        ].includes(name);
```

packages/respect-core/src/handlers/run.ts
DmitryAnansky: This is file level report, requested by API monitoring.

packages/respect-core/src/modules/flow-runner/run-step.ts
```diff
@@ -32,7 +33,6 @@ import type {
 import type { ParameterWithoutIn } from '../config-parser';
 
 const logger = DefaultLogger.getInstance();
-
 const MAX_STEPS = parseInt(process.env.RESPECT_MAX_STEPS || '2000', 10);
```
tatomyr, line 36: Please replace the 'magic' number.

tatomyr (approved):
LGTM

PR #2018, opened 2025-03-19
rename all *.js snapshots to *.txt

andriimredocly:
## What/Why/How?

Rename the E2E tests snapshots files from *.js to *.txt as we use toMatchFileSnaphot now and the snapshot is plain text - not the JS export string

## Testing

232 tests and 232 renames, all tests passed

## Check yourself

## Security

image


.prettierignore
```diff
@@ -1,9 +1,7 @@
 coverage/
 lib/
 output/
-*snapshot.js
 packages/core/src/rules/__tests__/fixtures/invalid-yaml.yaml
 packages/respect-core/src/modules/runtime-expressions/abnf-parser.js
 benchmark/api-definitions/
 LICENSE.md
```
tatomyr, line 7: Does it make sense to add this?

```suggestion
LICENSE.md
snapshot*.txt
```

tatomyr (approved):
How do we update the snapshots? Could you update the [CONTRIBUTING guide](https://github.com/Redocly/redocly-cli/blob/v2/CONTRIBUTING.md#e2e-tests) if needed?
Didn't you also plan to resolve the following fixme?

```ts
  // FIXME: remove after migration
  // This was needed to fix the issue with unescaped quotes in the output
  // as the snapshot migrated from JS exports string to a plain text
  const cleanedFromUnescapedQuotes = cleanedFromVersion.replace(/`/g, '\\`');
```

.prettierignore
andriimredocly: This can be done - make sure no txt formatting will be done

andriimredocly:
> How do we update the snapshots? Could you update the [CONTRIBUTING guide](https://github.com/Redocly/redocly-cli/blob/v2/CONTRIBUTING.md#e2e-tests) if needed? Didn't you also plan to resolve the following fixme in this PR?
> 
> ```ts
>   // FIXME: remove after migration
>   // This was needed to fix the issue with unescaped quotes in the output
>   // as the snapshot migrated from JS exports string to a plain text
>   const cleanedFromUnescapedQuotes = cleanedFromVersion.replace(/`/g, '\\`');
> ```

The doc on how to update the snapshots is correct - no need to update it

PR #2033, opened 2025-03-31
refactor: use stderr for all output except for the output for program…

andriimredocly:
## What/Why/How?

Refactor the CLI logger to use stderr for all text output except the output that the user may want to use programmatically - it is sent to stdout. Use in cli package the logger from openapi-core

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/commands/split/__tests__/index.test.ts
+++ b/packages/cli/src/commands/split/__tests__/index.test.ts
@@ -20,7 +20,6 @@ describe('split', () => {
       const actual = await vi.importActual('node:process');
       return {
         ...actual,
-        stderr: { write: vi.fn() },
       };
     });
     vi.mock('node:fs', async () => {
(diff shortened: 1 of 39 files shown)


packages/core/src/logger.ts
```diff
@@ -15,11 +15,16 @@ export const colorize = new Proxy(colorette, {
...
   protected stderr(str: string) {
     return process.stderr.write(str);
   }
 
+  protected stdout(str: string) {
+    return process.stdout.write(str);
+  }
+
   info(str: string) {
     return isBrowser ? console.log(str) : this.stderr(str);
```
tatomyr, line 29:
```suggestion
    return isBrowser ? console.info(str) : this.stderr(str);
```

tatomyr (approved):

tatomyr (approved):

packages/cli/src/commands/split/__tests__/index.test.ts
```diff
@@ -52,7 +53,7 @@ describe('split', () => {
       version: 'cli-version',
     });
 
-    expect(process.stderr.write).toBeCalledTimes(2);
+    expect(vi.mocked(process.stderr.write)).toBeCalledTimes(2);
```
tatomyr, line 56: Should we expect mocked?

tatomyr (approved):

PR #2036, opened 2025-04-01
refactor: use same output method for commands

andriimredocly:
## What/Why/How?

The lint and stats outputs should be sent to the same output stream, disregarding the requested format

## Check yourself

## Security


tatomyr (approved):
Please also add an explanation somewhere in the CONTRIBUTING guide how do we use the logger and what's being logged to stdout/stderr/browser console &c. (and a couple of words regarding the snapshot testing approach).

tatomyr (approved):

PR #2040, opened 2025-04-04
refactor: error handling

andriimredocly:
## What/Why/How?

The AbortFlowError was introduced to address a fundamental issue in our error handling approach. Previously, the codebase was using regular errors as a mechanism to abort execution flows, which were then silently caught and ignored in the top-level catch block. This approach had several drawbacks:
It mixed two distinct concerns: actual error conditions that should be reported to users, and flow control mechanisms that were intentionally stopping execution.
When we moved to centralized error logging, these previously silent errors suddenly started appearing in logs, causing test failures and potentially confusing users.
The code's intent was unclear - it wasn't obvious whether an error was thrown because something went wrong or because we wanted to stop execution.
The AbortFlowError provides a clear distinction between these two cases. By using a dedicated error type for flow control, we can handle these cases differently in our centralized error handler - logging actual errors while silently handling flow control errors. This approach maintains the existing functionality while making the code's intent more explicit and improving the overall error handling architecture.

## Check yourself

## Security


packages/cli/src/utils/error.ts
```diff
@@ -0,0 +1,7 @@
+import { HandledError } from '@redocly/openapi-core';
+
+export class AbortFlowError extends Error {}
```
tatomyr, line 3: Could you leave a comment on the purpose of the error?

packages/cli/src/wrapper.ts
```diff
@@ -60,7 +67,25 @@ export function commandWrapper<T extends CommandOptions>(
       }
       code = 0;
     } catch (err) {
-      // Do nothing
+      if (err instanceof AbortFlowError) {
+        // do nothing
+      } else if (err instanceof HandledError) {
+        if (
```
tatomyr, line 73: Let's leave a `FIXME:` so we won't forget to improve this later.

packages/core/src/error.ts
```diff
@@ -0,0 +1 @@
+export class HandledError extends Error {}
```
tatomyr, line 1: Move to -> `packages/core/src/utils/error.ts`?

tatomyr (approved):

PR #2044, opened 2025-04-08
fix: improve performance by pre-computing config type tree

RomanHotsiy:
## What/Why/How?

In one of the dependencies, we call `bundleConfig` and `resolveConfigFileAndRefs` multiple times (2k+ times each).
It looks like normalizeTypes is slow, and moving it into the global scope dramatically speeds up the calls. We save around 30 seconds of build time 🎉  

## Reference

Discovered while perf-testing one of the customer build times.

## Check yourself

## Security


tatomyr (approved):
LGTM. E2E and smokes are failing for a different reason (will be fixed separately).

PR #2071, opened 2025-04-28
fix: smoke test in yarn env

DmitryAnansky:
## What/Why/How?

Updated Yarn smoke tests with cache clean and retries due its instability.

## Check yourself

## Security

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -200,7 +200,7 @@ jobs:
             if (bash ./__tests__/smoke/run-smoke.sh "yarn add ./redocly-cli.tgz" "yarn") {
               break
             }
-            Start-Sleep -Seconds 10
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke.yaml
```diff
@@ -102,7 +107,12 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 22
-      - run: bash ./__tests__/smoke/run-smoke.sh "yarn add redoc ./redocly-cli.tgz" "yarn"
+      - name: Clear Yarn Cache
+        run: yarn cache clean
```
DmitryAnansky, line 111: I will also test with just sleep

DmitryAnansky:
> Could you try if sleep without retries also could do the trick?

tried with 10 sec sleep - not working, lets keep retry approach + sleep

PR #2090, opened 2025-05-14
feat(respect): implementing digest auth with x-security

DmitryAnansky:
## What/Why/How?
Adding support for `digest` auth scheme.

```
        x-security:
          - scheme:
              type: http
              scheme: digest
            values:
              username: dimaananskyi
              password: secret
```

This changes the way fetch performed when Step have `x-security` with `digest` scheme described.
In this case fetch will be called two times during Step execution to establish required data exchange.

`Authorization` header for `digest` will override other `Authorization` headers, that was described by other `x-security` schemes.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/respect/x-security-schema-required-values.ts
+++ b/packages/core/src/rules/respect/x-security-schema-required-values.ts
@@ -41,7 +41,7 @@ export const XSecuritySchemaRequiredValues: Arazzo1Rule = () => {
           const { type } = scheme;
           const authType = type === 'http' ? scheme.scheme : type;
 
-          if (authType === 'mutualTls') {
+          if (schemeName === 'mutualTLS') {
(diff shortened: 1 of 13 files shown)


packages/core/src/rules/respect/x-security-schema-required-values.ts
```diff
@@ -39,7 +39,7 @@ export const XSecuritySchemaRequiredValues: Arazzo1Rule = () => {
           const { type } = scheme;
           const schemeName = type === 'http' ? scheme.scheme : type;
 
-          if (schemeName === 'mutualTls') {
+          if (schemeName === 'mutualTLS') {
```
DmitryAnansky, line 42: Fixed the prev added  issue.

yarokon (approved):

PR #2096, opened 2025-05-15
feat: x-security usage monitoring

DmitryAnansky:
## What/Why/How?

- fix rule validation
- check usage of x-security extension

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/utils/miscellaneous.ts
+++ b/packages/cli/src/utils/miscellaneous.ts
@@ -532,15 +532,23 @@ export function cleanColors(input: string): string {
   return input.replace(/\x1b\[\d+m/g, '');
 }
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/miscellaneous.ts
```diff
@@ -536,7 +538,8 @@ export async function sendTelemetry(
   has_config: boolean | undefined,
   spec_version: string | undefined,
   spec_keyword: string | undefined,
-  spec_full_version: string | undefined
+  spec_full_version: string | undefined,
+  respect_x_security_auth_types: string[] | undefined
```
tatomyr, line 542: It looks like it'd be better to arrange all information regarding spec into a single object, what do you think?

PR #2110, opened 2025-05-21
feat(respect): workflow x-security support

DmitryAnansky:
## What/Why/How?

Add support for `x-security` on workflow level.
Add that described there should be merged to step level `x-security` in the similar to `parameters` way.

## Check yourself

## Security


packages/core/src/rules/respect/x-security-scheme-required-values.ts
```diff
@@ -0,0 +1,95 @@
...
+  digest: ['username', 'password'],
+  bearer: ['token'],
+  oauth2: ['accessToken'],
+  openIdConnect: ['accessToken'],
+  mutualTLS: [],
+} as const;
+
+type AuthType = keyof typeof REQUIRED_VALUES_BY_AUTH_TYPE;
+
+interface SecuritySchema {
```
DmitryAnansky, line 18: SecurityScheme

DmitryAnansky: check ResolveSecurity union type!

packages/cli/src/wrapper.ts
```diff
@@ -6,17 +6,14 @@ import {
...
-} from './utils/miscellaneous.js';
+import { loadConfigAndHandleErrors } from './utils/miscellaneous.js';
+import { sendTelemetry, collectXSecurityAuthTypes } from './utils/telemetry.js';
 import { lintConfigCallback } from './commands/lint.js';
 import { AbortFlowError, exitWithError } from './utils/error.js';
 
 import type { Arguments } from 'yargs';
 import type { Config, CollectFn, ArazzoDefinition } from '@redocly/openapi-core';
-import type { ExitCode } from './utils/miscellaneous.js';
+import type { ExitCode } from './utils/telemetry.js';
```
tatomyr, line 16: I believe ExitCode should remain in miscelaneous, as it could be used not only for telemetry.

tatomyr: Could you also move the `collectSpecData` callback to `utils/telemetry.ts`?

packages/respect-core/src/modules/flow-runner/resolve-x-security-parameters.ts
```diff
@@ -8,27 +8,105 @@ import type { Step, RuntimeExpressionContext } from '../../types.js';
...
+      // Remove any existing scheme with the same key from orderedSecuritySchemes
+      const existingIndex = orderedSecuritySchemes.findIndex(
+        (s) => getSecuritySchemeKey(s) === key
+      );
+      if (existingIndex !== -1) {
+        orderedSecuritySchemes.splice(existingIndex, 1);
+      }
+      orderedSecuritySchemes.push(security);
+    }
+  }
```
tatomyr: Looks too complex to me. Could we just combine the schemes without mutation?

packages/respect-core/src/modules/flow-runner/resolve-x-security-parameters.ts
```diff
@@ -8,27 +8,105 @@ import type { Step, RuntimeExpressionContext } from '../../types.js';
...
+
+  // Convert array to parameters and process them
+  return orderedSecuritySchemes
     .map((security) => {
       const scheme =
         'schemeName' in security
           ? (operation?.securitySchemes?.[security.schemeName] as Oas3SecurityScheme)
           : security.scheme;
 
+      if ('schemeName' in security && !scheme) {
```
tatomyr, line 61: I wouldn't throw inside `map`. Can we check that before? Or after?

packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -336,34 +341,34 @@ export class ApiFetcher implements IFetcher {
...
         headers: updatedHeaders,
       });
 
       responseTime = Math.ceil(performance.now() - startTime);
-      res = await result.text();
+      responseBody = await fetchResult.text();
       // REGULAR FETCH
     } else {
-      result = await wrappedFetch(urlToFetch, fetchParams);
+      fetchResult = await wrappedFetch(urlToFetch, fetchParams);
```
tatomyr, line 353: Could you put the comment inside the `else` block?

yarokon (approved):

packages/respect-core/src/modules/flow-runner/resolve-x-security-parameters.ts
tatomyr: Yes, but `map` is meant for transformation, not control flow or validation. Throwing an error inside a map makes code less predictable and harder to debug.

DmitryAnansky: Do you suggest to filter al scheme before?

DmitryAnansky: Actually its more than this - we are checking there that the scheme was resolved when described with schemeName, so no before filtering is possible.

tatomyr: Then we can check it after, right?

packages/cli/src/wrapper.ts
DmitryAnansky: Lets discuss `collectSpecData` move, as in current implementation it will require passing more params to each command handler.

PR #2111, opened 2025-05-23
feat(respect): hide in logs authorization header generated by x-security

DmitryAnansky:
## What/Why/How?

- Hide generated by x-security Authorization headers (`basic`, `bearer`, `oath2`, `openIdConnect` )
- Hide header for `digest` only when username value is an input with format:`password`
- Hide params for `apiKey` only when value is an input with format:`password`

## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/modules/context-parser/get-security-parameters.ts
+++ b/packages/respect-core/src/modules/context-parser/get-security-parameters.ts
@@ -1,5 +1,3 @@
-import { addSecretFields } from '../flow-runner/context/create-test-context.js';
-
 import type {
   ApiKeyAuth,
   BasicAuth,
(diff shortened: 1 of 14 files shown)


packages/respect-core/src/modules/context-parser/get-security-parameters.ts
```diff
@@ -6,9 +8,13 @@ import type {
   OpenIDAuth,
   ResolvedSecurity,
 } from '@redocly/openapi-core';
+import type { TestContext } from '../../types';
 import type { ParameterWithIn } from './parse-parameters';
 
-export function getSecurityParameters(security: ResolvedSecurity): ParameterWithIn | undefined {
+export function getSecurityParameters(
+  ctx: TestContext,
+  security: ResolvedSecurity
```
tatomyr, line 16:
```suggestion
  security: ResolvedSecurity,
  ctx: TestContext
```

PR #2115, opened 2025-05-28
fix: resolve scheme from digest schemeName

DmitryAnansky:
## What/Why/How?

Fixed the issue when resolving scheme for digest schemeName x-security.

## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/utils/api-fetcher.ts
+++ b/packages/respect-core/src/utils/api-fetcher.ts
@@ -283,7 +283,7 @@ export class ApiFetcher implements IFetcher {
         const scheme =
           'schemeName' in security && security.schemeName
             ? openapiOperation?.securitySchemes?.[security.schemeName]
-            : (security.scheme as Oas3SecurityScheme);
+            : security.scheme;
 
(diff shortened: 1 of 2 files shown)


packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -278,12 +279,14 @@ export class ApiFetcher implements IFetcher {
...
       .reverse()
-      .find(
-        (security) =>
-          'scheme' in security &&
-          security.scheme?.type === 'http' &&
-          security.scheme?.scheme === 'digest'
-      );
+      .find((security) => {
+        const scheme =
+          'schemeName' in security && security.schemeName
```
yarokon, line 284: `'schemeName' in security` and` security.schemeName` looks like the same, you can keep one.

```ts
const scheme = 'schemeName' in security
    ? openapiOperation?.securitySchemes?.[security.schemeName]
    : security.scheme as Oas3SecurityScheme;

  return scheme?.type === 'http' && scheme.scheme === 'digest';
  ```

Issue #2104, opened 2025-05-19
Redocly lint fails with an error 'Expected type `object,null` but got `object`.'


ZyzychO:
**Describe the bug**

With the following schema in openapi: 3.1.1 

```yaml
    pet:
      title: petapi
      type: object
      description: this is pet api
      properties:
        constructor:
          title: petapi
          type:
            - object
            - 'null'
```

Redocly CLI lint command fails with an error message:  'Expected type `object,null` but got `object`.' If you change the 'type' into 'types', lint is passing.

**To Reproduce**

1. Use the following openapi.yaml file: 

```yaml
openapi: 3.1.1
info:
  title: Pet API
  version: 1.0.0
  description: A simple API for pet-related contracts.

paths:
  /pets:
    get:
      summary: Get pet contract information
      operationId: getPet
      responses:
        '200':
          description: Successful response with pet contract
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/pet'

components:
  schemas:
    pet:
      title: petapi
      type: object
      description: this is pet api
      properties:
        constructor:
          title: petapi
          type:
            - object
            - 'null'
```
2. Execute Redocly lint command
3. See error

**Expected behavior**

Lint should pass

**Logs**

**OpenAPI description**

Configuration in redocly.yaml file:
```yaml

extends:
  - minimal
```

**Redocly Version(s)**

1.33.0

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_fy40zw)

tatomyr:
@ZyzychO as an immediate workaround, you can change the schema the following way to eliminate the error (it's an alternative way of describing the same data structure):

```yaml
    pet:
      title: petapi
      type: object
      description: this is pet api
      properties:
        constructor:
          title: petapi
          oneOf:
            - type: object
            - type: 'null'

```

> If you change the 'type' into 'types', lint is passing.

I believe it doesn't matter how you change the name of that property. Redocly CLI specifically check only for the `type` property, so the error disappears because there's no such one.

tatomyr:
It seems that support for that syntax was never implemented.

adamaltman:
This is a valid syntax we have other users that use it. We should update the priority on this issue. It is particularly common with null and another type since they eliminated nullable in 3.1.

adamaltman:
Is the issue specific to

```
- object
- 'null'
```

[...]

PR #2126, opened 2025-06-04
fix: fixed the wrong validation of constructor property

RomanHotsiy:
## What/Why/How?

Because we used regular objects we mistakenly were getting `constructor` property validated incorrectly.

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/2104

## Testing
Added e2e.

## Check yourself

## Security

### DIFF
--- a/.changeset/thick-planets-fail.md
+++ b/.changeset/thick-planets-fail.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Fixed the incorrect validation logic for the `constructor` property.
+Fixed incorrect validation logic for the `constructor` property.
--- a/.changeset/thick-planets-fail.md
+++ b/.changeset/thick-planets-fail.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed the incorrect validation logic for the `constructor` property.
--- a/packages/core/src/types/index.ts
+++ b/packages/core/src/types/index.ts
@@ -47,7 +47,7 @@ type NormalizedResolveTypeFn = (value: any, key: string) => NormalizedPropType;
 export function listOf(typeName: string) {
   return {
     name: `${typeName}List`,
-    properties: Object.create(null),
+    properties: {},
     items: typeName,
   };
 }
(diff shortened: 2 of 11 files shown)


packages/core/src/types/index.ts
```diff
@@ -47,22 +47,22 @@ type NormalizedResolveTypeFn = (value: any, key: string) => NormalizedPropType;
...
-    properties: {},
+    properties: Object.create(null),
     additionalProperties: () => typeName,
   };
 }
 
 export const SpecExtension: NormalizedNodeType = {
   name: 'SpecExtension',
-  properties: {},
+  properties: Object.create(null),
```
tatomyr: We use `properties: {}` in lots of other node types. Don't we need to update them all?

tatomyr: Or it'd be better to check objects for own properties?

RomanHotsiy: I started from own properties but it's too many places. Let's update everywhere. Where else you can see it?

RomanHotsiy: I just checked. I think simpler to check for own properties.

tatomyr: Basically in most of our node types:
image

However, maybe you're right and this would be easier to do. If I'm not mistaken, this syntax means 'no properties'. Let's define a factory method `Empty = () => Object.create(null)` and use it everywhere.

RomanHotsiy: Updated

packages/core/src/utils.ts
```diff
@@ -296,6 +296,10 @@ export function dequal(foo: any, bar: any): boolean {
   return foo !== foo && bar !== bar;
 }
 
+export function getOwn(obj: Record<string, any>, key: string) {
```
tatomyr, line 299: Wouldn't it be more idiomatic if we swapped the parameters? Then it would read like 'getOwn key [from] obj'.

RomanHotsiy: I don't agree.

I just did a quick code search across opensource repos: 
image

.changeset/thick-planets-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed the incorrect validation logic for the `constructor` property.
```
JLekawa:
```suggestion
Fixed incorrect validation logic for the `constructor` property.
```

JLekawa (approved):

packages/core/src/utils.ts
tatomyr: Ok, convinced 🙂.

tatomyr (approved):

Issue #904, opened 2022-10-10
refactor: the `styleguide` object was eliminated in 109.


adamaltman:
I think we should refactor the internal code to not have `styleguide` as it makes reading some of the code and tests confusing. I thought, for example, that Roman made a mistake here:

        The `styleguide` object was eliminated in 109. 
```suggestion
  it('should default to the extends from the main config if no extends defined', async () => {
    const rawConfig: RawConfig = {
      apis: {
        petstore: {
          root: 'some/path',
          rules: {
            'operation-4xx-response': 'error',
          },
        },
      },
      extends: ['minimal'],
      rules: {
        'operation-2xx-response': 'warn',
      },
    };

    const { apis } = await resolveConfig(rawConfig, configPath);
    expect(apis['petstore'].rules).toBeDefined();
    expect(apis['petstore'].rules?.['operation-2xx-response']).toEqual('warn'); // form minimal ruleset
  });
```

_Originally posted by @adamaltman in https://github.com/Redocly/redocly-cli/pull/902#discussion_r991438515_

PR #2120, opened 2025-06-03
feat: wip, implement extends bundling and support for array in plugins

RomanHotsiy:
## What/Why/How?
WIP

## Check yourself

- [ ] Code changed? - Tested with redoc (internal)
- [ ] All new/updated code is covered with tests
- [ ] New package installed? - Tested in different environments (browser/node)

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

Issue #1695, opened 2024-08-29
'remove-unused-components' decorator is not working at the per-API level.


ZyzychO:
**Describe the bug**

When you configure decorator: 'remove-unused-components' per API level it will not remove the unused 
**To Reproduce**

1. Given this `redocly.yaml` file

```yml
apis:
  testclipush@v1:
    root: redint.yaml
    decorators:
      remove-x-internal: on
      remove-unused-components: on


extends: 
  -  minimal


# decorators:
#   remove-unused-components: on
```

2. And this OpenAPI file
[redint.yaml.zip](https://github.com/user-attachments/files/16796310/redint.yaml.zip)

3. Run 'redocly bundle' command
4. Observe that unused component 'StartDate' is not removed from the API definition

**Redocly Version(s)**

1.21.1

tatomyr:
But the `remove-unused-components` decorator does work on the root level, right @ZyzychO?

ZyzychO:
Yes, the decorator works on the root level. @tatomyr

tatomyr:
Related: https://github.com/Redocly/redocly-cli/issues/1323

PR #2131, opened 2025-06-11
chore!: refactor config loading and usage

tatomyr:
## What/Why/How?

Main changes: 

- Refactored [config/load.ts](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-91233afeb25fea2c651ad22452d823830a0c6e8378deba81a1109b57cdae3fb6) (removed `addCConfigMetadata` and using `resolveConfig` directly instead)
- Refactored [config/config-resolvers.ts](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-32a74d0bc02f339c875f10b6706077b6b16a1838db334faaac75b96c681f700d) (removed `getMergedRawStyleguideConfig`; now an api-level config is getting merged with the root config inside `resolveAndMergeNestedGovernanceConfig` using the `mergeExtends` function)
- Refactored [config/config.ts](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-47607adc7025266f655cda3df6abbd5cc9f1fbf4e9f493c7b0ba6ed6c3350614):
  - Combined `StyleguideConfig` with `Config`
  - changed the `Config` structure and renamed some of its fields for better consistency: now the config contains  `rawConfig` -- the actual config bundled into json; `resolvedConfig` -- config with resolved plugins and extends. Whenever we need to use a specific config for an alias, we can use `config.forAlias(api)` to create a new `Config` instance.
  - removed the imperative `recommendedFallback` field -- instead we apply the default fallback if there's no raw config provided (which means `redocly.yaml` does not exist)
- Now `lintDocument` and `bundleDocument` accept the regular config and an alias instead of `StyleguideConfig` (https://github.com/Redocly/redocly-cli/pull/2131/files#diff-6249eec9b70b0e554ea9c353ce79872a0f45795ab3e0ea36e4fa3b0c408fc537, https://github.com/Redocly/redocly-cli/pull/2131/files#diff-6d58493339aa21bcd72df7f220f88883133ead11df05345e8fb369b77210cbea)
- Refactored [config/utils.ts](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-3e1b77fc75cd43c6d3286b58a4d08151fe1ae530563c04884376972c04155f8d): removed `getMergedConfig` (now the needed config is being precalculated when creating the config instead of creating it each time we lint/bundle a file listed in the `apis` section), `extractFlatConfig`,  and config transformers that were used to maintain backward compat with deprecated config versions
- Removed [redundant types](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-623912bbe54cdd04399ad03ecf4c603f03682cf4d64116197426943f96811cf9) (`StyleguideRawConfig`, `ResolvedStyleguideConfig`, `PluginStyleguideConfig`, `Api`, `FlatApi`, `RulesFields` and others)
- Removed all mentions of styleguide from the codebase. The idea is that what we previously had as `styleguide` became governance config (rules, decorators, preprocessors, oas3rules, oas3decorators, oas3preprocessors, ...) or normalized governance config (the same but in the form that is used in lint and bundle)
- Removed the `benchmark/` folder. It requires a lot of work to make it compatible with the new changes, and it is not used anyway. If we were to make it work, I'd rather rewrite it from scratch. Anyway, we'll have it in the `v1` branch for reference if needed.
- Improved `redocly.yaml` types consistency in [types/redocly-yaml.ts](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-2528117ea3cbe25dfbfd0583c9b11ba7baa7ef347184dfba32a73ec486aa1a5e)
- Fixed various types and test data inconsistencies

Most of the other changes are test files/snapshots.

**What's left**:
- sort out `loadConfig`, `getConfig`, `resolveConfig` &c. (in a separate PR)
- remove `processRawConfig`
- refine config type names
- remove resolving extends/plugins in `resolveAndMergeNestedGovernanceConfig` after merging https://github.com/Redocly/redocly-cli/pull/2120 (this should simplify the code)

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/904

## Check yourself

- [x] Code changed? - Tested with redoc (internal)

## Security


packages/core/src/config/load.ts
```diff
@@ -86,7 +59,11 @@ export async function loadConfig(
     }
   }
 
-  return config;
+  return {
+    ...config,
+    document,
+    resolvedRefMap,
```
tatomyr: Extendnig Config with `document` and `resolvedRefMap` so we don't have to use `processRawConfig` anymore.

packages/core/src/config/load.ts
```diff
@@ -106,54 +83,44 @@ export function findConfig(dir?: string): string | undefined {
...
-    configPath?: string;
-    externalRefResolver?: BaseResolver;
-  } = {}
-): Promise<{
-  rawConfig: RawConfig;
+export async function getConfig(options: {
+  configPath?: string;
+  externalRefResolver?: BaseResolver;
+}): Promise<{
+  rawConfig?: RawUniversalConfig;
```
tatomyr, line 94: Made `rawConfig` optional. If it's `undefined`, then we apply the default `recommeded` ruleset.

packages/core/src/config/load.ts
```diff
@@ -106,54 +83,44 @@ export function findConfig(dir?: string): string | undefined {
...
   document?: Document;
-  parsed?: Document['parsed'];
   resolvedRefMap?: ResolvedRefMap;
 }> {
-  const { configPath = findConfig(), externalRefResolver = new BaseResolver() } = options;
```
tatomyr, line 120: We don't need to call `findConfig` since we have already called it before calling `getConfig`.

packages/core/src/config/load.ts
```diff
@@ -106,54 +83,44 @@ export function findConfig(dir?: string): string | undefined {
...
 
 type CreateConfigOptions = {
-  extends?: string[];
   configPath?: string;
   externalRefResolver?: BaseResolver;
 };
 
 export async function createConfig(
-  config: string | RawUniversalConfig,
+  config?: string | RawUniversalConfig,
```
tatomyr, line 127: If `config === undefined`, then the default `recommended` fallback will be applied.

packages/core/src/config/config-resolvers.ts
```diff
@@ -73,43 +73,63 @@ export async function resolveConfigFileAndRefs({
...
-      apis,
-      styleguide,
+  const { plugins: _plugins, extends: _extends, apis: _apis, ...rest } = config;
+  const resolvedConfig: ResolvedConfig = {
+    ...rest,
+    ...rootGovernanceConfig,
+    apis,
+  };
+
+  return {
```
tatomyr: Returning config as an object instead of a class instance. It was somewhat confusing that we had that class and some people used it for instantiating the config (which might work incorrectly).

packages/cli/src/commands/build-docs/utils.ts
```diff
@@ -38,10 +39,10 @@ export function getObjectOrJSON(
...
       break;
     default: {
-      if (config) {
-        logger.info(`Found ${config.configFile} and using theme.openapi options\n`);
-
-        return config.theme.openapi ? config.theme.openapi : {}; // FIXME: theme is deprecated (2.0)
+      if (config?.configPath) {
+        logger.info(`Found ${config.configPath} and using theme.openapi options\n`);
+        const apiConfig = alias ? config.resolvedConfig.apis?.[alias] : config.resolvedConfig;
+        return apiConfig?.theme.openapi ? apiConfig.theme.openapi : {}; // FIXME: theme is deprecated (2.0)
```
RomanHotsiy, line 45: should be removed?

packages/cli/src/commands/bundle.ts
```diff
@@ -41,31 +46,27 @@ export async function handleBundle({
 }: CommandArgs<BundleOptions>) {
   const removeUnusedComponents =
     argv['remove-unused-components'] ||
-    config.rawConfig?.styleguide?.decorators?.hasOwnProperty('remove-unused-components');
+    config.resolvedConfig?.decorators?.hasOwnProperty('remove-unused-components'); // FIXME: also on `apis` level
```
RomanHotsiy: need to be fixed?

RomanHotsiy: oh, I see it was there before so we can fix it separately to reduce the scope of this pr

packages/cli/src/commands/bundle.ts
```diff
@@ -41,31 +46,27 @@ export async function handleBundle({
...
     try {
       const startedAt = performance.now();
-      const resolvedConfig = getMergedConfig(config, alias);
-      const { styleguide } = resolvedConfig;
+      const governanceConfig = getGovernanceConfig(config, alias);
 
-      styleguide.skipPreprocessors(argv['skip-preprocessor']);
-      styleguide.skipDecorators(argv['skip-decorator']);
+      governanceConfig.skipPreprocessors(argv['skip-preprocessor']);
+      governanceConfig.skipDecorators(argv['skip-decorator']);
```
RomanHotsiy: I think we should make the whole Config a class and remove the GovernanceConfig

I think it wold be nicer if it looks like this:

```
const aliasConfig = config.for(alias);
aliasConfig.skipPreprocessors(argv['skip-preprocessor']);
//...
... = bundle({..., config: aliasConfig })
```

where config.for(alias) returns a new config instance with the correct internal state.

packages/cli/src/commands/join.ts
```diff
@@ -106,25 +107,27 @@ export async function handleJoin({
...
   ]);
-  config.styleguide.skipPreprocessors(Array.from(preprocessors));
+  rootGovernance.skipPreprocessors(Array.from(preprocessors));
 
   const bundleResults = await Promise.all(
     documents.map((document) =>
       bundleDocument({
         document,
-        config: config.styleguide,
+        config: config,
```
RomanHotsiy:
```suggestion
        config,
```

packages/cli/src/utils/miscellaneous.ts
```diff
@@ -526,6 +531,7 @@ export function cleanColors(input: string): string {
   return input.replace(/\x1b\[\d+m/g, '');
 }
 
+// TODO: remove this
```
RomanHotsiy: need to remove?

packages/core/src/bundle.ts
```diff
@@ -157,14 +158,15 @@ export async function bundleDocument(opts: {
   } = opts;
   const specVersion = detectSpec(document.parsed);
   const specMajorVersion = getMajorSpecVersion(specVersion);
-  const rules = config.getRulesForSpecVersion(specMajorVersion);
+  const governanceConfig = getGovernanceConfig(config, alias);
```
RomanHotsiy: do we need to send the alias from the top or should we just pass the correct config instance ? (like based on my idea above)

packages/core/src/config/config-resolvers.ts
RomanHotsiy: This is a valid point 🤔

RomanHotsiy: but maybe we should just not export this class anymore directly.
Only loadConfig and crateConfig will be exported.

packages/core/src/config/config-resolvers.ts
```diff
@@ -436,54 +479,54 @@ async function resolveAndMergeNestedStyleguideConfig({
...
         resolver,
         parentConfigPaths: [...parentConfigPaths, resolvedConfigPath],
         extendPaths,
       });
-    }) || []
+    })
   );
 
-  const { plugins: mergedPlugins = [], ...styleguide } = mergeExtends([
+  const { plugins: mergedPlugins = [], ...mergedConfig } = mergeExtends([
```
RomanHotsiy, line 478: can do a single structured clone on the top level

packages/core/src/config/config-resolvers.ts
```diff
@@ -367,67 +387,90 @@ export async function resolveApis({
...
+    ...possiblyRootConfig
+  } = rootRawConfig || {};
+
   const plugins = isBrowser
     ? // In browser, we don't support plugins from config file yet
       [defaultPlugin]
     : getUniquePlugins(
         await resolvePlugins(
-          [...(styleguideConfig?.plugins || []), defaultPlugin],
+          [
```
RomanHotsiy, line 435: I think we should not support plugins on `apis` level, only globally. This would simplify this part.

We need to bundle config first to resolve all extends and then instantiate plugins from the root level.

packages/core/src/config/config-resolvers.ts
```diff
@@ -436,54 +479,54 @@ async function resolveAndMergeNestedStyleguideConfig({
...
+    extendPaths: mergedConfig.extendPaths?.filter((path) => path && !isAbsoluteUrl(path)),
     plugins: getUniquePlugins(mergedPlugins),
-    recommendedFallback: styleguideConfig?.recommendedFallback,
-    doNotResolveExamples: styleguideConfig?.doNotResolveExamples,
   };
 }
 
-export async function resolveStyleguideConfig(opts: {
-  styleguideConfig?: StyleguideRawConfig;
+export async function resolveGovernanceConfig<
```
RomanHotsiy, line 496: I think this function is redundant, it's only wrapping the groupAssertionRules

packages/core/src/config/config.ts
```diff
@@ -373,18 +387,28 @@ export class StyleguideConfig {
   }
 }
 
+// TODO: left for backward compatibility with Redoc. Replace with a type along with replacing Redoc
```
RomanHotsiy: can we publish new Redoc to fix that?

packages/core/src/config/load.ts
```diff
@@ -106,54 +84,44 @@ export function findConfig(dir?: string): string | undefined {
...
 }
 
-export async function getConfig(
-  options: {
-    configPath?: string;
-    externalRefResolver?: BaseResolver;
-  } = {}
-): Promise<{
-  rawConfig: RawConfig;
+export async function getConfig(options: {
```
RomanHotsiy, line 90: I think we need some renames here...
we have getConfig, loadConfig and createConfig. Either one has to be removed or we come up with better names.

RomanHotsiy: oh we also have `resolveConfig` but also `resolveConfigFileAndRefs`

packages/core/src/config/load.ts
```diff
@@ -106,54 +84,44 @@ export function findConfig(dir?: string): string | undefined {
...
   options?: CreateConfigOptions
 ): Promise<Config> {
-  return addConfigMetadata({
-    rawConfig: transformConfig(
-      typeof config === 'string' ? (parseYaml(config) as RawConfig) : config
-    ),
-    ...options,
-  });
+  const rawConfig = typeof config === 'string' ? (parseYaml(config) as RawUniversalConfig) : config;
+  return await resolveConfig({ rawConfig, ...options });
```
RomanHotsiy: I think this has to rename. `resolve` in openapi-core has a different meaning.

packages/core/src/index.ts
```diff
@@ -62,23 +62,23 @@ export { type NormalizedNodeType, type NodeType, normalizeTypes } from './types/
 export { Stats } from './rules/other/stats.js';
 
 export {
-  type RawConfigProcessor,
+  type RawConfigProcessor, // FIXME: remove this
```
RomanHotsiy, line 65: remove?

packages/core/src/index.ts
```diff
@@ -62,23 +62,23 @@ export { type NormalizedNodeType, type NodeType, normalizeTypes } from './types/
...
   type ResolvedConfig,
+  type ResolvedApi,
```
RomanHotsiy: should it be ResolvedApiConfig?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -180,40 +180,45 @@ const builtInRules = [
 
 type BuiltInRuleId = typeof builtInRules[number];
 
-const ConfigStyleguide: NodeType = {
-  properties: {
-    extends: {
```
RomanHotsiy, line 185: so you removed extends, I think we need to make it as on my branch

packages/core/src/config/config.ts
tatomyr: Not now. First we have to release openapi-core 2.0-next (without using `new Config` in the bundler), then update it in Redoc, and then we can cut a release and reuse new Redoc here.

packages/core/src/config/load.ts
tatomyr: Let's do that in a separate PR.

packages/cli/src/commands/bundle.ts
tatomyr: Yeah. This is likely related to https://github.com/Redocly/redocly-cli/issues/1695. Left a FIXME to address it separately.

packages/core/src/index.ts
tatomyr: Not now. I've removed its usage in this repo, but it's still used in other products. Left a FIXME to handle that separately.

tatomyr: I'm going to resolve all added FIXMEs before the GA.

packages/core/src/types/redocly-yaml.ts
tatomyr: Yes, it overrides what is declared in the config package. I've added it there just a week ago 😅. Let's do it in your PR. I'll revert it here.

packages/cli/src/commands/build-docs/utils.ts
tatomyr: The `theme` is used extensively across the project, so I'd like to remove it in a separate PR. We can even publish a `next` version and then remove `theme` in CLI before going GA.

RomanHotsiy: Let's solve in a separate PR.

packages/core/src/config/config-resolvers.ts
RomanHotsiy: Let's solve as part of another PR

packages/core/src/config/load.ts
RomanHotsiy: Let's handle in another PR

__tests__/check-config/invalid-config--lint-config-error/snapshot.txt
```diff
@@ -14,3 +14,4 @@ Error was generated by the configuration struct rule.
 
 ❌ Your config has 1 error.
 
+
```
DmitryAnansky, line 17: Why there some empty lines, is it product of prettier?

packages/cli/src/wrapper.ts
```diff
@@ -61,13 +60,9 @@ export function commandWrapper<T extends CommandOptions>(
...
       }
-      const config: Config = (await loadConfigAndHandleErrors({
-        configPath: argv.config,
-        customExtends: argv.extends as string[] | undefined,
-        processRawConfig: lintConfigCallback(argv as T & Record<string, undefined>, version),
-      })) as Config;
-      telemetry = config.telemetry;
-      hasConfig = !config.styleguide.recommendedFallback;
+      const config: Config = await loadConfigAndHandleErrors(argv as Exact<T>, version);
+      telemetry = config.resolvedConfig.telemetry;
```
DmitryAnansky, line 64: Can `config` have `resolvedConfig.telemetry` in case there will be no `config.rawConfig`?
Will it be empty `resolvedConfig={}`?

packages/cli/src/utils/miscellaneous.ts
```diff
@@ -526,24 +530,12 @@ export function cleanColors(input: string): string {
   return input.replace(/\x1b\[\d+m/g, '');
 }
 
-export function checkForDeprecatedOptions<T>(argv: T, deprecatedOptions: Array<keyof T>) {
```
DmitryAnansky, line 529: We might have deprecated options in the future, should we keep this logic?

DmitryAnansky (approved):

__tests__/check-config/invalid-config--lint-config-error/snapshot.txt
tatomyr: No, fixed output [here](https://github.com/Redocly/redocly-cli/pull/2131/files#diff-b781401ffbc8754cb3f94aa3ae701de53a9bd3e93aec59c8a21e0b605fc2a8b7R360).

packages/cli/src/wrapper.ts
tatomyr: > Will it be empty resolvedConfig={}?

Yes.

packages/cli/src/utils/miscellaneous.ts
tatomyr: Then we can reintroduce it. I already removed one more deprecation notifier before 😅. Keeping dead code is not a good option anyways.

RomanHotsiy (approved):

Issue #2118, opened 2025-06-02
Join produces valid yaml but invalid semantics


AndreasBergmeier6176:
**Describe the bug**

Join produces invalid OpenAPI doc. Especially bad because it does not notify user that the result will very likely be unexpected.

**To Reproduce**

1. Run `redocly join` with these two (non-conflicting) files:
```yaml
openapi: 3.1.1

paths:
  /foo:
    get:
      summary: Get Foo
servers:
  - url: https://foo.com/api/v1/first
```
and
```yaml
openapi: 3.1.1

paths:
  /bar:
    get:
      summary: Get bar
servers:
  - url: https://foo.com/api/v1/second
```

2. See invalid result

```yaml
openapi: 3.1.1

paths:
  /bar:
    get:
      summary: Get bar
  /foo:
    get:
      summary: Get Foo
servers:
  - url: https://foo.com/api/v1/first
  - url: https://foo.com/api/v1/second
```

This OpenAPI cannot be used to generate anything since it is just plain invalid.

**Expected behavior**

Either it needs to be smart enough to detect common anchored server URLs and output:

```yaml
openapi: 3.1.1

paths:
  /first/foo:
    get:
      summary: Get Foo
  /second/bar:
    get:
      summary: Get bar
servers:
  - url: https://foo.com/api/v1
```

or it needs to detect that it cannot possibly join these files to a valid output and issue an error:

`Cannot join documents containing different servers URLs to a valid output. If you want the operation to happen anyways use commandline option --ignore-semantics`

jeremyfiel:
Couple things

- the yaml is valid
- a definition can have more than one unique basepath defined, that's why it's an array of servers 


Can you explain in more detail what you are generating and why you think nothing _useful_ can be generated?

Where is the bug? The implementation may not produce what you require, but I'm not sure a bug exists. 

The `join` command is not a code generator, Redoc is probably considered the main output target. It is understandable that the definition may be used for other purposes than docs.

AndreasBergmeier6176:
Ok, we probably need to create an own semantic tool then if that is out of scope for redocly.
We have multiple teams that have this requirement.

RomanHotsiy:
I think it can work in a simpler way by leveraging `servers` on the path level:

```
openapi: 3.1.1

paths:
  /bar:
    servers:
      - url: https://foo.com/api/v1/second
    get:
      summary: Get bar
  /foo:
    servers:
      - url: https://foo.com/api/v1/first
    get:
[...]

PR #2133, opened 2025-06-12
fix: improve join command servers handling

RomanHotsiy:
## What/Why/How?
This pull request resolves an issue with the join command where it would incorrectly merge OpenAPI documents with different servers configurations, leading to a semantically invalid result.

The updated logic now handles server merging based on whether the input files share the same server definitions.

## Reference
Resolves https://github.com/Redocly/redocly-cli/issues/2118

## Check yourself

## Security

### DIFF
--- a/__tests__/join/with-metadata/snapshot.txt
+++ b/__tests__/join/with-metadata/snapshot.txt
@@ -10,9 +10,6 @@ info:
   license:
     name: Apache 2.0
     url: http://www.apache.org/licenses/LICENSE-2.0.html
-servers:
-  - url: http://redocly-example.com:8081
-  - url: http://redocly-example.com:8080
 tags:
   - name: user
     x-displayName: user
@@ -20,6 +17,12 @@ tags:
     x-displayName: pet
 paths:
   /GETUser/{userId}:
+    servers:
+      - url: /test
+      - url: /pets
+        description: another description
+      - url: /pet
+        description: some description
     summary: getUser
     description: getUser
     parameters:
@@ -29,12 +32,6 @@ paths:
         in: header
         schema:
           description: string
-    servers:
-      - url: /test
-      - url: /pets
-        description: another description
-      - url: /pet
-        description: some description
     post:
       tags:
         - user
--- a/packages/cli/src/__tests__/commands/join.test.ts
+++ b/packages/cli/src/__tests__/commands/join.test.ts
@@ -17,9 +17,17 @@ import {
 } from '../../utils/miscellaneous.js';
 import { exitWithError } from '../../utils/error.js';
 import { configFixture } from '../fixtures/config.js';
-import { firstDocument, secondDocument, thirdDocument } from '../fixtures/join/documents.js';
+import {
+  firstDocument,
+  secondDocument,
+  thirdDocument,
+  serverAndPaths,
+  anotherServerAndPaths,
+} from '../fixtures/join/documents.js';
 
 describe('handleJoin', () => {
+  let writeToFileByExtensionSpy: any;
+
   beforeEach(() => {
     vi.mock('../../utils/miscellaneous.js');
     vi.mock('../../utils/error.js');
@@ -30,7 +38,9 @@ describe('handleJoin', () => {
       async (entrypoints) => entrypoints?.map((path: string) => ({ path })) ?? []
     );
     vi.mocked(sortTopLevelKeysForOas).mockImplementation((document) => document);
-    vi.mocked(writeToFileByExtension).mockImplementation(() => {});
+    writeToFileByExtensionSpy = vi
+      .mocked(writeToFileByExtension)
+      .mockImplementation(() => undefined);
 
     vi.mock('colorette');
     vi.mocked(yellow).mockImplementation((text) => text as string);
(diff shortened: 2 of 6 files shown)


__tests__/join/with-metadata/snapshot.txt
```diff
@@ -10,16 +10,19 @@ info:
   license:
     name: Apache 2.0
     url: http://www.apache.org/licenses/LICENSE-2.0.html
-servers:
```
tatomyr, line 13: Please add one more path item that doesn't have its own servers to make sure the root servers will be applied to it.

packages/cli/tsconfig.json
```diff
@@ -5,6 +5,6 @@
     "outDir": "lib"
   },
   "references": [{ "path": "../core" }, { "path": "../respect-core" }],
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "src/__tests__/fixtures/join/documents.ts1"],
```
tatomyr: `documents.ts1` looks wrong.

packages/cli/src/commands/join.ts
```diff
@@ -374,13 +377,17 @@ export async function handleJoin({
...
+          collectPathServers(servers, path);
+        }
 
         for (const field of keysOf(pathItem)) {
           if (operationsSet.has(field as OPENAPI3_METHOD)) {
             collectPathOperation(pathItem, path, field as OPENAPI3_METHOD);
           }
           if (field === 'servers') {
-            collectPathServers(pathItem, path);
+            // already processed
```
tatomyr: Can you just remote the `if` statement?

packages/cli/src/commands/join.ts
```diff
@@ -374,13 +377,17 @@ export async function handleJoin({
         }
 
         const pathItem = paths[path] as Oas3PathItem;
+        const servers = serversAreTheSame ? pathItem.servers : pathItem.servers || rootServers;
```
tatomyr: Do we need to calculate if the servers are the same? Maybe it's better to just spread the root servers to the path item level in each document and eliminate the root servers in the joined document at all? Technically it will be the same but less code.

RomanHotsiy: but also worse result. I would try to keep the resulting file look better.

__tests__/join/with-metadata/snapshot.txt
RomanHotsiy: done

RomanHotsiy: created a separate test

packages/cli/src/__tests__/commands/join.test.ts
```diff
@@ -51,6 +61,10 @@ describe('handleJoin', () => {
       );
   });
 
+  afterEach(() => {
```
tatomyr, line 75: Are you sure you need this? We've configured Vitest to [restore all mocks](https://vitest.dev/config/#restoremocks): https://github.com/Redocly/redocly-cli/blob/main/vitest.config.ts#L49. Isn't that enough?

packages/cli/src/commands/join.ts
```diff
@@ -169,6 +169,19 @@ export async function handleJoin({
...
+    firstDocServers &&
+    documents.slice(1).every((doc) => {
+      // include only documents with paths
+      if (Object.keys(doc.parsed.paths || {}).length === 0) {
+        return true;
+      }
+      return doc.parsed.servers?.every((server: Oas3Server) =>
+        firstDocServers?.find((firstDocServer: Oas3Server) => firstDocServer.url === server.url)
+      );
+    });
```
tatomyr: A little bit easier to read for me (no magic numbers like 0 or 1 🙂): 

```suggestion
  const [first, ...others] = documents ?? [];
  const serversAreTheSame = others.every(({ parsed: { paths, servers } }) => {
     // include only documents with paths
     if (!paths || isEmptyObject(paths || {})) {
       return true;
     }
     return servers?.every((server: Oas3Server) =>
       first.parsed.servers?.find(({ url }: Oas3Server) => url === server.url)
     );
  })
```

__tests__/commands.test.ts
```diff
@@ -335,7 +336,10 @@ describe('E2E', () => {
 
       test.each(testDirNames)('test: %s', async (dir) => {
         const testPath = join(__dirname, `join/${dir}`);
-        const args = getParams(indexEntryPoint, ['join', ...entrypoints]);
+        let args = getParams(indexEntryPoint, ['join', ...entrypoints]);
+        if (dir === 'three-files-one-without-servers') {
```
tatomyr: Could you move this to a separate `describe`/`test`? Or put a FIXME so I'll do that later. I try to avoid conditions in tests so they are easier to debug and maintain.

packages/cli/src/commands/join.ts
RomanHotsiy: done

__tests__/commands.test.ts
```diff
@@ -392,12 +392,19 @@ describe('E2E', () => {
 
       test.each(testDirNames)('test: %s', async (dir) => {
         const testPath = join(__dirname, `join/${dir}`);
-        const args = getParams(indexEntryPoint, ['join', ...entrypoints]);
+        let args = getParams(indexEntryPoint, ['join', ...entrypoints]);
```
tatomyr: Why it's changed to `let`?

tatomyr (commented):
Could you resolve the conflict?

packages/cli/src/__tests__/commands/join.test.ts
RomanHotsiy: yes, breaks without it.

__tests__/commands.test.ts
RomanHotsiy: done

PR #2135, opened 2025-06-12
fix: guard against race conditions when resolving $refs

vlindhol:
## What/Why/How?

We were using a global AJV instance shared for all documents, which might lead to race conditions.

Add a `WeakMap` holding an AJV instance for each document, and add some extra deduplication logic to prevent trying to look up the same reference multiple times *within* a document, just in case.

## Reference

?

## Testing

Race conditions are hard to trigger reliably, I think the best I can do right now is say "all existing tests must pass".

I have failures triggered by processing a large amount of .yaml files at once at work, but can't post those files publicly :/

Example of the type of errors I get:

```
Example validation errored: can't resolve reference #/components/schemas/Product from <filename>
```

as if local references can't be resolved! **After applying this fix those errors go away**.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/resolve.ts
+++ b/packages/core/src/resolve.ts
@@ -234,9 +234,9 @@ export async function resolveDocument(opts: {
   const { rootDocument, externalRefResolver, rootType } = opts;
   const resolvedRefMap: ResolvedRefMap = new Map();
   const seenNodes = new Set<string>(); // format "${type}::${absoluteRef}${pointer}"
-  
+
   const resolvePromises: Array<Promise<void>> = [];
-  
+
   const resolvingRefs = new Set<string>();
   /** Ensures a ref is resolved only once, preventing race conditions and duplicate work */
   function resolveRefOnce<T>(options: {
@@ -246,13 +246,13 @@ export async function resolveDocument(opts: {
     onError?: (error: any) => void;
   }): boolean {
     const { refId, resolverFn, onSuccess, onError } = options;
-    
+
     if (resolvingRefs.has(refId) || resolvedRefMap.has(refId)) {
       return false; // Already resolving or resolved
     }
 
     resolvingRefs.add(refId);
-    
+
     const promise = resolverFn()
       .then((result) => {
         resolvingRefs.delete(refId);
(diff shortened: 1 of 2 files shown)


tatomyr:
Thanks for the contribution! We'll take a look at your PR in a bit.

vlindhol:
@tatomyr Great! Any updates on this? 😇

tatomyr:
@vlindhol sorry, haven't had a chance to look at it yet. We're currently busy with preparing the next version release. Will review it after we finish that.

packages/core/src/resolve.ts
```diff
@@ -236,6 +236,50 @@ export async function resolveDocument(opts: {
   const seenNodes = new Set<string>(); // format "${type}::${absoluteRef}${pointer}"
 
   const resolvePromises: Array<Promise<void>> = [];
+
+  const resolvingRefs = new Set<string>();
+  /** Ensures a ref is resolved only once, preventing race conditions and duplicate work */
+  function resolveRefOnce<T>(options: {
```
RomanHotsiy, line 242: I don't understand why we need it.

There is already cache involved in BaseRefResolver https://github.com/Redocly/openapi-cli/blob/main/packages/core/src/resolve.ts#L178

It's safe against race-conditions, it should never resolve same ref twice as soon as BaseRefResolver is reused.

packages/core/src/rules/ajv.ts
```diff
@@ -5,13 +5,12 @@ import type { Location } from '../ref-utils.js';
...
-let ajvInstance: Ajv | null = null;
-
-export function releaseAjvInstance() {
-  ajvInstance = null;
-}
+// Use per-resolve-function AJV instances to prevent race conditions
+const ajvInstanceCache = new WeakMap<ResolveFn, Ajv>();
 
 function getAjv(resolve: ResolveFn, allowAdditionalProperties: boolean) {
+  let ajvInstance = ajvInstanceCache.get(resolve);
```
RomanHotsiy, line 12: why do we need this? I don't understand

RomanHotsiy (changes requested):
Those changes are very hard to undertand.

I think neither of us have good understanding of what the problem is.

It would be great if you can prepare some minimal reproducible sample.
Can you reproduce this error on your files reliable? If you can't share your files maybe you would be open to screenshare and do a debug session or we can add more logs and you then would run it and share the output with us.

I'm not ready to merge it without understanding how it fixes the problem and which problem in particular.

tatomyr:

PR #2134, opened 2025-06-12
docs(all): add per-api confgurable rules example

JLekawa:
## What/Why/How?

Adds per-api configurable rule example

Closes #15771

## Check yourself

- [ ] Code changed? - Tested with redoc (internal)
- [x] All new/updated code is covered with tests
- [ ] New package installed? - Tested in different environments (browser/node)

## Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

PR #2138, opened 2025-06-14
docs: reuse code snippet to show configurable rule per api

adamaltman:
## What/Why/How?

Uses code-snippet for reuse of code snippets.

## Reference

- Replaces https://github.com/Redocly/redocly-cli/pull/2134
- Closes https://github.com/Redocly/redocly/issues/15771

## Check yourself

## Security


tatomyr:
@adamaltman I’m wondering -- is it really good practice to allow linting different APIs with different rules in the same repository? I'd think the best way to keep API docs consistent is to apply the same rules across the entire repo.
What’s more, when an API document is referenced multiple times under different aliases in the `apis` section, it creates ambiguity about which rules should apply when linting that file without explicitly specifying an alias (as is the case when running `redocly lint`, `redocly lint ./some-file.yaml`, or in the VS Code extension).

adamaltman:
@tatomyr this isn't about best practice, it's about completing [this issue](https://github.com/Redocly/redocly-cli/pull/2134) because people are requesting per-API configurable rules examples. 

The "entire repo" could be a monorepo, and could contain a variety of APIs that require different rules:

- **Vendor APIs:** For example, someone consume's GitHub's API and compliance requires documentation of every API operation consumed, but the developers can't influence the design of it.
- **Legacy APIs:** Such as internal APIs that pre-date standards.

I think the ambiguity is not unclear for running `redocly lint` because it should run it for **all** of the APIs. It is less clear for the VS Code Extension, but I think we should also run it for all APIs and report all unique problems (rather than duplicate problems).

tatomyr:
Yes, but this PR made me think about best practices.

Thanks for the clarification. Regarding vendor APIs, does it makes sense to lint that docs at all?

adamaltman:
Regarding vendor APIs, possibly not. Maybe it makes sense if you generate an SDK for it to meet certain minimums like `struct`.

Regarding legacy APIs, yes, probably it does.

JLekawa (approved):

PR #2143, opened 2025-06-20
docs: update readme with respect information

Marshevskyy:
## What/Why/How?

docs: update readme with Respect information

## Check yourself

## Security


packages/respect-core/README.md
```diff
@@ -6,12 +6,47 @@ Continuous API monitoring powered by OpenAPI Arazzo workflows.
...
+2. **Generate test workflows** from your OpenAPI spec:
+
+   ```sh
+   redocly generate-arazzo openapi.yaml
+   ```
+
+3. **Run your first test**:
+
+   ```sh
+   redocly respect auto-generated.arazzo.yaml --verbose
```
DmitryAnansky: Maybe `packages/respect-core` ReadMe is not the best place to describe CLI command.
This part is more suitable for `packages/cli` as it actually contains the entry point of commands execution.
What do you think?

JLekawa (approved):

packages/respect-core/README.md
Marshevskyy: Good point. I was hesitating about whether we need to include this information here. What do you think the Getting Started section should look like in this README? Or should we remove it completely?

DmitryAnansky: Maybe add a section under
```
## Common tasks
```
in packages/cli ReadMe?

Marshevskyy: hmm, maybe just remove it from here? In the main readme of cli I've added `Quick start`

Marshevskyy: removed

PR #2145, opened 2025-06-23
fix(project): cli consider residency setting from Redocly config file

DmitryAnansky:
## What/Why/How?
CLI should consider [residency setting](https://redocly.com/docs/realm/config/residency) from redocly.yaml during login and token verification process.

## Reference
Closes: https://github.com/Redocly/redocly/issues/16251

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/utils/telemetry.ts
+++ b/packages/cli/src/utils/telemetry.ts
@@ -6,7 +6,7 @@ import { version } from './package.js';
 import { getReuniteUrl } from '../reunite/api/index.js';
 
 import type { ExitCode } from './miscellaneous.js';
-import type { ArazzoDefinition } from '@redocly/openapi-core';
+import type { ArazzoDefinition, Config } from '@redocly/openapi-core';
 import type { ExtendedSecurity } from 'respect-core/src/types.js';
 import type { Arguments } from 'yargs';
 import type { CommandOptions } from '../types.js';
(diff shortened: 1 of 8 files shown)


packages/cli/src/wrapper.ts
```diff
@@ -82,7 +83,9 @@ export function commandWrapper<T extends CommandOptions>(
       }
     } finally {
       if (process.env.REDOCLY_TELEMETRY !== 'off' && telemetry !== 'off') {
+        const residency = (argv.residency as string) || config?.rawConfig?.residency;
         await sendTelemetry({
+          residency,
```
tatomyr: We can pass config here and calculate the related fields inside `sendTelemetry` (including `residency`, `hasConfig`, and maybe some other fields).

DmitryAnansky: Sure, we can do this too. Updated

packages/cli/src/utils/telemetry.ts
```diff
@@ -60,9 +60,11 @@ export async function sendTelemetry({
       ...args
     } = argv;
     const event_time = new Date().toISOString();
+    const residency = (argv.residency as string) || config?.rawConfig?.residency;
+    const has_config = !!config?.rawConfig;
```
tatomyr:
```suggestion
```

Could you calculate it directly in the `data` object?


```ts
has_config: typeof config?.rawConfig === 'undefined' ? 'no' : 'yes',
```

This syntax better corresponds to the original config check syntax BTW: https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/config/config-resolvers.ts#L88

packages/core/src/config/types.ts
```diff
@@ -219,6 +219,7 @@ export type RawUniversalConfig = Omit<Partial<RedoclyConfig>, 'apis' | 'plugins'
 export type ResolvedConfig = Omit<RawUniversalConfig, 'apis' | 'plugins'> &
   ResolvedGovernanceConfig & {
     apis?: Record<string, ResolvedApiConfig>;
+    residency?: string;
```
tatomyr: Please put a FIXME so we can infer `residency` from the config package.

Issue #925, opened 2022-11-07
nullable in schema object must have a type sibling


adamaltman:
**Describe the bug**

From 3.0 about `nullable`:

> A true value adds "null" to the allowed type specified by the type keyword, only if type is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of null as a value. A false value leaves the specified or default type unmodified. The default value is false.

**To Reproduce**

This schema should not be valid because `foo` definition has `nullable` and that does not have an explicit sibling of `type`. 

```yaml
schema:
  type: object
  properties:
    foo:
       nullable: true
```

People also seem to think this is valid, but it isn't:

```yaml
schema:
  type: object
  properties:
    foo:
       nullable: true
       allOf:
          $ref: # some reference
```

**Expected behavior**

Let the user know where incorrect use of `nullable` has happened.

**Logs**

N/A

**OpenAPI definition**

OpenAPI 3.0.

**Redocly Version(s)**

1.0.0-beta.112

**`Node.js` Version(s)**

16

**Additional context**

n/a

adamaltman:
@AntonKozachuk you didn't close this issue. You should have put in the reference of the description "Closes #925" or "Fixes #925"

jroyal:
I want to preface this by saying that i'm not sure what the correct answer is, but I think the recent change in #938 broke validation for a specific use case.  My understanding of the spec is that the following should be ok because its not a ref, but the new validator says that type needs to be specified. 

```
api-response-single:
      type: object
      allOf:
        - $ref: "#/components/schemas/api-response-common"
        - properties:
            result:
              nullable: true
              anyOf:
                - type: object
                - type: string
  ```

Curious to hear your thoughts on it?

**Comment by marqu3z** [comment]
It seems like https://github.com/Redocly/redocly-cli/pull/938 is a breaking change.

Schemas like 
```
 veryBigNumber:
      allOf:
        - $ref: '#/components/schemas/BigInt'
        - nullable: true
```

Are not valid anymore even though the ref schema has a `type` field declared.

@adamaltman is this the intended behavior?

alexwiese:
I don't think this is actually correct.

The spec doesn't say anything about it being invalid, just that it doesn't add `null` as a possible value for that type.

[...]

PR #938, opened 2022-11-16
fix: nullable in schema object must have a type sibling

AntonKozachuk:
## What/Why/How?
We need to let the user know where incorrect use of nullable has happened.

## Reference
#925 

## Testing
Locally.

## Screenshots (optional)
N/a

## Check yourself

- [ ] Code is linted
- [ ] Tested with redoc/reference-docs/workflows
- [ ] All new/updated code is covered with tests

## Security

- [ ] Security impact of change has been considered
- [ ] Code follows company security practices and guidelines

PR #2146, opened 2025-06-24
feat: add nullable-type-sibling rule instead of validating the nullable field in the struct rule

tatomyr:
## What/Why/How?

- Reverted the previous changes made in https://github.com/Redocly/redocly-cli/pull/938
- Added a new rule to cover the `nullable` field validation

This gives users better control over APIs.

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/925

## Testing

Unit/e2e tests

## Check yourself

## Security


docs/@v2/rules/oas/nullable-type-sibling.md
```diff
@@ -0,0 +1,72 @@
...
+    Incorrect:
+      nullable: true
+    ReferencingATypeButStillIncorrect:
+      nullable: true
+      allOf:
+        - $ref: '#/components/schemas/SomeType'
+    SomeType:
+      type: string
+
+```
```
JLekawa, line 53:
```suggestion
```

docs/@v2/rules/oas/nullable-type-sibling.md
```diff
@@ -0,0 +1,72 @@
...
+
+Example of a **correct** usage:
```
JLekawa, line 55:
```suggestion
Example of usage:
```

docs/@v2/rules/oas/nullable-type-sibling.md
tatomyr: The `ReferencingATypeButStillIncorrect` one is a tricky case of incorrect usage that we sometimes get reported as a correct one, so I wanted to explicitly state it in the docs so users don't get confused. Can we somehow get this example included?

tatomyr: @JLekawa ☝️

.changeset/fuzzy-ducks-exist.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Extracted `nullable` validation from the `struct` rule into a new `nullable-type-sibling` rule for OpenAPI 3.0. This allows users to disable `nullable` validation separately from other structural checks.
```
adamaltman, line 6: Should it be `spec-nullable-type-sibling` since it is part of the spec?

docs/@v2/rules/oas/nullable-type-sibling.md
adamaltman: I agree @tatomyr  we need to show an incorrect example.

docs/@v2/rules/oas/nullable-type-sibling.md
```diff
@@ -0,0 +1,72 @@
...
+
+```yaml
+components:
+  schemas:
+    Correct:
+      type: string
+      nullable: true
```
adamaltman, line 62: Maybe we should also show a correct example when using `allOf`.

.changeset/fuzzy-ducks-exist.md
tatomyr: I'm currently working on the `spec` ruleset PR and noticed we have a few other spec-related rules that are not prefixed with `spec-` (for instance `operation-parameters-unique`). I don't think we'll rename all of them, so I don't see any reason to introduce this prefix at all.

docs/@v2/rules/oas/nullable-type-sibling.md
tatomyr: Like this?

```yaml
    CorrectNullableWithAllOf:
      type: object
      nullable: true
      allOf:
        - type: object
          properties:
            name:
              type: string
        - type: object
          properties:
            age:
              type: number
```

JLekawa (approved):

docs/@v2/rules/oas/nullable-type-sibling.md
tatomyr: Added.

PR #2184, opened 2025-07-18
chore: add internet ckecker before telemetry

SerhiiOselskyi:
## What/Why/How?
Revert changes for BatchExporter and add internet checker before any interaction with telemetry.
## Check yourself

## Security


packages/cli/src/utils/network-check.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Checks if the system has internet connectivity by attempting to reach a reliable endpoint
+ * @param timeout - Timeout in milliseconds for the connectivity check
+ * @returns Promise<boolean> - true if internet is available, false otherwise
+ */
+export async function hasInternetConnectivity(
+  timeout: number = DEFAULT_FETCH_TIMEOUT
+): Promise<boolean> {
+  // Skip network check if explicitly disabled
+  if (process.env.REDOCLY_OFFLINE === 'true' || process.env.CI === 'true') {
```
tatomyr: There's no such thing as REDOCLY_OFFLINE.

packages/cli/src/utils/network-check.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Use a reliable endpoint for connectivity check
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), timeout);
+
+    const response = await fetch('https://otel.cloud.redocly.com', {
+      method: 'GET',
+      signal: controller.signal,
+    });
+
+    clearTimeout(timeoutId);
```
tatomyr: Could you simply reuse fetch from fetch-with-timeout?

packages/cli/src/utils/network-check.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (process.env.REDOCLY_OFFLINE === 'true' || process.env.CI === 'true') {
+    return false;
+  }
+
+  try {
+    // Use a reliable endpoint for connectivity check
+    const controller = new AbortController();
+    const timeoutId = setTimeout(() => controller.abort(), timeout);
+
+    const response = await fetch('https://otel.cloud.redocly.com', {
```
tatomyr: It'd be great to reuse the const from `otel.ts` (or vice versa -- export this domain and reuse it there).

JLekawa (approved):

packages/cli/src/utils/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const OTEL_URL = 'https://otel.cloud.redocly.com';
```
tatomyr, line 1: Does it make sense to move the `DEFAULT_FETCH_TIMEOUT` here as well?

SerhiiOselskyi: Sure, why not

packages/cli/src/utils/fetch-with-timeout.ts
```diff
@@ -1,4 +1,5 @@
 import { getProxyAgent } from './proxy-agent.js';
+import { Agent } from 'undici';
```
DmitryAnansky, line 2: should we have "undici": "^6.21.1" in cli package.json dependencies then ?

SerhiiOselskyi: Added. It appears in the main node_modules, so it works fine locally and in testing.

tatomyr: Strange it didn't fail in smoke tests.

packages/cli/src/utils/telemetry.ts
```diff
@@ -54,6 +55,12 @@ export async function sendTelemetry({
     if (!argv) {
       return;
     }
+
+    const hasInternet = await hasInternetConnectivity(1000);
```
DmitryAnansky: Can you name somehow `1000` or add comment?

tatomyr: Maybe rename the function so the argument's meaning become obvious? Something like that?

```suggestion
    const hasInternet = await respondWithinMs(1000);
```

tatomyr (approved):

PR #2186, opened 2025-07-20
chore: speed up test by removing slow external call

RomanHotsiy:
## What/Why/How?

One test is slow as example.com responds very slow.

## Check yourself

## Security

### DIFF
--- a/resources/petstore-with-errors.yaml
+++ b/resources/petstore-with-errors.yaml
@@ -29,12 +29,10 @@ paths:
             example:
               summary: A foo example
               value: { 'foo': 'bar' }
-              externalValue: 'http://example.org/foo.json'
             examples:
               foo:
                 summary: A foo example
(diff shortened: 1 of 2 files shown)


DmitryAnansky (approved):

resources/petstore-with-errors.yaml
```diff
@@ -29,12 +29,10 @@ paths:
             example:
               summary: A foo example
               value: { 'foo': 'bar' }
-              externalValue: 'http://example.org/foo.json'
```
tatomyr, line 32: Have you considered replacing it with something that will fail immediately instead (e.g., `localhost`)?

RomanHotsiy: yes, but it doesn't make sense. this doesn't test anything meaningful.

tatomyr (approved):

PR #2187, opened 2025-07-20
fix: improve perf of config parsing

RomanHotsiy:
## What/Why/How?

The `normalizeVisitors` has a cubic asymptotic complexity.
It is very slow for config type tree which has 1000+ types in it. So resolving config multiple times  is very slow. I moved the normalizeVisitors call to be static so it's not recomputed every time.

I also fixed the condition in `normalizeVisitors` to improve the performance .

## Reference

Noticed while debugging performance of one project.

## Testing

No changes, all unit and e2e tests passing.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/visitors.ts
+++ b/packages/core/src/visitors.ts
@@ -575,7 +575,6 @@ export function normalizeVisitors<T extends BaseVisitor>(
       if (parentContext) {
         addWeakNodes(ruleConf, parentContext.type, types[typeName], parentContext);
       }
-      3;
       if (visitorEnter || isObjectVisitor) {
         if (visitorEnter && typeof visitorEnter !== 'function') {
           throw new Error('DEV: should be function');


JLekawa (approved):

packages/core/src/visitors.ts
```diff
@@ -555,21 +556,26 @@ export function normalizeVisitors<T extends BaseVisitor>(
...
-      if (typeof typeVisitor === 'object') {
+      if (hasNestedVisitors) {
         normalizeVisitorLevel(ruleConf, typeVisitor, context, depth + 1);
       }
 
       if (parentContext) {
         addWeakNodes(ruleConf, parentContext.type, types[typeName], parentContext);
       }
-
+      3;
```
tatomyr: What is it?

RomanHotsiy:
```suggestion
```

tatomyr:
Tested internally: https://github.com/Redocly/redocly/pull/16846.

PR #2202, opened 2025-07-21
chore!: remove processRawConfig from loadConfig

tatomyr:
## What/Why/How?

Removed the deprecated `processRawConfig` from `loadConfig`. 
All the arguments that `processRawConfig` was receiving before, are in the `Config` instance now. 
There should be no user impact (expect for internal usage).

## Check yourself

## Security

### DIFF
--- a/packages/core/src/config/load.ts
+++ b/packages/core/src/config/load.ts
@@ -1,6 +1,5 @@
 import * as fs from 'node:fs';
 import * as path from 'node:path';
-import { ConfigValidationError } from './utils.js';
 import { resolveConfig } from './config-resolvers.js';
 import {
   BaseResolver,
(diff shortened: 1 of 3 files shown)


packages/core/src/config/load.ts
```diff
@@ -11,28 +10,14 @@ import {
...
     customExtends?: string[];
-    /** Deprecated */
-    processRawConfig?: RawConfigProcessor;
     externalRefResolver?: BaseResolver;
   } = {}
 ): Promise<Config> {
-  const {
-    configPath = findConfig(),
-    customExtends,
-    processRawConfig,
```
DmitryAnansky, line 33: Please make sure you adapt mono-repo when this changes applied.
I can see some usage in the codebase.

PR #2207, opened 2025-07-22
chore: remove duplications

tatomyr:
## What/Why/How?

WIP

## Check yourself

## Security


packages/cli/src/commands/split/index.ts
```diff
@@ -148,19 +156,23 @@ function traverseDirectoryDeepCallback(
 }
 
 export function crawl(object: unknown, visitor: (node: Record<string, unknown>) => void) {
-  if (!isObject(object)) return;
```
tatomyr, line 151: The previous `isObject` was also allowing arrays.
It was working because `Object.keys` for arrays produce an array of its keys 😄.

PR #2208, opened 2025-07-22
refactor!: replace enums with types

tatomyr:
## What/Why/How?

Addressing FIXMES before 2.0 GA: 
- Replaced `SpecVersion`, `SpecMajorVersion`, `OPENAPI3_METHOD`, and `OPENAPI3_COMPONENT` enums with types
- Removed `OasVersion` enum

## Check yourself

## Security

### DIFF
--- a/packages/core/src/bundle.ts
+++ b/packages/core/src/bundle.ts
@@ -28,11 +28,6 @@ import type { OasRef } from './typings/openapi.js';
 import type { Document, ResolvedRefMap } from './resolve.js';
 import type { CollectFn } from './utils.js';
 
-export enum OasVersion {
-  Version2 = 'oas2',
-  Version3_0 = 'oas3_0',
-  Version3_1 = 'oas3_1',
-}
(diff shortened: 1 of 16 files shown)


JLekawa (approved):

packages/core/src/bundle.ts
```diff
@@ -402,22 +389,22 @@ function makeBundleVisitor(
...
+        } else if (version === 'oas2') {
           components = root;
-        } else if (version === SpecMajorVersion.Async2) {
+        } else if (version === 'async2') {
           components = root.components = root.components || {};
-        } else if (version === SpecMajorVersion.Async3) {
+        } else if (version === 'async3') {
           components = root.components = root.components || {};
-        } else if (version === SpecMajorVersion.Arazzo1) {
+        } else if (version === 'arazzo1') {
```
DmitryAnansky, line 400: Could you please explain why do you want to avoid using some const here is all those places?

PR #2209, opened 2025-07-23
fix: fix resolved config had extra transformations

RomanHotsiy:
## What/Why/How?

`resolvedConfig` had assertions transformed for internal use but resolvedConfig is supposed to be "raw" (with only refs and extends resolved).

I moved the grouping into the `Config` instance.

## Reference

Noticed while testing in Realm.

## Testing
All tests pass, had to adjust snapshots and one test.

## Check yourself

## Security

### DIFF
--- a/.changeset/modern-fishes-float.md
+++ b/.changeset/modern-fishes-float.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Fixed config resolver grouping assertions instead of returning unchanged rules.
+Fixed and issue where the config resolver grouped assertions instead of returning unchanged rules.


packages/core/src/config/__tests__/fixtures/resolve-config/local-config-with-wrong-custom-function.yaml
```diff
@@ -6,8 +6,9 @@ rules:
     property: description
     message: Tag description must have at least 3 words.
     severity: error
-    test-plugin/checkWordsCount2:
-      min: 3
+    assertions:
+      test-plugin/checkWordsCount2:
+        min: 3
```
tatomyr: Old format. It should be 

```yaml
    subject: 
      type: Tag
      property: description
    message: Tag description must have at least 3 words.
    severity: error
    assertions:
      test-plugin/checkWordsCount2:
        min: 3
```

packages/core/src/config/__tests__/config-resolvers.test.ts
```diff
@@ -286,21 +291,23 @@ describe('resolveConfig', () => {
       );
     }
 
-    expect(asserts['test-plugin/checkWordsCount' as keyof Asserts]).toBeDefined();
+    // expect(asserts['test-plugin/checkWordsCount' as keyof Asserts]).toBeDefined();
```
tatomyr: Remove the comment?

packages/core/src/config/__tests__/config-resolvers.test.ts
```diff
@@ -259,11 +261,13 @@ describe('resolveConfig', () => {
     const rootOrApiRawConfig = {
       extends: ['local-config-with-custom-function.yaml'],
     };
-    const { plugins } = await resolveConfig({
+    const { plugins, resolvedConfig } = await resolveConfig({
       rawConfigDocument: makeDocument(rootOrApiRawConfig, configPath),
       configPath,
     });
 
+    new Config(resolvedConfig, { plugins });
```
tatomyr, line 270: What does this do?

packages/core/src/config/__tests__/config-resolvers.test.ts
RomanHotsiy: yes

packages/core/src/config/__tests__/config-resolvers.test.ts
RomanHotsiy: The assertions registration happens inside the constructor now, I'll add comment

packages/core/src/config/__tests__/fixtures/resolve-config/local-config-with-wrong-custom-function.yaml
RomanHotsiy: I will fix now.

packages/core/src/config/__tests__/config-resolvers.test.ts
tatomyr: Yes, but I don't see any related expects in the test. Or am I'm missing something?

packages/core/src/config/__tests__/config-resolvers.test.ts
RomanHotsiy: this was relying on the previous test I think

.changeset/modern-fishes-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed config resolver grouping assertions instead of returning unchanged rules.
```
JLekawa:
```suggestion
Fixed and issue where the config resolver grouped assertions instead of returning unchanged rules.
```

tatomyr (approved):

PR #2213, opened 2025-07-24
fix: fix plugins in resolvedConfig

RomanHotsiy:
## What/Why/How?

`plugins` in `resolvedConfig` were not reflecting all the nested plugins. While other rules are merged I think plugins should be merged too.

## Reference
It caused a bug in Realm.

## Testing
Updated tests.
Tested with Realm.

## Check yourself

## Security

### DIFF
--- a/.changeset/funny-mayflies-knock.md
+++ b/.changeset/funny-mayflies-knock.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": major
+---
+
+Replaced the `SpecVersion`, `SpecMajorVersion`, `OPENAPI3_METHOD`, and `OPENAPI3_COMPONENT` enums with types for improved flexibility and type safety.
+Removed the unused `OasVersion` enum.
--- a/packages/cli/src/__tests__/commands/join.test.ts
+++ b/packages/cli/src/__tests__/commands/join.test.ts
@@ -4,7 +4,6 @@ import {
   detectSpec,
   getTotals,
   loadConfig,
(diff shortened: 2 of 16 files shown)


packages/core/src/config/types.ts
```diff
@@ -225,6 +225,7 @@ export type RawUniversalConfig = Omit<Partial<RedoclyConfig>, 'apis' | 'plugins'
 export type ResolvedConfig = Omit<RawUniversalConfig, 'apis' | 'plugins'> &
   ResolvedGovernanceConfig & {
     apis?: Record<string, ResolvedApiConfig>;
+    plugins?: string[];
```
yarokon, line 228: Form tests I see that plugins always defined. Should we change type or update logic?

RomanHotsiy: yes, will fix now

RomanHotsiy: Fixed logic and updated tests.

tatomyr (approved):

packages/core/src/config/__tests__/load.test.ts
```diff
@@ -7,6 +7,7 @@ import { BaseResolver } from '../../resolve.js';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { createRequire } from 'node:module';
```
tatomyr: Do we need this?

PR #2215, opened 2025-07-24
chore!: remove the deprecated theme config property

tatomyr:
## What/Why/How?

Removed support for the deprecated `theme` property of Redocly config.
All the properties of `theme` are now available in the config root.

## Check yourself

## Security


JLekawa (approved):

JLekawa:
Closes [#15452](https://github.com/Redocly/redocly/issues/15452)

adamaltman (approved):

javierjulio:
This doesn't seem to work as expected in preview only. Running `redocly check-config` reports that my config is valid. When I run `redocly preview` the page reports an error in a modal from clicking an alert near the top right of the screen.

Screenshot 2025-07-25 at 11 06 26 AM

tatomyr:
@javierjulio what's the version of Redocly CLI you're using?

javierjulio:
@tatomyr sorry, I should have clarified. This is regarding the new v2 release as I was working on that update today. Thanks.

tatomyr:
Interesting. That's looks like something on the Realm's side. What's the version of Realm? Or it's just latest?

cc @volodymyr-rutskyi @Redocly/lightsabers

javierjulio:
I assume it's just the latest because I installed redocly v2. I tried setting a different product for the preview command but same error came up.

tatomyr:
@javierjulio I notified the owners of Realm. For now, I suggest you still use the root `theme` and ignore the check-config warnings until the validation in Realm is fixed.

tatomyr:
@javierjulio what do you use that options for? Configuring Redoc?

AlexVarchuk:
@javierjulio  For now, you could use `codeSamples` option in `redocly.yaml`. The docs you could find [here](https://redocly.com/docs/realm/config/openapi/code-samples)

javierjulio:
@tatomyr I noticed with v2 that the preview was all new and that it had on a list of languages in the middle right of the page. I just wanted to customize the order or which languages would only be shown but that didn't work due to the result above.

Screenshot 2025-07-29 at 7 30 50 PM

@AlexVarchuk thank you. That configuration example does work. I went off [this configuration example which fails](https://redocly.com/docs/cli/configuration#example-redocly-configuration-file) because it uses `generateCodeSamples` whereas what you linked to has `codeSamples` which works.

tatomyr:
@javierjulio yes, v2 uses our new Redoc.

PR #2245, opened 2025-08-05
fix(respect): binary data is destroyed when consumed with text

DmitryAnansky:
## What/Why/How?

Response binary data is destroyed when consumed with .text() and it wasn't possible to recreate, for example png image when working with the `Respect` run function results.

## Screenshots (optional)

Screenshot 2025-08-06 at 15 44 21

## Check yourself

## Security


packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -358,6 +358,14 @@ export class ApiFetcher implements IFetcher {
     const [responseContentType] = fetchResult.headers.get('content-type')?.split(';') || [
       'application/json',
     ];
+
+    // Handle response body based on content type
+    if (isBinaryContentType(responseContentType)) {
+      const arrayBuffer = await fetchResult.arrayBuffer();
+      responseBody = arrayBufferToBase64(arrayBuffer);
```
RomanHotsiy: do we need encoding to base64?

DmitryAnansky: I think you are right, let me test one scenario and if it works - I remove the encoding.

DmitryAnansky: Ok, so we will keep initial response type, this requires some adjustments for verbose / json / har logs.
We also should use deepCopy instead os JSON parse - stringify.

packages/cli/src/utils/json-stringify-with-array-buffer.ts
```diff
@@ -0,0 +1,31 @@
+export function jsonStringifyWithArrayBuffer(obj: any, space?: string | number): string {
+  return JSON.stringify(
+    obj,
+    (key, value) => {
+      if (value instanceof ArrayBuffer) {
```
RomanHotsiy, line 7: we should add some size limit here, we probably don't want to serialize more than 1mb as it will be really inefficient. Better to show "File to big blah blah".

DmitryAnansky: Updated, please take a look

Screenshot 2025-08-08 at 11 30 33

RomanHotsiy (approved):

.changeset/unlucky-feet-give.md
```diff
@@ -0,0 +1,7 @@
+---
+"@redocly/respect-core": patch
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed binary response data in `Respect` results by properly encoding as base64.
```
tatomyr:
```suggestion
Fixed binary response data in `Respect` results by properly encoding it as base64.
```

packages/cli/src/utils/json-stringify-with-array-buffer.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (value.byteLength > MAX_SIZE) {
+          return {
+            __type: 'ArrayBuffer',
+            data: `File too large to serialize (${value.byteLength} bytes). Maximum allowed size is ${MAX_SIZE} bytes.`,
+            byteLength: value.byteLength,
+          };
+        }
+
+        // Convert ArrayBuffer to base64 string for JSON serialization
```
tatomyr: Please remove redundant comments.

packages/respect-core/src/modules/logger-output/verbose-logs.ts
```diff
@@ -18,7 +18,13 @@ export function getVerboseLogs({
   if (headerParams && Object.keys(headerParams).length > 0) {
     verboseLogs.headerParams = headerParams;
   }
-  if (body && (body instanceof FormData || body instanceof File || Object.keys(body).length > 0)) {
+  if (
+    body &&
+    (body instanceof FormData ||
+      body instanceof File ||
+      Object.keys(body).length > 0 ||
+      body instanceof ArrayBuffer)
```
tatomyr:
```suggestion
      body instanceof ArrayBuffer) ||
      Object.keys(body).length > 0
```

Could it be put that way?

packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -358,8 +357,18 @@ export class ApiFetcher implements IFetcher {
     const [responseContentType] = fetchResult.headers.get('content-type')?.split(';') || [
       'application/json',
     ];
+
+    // Handle response body based on content type
```
tatomyr: Do we need this comment?

packages/respect-core/src/utils/deepCopy.ts
```diff
@@ -0,0 +1,46 @@
...
+    return obj;
+  }
+
+  // Handle circular references
+  if (visited.has(obj)) {
+    return visited.get(obj);
+  }
+
+  // Handle special object types that should be returned as-is
+  if (
```
tatomyr: Is it the same as https://github.com/Redocly/redocly-cli/pull/2245/files#diff-54e095d192c7833f14cb467989885a55d3c5c9912c4454797737996ce1ac6fd8R46? 
Could you reuse it?

packages/respect-core/src/utils/deepCopy.ts
```diff
@@ -0,0 +1,46 @@
...
+      copiedArray[i] = deepCopy(obj[i], visited);
+    }
+    return copiedArray;
+  }
+
+  // Handle regular objects
+  const copied = {} as any;
+  visited.set(obj, copied);
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
tatomyr:
```suggestion
    if (Object.hasOwn(obj, key)) {
```

tatomyr (commented):
Had a quick look. Will take another round a bit later.

packages/respect-core/src/utils/deepCopy.ts
DmitryAnansky: Our compiler does not support this
```
Property 'hasOwn' does not exist on type 'ObjectConstructor'
```

packages/respect-core/src/utils/deepCopy.ts
DmitryAnansky: nor really, for verbose logs we would like to control the output and only display what we expect.

packages/respect-core/src/utils/deepCopy.ts
tatomyr: Well, then we have to update our compiler settings. Let's do that separately.

.changeset/unlucky-feet-give.md
tatomyr: Have you applied the change?

packages/respect-core/src/utils/deepCopy.ts
tatomyr: But please put a FIXME to change it.

packages/respect-core/src/modules/logger-output/verbose-logs.ts
DmitryAnansky: no, this logic would be different then.

packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -17,6 +17,7 @@ import { getResponseSchema } from '../modules/description-parser/index.js';
 import { collectSecretFields } from '../modules/flow-runner/index.js';
 import { parseWwwAuthenticateHeader } from './digest-auth/parse-www-authenticate-header.js';
 import { generateDigestAuthHeader } from './digest-auth/generate-digest-auth-header.js';
+import { isBinaryContentType } from './isBinaryContentType.js';
```
tatomyr: Please use kebab-case for filenames.

packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -358,8 +357,17 @@ export class ApiFetcher implements IFetcher {
     const [responseContentType] = fetchResult.headers.get('content-type')?.split(';') || [
       'application/json',
     ];
+
+    if (isBinaryContentType(responseContentType)) {
+      const arrayBuffer = await fetchResult.arrayBuffer();
+      responseBody = arrayBuffer;
```
tatomyr: Do we need arrayBuffer?

packages/respect-core/src/utils/deepCopy.ts
```diff
@@ -0,0 +1,46 @@
+export function deepCopy(obj: any, visited = new WeakMap()): any {
+  if (obj === null || typeof obj !== 'object') {
```
tatomyr:
```suggestion
  if (Array.isArray(obj) || isPlainObject(obj)) {
```

packages/respect-core/src/utils/deepCopy.ts
```diff
@@ -0,0 +1,46 @@
+export function deepCopy(obj: any, visited = new WeakMap()): any {
```
tatomyr: It's not always an object.

packages/respect-core/src/utils/deepCopy.ts
```diff
@@ -0,0 +1,45 @@
+import { isPlainObject } from '@redocly/openapi-core';
```
tatomyr: Please rename the file.

tatomyr (approved):

PR #2249, opened 2025-08-12
fix: pass externalRefResolver to nested workflow context

DmitryAnansky:
## What/Why/How?
Ensure `externalRefResolver`  and `skipLint` options are correctly passed to nested workflow contexts.
## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/modules/flow-runner/runner.ts
+++ b/packages/respect-core/src/modules/flow-runner/runner.ts
@@ -251,7 +251,6 @@ export async function resolveWorkflowContext(
   ctx: TestContext,
   config: Config
 ) {
-  const { logger } = ctx.options;
   const sourceDescriptionId =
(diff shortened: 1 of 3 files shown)


packages/respect-core/src/modules/flow-runner/runner.ts
```diff
@@ -282,6 +282,8 @@ export async function resolveWorkflowContext(
           envVariables: ctx.options.envVariables,
           logger,
           fetch: ctx.options.fetch,
+          externalRefResolver: ctx.options?.externalRefResolver,
```
tatomyr: Are `ctx.options` optional? I see it's used without the optional chaining elsewhere. Also, could we just spread all `...ctx.options` here instead of putting every single property?

DmitryAnansky: there are some extra info passed in this case, but lets do it, to keep it sync

tatomyr (approved):

PR #2254, opened 2025-08-12
feat: updated the exported `run` function in `Respect` to utilize the max steps count logic

DmitryAnansky:
## What/Why/How?

Updated the exported `run` function in `Respect` to utilize the max steps count logic.

## Check yourself

## Security


.changeset/thick-shrimps-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/respect-core": patch
+---
+
+Updated the exported `run` function in `Respect` to utilize the max steps count logic.
```
JLekawa:
```suggestion
Updated the exported `run` function in `Respect` to utilize the `maxSteps` count logic.
```

packages/respect-core/src/modules/flow-runner/runner.ts
```diff
@@ -70,12 +91,13 @@ async function runWorkflows(testDescription: TestDescription, options: AppOption
...
     // run dependencies workflows first
     if (workflow.dependsOn?.length) {
-      await handleDependsOn({ workflow, ctx, config: options.config });
+      await handleDependsOn({ workflow, ctx, config: options.config, executedStepsCount });
     }
 
     const workflowExecutionResult = await runWorkflow({
       workflowInput: workflow.workflowId,
       ctx,
+      executedStepsCount,
```
tatomyr, line 100: This looks like it could reside inside `ctx`.

packages/respect-core/src/modules/flow-runner/runner.ts
```diff
@@ -212,10 +236,12 @@ async function handleDependsOn({
   workflow,
   ctx,
   config,
+  executedStepsCount,
```
tatomyr, line 239: Same.

packages/respect-core/src/run.ts
```diff
@@ -32,6 +32,7 @@ export type RespectOptions = {
 };
 
 export async function run(options: RespectOptions): Promise<RunFileResult[]> {
+  const executedStepsCount = { value: 0 };
```
tatomyr, line 35: What is the need for making it an object? Wouldn't a bare number be sufficient?

packages/respect-core/src/run.ts
```diff
@@ -57,13 +59,15 @@ export async function run(options: RespectOptions): Promise<RunFileResult[]> {
...
+  executedStepsCount,
   collectSpecData,
 }: {
   options: RunOptions;
   startedAt: number;
+  executedStepsCount: ExecutedStepsCount;
   collectSpecData?: CollectFn;
 }): Promise<RunFileResult> {
-  const result = await runTestFile(options, collectSpecData);
+  const result = await runTestFile({ options, collectSpecData, executedStepsCount });
```
tatomyr, line 70: I think we can declare [this line](https://github.com/Redocly/redocly-cli/pull/2254/files#diff-506a59c0d66ea6dbc12dac72910481ff7f53671c50a900e998285c236816a088R35) here.

tatomyr: Or [here](https://github.com/Redocly/redocly-cli/pull/2254/files#diff-cdff60738b3360481be05ac2a341b18d47ced550d3beea9ef1cf1ecfcefcf89bR33).

tatomyr: Or even [here](https://github.com/Redocly/redocly-cli/pull/2254/files#diff-cdff60738b3360481be05ac2a341b18d47ced550d3beea9ef1cf1ecfcefcf89bR63).

tatomyr: And therefore pass the `executedStepsCount` down fewer floors.

DmitryAnansky: its better to keep it on the top level to have clear vision when it is started.

packages/respect-core/src/modules/flow-runner/runner.ts
DmitryAnansky: discussed offline

packages/respect-core/src/modules/flow-runner/runner.ts
DmitryAnansky: discussed offline

packages/respect-core/src/run.ts
DmitryAnansky: discussed offline

DmitryAnansky: number is passed by value and we needed object to pass by reference and mutate deeper in the path

tatomyr (approved):

PR #2262, opened 2025-08-14
fix: loadConfig function should work in browser

AlexVarchuk:
## What/Why/How?
When using `loadConfig` function to get the config for the browser environment, we get an error: "`path.extname` is not a function"
## Screenshots (optional)
image (16)

Screenshot 2025-08-14 at 14 53 34

## Check yourself

## Security

### DIFF
--- a/.changeset/popular-tomatoes-draw.md
+++ b/.changeset/popular-tomatoes-draw.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Improved browser compatibility for `loadConfig`.


packages/core/src/utils.ts
```diff
@@ -212,10 +211,8 @@ export function isCustomRuleId(id: string) {
 }
 
 export function doesYamlFileExist(filePath: string): boolean {
-  return (
-    (path.extname(filePath) === '.yaml' || path.extname(filePath) === '.yml') &&
-    !!fs?.existsSync?.(filePath)
-  );
+  const ext = filePath.substr(filePath.lastIndexOf('.'));
```
tatomyr, line 214: The `substr` is deprecated already. Also, this doesn't handle the case when a file is named `file` or `.env`. It would be better to create a utility function and write a couple of tests for it.

PR #2271, opened 2025-08-22
chore: export AsyncAPI types externally

tatomyr:
## What/Why/How?

Export AsyncAPI descriptions types to be able to refer to them externally and slightly rearranged exports.

 

## Security


packages/core/src/dequal.ts
```diff
@@ -0,0 +1,36 @@
+/**
```
tatomyr, line 1: Moved the code without changes.

DmitryAnansky (approved):

Issue #2280, opened 2025-09-04
Export `Oas3Visitor` type definitions


jooola:
**Is your feature request related to a problem? Please describe.**

Having access to e.g. the visitor config types definition, would greatly help when developing a local plugin. I always struggle to discover how the API is shaped and what possibilities I have to accomplish a transformation.

Other types for anything a user can plug into redocly.

**Describe the solution you'd like**

Export more types that would be helpful for developing local plugins.

**Describe alternatives you've considered**

I tried digging into the `Plugin` type that is exported, but I didn't manage to access all the types that I needed.

```ts
import type { Plugin as RedoclyPlugin } from "@redocly/openapi-core";

export type Plugin = RedoclyPlugin;
export type Decorators = NonNullable<Plugin["decorators"]>
export type Oas3DecoratorsSet = NonNullable<Decorators["oas3"]>
```

**Additional context**

```
 "@redocly/cli": "2.0.8",
```

tatomyr:
Hey @jooola, would [these types](https://github.com/Redocly/redocly-cli/pull/2282) work for you? Any other types you'd like to use?

jooola:
@tatomyr Thanks for such a quick answer, I responded in the PR directly: https://github.com/Redocly/redocly-cli/pull/2282#issuecomment-3258227847

tatomyr:
@jooola released in v2.1.0.

PR #2282, opened 2025-09-05
chore: export visitor type definitions to be able to use them in custom plugins

tatomyr:
## What/Why/How?

Improved developer experience for custom plugin authors by exposing more type definitions, making it easier to write and maintain custom rules and visitors.

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/2280

## Check yourself

## Security

### DIFF
--- a/.changeset/whole-brooms-taste.md
+++ b/.changeset/whole-brooms-taste.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Improved developer experience for custom plugin authors by exposing more type definitions, making it easier to write and maintain custom rules and visitors.
+Exposed additional type definitions, making it easier for custom plugin authors to write and maintain custom rules and visitors.


jooola:
On first look, there should be everything that I need in there.

Should we consider also adding the "Flat" visitor type? for example `Oas3FlatVisitor` ? Or does `Oas3Visitor` cover that case already? 

This way I can define my plugin in one file:
```ts
export default function LocalPlugins(): Plugin {
  return {
    id: "local",
    typeExtension: {
      oas3(types) {
        types = XNavPath(types);
        return types;
      },
    },
    rules: {
      oas3: {
        "ref-in-allowed-dir": RefInAllowedDir,
      },
    },
    decorators: {
      oas3: {
        "embed-code-examples": EmbedCodeExamples,
        "remove-enum-internal": RemoveEnumInternal,
        "merge-enum-internal": MergeEnumInternal,
        "merge-templates": MergeTemplates,
      },
    },
  };
}
```

And then, in another file, define e.g. a decorator:

```ts
const key = "x-enumInternal";

export const RemoveEnumInternal = (): Oas3FlatVisitor=> {
  return {
    Schema: {
      leave(schema: any) {
        if (schema.type == "string" && key in schema) {
          delete schema[key];
        }
      },
    },
  };
};

export const MergeEnumInternal = (): Oas3FlatVisitor => {
  return {
    Schema: {
      leave(schema: any) {
        if (schema.type == "string" && key in schema) {
          schema.enum ??= [];
          schema.enum.push(...schema[key]);
          delete schema[key];
        }
      },
    },
  };
};
```

jooola:
I might also need `NodeType`, see this other example:
```ts
export function XNavPath(types: Record<string, NodeType>) {
  return {
    ...types,
    Tag: {
      ...types.Tag,
      properties: {
        ...types.Tag.properties,
        "x-nav-path": {
          type: "string",
          description:
            "Path used to build the tags nested navigation in the API spec viewer. Uses a slash as the divider.",
          pattern: "^[a-zA-Z ]+(/[a-zA-Z ])*$",
          example: "Foo/Bar",
        },
      },
      required: [...types.Tag.required, "x-nav-path"],
    },
  };
}

```

jooola:
Actually would it be possible to export every type and nested type used for the plugins? I come to realize that I might also need `UserContext`, `Location`, and more.

tatomyr:
> Should we consider also adding the "Flat" visitor type?

No, it's not needed for writing plugins unless I'm missing something. Just use the regular visitor type.


> Actually would it be possible to export every type and nested type used for the plugins?

I'd like to keep it constrained if possible because some of the types are intended for internal usage. We can add more types if you need them later.

> might also need UserContext, Location, and more.

Good catch. I'll update the PR accordingly next week.

jooola:
I added the following files to my tsconfig to workaround the missing exports:
```
{
  "compilerOptions": {
    "paths": {
      "@redocly/openapi-core/types": ["./node_modules/@redocly/openapi-core/lib/types.ts"],
      "@redocly/openapi-core/visitors": ["./node_modules/@redocly/openapi-core/lib/visitors.ts"]
    }
  }
}
```

And I was able to type my plugin using only:
```
import type { Oas3Decorator } from "@redocly/openapi-core/visitors";
import type { Oas3Rule } from "@redocly/openapi-core/visitors";
import type { Oas3Operation, Oas3PathItem } from "@redocly/openapi-core";
import type { Plugin } from "@redocly/openapi-core";
import type { TypesExtensionFn } from "@redocly/openapi-core/lib/config/config";
 ```

So I think this PR is already a great improvement.

Regarding the `UserContext`, `Location` and more, I did not yet find a need for it, but the moment I will break down my functions into smaller ones to reduce complexity, I might need them.

File: .changeset/whole-brooms-taste.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Improved developer experience for custom plugin authors by exposing more type definitions, making it easier to write and maintain custom rules and visitors.
```
JLekawa:
```suggestion
Exposed additional type definitions, making it easier for custom plugin authors to write and maintain custom rules and visitors.
```

DmitryAnansky (approved):

PR #2286, opened 2025-09-08
fix: return on invalid ref pointer in remove-unused-components

jooola:
## What/Why/How?

When bundling a specification with `--remove-unused-components --dereferenced`, and the specification has an invalid reference:

```yml
  parameters:
    - $ref: "../parameters.yml#/QueryProjectCreatedBefore.fail"
```

The process fails with:

```
  $redocly bundle my-api --dereferenced --remove-unused-components --output dist/my-api.json
  bundling src/my-api.yml using configuration for api 'my-api'...
  Something went wrong when processing /work/openapi-spec/src/openapi.yml:

    - Cannot read properties of undefined (reading 'split')
```

This change ensures that when the absolute pointer cannot be split into 2 parts, the visitor returns and let redocly report the `Can't resolve $ref` error.

Note that not all invalid references run into this bug, for example the following reference will correctly be reported as an error:

```yml
  parameters:
    - $ref: "../parameters.yml#/QueryProjectCreatedBefore/fail"
```

## Reference

```yml
  parameters:
    - $ref: "../parameters.yml#/QueryProjectCreatedBefore.fail"
 ```

## Testing

Couldn't find a way to test this bug with the existing tests functions. To fully test this, 2 files are required, one referencing the other. 

## Check yourself

## Security

### DIFF
--- a/.changeset/eager-things-train.md
+++ b/.changeset/eager-things-train.md
@@ -2,4 +2,4 @@
 "@redocly/openapi-core": patch
 ---
 
-Fixed undefined variable use in the `remove-unused-components` decorator, which prevented an invalid reference error to be reported.
+Fixed undefined variable used in the `remove-unused-components` decorator, which prevented an invalid reference error from being reported.

### CODE REVIEW

**Review by RomanHotsiy — APPROVED** [review]

**Comment by tatomyr** [comment]
Thanks for contributing, @jooola! Could you add a unit or an E2E test to cover this case?

**Comment by jooola** [comment]
@tatomyr Done, I hope the test are in the correct location.

**Comment by tatomyr** [comment]
@jooola could you also add the corresponding fix for OAS2 [here](https://github.com/Redocly/redocly-cli/blob/3d7fa170072799e80ae3c6272265ebdbf30b55c5/packages/core/src/decorators/oas2/remove-unused-components.ts#L63)? Otherwise looks good to me.

File: .changeset/eager-things-train.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed undefined variable use in the `remove-unused-components` decorator, which prevented an invalid reference error to be reported.
```
JLekawa:
```suggestion
Fixed undefined variable used in the `remove-unused-components` decorator, which prevented an invalid reference error from being reported.
```

JLekawa (approved):

tatomyr:
Or we will do it ourselves 🙂. Anyway, thank you for your contribution!
cc @DmitryAnansky

el2iot2:
Heads up, a recent setup was working on ran afoul of (possibly? probably?) this change. Got an error like:
image

But in VSCode with redocly extension, it shows no errors and is valid:
image
(example invalid entry supplied to show validation was on and working)

Had several such `apis` with `$ref` usages side by side, and _only one_ was marked invalid. 

By commenting `remove-unused-components` in `redocly.yaml`, the validation error went away.

I suspect that the remove references somehow dropped the `securitySchemes` in a way that made it look invalid to a later pass??

Anyway...submitted for consideration.

CC @jooola

Issue #2287, opened 2025-09-09
Extend Respect to check defined `response` headers to matches what is defined in the provided OpenAPI description


DmitryAnansky:
**Is your feature request related to a problem? Please describe.**
In OpenAPI descriptions it is possible to describe [response headers](https://spec.openapis.org/oas/latest.html#response-object), like in this example:

```yaml
      responses:
        '200':
          description: successful operation
          headers:
            X-Rate-Limit:
              description: calls per hour allowed by the user
              schema:
                type: integer
                format: int32
            X-Expires-After:
              description: date in UTC when token expires
              schema:
                type: string
                format: date-time
```

Sometimes API do not return the response headers that are described in OpenAPI document, this can happen due to API issue, or when developer is working with mock APIs.

[Respect](https://redocly.com/docs/respect) already perform schema validation for described response body by default in case Step is connected to OpenAPI operation.

<img width="462" height="103" alt="Image" src="https://github.com/user-attachments/assets/86d0fbcd-d338-4060-a3a0-a94fd0db3f1a" />

**Describe the solution you'd like**

- Introduce the new check to [Respect](https://redocly.com/docs/respect), called `RESPONSE_HEADER_CHECK` and validate Step response headers to match OpenAPI description when Step is connected to operation by using `operationId`.
Response can contain some extra headers, this should be ok. But it must include required headers described in OpenAPI spec.
- Make this check to be configurable with `severity` cli options.

So if, for example OpenAPI response contains required headers, Respect should check that`X-Request-ID` and `X-Rate-Limit-Remaining` were returned from API.

```yaml
openapi: 3.1.0
info:
  title: Example API
  version: 1.0.0
paths:
  /items:
    get:
      summary: Get items
      responses:
        '200':
          description: Successful response
          headers:
            X-Request-ID:
              description: Correlation ID for the request
              required: true
              schema:
                type: string
            X-Rate-Limit-Remaining:
              description: Requests remaining in the current rate limit window
              required: false
              schema:
                type: integer
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
```

**Describe alternatives you've considered**

Do nothing.

PR #2288, opened 2025-09-09
feat: improve respect digest auth error handling

DmitryAnansky:
## What/Why/How?
During training sessions an interesting case was discovered, related to `digest` auth:
Screenshot 2025-09-09 at 14 47 54

The original issue when we get 200 response for the first digest call is related to custom mock implementation of the digest server. 
In short, when `digest` auth performed there are two calls:
- First should get 401 response with specific 'www-authenticate' header.
- Second should be send with some specific calculation, based on first call result.

All works as expected, but the screenshot displays not clear information for the user with the `network error`.
We improving this flow by adding two detailed errors for `digest`:
- STATUS_CODE_CHECK - when the response from the first call is not 401.
- RESPONSE_HEADER_CHECK - we introduce a new CHECK to provide more information about errors that happened due to some header issues. Also there is a [suggestion to check described headers from OpenAPI description.](https://github.com/Redocly/redocly-cli/issues/2287).

## Check yourself

## Security


.changeset/gold-houses-leave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/respect-core": patch
+"@redocly/cli": patch
+---
+
+Improve error reporting and handling for Digest authentication failures in Respect, providing clearer feedback when required headers or status codes are missing.
```
JLekawa:
```suggestion
Improved error reporting and handling for Digest authentication failures in Respect. Now users receive clearer feedback when required headers or status codes are missing.
```

JLekawa (approved):

packages/respect-core/src/utils/api-fetcher.ts
```diff
@@ -291,16 +292,32 @@ export class ApiFetcher implements IFetcher {
...
+        throw new StatusCodeError(
+          `Digest auth failed, expected 401 status code, but received ${first401Result.status} in the first response`
+        );
+      }
+
       if (!wwwAuthenticateHeader) {
         this.initVerboseResponseLogs({
           body: body401,
-          method: (method || 'get') as OperationMethod,
+          method: method as OperationMethod,
```
tatomyr: Do we still need to cast it?

packages/respect-core/src/modules/checks/checks-errors.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+}
+
+export class StatusCodeError extends ChecksErrors {
+  constructor(message: string, originalError?: Error) {
+    super('STATUS_CODE_CHECK', message, originalError);
+    this.name = 'StatusCodeError';
+  }
+}
+
```
tatomyr: Do we use the originalError anywhere? Let's keep it simple.

```suggestion
export class ChecksErrors extends Error {}

export class StatusCodeError extends ChecksErrors {}

```

tatomyr: You can also move this to the `./checks.ts` file itself.

packages/respect-core/src/utils/api-fetcher.ts
DmitryAnansky: unfortunately yes, as it is just a string

packages/respect-core/src/modules/checks/checks-errors.ts
DmitryAnansky: good, we can simplify.
Thx

tatomyr (approved):

PR #2333, opened 2025-09-29
feat: improve telemetry insights for spec version detection

AlbinaBlazhko17:
## What/Why/How?
Added new logic inside telemetry sender to send an `undefined` keyword only if spec version is not detected and send `unknown` keyword in other cases (ex., error occurred).

## Reference
Closes: [#18266](https://github.com/Redocly/redocly/issues/18266)

## Testing
Created unit test and tested manually.
## Check yourself

## Security

### DIFF
--- a/packages/cli/src/wrapper.ts
+++ b/packages/cli/src/wrapper.ts
@@ -28,7 +28,7 @@ export function commandWrapper<T extends CommandArgv>(
   return async (argv: Arguments<T>) => {
     let code: ExitCode = 2;
     let telemetry;
-    let specVersion: string | undefined = 'unknown';
+    let specVersion: string = 'unknown';
     let specKeyword: string | undefined;
     let specFullVersion: string | undefined;
     let config: Config | undefined;


packages/cli/src/wrapper.ts
```diff
@@ -71,6 +71,7 @@ export function commandWrapper<T extends CommandArgv>(
       if (err instanceof AbortFlowError) {
         // do nothing
       } else if (err instanceof HandledError) {
+        if (err.message.includes('Unsupported specification')) specVersion = 'undefined';
```
tatomyr: What about this:

```suggestion
        if (err.message.includes('Unsupported specification')) specVersion = 'unsupported';
```

Can we know which exact specification is it?

packages/cli/src/wrapper.ts
```diff
@@ -28,13 +32,17 @@ export function commandWrapper<T extends CommandArgv>(
   return async (argv: Arguments<T>) => {
     let code: ExitCode = 2;
     let telemetry;
-    let specVersion: string | undefined;
+    let specVersion: string | undefined = 'unknown';
```
tatomyr:
```suggestion
    let specVersion: string = 'unknown';
```

packages/cli/src/utils/telemetry.ts
```diff
@@ -102,6 +102,42 @@ export async function sendTelemetry({
   }
 }
 
+export function transformSpecVersionError(errorMessage: string): string {
```
tatomyr: I believe we don't need all those details as they simply duplicate what we already have in `specKeyword` and `specFullVersion`, don't we?

packages/cli/src/wrapper.ts
```diff
@@ -28,13 +32,17 @@ export function commandWrapper<T extends CommandArgv>(
...
     let specKeyword: string | undefined;
     let specFullVersion: string | undefined;
     let config: Config | undefined;
     const respectXSecurityAuthTypes: string[] = [];
     const collectSpecData: CollectFn = (document) => {
-      specVersion = detectSpec(document);
+      try {
+        specVersion = detectSpec(document);
+      } catch (err) {
+        specVersion = transformSpecVersionError(err.message);
```
tatomyr:
```suggestion
        specVersion = 'unsupported';
```

packages/cli/src/wrapper.ts
AlbinaBlazhko17: I did this change

packages/cli/src/wrapper.ts
AlbinaBlazhko17: I did this change

tatomyr (approved):

PR #2334, opened 2025-09-30
fix: improved separator display in respect workflows

DmitryAnansky:
## What/Why/How?

Fixed an issue where the Respect workflow separator did not render correctly in GitHub CI environments `───────[0m`.
GitHub CI terminal is cleaning single `/n` if there is no additional content. 
And it is not always cleans well this reset back to default attributes `\x1B[0m` leaving  `[0m` artifacts.

[Before](https://github.com/Redocly/redocly/actions/runs/18122127568/job/51568976483):

Screenshot 2025-09-30 at 19 48 35

[After: ](https://github.com/Redocly/redocly/actions/runs/18136432350/job/51617717388?pr=18307)

Screenshot 2025-09-30 at 19 49 05

## Check yourself

## Security

### DIFF
--- a/packages/core/src/__tests__/logger.test.ts
+++ b/packages/core/src/__tests__/logger.test.ts
@@ -45,7 +45,11 @@ describe('Logger in nodejs', () => {
 
     expect(spyingStdout).toBeCalledTimes(1);
     expect(spyingStdout).toBeCalledWith(expect.stringContaining('separator'));
-    expect(spyingStdout).toBeCalledWith(expect.stringMatching(/separator.*separator.*separator/));
+    expect(spyingStdout).toBeCalledWith(
+      colorize.gray(
+        'separatorseparatorseparatorseparatorseparatorseparatorseparatorseparatorseparato'
+      )
+    );
   });
 
   it('should call output with indent', () => {


packages/respect-core/src/modules/logger-output/helpers.ts
```diff
@@ -29,7 +29,7 @@ export function indent(str: string, level: number) {
 }
 
 export function printWorkflowSeparatorLine(logger: LoggerInterface) {
-  logger.printSeparator('\u2500');
```
DmitryAnansky, line 32: I will probably revert to this line format

packages/core/src/__tests__/logger.test.ts
```diff
@@ -37,14 +37,15 @@ describe('Logger in nodejs', () => {
...
+    expect(spyingStdout).toBeCalledWith(' \n');
   });
 
   it('should call "process.stdout.write" for output printSeparator', () => {
     logger.printSeparator('separator');
 
     expect(spyingStdout).toBeCalledTimes(1);
-    expect(spyingStdout).toBeCalledWith(expect.stringContaining(colorize.gray('separator')));
+    expect(spyingStdout).toBeCalledWith(expect.stringContaining('separator'));
+    expect(spyingStdout).toBeCalledWith(expect.stringMatching(/separator.*separator.*separator/));
```
tatomyr: Why do you need matching? Isn't it as simple string?

DmitryAnansky: it will be much longer.
This is just for shorter match to confirm it will be repeated several times.

DmitryAnansky: Can be exact value, no strong arguments against it, will apply suggestion.

tatomyr (approved):

Issue #2316, opened 2025-09-22
Support 3.2.0 and 3.1.2 OpenAPI Drafts


DmitryAnansky:
**Is your feature request related to a problem? Please describe.**

There is a new [OpenAPI draft release](https://github.com/OAI/OpenAPI-Specification/pull/4600) and we should add support for it.

**Describe the solution you'd like**

Implements new features/changes described in the draft.

**Describe alternatives you've considered**

**Additional context**

tatomyr:
Migration guide: https://learn.openapis.org/upgrading/v3.1-to-v3.2.html

Related: https://github.com/Redocly/redocly/issues/15509

tatomyr:
A ticket for new rules: https://github.com/Redocly/redocly-cli/issues/2342.

Additional rules to implement:
- [x] encodings are mutually exclusive (https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-11)
- [x] defaultMapping fallback (https://spec.openapis.org/oas/v3.2.0.html#composition-and-inheritance-polymorphism)
- [x] some example values are mutually exclusive (https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-15)

tatomyr:
Resolved in https://github.com/Redocly/redocly-cli/pull/2364.

PR #2119, opened 2025-06-03
wip for OAS 3.2 updates

jeremyfiel:
## What/Why/How?

this is a running update of features for OAS 3.2 support. 

https://github.com/OAI/OpenAPI-Specification/pulls?q=is%3Apr+milestone%3Av3.2.0+is%3Amerged
## Check yourself

- [ ] Code changed? - Tested with redoc (internal)
- [ ] All new/updated code is covered with tests
- [ ] New package installed? - Tested in different environments (browser/node)

## Security

- [ ] Security impact of change has been considered
- [ ] Code follows company security practices and guidelines

PR #2335, opened 2025-09-30
feat: introduce basic OAS3.2 support

tatomyr:
## What/Why/How?

Added basic support for OAS3.2 features:

- [x] [tags](https://learn.openapis.org/specification/tags)
- [x] [http methods](https://learn.openapis.org/specification/http-methods)
- [x] [media types](https://learn.openapis.org/specification/media-types)
- [x] queryString
- [x] servers
- [x] [security enhancements](https://learn.openapis.org/specification/security)
- [x] discriminator
- [x] $self partial support
- [x] examples
- [x] updated the `join` command to work with new tags and take `query` methods into account

## Reference

Partially addresses https://github.com/Redocly/redocly-cli/issues/2316.

Continuation of https://github.com/Redocly/redocly-cli/pull/2119 (cudos to @jeremyfiel!).

## Testing

Added basic E2E tests. More tests to be added as we enhance the OAS3.2 support.

## Check yourself

## Security

## Next steps

- Documentation to be updated
- Rules to be added
- More tests to be added
- Actual support for $self

### DIFF
--- a/packages/core/src/types/oas3_2.ts
+++ b/packages/core/src/types/oas3_2.ts
@@ -13,7 +13,7 @@ const Root: NodeType = {
 const Tag: NodeType = {
   ...Oas3_1Types.Tag,
   properties: {
-    ...Oas3_1Types.Tag.properties;
+    ...Oas3_1Types.Tag.properties,
     kind: { type: 'string' },
     parent: { type: 'string' },
     summary: { type: 'string' },


docs/@v2/rules/ruleset-templates.md
```diff
@@ -297,6 +297,7 @@ rules:
   required-string-property-missing-min-length: off
   response-contains-header: off
   scalar-property-missing-example: off
+  no-duplicated-tag-names: warn
```
tatomyr, line 300: Missed when originally added.

CONTRIBUTING.md
```diff
@@ -18,9 +18,14 @@ Before submitting your contribution though, please make sure to take a moment an
 
 ## Issue reporting guidelines
 
-- Before opening a new issue, try to make sure the same problem or idea hasn't already been reported. You can do that on the [Issues page](https://github.com/Redocly/redocly-cli/issues) in the repository and using the filter `is:issue` combined with some keywords relevant to your idea or problem. It helps us notice that more people have the same issue or use-case, and reduces the chance of getting your issue marked as a duplicate. Plus, you can even find some workarounds for your issue in the comments of a previously reported one!
+- Before opening a new issue, try to make sure the same problem or idea hasn't already been reported.
+  You can do that on the [Issues page](https://github.com/Redocly/redocly-cli/issues) in the repository and using the filter `is:issue` combined with some keywords relevant to your idea or problem.
+  It helps us notice that more people have the same issue or use-case, and reduces the chance of getting your issue marked as a duplicate.
```
DmitryAnansky:
```suggestion
  It helps us notice that more people have the same issue or use case, and reduces the chance of getting your issue marked as a duplicate.
```

packages/core/src/typings/swagger.ts
```diff
@@ -49,6 +49,7 @@ export interface Oas2PathItem {
   options?: Oas2Operation;
   head?: Oas2Operation;
   patch?: Oas2Operation;
+
```
DmitryAnansky, line 52: looks like space was accidental, please remove if not needed.

packages/core/src/rules/other/stats.ts
```diff
@@ -1,4 +1,4 @@
-import type { Oas3Parameter, OasRef, Oas3Tag } from '../../typings/openapi.js';
+import type { Oas3Parameter, OasRef, Oas3Tag, Oas3_2Tag } from '../../typings/openapi.js';
```
DmitryAnansky, line 1: Just a question, why we have `Oas3Tag` and `Oas3_2Tag` and don't have `Oas3_1Tag` ?
Were there no changes?

packages/core/src/typings/swagger.ts
tatomyr: Added it to separate method names from other properties.

packages/cli/src/commands/join.ts
```diff
@@ -220,14 +225,15 @@ export async function handleJoin({
...
+      oasVersion,
     };
     if (tags) {
       populateTags(context);
     }
     collectExternalDocs(openapi, context);
     collectPaths(openapi, context, serversAreTheSame);
     collectComponents(openapi, context);
-    collectWebhooks(oasVersion!, openapi, context);
+    collectWebhooks(openapi, context);
```
jeremyfiel, line 236: why did you change this from non-null to include  null?

the old one was 
`oasVersion!` 
the new one is 
`oasVersion: Extract<SpecVersion, 'oas3_0' | 'oas3_1' | 'oas3_2'> | null;`

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -5,10 +5,13 @@ import type { Oas3Decorator } from '../../visitors.js';
 import type {
   Oas3Definition,
   Oas3_1Definition,
+  Oas3_2Definition,
   Oas3Components,
   Oas3_1Components,
 } from '../../typings/openapi.js';
 
+type AnyOas3Definition = Oas3Definition | Oas3_1Definition | Oas3_2Definition;
```
DmitryAnansky, line 13: There are several places where this type definition repeated, can we define in one place and reuse it?

packages/cli/src/commands/join.ts
```diff
@@ -281,7 +290,11 @@ export async function handleJoin({
         );
       } else if (!tagDuplicate) {
         // Instead add tag to joinedDef if there no duplicate;
-        tag['x-displayName'] = tag['x-displayName'] || tag.name;
+        if (oasVersion === 'oas3_0' || oasVersion === 'oas3_1') {
+          (tag as Oas3Tag)['x-displayName'] = (tag as Oas3Tag)['x-displayName'] || tag.name;
+        } else if (oasVersion === 'oas3_2') {
+          (tag as Oas3_2Tag).summary = (tag as Oas3_2Tag).summary || tag.name;
```
jeremyfiel, line 296: this is probably an enhancement, but maybe consider how to group with `parent` or `kind`

packages/core/src/rules/other/stats.ts
tatomyr: `Oas3Tag` is the same in 3.0 and 3.1 but was changed in 3.2.

packages/core/src/decorators/oas3/remove-unused-components.ts
tatomyr: They are mostly scoped to rules. Some rules need that, some don't, some use other combinations of definition types. I'd leave it as it is.

packages/cli/src/commands/split/types.ts
```diff
@@ -1,6 +1,11 @@
-import type { Oas2Definition, Oas3_1Definition, Oas3Definition } from '@redocly/openapi-core';
+import type {
+  Oas2Definition,
+  Oas3Definition,
+  Oas3_1Definition,
+  Oas3_2Definition,
+} from '@redocly/openapi-core';
 
-export type Definition = Oas3_1Definition | Oas3Definition | Oas2Definition;
+export type Definition = Oas2Definition | Oas3Definition | Oas3_1Definition | Oas3_2Definition;
```
DmitryAnansky, line 8: Minor, also can use shared type:
```type AnyOas3Definition = Oas3Definition | Oas3_1Definition | Oas3_2Definition;```

File: packages/core/src/rules/common/no-http-verbs-in-paths.ts [hunk 9]
```diff
@@ -5,7 +5,7 @@ import type { Oas2PathItem } from '../../typings/swagger.js';
 import type { Oas3PathItem } from '../../typings/openapi.js';
 import type { UserContext } from '../../walk.js';
 
-const httpMethods = ['get', 'head', 'post', 'put', 'patch', 'delete', 'options', 'trace'];
+const httpMethods = ['get', 'head', 'post', 'put', 'patch', 'delete', 'options', 'trace', 'query'];
```
jeremyfiel, line 8: how can we find `additionalOperations`?  I think it will be difficult

__tests__/commands.test.ts
```diff
@@ -398,6 +398,7 @@ describe('E2E', () => {
         'references-in-parameters',
         'ignore-decorators',
         'multi-references-to-one-file',
+        'oas3.2',
```
DmitryAnansky, line 401: Does join works differently for oas3.2 ?
If you checking some specific case, maybe we could find better name for this folder?

packages/cli/src/commands/split/types.ts
tatomyr: Not needed in CLI :)

packages/core/src/types/oas3_2.ts
```diff
@@ -0,0 +1,220 @@
...
+};
+
+const Tag: NodeType = {
+  ...Oas3_1Types.Tag,
+  properties: {
+    name: { type: 'string' },
+    description: { type: 'string' },
+    externalDocs: 'ExternalDocs',
+    'x-traitTag': { type: 'boolean' },
+    'x-displayName': { type: 'string' }, // deprecated
```
jeremyfiel:
```suggestion
    ...Oas3_1Types.Tag.properties;
```

no need to redefine the existing type

__tests__/commands.test.ts
tatomyr: Yes, it's different for tags, also checking support for `query` methods. It's just general check, similar to `oas3.1`.

packages/cli/src/commands/split/types.ts
DmitryAnansky: Can be added later.

DmitryAnansky (approved):
Left minor comments.

__tests__/commands.test.ts
jeremyfiel: I wonder if Join should consider the new tag properties because someone out there will definitely use the tags to define join characteristics

packages/cli/src/commands/join.ts
tatomyr: Yeah, looks like an enhancement.

packages/core/src/types/oas3_2.ts
tatomyr: I actually wanted to remove `x-displayName` as it's not needed anymore, but haven't decided on that yet.

packages/cli/src/commands/join.ts
tatomyr: Good catch. The type change is correct but it helped me to find a missing condition inside the function!

packages/core/src/rules/common/no-http-verbs-in-paths.ts
tatomyr: I don't think the rule is about catching every declared method. It just shouldn't interfere with the common method names.

docs/@v1/rules/ruleset-templates.md
```diff
@@ -286,47 +286,6 @@ rules:
   response-contains-header: off
   scalar-property-missing-example: off
   no-required-schema-properties-undefined: off
-oas3_1Rules:
```
tatomyr, line 289: Redundant.

__tests__/commands.test.ts
tatomyr: @jeremyfiel good question. Do you see a potential example?

PR #2338, opened 2025-10-02
feat: added login verbose option

DmitryAnansky:
## What/Why/How?

Introduced `verbose` option for `login` command that can help users to debug login problems.
## Check yourself

## Security


JLekawa (approved):

packages/cli/src/auth/oauth-client.ts
```diff
@@ -19,24 +19,29 @@ export class RedoclyOAuthClient {
...
-    this.key = crypto.createHash('sha256').update(`${homeDirPath}${CREDENTIALS_SALT}`).digest();
-    this.iv = crypto.createHash('md5').update(homeDirPath).digest();
+      this.key = crypto.createHash('sha256').update(`${homeDirPath}${CREDENTIALS_SALT}`).digest();
+      this.iv = crypto.createHash('md5').update(homeDirPath).digest();
 
-    mkdirSync(this.credentialsFolderPath, { recursive: true });
+      mkdirSync(this.credentialsFolderPath, { recursive: true });
+    } catch (error) {
+      logger.error('Failed to initialize OAuth client.\n');
+      throw error;
```
tatomyr: I wouldn't add additional try-catch nesting here. We can just catch the error inside [handleLogin](https://github.com/Redocly/redocly-cli/pull/2338/files) and log it to the console in verbose mode. WDYT?

DmitryAnansky: Just wanted to emphasize this message `Failed to initialize OAuth client.`
Later this can be login method issue.

packages/core/src/logger.ts
```diff
@@ -41,6 +42,15 @@ class Logger implements LoggerInterface {
     return isBrowser ? console.error(str) : process.stderr.write(colorize.red(str));
   }
 
+  errorWithStack(str: string | Error) {
```
tatomyr: Do we really need this method? Why not simply output the stacktrace where we need it?

PR #2349, opened 2025-10-07
feat: add test for missing @redocly/config type

AlbinaBlazhko17:
## What/Why/How?
We faced the issue when updating the version of openapi-core in other products. The problem was that we forgot to add new types in `@redocly/config`. 

I created a unit test, which will fail if we forget to add types.

## Testing
Manually.
## Screenshots (optional)
Screenshot 2025-10-07 at 13 48 41

## Check yourself

## Security

### DIFF
--- a/packages/core/src/config/__tests__/config.test.ts
+++ b/packages/core/src/config/__tests__/config.test.ts
@@ -1,4 +1,5 @@
 import { type SpecVersion } from '../../oas-types.js';
+import { rootRedoclyConfigSchema } from '@redocly/config';
 import { Config } from '../config.js';
 import * as utils from '../../utils.js';
 import * as jsYaml from '../../js-yaml/index.js';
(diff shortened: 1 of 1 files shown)


packages/core/src/config/__tests__/config.test.ts
```diff
@@ -239,3 +241,44 @@ describe('generation ignore object', () => {
     expect(config).toMatchSnapshot();
   });
 });
+
+describe('rootRedoclyConfigSchema synchronization', () => {
```
tatomyr, line 245: I'm reluctant to add a unit test for this here. Imagine you just want to add a property--but you wouldn't be able to do it without releasing a new config version. Can you move it to something like a smoke test (so it's not required for merging PRs)?

tatomyr: Maybe term 'intergration test' is more relevant to our case though.

tatomyr:
Closing this in favour of https://github.com/Redocly/redocly/pull/20095.

PR #2358, opened 2025-10-10
fix: no-http-verbs-in-paths should not error on the verb query

tatomyr:
## What/Why/How?

Fixed an issue where the `no-http-verbs-in-paths` rule was incorrectly flagging path names containing the verb `query`.

## Reference

Fixes https://redoc-ly.slack.com/archives/C013VB35DB4/p1760032419326879 (internal).

## Check yourself

## Security

### DIFF
--- a/.changeset/giant-years-admire.md
+++ b/.changeset/giant-years-admire.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed an issue where the `no-http-verbs-in-paths` rule was incorrectly flagging path names containing the verb `query`.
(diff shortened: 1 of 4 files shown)


jeremyfiel:
any reason you didn't add a check for `Root === 3.x` and apply the rule according to the spec version?  Will you create another rule for 3.2 ?

tatomyr:
@jeremyfiel  there's no http verb 'query': https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods. It's only a new new operation name in OAS.

jeremyfiel:
The RFC is just about finished going through the approval process. 

https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html

tatomyr:
Yeah, let's get back to it when it's approved then.

PR #2362, opened 2025-10-13
fix: resolve sibling refs alongside $ref

vadyvas:
## What/Why/How?

Fix handling of description/summary when they are $ref siblings next to a schema $ref so they resolve correctly and override the target schemas.

Why: Previously, bundling left description as a $ref object when used alongside another $ref, matching the bug described in the linked issue.

```yaml
# openapi.yaml
openapi: 3.1.0
paths:
  /test:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  field:
                    description:
                      $ref: './components/schemas/description.md'
                    $ref: '#/components/schemas/Child'

components:
  schemas:
    Child:
      type: object
      properties:
        name:
          type: string
```

## Reference

Fixes: [СLI issue #1907](https://github.com/Redocly/redocly-cli/issues/1907)

## Testing

- Added unit and e2e tests
- Existing suites pass locally.

## Check yourself

## Security


packages/core/src/walk.ts
```diff
@@ -339,6 +339,11 @@ export function walkDocument<T extends BaseVisitor>(opts: {
               propType = { name: 'scalar', properties: {} };
             }
 
+            if (isRef(node[propName]) && type.name === 'Schema' && propType?.name === 'scalar') {
```
RomanHotsiy: we should not hardcode name of the type here

vadyvas: - removed the hardcoded Schema check and now resolve scalar sibling $refs generically
- added a unit test for this case

packages/core/src/walk.ts
tatomyr, line 310: Could you check if the fix could be applied somewhere here instead?

vadyvas: I think the logic from lines 342-345 cannot be moved to lines 307-309 because they serve different purposes:
Lines 307-309: Collection phase - gather property names into the props array,
Lines 342-345: Processing phase - handle individual properties.

The variable needed for the processing logic (propType) are only available inside the loop, not during the collection phase.

@tatomyr I believe, current solution is architecturally correct: collect all properties first, then process each one with proper context.

.changeset/smart-points-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Allow resolving `description`/`summary` sibling refs alongside `$ref` in Schema contexts.
```
tatomyr:
```suggestion
Allow resolving references inside other properties alongside the `$ref` keyword.
```

tatomyr (approved):

.changeset/smart-points-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Allow resolving references inside other properties alongside the `$ref` keyword.
```
JLekawa:
```suggestion
Fixed an issue where references were not resolved when used inside other properties alongside the `$ref` keyword.
```

PR #2365, opened 2025-10-14
feat: introduced minimal bundleOas export function

DmitryAnansky:
## What/Why/How?

- Added new export `bundleOas` that support only OAS bundling. Minimized bundle size with only `bundleOas` and empty config should be near ~ `299.83KiB`.
- Reorganized `utils` to allow importing isolated functions.
- Removed `jest-matcher-utils` from `respect-core`.
- Migrated from `minimatch` to `picomatch`.

## Check yourself

## Security

### DIFF
--- a/.changeset/old-steaks-talk.md
+++ b/.changeset/old-steaks-talk.md
@@ -1,7 +1,5 @@
 ---
-"@redocly/respect-core": minor
 "@redocly/openapi-core": minor
-"@redocly/cli": minor
 ---
 
 Added new `bundleOas` export: a lightweight function for bundling OpenAPI specifications.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -10,11 +10,6 @@
       "require": "./lib/index.js",
       "types": "./lib/index.d.ts"
     },
-    "./lib/config/config": {
-      "import": "./lib/config/index.js",
-      "require": "./lib/config/index.js",
-      "types": "./lib/config/index.d.ts"
-    },
     "./lib/bundle-oas": {
       "import": "./lib/bundle/bundle-oas.js",
       "require": "./lib/bundle/bundle-oas.js",


package.json
```diff
@@ -67,10 +67,12 @@
     "outdent": "^0.7.1",
     "pegjs": "0.11.0-master.b7b87ea",
     "prettier": "^2.1.2",
+    "rollup-plugin-bundle-stats": "^4.21.5",
```
DmitryAnansky: todo: remove

package.json
```diff
@@ -67,10 +67,12 @@
...
     "slackify-markdown": "^4.3.1",
     "ts-loader": "^8.0.2",
     "tsx": "^4.19.3",
     "typescript": "5.5.3",
+    "vite": "^7.1.9",
```
DmitryAnansky: todo: remove

vite.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from 'vite';
```
DmitryAnansky: todo: remove

packages/core/src/bundle/bundle-document.ts
```diff
@@ -0,0 +1,137 @@
...
+  fileDependencies: Set<string>;
+  rootType: NormalizedNodeType;
+  refTypes?: Map<string, NormalizedNodeType>;
+  visitorsData: Record<string, Record<string, unknown>>;
+};
+
+export async function bundleDocument(opts: {
+  document: Document;
+  config: Config;
+  types: Record<string, NodeType>;
```
DmitryAnansky, line 39: @RomanHotsiy / @tatomyr 
JFYI: this will be breaking change for users, who imports bundleDocument.

DmitryAnansky: E.G. => https://github.com/Redocly/redocly/pull/18592/files
packages/openapi-language-server/server/bundle.ts

Before:
```
  const { bundle, problems, fileDependencies, rootType, refTypes } = await bundleDocument({
    document,
    config,
    externalRefResolver,
  });
 ```
 
 After
 ```
   const { bundle, problems, fileDependencies, rootType, refTypes } = await bundleDocument({
    document,
    config,
    externalRefResolver,
    types: getTypes(detectSpec(document.parsed)),
  });
  ```

File: packages/cli/src/__tests__/commands/join.test.ts [hunk 5]
```diff
@@ -158,6 +158,8 @@ describe('handleJoin', () => {
 
   it('should call exitWithError if mixing OpenAPI 3.0 and 3.1', async () => {
     vi.mocked(detectSpec)
+      .mockImplementationOnce(() => 'oas3_0')
```
tatomyr, line 161: Why the join test has been changed? I don't see any related changes in the code.

packages/core/src/__tests__/bundle-oas.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        get: () => '',
+      },
+    })
+  );
+
+  expect.addSnapshotSerializer(yamlSerializer);
+
+  it('should bundle external refs', async () => {
+    const { bundle: res, problems } = await bundleOas({
+      config: await createConfig({}),
```
tatomyr: Here's where we can use `loadConfig` from bundle-oas, I suppose.

packages/core/src/bundle/bundle-oas.ts
```diff
@@ -0,0 +1,95 @@
...
+    throw new Error('Document or reference is required.\n');
+  }
+
+  const document =
+    doc === undefined ? await externalRefResolver.resolveDocument(base, ref!, true) : doc;
+
+  if (document instanceof Error) {
+    throw document;
+  }
+  opts.collectSpecData?.(document.parsed);
```
tatomyr: Is it needed? The `bundleOas` function is going to be used as an external dependency and the collecting doesn't make sense to me.

packages/core/src/bundle/bundle-oas.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  return bundleDocument({
+    document,
+    ...opts,
+    externalRefResolver,
+    types,
+  });
+}
+
+export function loadConfig(
```
tatomyr: If this is intended only for `bundleOas`, it makes more sense to inject the config directly into it rather than expose yet another `loadConfig`.

packages/core/src/bundle/bundle-oas.ts
```diff
@@ -0,0 +1,95 @@
...
+export type CoreBundleOptions = {
+  externalRefResolver?: BaseResolver;
+  config: Config;
+  dereference?: boolean;
+  base?: string | null;
+  removeUnusedComponents?: boolean;
+  keepUrlRefs?: boolean;
+};
+
+export async function bundleOas(
```
tatomyr, line 25: What if we simply import `bundleDocument` (along with the necessary OAS types and utils) where it's used and create this wrapper there instead of duplicating code here?

packages/core/src/bundle/bundle-oas.ts
tatomyr: BTW, this is just a dummy config, and we're not loading anything here.

packages/core/package.json
```diff
@@ -16,9 +16,14 @@
       "types": "./lib/config/index.d.ts"
     },
     "./lib/bundle": {
```
tatomyr: This export is for the old Redoc and can be removed/replaced.

packages/core/src/bundle/bundle-oas.ts
DmitryAnansky: @AlexVarchuk, 
Could you please check on your side, if you can just use `bundleDocument` ?

packages/core/src/bundle/bundle-oas.ts
DmitryAnansky: removed

packages/core/src/bundle/bundle-oas.ts
DmitryAnansky: Lets keep is as it is for now, this can be changed later.

.changeset/old-steaks-talk.md
```diff
@@ -0,0 +1,7 @@
+---
+"@redocly/respect-core": minor
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
```
tatomyr: The change only affects openapi-core, isn't it?

DmitryAnansky: it affect all 3

packages/core/package.json
tatomyr: This export also looks redundant.

tatomyr: I believe it was for the old Redoc as well.

DmitryAnansky: Maybe, just don't know who might use it, it was there before.

PR #2369, opened 2025-10-15
feat: add new linting rules for compositions keys oneOf, anyOf, allOf

### LINKED ISSUE #210: Check for incorrect discriminator use

Yes: users need to get feedback about incorrect discriminator use.

`no-discriminator-property-in-object`

Problem message:
> The discriminator property `{insert the property name}` must exist in schemas {...names}.

None.

**Additional context**
Here is a minimal example:

```yaml
components:
  schemas:
    Problematic:
      type: object
      properties:
        data:
          discriminator:
            propertyName: objectType
          oneOf:
          - $ref: '#/components/schemas/A'
          - $ref: '#/components/schemas/B'
        objectType:
          type: string
    A:
      type: object
      properties:
        message:
          type: string
    B:
      type: object
      properties:
        message:
          type: string
```

Notice that `objectType` is defined in the `Problematic` schema and not in `A` and `B`.

### LINKED ISSUE #2326: New linting rules for `oneOf`/`anyOf`/`allOf`
In the current implementation we didn't handle illogical use of these schema composition keywords (`oneOf`, `anyOf`,`allOf`). In some cases it caused a validation error while sending a request to mock server.

### **`oneOf` rules:**
  1. If oneOf contains the same properties with the same type, which are required and additionalProperty doesn't set to false → show warning
  2. If oneOf is used with nullable type (type array containing 'null' or nullable: true), but schemas are not mutually exclusive → show error
  3. If oneOf schemas have overlapping properties without discriminator property → show warning
  4. If oneOf schemas have overlapping properties without discriminator property → show warning
  5. If oneOf contains only one schema → show warning to use direct schema
  6. If oneOf contains duplicate schemas → show warning
  7. If oneOf has missing discriminator mapping when using $ref → show warning

### **`anyOf` rules:**
1. If anyOf contains required properties → show warning to restructure schema
2. If anyOf contains required properties → show warning to restructure schema
3. If anyOf contains only one schema → show warning to use direct schema

### **`allOf` rules:**

1. If allOf contains conflicting property types for same property → show error

**Additional context**

When user send a request with this schema:
```
 defaultPaymentInstrument:
    type:
      - 'object'
      - 'null'
    oneOf:
      - $ref: ./VaultedInstrument.yaml
      - $ref: ./AlternativePaymentInstrument.yaml
      - $ref: ./CashInstrument.yaml
      - $ref: ./CheckInstrument.yaml
```
He received an error:

> "message": "Error validating body. data/defaultPaymentInstrument must match exactly one schema in oneOf, data/leadSource/referrer must match format \"url\""

AlbinaBlazhko17:
## What/Why/How?
Added new linting rules to handle illogical use of composition keys: `oneOf`, `anyOf`, `allOf`.

## Reference
Resolves #2326 
Resolves #210 

## Testing
Locally with different specs.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Default `recommended` presets will surface new warnings on many existing API descriptions; overlap detection is intentionally conservative but still affects broad schema linting paths.
> 
> **Overview**
> Introduces **`no-illogical-composition-keywords`**, a new OpenAPI 3.x lint rule that flags illogical or ambiguous use of **`oneOf`**, **`anyOf`**, and **`allOf`** (single-member lists, duplicates, empty `{}` members, overlapping `oneOf` branches, nullable conflicts, and discriminator misuse such as inline members or optional discriminator properties on 3.0/3.1).
> 
> The rule is registered in the OAS3 ruleset, enabled as **`warn`** in **`recommended`** and **`error`** in **`recommended-strict`** / **`all`** (off in **`minimal`** and **`spec`**). Documentation, Spectral migration mapping, sidebar, and a changeset note that existing specs may see new findings.
> 
> The OpenAPI type tree now exposes **`AllOf`**, **`AnyOf`**, and **`OneOf`** as distinct visitor/configurable-rule subjects (via named `listOf` types), with a large unit test suite and updated e2e lint snapshots.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d10c389dc64f865d214fe4c2e34fb7c3d5aaf94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+
+export const NoIllogicalCompositionKeywords: Oas3Rule = (): Oas3Visitor => {
+  return {
+    Schema: {
+      skip(node) {
+        return !node.oneOf && !node.anyOf && !node.allOf;
+      },
+      enter(schema, { report, location, resolve }) {
+        // Helper function to get the composition keyword and schemas
+        const compositionData = (() => {
```
tatomyr: Please use correct naming so you don't need the comments, e.g.:


```suggestion
        const getSchemaCompositionData = (() => {
```

or something similar.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+type ReturnType = {
+  isExclusive: boolean;
+  reason?: string;
+};
+
+export const NoIllogicalCompositionKeywords: Oas3Rule = (): Oas3Visitor => {
+  return {
+    Schema: {
+      skip(node) {
+        return !node.oneOf && !node.anyOf && !node.allOf;
```
tatomyr: This would be easier to read, as you don't have to apply negation 3 times:


```suggestion
        return !(node.oneOf || node.anyOf || node.allOf);
```

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+          if (schema.oneOf && Array.isArray(schema.oneOf)) {
+            return { keyword: 'oneOf' as const, schemas: schema.oneOf };
+          }
+          if (schema.anyOf && Array.isArray(schema.anyOf)) {
+            return { keyword: 'anyOf' as const, schemas: schema.anyOf };
+          }
+          if (schema.allOf && Array.isArray(schema.allOf)) {
+            return { keyword: 'allOf' as const, schemas: schema.allOf };
+          }
+          return null;
```
tatomyr: It's literally undefined 🙂 

```suggestion
          return undefined;
```

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+  return {
+    Schema: {
+      skip(node) {
+        return !node.oneOf && !node.anyOf && !node.allOf;
+      },
+      enter(schema, { report, location, resolve }) {
+        // Helper function to get the composition keyword and schemas
+        const compositionData = (() => {
+          if (schema.oneOf && Array.isArray(schema.oneOf)) {
+            return { keyword: 'oneOf' as const, schemas: schema.oneOf };
```
tatomyr: Technically, your schema could contain `oneOf` and `allOf` simultaneously. The code doesn't handle this case.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+        })();
+
+        if (!compositionData) return;
+
+        const { keyword, schemas } = compositionData;
+
+        // Check for minimum schema count (oneOf and anyOf require at least 2)
+        if ((keyword === 'oneOf' || keyword === 'anyOf') && schemas.length < 2) {
+          report({
+            message: `Schema object \`${keyword}\` should contain at least 2 schemas. Use the schema directly instead.`,
```
tatomyr: I'd simply use quotes:

```suggestion
            message: `Schema object '${keyword}' should contain at least 2 schemas. Use the schema directly instead.`,
```

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+            return { keyword: 'allOf' as const, schemas: schema.allOf };
+          }
+          return null;
+        })();
+
+        if (!compositionData) return;
+
+        const { keyword, schemas } = compositionData;
+
+        // Check for minimum schema count (oneOf and anyOf require at least 2)
```
tatomyr: You don't need comments like this as the intention should be clear enough from the condition and the report message.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+              message: `Schema in \`${keyword}\` is empty, which provides no validation constraints.`,
+              location: location.child([keyword, String(i)]),
+            });
+            return;
+          }
+        }
+
+        // Check for duplicate schemas
+        for (let i = 0; i < schemas.length - 1; i++) {
+          if (dequal(Object.values(schemas[i]), Object.values(schemas[i + 1]))) {
```
tatomyr: This will only check for duplication in the consecutive schemas, right? But the duplication could reside anywhere in the array.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+          if (schema.anyOf && Array.isArray(schema.anyOf)) {
+            return { keyword: 'anyOf' as const, schemas: schema.anyOf };
+          }
+          if (schema.allOf && Array.isArray(schema.allOf)) {
+            return { keyword: 'allOf' as const, schemas: schema.allOf };
+          }
+          return null;
+        })();
+
+        if (!compositionData) return;
```
tatomyr: This duplicates the `skip()` functianality.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
tatomyr: Oh, wait, it's an IIFE? Then it's a bit of overengineering. You can do simply this:

```ts
const oneOfs = schema.oneOf
```

if you prefer to have a separate variable, and then just do every needed check. Don't be afraid to repeat yourself.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,495 @@
...
+          });
+          return;
+        }
+
+        // Check for empty schemas (applies to all composition keywords)
+        for (let i = 0; i < schemas.length; i++) {
+          const resolvedSchema = resolveSchema(schemas[i], resolve);
+          if (isEmptyObject(resolvedSchema)) {
+            report({
+              message: `Schema in \`${keyword}\` is empty, which provides no validation constraints.`,
```
tatomyr:
```suggestion
              message: `Schema is empty.`,
```

I just highlight the correct key, and this would be enough. Try keeping the report messages concise and clear.

tatomyr (changes requested):
Had a quick look. I suggest to reorganise the checks, so they cover fist for oneOfs, then anyOfs, and then allOfs. Just duplicate the checks when needed. 
Will have another look after you implement those changes.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Thanks, it's really looks better

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: You are right

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: This functionality at all doesn't resolve all nested composition keywords, because we consider, that every composition keyword is a new scope of validation.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Yeah, it is duplication, because it doesn't see proper type, even we check in skip method.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: I was thinking about using a separate variable for every composition keyword, but in that case, i need to duplicate checks (duplicated schemas, length of schemas) for every composition keyword and the only difference will be in the warning message. So, the purpose of the IIFE is to return the schema and string, and after that to easily retrieve them in the report message and checks.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Thanks, i miss it after refactoring

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Get it, fixed it

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Thanks, it looks better

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Yeah, you are right. I just left all comments to simplify the review of PR, i will remove it after finalizing the functionality.

AlbinaBlazhko17:
@cursor review

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,508 @@
...
+
+  // A property required by both sides distinguishes nothing, no matter where it is declared.
+  const leftRequiresUndeclared = [...leftRequired].some(
+    (name) => !getOwn(rightProperties, name) && !rightRequired.has(name)
+  );
+  const rightRequiresUndeclared = [...rightRequired].some(
+    (name) => !getOwn(leftProperties, name) && !leftRequired.has(name)
+  );
+  return leftRequiresUndeclared && rightRequiresUndeclared;
+}
```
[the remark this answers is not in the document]
AlbinaBlazhko17: It is behavior, that I want to introduce: if the two schemas have at least one different property, which is in required - two schemas are mutually exclusive. We don't care about other props, which will be added with `additionalProperties: true`.

AlbinaBlazhko17:
@cursor review

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,206 @@
...
+- A `oneOf` or `anyOf` with fewer than two schemas, unless the schema declares a `discriminator`.
+- An `allOf` with fewer than two schemas that neither declares another keyword of its own nor extends a discriminated schema.
+- A schema repeated inside the same keyword.
+- An empty schema (`{}`) used as a member.
+- Two `oneOf` schemas that a single value can match at the same time.
+- A nullable schema whose `oneOf` also accepts `null`.
+- A `discriminator` whose property is missing from `required` in every member schema.
+- An inline `oneOf` or `anyOf` member that a `discriminator` cannot select.
+
+| OAS | Compatibility |
```
tatomyr, line 21: I guess the same could be extended later to the AsyncAPI spec.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,206 @@
...
+The most common version of this is nullability: if a referenced schema already accepts `null` and the `oneOf` also lists `type: 'null'`, a null value matches both branches.
+The same ambiguity appears one level up, when the schema holding the `oneOf` is itself nullable and a member accepts `null` too.
+
+Deciding whether two arbitrary schemas overlap is not solvable in general, so the comparison stays deliberately narrow.
+It reads `type`, `nullable`, `enum`, `const`, `properties`, `required`, and `additionalProperties: false`.
+A `const` counts as a single-value `enum`, so one member can use `enum` and the other `const`.
+When a member uses any other constraint, such as `not`, `pattern`, `minimum`, or a nested `allOf`, that constraint may be what separates the schemas, so the rule reports nothing for the pair.
+
+A `discriminator` names the property that tells the members apart, so the rule trusts it and checks only what the specification requires.
+The property must be listed in `required` in every member schema, because a value can otherwise omit it and nothing decides which schema applies.
```
tatomyr, line 61: I believe this is already covered by another rule (https://redocly.com/docs/cli/rules/oas/spec-discriminator-defaultMapping), at least for OAS 3.2 (where it's a specification requirement).

packages/core/src/types/oas3_1.ts
```diff
@@ -140,9 +140,9 @@ const Schema: NodeType = {
         };
       }
     },
-    allOf: listOf('Schema'),
-    anyOf: listOf('Schema'),
-    oneOf: listOf('Schema'),
+    allOf: listOf('Schema', { name: 'AllOf' }),
```
tatomyr, line 143: Maybe let's just introduce the `AllOF` type directly?


```suggestion
    allOf: 'AllOf',
```

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,497 @@
...
+  const leftTypes = getTypeSet(left);
+  const rightTypes = getTypeSet(right);
+  if (!leftTypes || !rightTypes) return undefined;
+
+  return [...leftTypes].filter((type) => rightTypes.has(type));
+}
+
+function schemaAllowsNull(schema: CompositionSchema): boolean {
+  return isNullable(schema) || getTypeSet(schema)?.has('null') === true;
+}
```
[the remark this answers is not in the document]
AlbinaBlazhko17: It is expected. We want to carry case, when parent and child have null.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,497 @@
...
+  'deprecated',
+  'readOnly',
+  'writeOnly',
+  'externalDocs',
+  'xml',
+]);
+
+export const NoIllogicalCompositionKeywords: Oas3Rule = () => {
+  return {
+    AllOf(members, ctx) {
```
tatomyr, line 67: Wouldn't it just be easier to start a the Schema level and elevate nested allOfs/oneOfs through `resolveSchema` or a similar helper? Then we wouldn't need the AllOf/OneOf node types at all (at least it seems so).

And it would be easier to judge the logical order of the schema in its own context, wouldn't it?

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Yeap, we can do it, but need more investigation with the specific cases usage of those keywords in Async.

packages/core/src/types/oas3_1.ts
AlbinaBlazhko17: It cannot, i faced the issues, because `listOf` adds `List` to each schema and in consequence instead of `AllOf`, `AnyOf` and `OneOf` visitors i received 3 same visitors named `SchemaList`.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
tatomyr: Sure, not in this PR.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Fair points. Let me explain why i introduced new visitors:
1. I reviewed previous implementation and it looks overcomplicated and difficult to read with the skip visitor and with the logic inside `Schema` visitor, where we pick what branch of check we need to run.
2. Previous implementation had one issue, where i check what composition key i have inside the schema to pick the checks and error message and the issue was, that one schema can have more than one composition keyword. e.g. `oneOf` + `allOf`. So, i needed to add a logic to handle and resolve both, which could be overcomplicated and we can solve this just using visitors.
3. When i implemented this rule with visitors it became more readable and easily to understand and have all necessary data to compare, such as all schemas, that are inside the keyword and parent schema using `ctx.parent` . The visitor pattern fixes the issue, which i described in second point.

packages/core/src/types/oas3_1.ts
tatomyr: You can omit the `listOf` helper and create the type directly if needed.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Good catch, let me investigate this.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,497 @@
...
+  return isPlainObject(resolveSchema(member, resolve)?.schema.discriminator);
+}
+
+function reportEmptyMembers(
+  members: CompositionSchema[],
+  keyword: CompositionKeyword,
+  ctx: UserContext
+) {
+  for (const [index, member] of members.entries()) {
+    if (isPlainObject(member) && Object.keys(member).length === 0) {
```
tatomyr, line 154: Shall we check if it's `isRef`?

.changeset/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/cli': minor
```
tatomyr, line 2:
```suggestion
'@redocly/cli': minor
'@redocly/openapi-core': minor
```

.changeset/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Added the `no-illogical-composition-keywords` rule.
+
+**Note**: the rule is set to `warn` in the `recommended` ruleset and to `error` in `recommended-strict`, so existing API descriptions may report new problems.
```
JLekawa:
```suggestion
**Note**: the rule is set to `warn` in the `recommended` ruleset and to `error` in `recommended-strict`. Existing API descriptions may report new problems.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,206 @@
...
+The rule reports:
+
+- A `oneOf` or `anyOf` with fewer than two schemas, unless the schema declares a `discriminator`.
+- An `allOf` with fewer than two schemas that neither declares another keyword of its own nor extends a discriminated schema.
+- A schema repeated inside the same keyword.
+- An empty schema (`{}`) used as a member.
+- Two `oneOf` schemas that a single value can match at the same time.
+- A nullable schema whose `oneOf` also accepts `null`.
+- A `discriminator` whose property is missing from `required` in every member schema.
+- An inline `oneOf` or `anyOf` member that a `discriminator` cannot select.
```
JLekawa:
```suggestion
- `oneOf` or `anyOf` with fewer than two schemas, unless the schema declares a `discriminator`
- `allOf` with fewer than two schemas that neither declares another keyword of its own nor extends a discriminated schema
- schemas repeated inside the same keyword
- empty schemas (`{}`) used as members
- two `oneOf` schemas that a single value can match at the same time
- schemas that accept null alongside a `oneOf` member that also accepts null
- member schemas that omit the `discriminator` property from required
- inline `oneOf` or `anyOf` members that a `discriminator` cannot select
```

AlbinaBlazhko17:
@cursor review

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,209 @@
...
+
+style OneOf fill:#codaf9,stroke:#0044d4,stroke-width:5px
+style AnyOf fill:#codaf9,stroke:#0044d4,stroke-width:5px
+style AllOf fill:#codaf9,stroke:#0044d4,stroke-width:5px
+```
+
+## API design principles
+
+`oneOf` means "exactly one".
+When a value matches two of the listed schemas, no tool can tell which one was intended, and validators, code generators, and documentation all disagree about the result.
```
JLekawa:
```suggestion
When a value matches two of the listed schemas, no tool can tell which one was intended.
Validators, code generators, and documentation all disagree about the result.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Thanks for the suggestion. I added support of `defaultMapping` to the current rule. I am also considered about duplication between spec rule and this rule and decided to keep both, because we don't know if user use both rules or just `no-illogical-composition-keywords` alone. If user will have two duplicated errors he can just fix one issue and remove two errors or add these to lint ignore.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,209 @@
...
+
+The most common version of this is nullability: if a referenced schema already accepts `null` and the `oneOf` also lists `type: 'null'`, a null value matches both branches.
```
JLekawa:
```suggestion
The most common version of this issue is nullability: if a referenced schema already accepts `null` and the `oneOf` also lists `type: 'null'`, a null value matches both branches.
```

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: It has a separate check for the inline schemas.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,209 @@
...
+      oneOf:
+        - $ref: '#/components/schemas/Cat'
+    Animal:
+      allOf:
+        - $ref: '#/components/schemas/Cat'
+        - $ref: '#/components/schemas/Cat'
+        - {}
+```
+
+> `Pet` wraps a single schema, and `Animal` repeats one schema and adds an empty one that matches any value.
```
JLekawa:
```suggestion
`Pet` wraps a single schema, and `Animal` repeats one schema and adds an empty one that matches any value.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,209 @@
...
+        petType:
+          type: string
+    Dog:
+      type: object
+      properties:
+        petType:
+          type: string
+```
+
+> Add `petType` to `required` in both `Cat` and `Dog` to fix this.
```
JLekawa:
```suggestion
Add `petType` to `required` in both `Cat` and `Dog` to fix this.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,209 @@
...
+      oneOf:
+        - $ref: '#/components/schemas/Cat'
+        - type: object
+          properties:
+            petType:
+              type: string
+          required: [petType]
+```
+
+> Move the inline schema into `components/schemas` and reference it with a `$ref`.
```
JLekawa:
```suggestion
Move the inline schema into `components/schemas` and reference it with a `$ref`.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+style AnyOf fill:#codaf9,stroke:#0044d4,stroke-width:5px
+style AllOf fill:#codaf9,stroke:#0044d4,stroke-width:5px
+```
+
+## API design principles
+
+`oneOf` means "exactly one".
+When a value matches two of the listed schemas, no tool can tell which one was intended, and validators, code generators, and documentation all disagree about the result.
+
+The most common version of this is nullability: if a referenced schema already accepts `null` and the `oneOf` also lists `type: 'null'`, a null value matches both branches.
```
JLekawa:
```suggestion
One of the most common ambiguous patterns is nullability.
If a referenced schema already accepts `null` and the `oneOf` also lists `type: 'null'`, a null value matches both branches.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+The same ambiguity appears one level up, when the schema holding the `oneOf` is itself nullable and a member accepts `null` too.
+
+Deciding whether two arbitrary schemas overlap is not solvable in general, so the comparison stays deliberately narrow.
+It reads `type`, `nullable`, `enum`, `const`, `properties`, `required`, and `additionalProperties: false`.
+A `const` counts as a single-value `enum`, so one member can use `enum` and the other `const`.
```
JLekawa:
```suggestion
A `const` counts as a single-value `enum`.
One member can use `enum` and the other `const`.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+When a member uses any other constraint, such as `not`, `pattern`, `minimum`, or a nested `allOf`, that constraint may be what separates the schemas, so the rule reports nothing for the pair.
```
JLekawa:
```suggestion
When a member uses any other constraint, such as `not`, `pattern`, `minimum`, or a nested `allOf`, that constraint may be what separates the schemas.
In such cases, the rule reports nothing for the pair.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+
+A `discriminator` names the property that tells the members apart, so the rule trusts it and checks only what the specification requires.
```
JLekawa:
```suggestion
A `discriminator` names the property that tells the members apart.
The rule trusts it and checks only what the specification requires.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+
+> `Pet` wraps a single schema, and `Animal` repeats one schema and adds an empty one that matches any value.
+
+## Related rules
+
+- [no-schema-type-mismatch](../common/no-schema-type-mismatch.md)
+- [no-required-schema-properties-undefined](../common/no-required-schema-properties-undefined.md)
+- [spec-discriminator-defaultMapping](./spec-discriminator-defaultMapping.md) —
+  reports an optional discriminator property on OAS 3.2, where the specification requires it.
+  This rule covers OAS 3.0 and 3.1 instead, so enable both to check every version.
```
JLekawa:
```suggestion
- [spec-discriminator-defaultMapping](./spec-discriminator-defaultMapping.md) — on OAS 3.2, flags a discriminator whose `propertyName` is optional but has no `defaultMapping` required by OAS 3.2.
  `no-illogical-composition-keywords` checks the same issue on OAS 3.0 and 3.1, where `defaultMapping` doesn't exist.
  Enable both rules to cover every OAS version.
```

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Resources
+
```
tatomyr, line 221: Maybe mention the corresponding Adam's article here?

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,531 @@
...
+      if (schemaAllowsNull(parentSchema)) {
+        reportNullableParent(members, ctx);
+      }
+
+      reportAmbiguousMembers(members, hasDiscriminator ? discriminator : undefined, ctx);
+    },
+  };
+};
+
+// `struct` is not guaranteed to have rejected a malformed composition first.
```
tatomyr: Could you check it once again? It seems to me that it's impossible that the  parent is not an object.

packages/core/src/types/oas3.ts
```diff
@@ -914,6 +911,10 @@ export const Oas3Types = {
   Link,
   Logo,
   Schema,
+  // Registered so rules and configurable rules can target these node names.
```
tatomyr: This comment could be equally applied to every member of this object 😄.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+> Move the inline schema into `components/schemas` and reference it with a `$ref`.
+
+Example of an **incorrect** `discriminator`, where `petType` is optional:
+
+```yaml
+components:
+  schemas:
+    Pet:
```
tatomyr: Could you use Cafe example instead?

tatomyr (commented):
Almost there.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Thanks, added.

packages/core/src/types/oas3.ts
AlbinaBlazhko17: Yeap, forget to remove.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Replaced in docs and unit tests.

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Yeap, you are right. This weird comment is left from previous implementations. The purpose of this function is to avoid casts, because `ctx.parent` typed as `any`. So, i removed the function and just cast everywhere to proper type to avoid misunderstanding.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Two points looks wrong for me:
1. "nullable schemas whose `oneOf` also accepts `null`." - I mean, that if parent schema has a `null` and the sibling schema inside `oneOf` also accepts null we report a warning. So, i will say the proper sentence sounds:
> - schemas that accept null alongside a oneOf member that also accepts null

2. "discriminators whose property is missing from `required` in every member schema" - `discriminators` word looks weird for me, because we have one discriminator in one schema, that accepts one prop. So, the proper sentence sounds:
> - member schemas that omit the discriminator property from required

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
```diff
@@ -0,0 +1,525 @@
...
+  'readOnly',
+  'writeOnly',
+  'externalDocs',
+  'xml',
+]);
+
+export const NoIllogicalCompositionKeywords: Oas3Rule = () => {
+  return {
+    AllOf(members, ctx) {
+      const parentSchema = ctx.parent as CompositionSchema | undefined;
```
tatomyr: How could a parent be `undefined`?

docs/@v2/rules/oas/no-illogical-composition-keywords.md
```diff
@@ -0,0 +1,222 @@
...
+```yaml
+components:
+  schemas:
+    Beverage:
+      type: object
+      properties:
+        category:
+          type: string
+          const: beverage
+      required: [category]
```
tatomyr: I'd suggest against mixing json/yaml syntax in the examples unless necessary. Albeit they are valid, it's better to stick to the most consistent approach in examples to confuse users less.

tatomyr (approved):
Left minor comments, otherwise LGTM.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
JLekawa: Thank you! Updated and applied.

JLekawa (approved):

packages/core/src/rules/oas3/no-illogical-composition-keywords.ts
AlbinaBlazhko17: Removed this and guard. Added because `ctx.parent` has an `any` type and wanted to have additional check.

docs/@v2/rules/oas/no-illogical-composition-keywords.md
AlbinaBlazhko17: Thanks, changed.

Issue #2345, opened 2025-10-06
Proxy settings do not take effective when using respect


sareykaj:
**Describe the bug**

We are trying to execute an Arazzo workflow where the API needs to be accessed through a HTTP proxy server.
We are using the docker image like this:

`docker run -e HTTP_PROXY=$HTTP_PROXY -e HTTPS_PROXY=$HTTPS_PROXY -e http_proxy=$HTTP_PROXY -e https_proxy=$HTTPS_PROXY -v $PWD:/spec redocly/cli:2.2.2 respect test.arazzo.yaml --verbose ....`

And we get the error: 

```
...
    ✗ failed network request
 Tests exited with error 
```

**To Reproduce**
- only possible in our environment

**Expected behavior**

Network error should not occur.

**Logs**

see above

**OpenAPI description**

**Redocly Version(s)**

`2.2.2` docker image

**`Node.js` Version(s)**

n.a.

**OS, environment**

Linux

**Additional context**

none

DmitryAnansky:
@sareykaj 
Thanks for the report — really appreciate you bringing this to our attention.
We’ll investigate and follow up here with what we find.

tatomyr:
@sareykaj the fix has been released. Could you check if it works on your end?

PR #2371, opened 2025-10-16
fix: applied proxy to respect requests

DmitryAnansky:
## What/Why/How?

Applied proxy to respect requests.

```
HTTP_PROXY={URL} redocly respect arazzo.yaml --verbose
```

## Reference
Closes: https://github.com/Redocly/redocly-cli/issues/2345

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/commands/respect/connection-client.ts
+++ b/packages/cli/src/commands/respect/connection-client.ts
@@ -56,8 +56,19 @@ export function withConnectionClient(mtlsCerts: MtlsCerts = {}) {
     return fetch;
   }
 
-  return (input: string | URL | RequestInfo, init?: RequestInit) => {
-    const url = typeof input === 'string' ? input : 'url' in input ? input.url : input.toString();
+  return (input: RequestInfo, init?: RequestInit) => {
+    const url =
+      typeof input === 'string'
+        ? input
+        : input instanceof URL
+        ? input.toString()
+        : 'url' in input
+        ? input.url
+        : undefined;
+
+    if (!url) {
+      throw new Error('Invalid input URL');
+    }
 
     return fetch(input, {
       ...init,


.changeset/tender-pigs-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/respect-core": patch
```
tatomyr:
```suggestion
"@redocly/cli": minor
```

tatomyr: I don't see any changes in respect-core.

packages/cli/src/commands/respect/connection-client.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // Only proxy
+  if (proxyUrl) {
+    return new ProxyAgent({ uri: proxyUrl });
+  }
+
+  return undefined;
+}
+
+export function withConnectionClientIfNeeded(mtlsCerts: MtlsCerts = {}) {
```
tatomyr: What about this?

```suggestion
export function withConnectionClient(mtlsCerts: MtlsCerts = {}) {
```

packages/cli/src/commands/respect/connection-client.ts
```diff
@@ -0,0 +1,67 @@
...
+  const proxyUrl = getProxyUrl();
+
+  if (!mtlsCerts && !proxyUrl) {
+    return fetch;
+  }
+
+  return function fetchWithDispatcher(input: string | URL | RequestInfo, init?: RequestInit) {
```
tatomyr: Where is this function used?

DmitryAnansky: This will be used in respect api-fetcher.
I add a name to it just to describe better what I intend to do, can be anonymous though.

packages/cli/src/commands/respect/connection-client.ts
```diff
@@ -0,0 +1,67 @@
...
+    const url = typeof input === 'string' ? input : 'url' in input ? input.url : input.toString();
```
tatomyr: This is a bit hard to read. What if we do it this way:  
```suggestion
    const url = typeof input === 'string' 
      ? input 
      :  input instanceof URL 
      ? input.toString()
      :  input.url
```

PR #2377, opened 2025-10-20
chore: add support for entity files node types and validation

kamil1094:
## What/Why/How?
Added 2 new methods to support validation and autocompletion of catalog entities files:
- `createEntityTypes` - method creating node types for entity files validation and autocompletion
- `lintEntityFile` - method for validating entity file, returning problems for single object files and for files containing arrays of entities

## Check yourself

## Security


packages/cli/src/__tests__/fixtures/config.ts
```diff
@@ -25,6 +25,7 @@ export const configFixture: Config = {
     async3: {},
     arazzo1: {},
     overlay1: {},
+    catalogEntity1: {},
```
tatomyr: I don't think it's appropriate to add catalog entities format to the list of API specifications. I believe it should reside separately, more like we treat `redocly.yaml` format, so we don't mix different conceptions together.

packages/core/src/__tests__/consts.ts
```diff
@@ -0,0 +1,356 @@
...
+  'relatesTo',
+  'hasMember',
+  'memberOf',
+  'triggers',
+  'triggeredBy',
+  'returns',
+  'returnedBy',
+] as const;
+
+export const userMetadataSchema = {
```
tatomyr: It's better to use the json schema adapter instead of declaring the types. Why not just reuse the schemas  declared in `@redocly/config` package?

packages/cli/src/__tests__/fixtures/config.ts
kamil1094: Fixed

packages/core/src/__tests__/consts.ts
kamil1094: I've made some updates to the entity schema in `@redocly/config`. Once they are published I'll switch to using schemas from `@redocly/config`.

kamil1094: Updated

JLekawa (approved):

packages/core/src/__tests__/entity-yaml.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        if ('properties' in resolvedType) {
+          expect(resolvedType.properties).toHaveProperty('metadata');
+        }
+      }
+    }
+
+    if (typeof entityFileArrayNode.items === 'function') {
+      const unknownEntity = { key: 'unknown', title: 'Unknown' };
+      const resolvedType = entityFileArrayNode.items(unknownEntity, 'root');
+      console.log('Resolved type for entity without type:', resolvedType);
```
vadyvas: I think, it should be removed

packages/core/src/__tests__/lint.test.ts
```diff
@@ -1959,4 +1968,365 @@ describe('lint', () => {
 
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
   });
+
+  describe('lintEntityFile', () => {
+    it('should lint a valid user entity file', async () => {
+      const entityYaml = `
```
tatomyr: I'd suggest using outdent.

packages/core/src/rules/common/__tests__/entity-key-valid.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import { entityFileSchema, entityFileDefaultSchema } from '@redocly/config';
+
+function lintEntityKey(source: string): WalkContext['problems'] {
+  const document = makeDocumentFromString(source, '/test.yaml');
+  const entityTypes = createEntityTypes(entityFileSchema, entityFileDefaultSchema);
+  const types = normalizeTypes(entityTypes);
+
+  const ctx: WalkContext = {
+    problems: [],
+    specVersion: 'oas3_0',
```
tatomyr: This doesn't seem right. You are processing entities, which shouldn't know about the OpenAPI specification.

packages/core/src/rules/common/entity-key-valid.ts
```diff
@@ -0,0 +1,46 @@
...
+          report({
+            message: 'Entity `key` must be a string.',
+            location: location.child(['key']),
+          });
+          return;
+        }
+
+        if (key.length < MIN_KEY_LENGTH) {
+          report({
+            message: `Entity \`key\` must be at least ${MIN_KEY_LENGTH} characters long.`,
```
tatomyr:
```suggestion
            message: `Entity 'key' must be at least ${MIN_KEY_LENGTH} characters long.`,
```

Let's use single quotes for property names here and in other places.

packages/core/src/lint.ts
```diff
@@ -193,3 +198,83 @@ export async function lintConfig(opts: {
...
+  };
+
+  const entityTypes = createEntityTypes(entitySchema, entityDefaultSchema);
+  const types = normalizeTypes(entityTypes);
+
+  let rootType = types.EntityFileDefault;
+  if (Array.isArray(config.document.parsed)) {
+    rootType = types.EntityFileArray;
+  } else if (isPlainObject(config.document.parsed)) {
+    const ENTITY_DISCRIMINATOR_NAME = 'type';
```
tatomyr: Why declare it 2 times? (See https://github.com/Redocly/redocly-cli/pull/2377/files#diff-ae941b646daa95dba07dd9c55aa01b19187c8ec95db3ef02710dbadf34b62c08R11.)

kamil1094: Updated

packages/core/src/rules/common/__tests__/entity-key-valid.test.ts
kamil1094: What should I do then? Can i just change it to `specVersion: 'entity' as SpecVersion,` ?

tatomyr: I'm not sure TBH. I remember that there was a similar place where the version was used incorrectly. Let's leave a FIXME so I can refactor all of those together later.

kamil1094: okay, done 👍

packages/core/src/__tests__/entity-yaml.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    expect(entityTypes).toHaveProperty('EntityFileDefault');
+    expect(entityTypes).toHaveProperty('EntityFileArray');
+  });
+
+  it('should resolve entity type based on discriminator for array items', () => {
+    const entityTypes = createEntityTypes(entityFileSchema, entityFileDefaultSchema);
+    const normalizedTypes = normalizeTypes(entityTypes);
+
+    const entityFileArrayNode = normalizedTypes['EntityFileArray'];
+    if (typeof entityFileArrayNode.items === 'function') {
```
vadyvas: I think it's better to avoid conditional checks inside tests

packages/core/src/lint.ts
```diff
@@ -193,3 +198,82 @@ export async function lintConfig(opts: {
...
+  };
+
+  const entityTypes = createEntityTypes(entitySchema, entityDefaultSchema);
+  const types = normalizeTypes(entityTypes);
+
+  let rootType = types.EntityFileDefault;
+  if (Array.isArray(config.document.parsed)) {
+    rootType = types.EntityFileArray;
+  } else if (isPlainObject(config.document.parsed)) {
+    const typeValue = (config.document.parsed as Record<string, unknown>)[
```
vadyvas: You can avoid type cast here

packages/core/src/types/entity-yaml.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function createEntityTypes(
+  entitySchema: JSONSchema,
+  entityDefaultSchema: JSONSchema
+): Record<string, NodeType> {
+  const defaultNodeTypes = getNodeTypesFromJSONSchema('EntityFileDefault', entityDefaultSchema);
+
+  const namedNodeTypes = getNodeTypesFromJSONSchema('EntityFile', entitySchema);
+
+  const arrayNodeType = {
```
tatomyr, line 17: Have you tried fixing `getNodeTypesFromJSONSchema` instead? The code is okay, but I wonder if we'd benefit from making the adapter more generic for other similar cases.

kamil1094: Yes. I've tried but it would be quite hard to decide what is a default schema like I have `EntityFileDefault` in my custom method. We would need to use first item as default or something like that which is not as obvious as in my custom method.

packages/core/src/lint.ts
kamil1094: Thanks! Fixed.

packages/core/src/__tests__/entity-yaml.test.ts
kamil1094: Fixed.

packages/core/src/__tests__/lint.test.ts
```diff
@@ -1959,4 +1968,365 @@ describe('lint', () => {
...
+          (p) =>
+            p.message.toLowerCase().includes('enum') ||
+            p.message.includes('invalidRelationType') ||
+            p.message.includes('type')
+        )
+      ).toBe(true);
+    });
+
+    it('should validate type-specific metadata fields correctly', async () => {
+      const apiOperationYaml = `
```
tatomyr: Could you use outdent consistently?

packages/core/src/__tests__/entity-yaml.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      key: 'john-doe',
+      title: 'John Doe',
+      metadata: { email: 'john@example.com' },
+    };
+    const resolvedType = (entityFileArrayNode.items as ResolveTypeFn)(userEntity, 'root');
+
+    expect(resolvedType).toBeTruthy();
+    if (resolvedType && typeof resolvedType === 'object' && 'name' in resolvedType) {
+      expect(resolvedType.name).toBe('user');
+      if ('properties' in resolvedType) {
```
tatomyr: Please avoid `if` statements (especially nested) in tests.

packages/core/src/__tests__/lint.test.ts
```diff
@@ -1959,4 +1968,365 @@ describe('lint', () => {
...
+        metadata:
+          email: john@example.com
+        tags:
+          - engineering
+      `;
+
+      const document = makeDocumentFromString(entityYaml, '/entity.yaml');
+
+      const problems = await lintEntityFile({
+        config: { document } as Config,
```
tatomyr: Wait, it shouldn't be like this... Entity files aren't config.

packages/core/src/lint.ts
```diff
@@ -193,3 +198,80 @@ export async function lintConfig(opts: {
...
+
+  const normalizedVisitors = normalizeVisitors(rules, types);
+  const resolvedRefMap = await resolveDocument({
+    rootDocument: config.document,
+    rootType,
+    externalRefResolver,
+  });
+
+  walkDocument({
+    document: config.document,
```
tatomyr:
```suggestion
    document,
```

packages/core/src/lint.ts
```diff
@@ -193,3 +198,80 @@ export async function lintConfig(opts: {
 
   return ctx.problems;
 }
+
+export async function lintEntityFile(opts: {
+  config: Config;
```
tatomyr: You likely don't need config at all here. All you need is the document to lint.
In regular linter it's needed because the rules are taken from the config, and in `lintConfig` it's literally the document to lint 🙂.

packages/core/src/lint.ts
```diff
@@ -193,3 +198,80 @@ export async function lintConfig(opts: {
...
+  entitySchema: JSONSchema;
+  entityDefaultSchema: JSONSchema;
+  severity?: ProblemSeverity;
+  externalRefResolver?: BaseResolver;
+}) {
+  const {
+    config,
```
tatomyr:
```suggestion
    document,
```

packages/core/src/lint.ts
```diff
@@ -193,3 +198,80 @@ export async function lintConfig(opts: {
...
+    entitySchema,
+    entityDefaultSchema,
+    severity,
+    externalRefResolver = new BaseResolver(),
+  } = opts;
+
+  if (!config.document) {
+    throw new Error('Config document is not set.');
```
tatomyr:
```suggestion
  if (document) {
    throw new Error('Entity document is not set.');
```

packages/core/src/lint.ts
```diff
@@ -193,3 +198,80 @@ export async function lintConfig(opts: {
...
+  }
+
+  const ctx: WalkContext = {
+    problems: [],
+    specVersion: 'entity' as SpecVersion, // FIXME: this should be proper SpecVersion
+    config,
```
tatomyr:
```suggestion
```

And so on...

packages/core/src/__tests__/lint.test.ts
kamil1094: Fixed

packages/core/src/__tests__/entity-yaml.test.ts
kamil1094: Fixed.

tatomyr (changes requested):
I took another look and found that you used `config` improperly. Left another portion of comments. Sorry, I didn't notice that in the previous run.

packages/core/src/lint.ts
kamil1094: Got it. I think I've fixed all the places.

packages/core/src/rules/catalog-entity/entity-key-valid.ts
```diff
@@ -0,0 +1,46 @@
+import type { UserContext } from '../../walk.js';
```
tatomyr, line 1: Could you move the rule out of the `common/` folder as it's for rules that apply to all specifications. You can create a folder like `entity/` or `special/` for it.

packages/core/src/rules/common/entity-key-valid.ts
```diff
@@ -0,0 +1,46 @@
+import type { UserContext } from '../../walk.js';
+import type { Oas3Rule } from '../../visitors.js';
+
+const validKeyPattern = /^[a-z0-9]+(?:-[a-z0-9]+)*$/;
+const MIN_KEY_LENGTH = 2;
+const MAX_KEY_LENGTH = 150;
+
+export const EntityKeyValid: Oas3Rule = () => {
```
tatomyr: It's not OAS 3 rule. You can use `any` for the time (but please put a FIXME to update that later), or better declare a special type with only the `any` visitor (as you don't use anything else).

tatomyr:
Please confirm you've checked the changes in our internal products, and check other checkmarks if needed:

image

packages/core/src/rules/catalog-entity/entity-key-valid.ts
kamil1094: Sure. Done.

packages/core/src/rules/common/entity-key-valid.ts
kamil1094: Sure. I introduced `catalogEntityRule` that uses `baseVisitor`.

packages/core/src/rules/catalog-entity/entity-key-valid.ts
tatomyr, line 1: We try to use kebab-case for file naming. Please rename the folder.

packages/core/src/rules/catalogEntity/__tests__/entity-key-valid.test.ts
```diff
@@ -0,0 +1,126 @@
+import { describe, it, expect } from 'vitest';
+import { makeDocumentFromString } from '../../../resolve.js';
+import { Config } from '../../../config/index.js';
```
tatomyr: Have you forgotten to remove this?

packages/core/src/__tests__/lint.test.ts
```diff
@@ -5,9 +5,17 @@ import { BaseResolver } from '../resolve.js';
...
+  rootRedoclyConfigSchema,
+  entityFileDefaultSchema,
+  entityFileSchema,
+} from '@redocly/config';
 import { createConfigTypes } from '../types/redocly-yaml.js';
 import { fileURLToPath } from 'node:url';
+import { describe, it, expect } from 'vitest';
+import { lintEntityFile } from '../lint.js';
+import { makeDocumentFromString } from '../resolve.js';
+import { type Config } from '../config/index.js';
```
tatomyr: Forgot to remove?

packages/core/src/index.ts
```diff
@@ -148,3 +156,4 @@ export type {
   ResolvedSecurity,
 } from './typings/arazzo.js';
 export type { StatsAccumulator, StatsName } from './typings/common.js';
+export { getNodeTypesFromJSONSchema } from './types/json-schema-adapter.js';
```
tatomyr: Where do you want to use it?

tatomyr (approved):
Left minor comments. Otherwise looks good.

packages/core/src/rules/catalog-entity/entity-key-valid.ts
kamil1094: Fixed

packages/core/src/index.ts
kamil1094: Removed.

PR #2379, opened 2025-10-22
chore: update release pipelines to use trusted publisher instead of NPM_TOKEN

tatomyr:
## What/Why/How?

Updated the release pipeline to follow the trusted publishers workflow instead of using npm token:

- moved snapshot-release to `release.yalm` as we cannot configure two release action file in the same repo
- updated NodeJS version to 24 so we use NPM v11 (required for the trusted publirers workflow)
- updated smoke tests accordingly (to use node 24)
- added content permissions for github token
- updated action/checkout to v4

## Reference

https://github.com/Redocly/redocly/issues/18692

https://docs.npmjs.com/trusted-publishers

## Testing

Tested by publishing a snapshot version to NPM

## Screenshots (optional)

image

## Check yourself

## Security

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -17,7 +17,7 @@ jobs:
   latest-vs-next:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
           node-version: 24
@@ -50,7 +50,7 @@ jobs:
     if: github.head_ref == 'changeset-release/main'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
         with:
           node-version: 24
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,7 +25,7 @@ jobs:
       publishedPackages: ${{ steps.changesets.outputs.publishedPackages }}
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup Node.js
         uses: actions/setup-node@v3
@@ -115,7 +115,7 @@ jobs:
     continue-on-error: true
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-node@v3
       - name: Install CLI from npm
         run: npm i -g @redocly/cli@latest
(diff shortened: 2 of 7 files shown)


.github/workflows/performance.yaml
```diff
@@ -1,5 +1,9 @@
 name: Performance Benchmark
 
+permissions:
+  contents: read
+  pull-requests: write
```
tatomyr, line 5: Needed for leaving PR comments.

.github/workflows/release.yaml
```diff
@@ -170,6 +176,82 @@ jobs:
         uses: slackapi/slack-github-action@v1.25.0
         with:
           channel-id: C019K52TC0L #releases
-          payload-file-path: './output/release-message.json'
+          payload-file-path: ./output/release-message.json
         env:
           SLACK_BOT_TOKEN: ${{ secrets.RELEASES_SLACK_BOT_TOKEN }}
+
+  release-snapshot:
```
tatomyr, line 183: Copy-pasted the job from the deleted snapshot-release.yaml.

.github/workflows/smoke.yaml
```diff
@@ -26,7 +29,7 @@ jobs:
           path: __tests__/smoke/
           key: cache-${{ github.run_id }}-${{ github.run_attempt }}
 
-  run-smoke--npm--node-22:
```
tatomyr, line 29: Reorganized the jobs to be grouped by NodeJS version, and added a test suite for v24.

.github/workflows/smoke.yaml
```diff
@@ -75,10 +83,71 @@ jobs:
...
+        with:
+          node-version: 24
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
+      - run: bash ./__tests__/smoke/run-smoke.sh "pnpm add ./redocly-cli.tgz" "pnpm run"
 
-  run-smoke--yarn--node-22:
+  # Skipped due to a known Windows + NodeJS v24 issue: https://github.com/nodejs/node/issues/56645
```
tatomyr, line 110: Let's investigate what's wrong here. 
cc @Redocly/dark-side

.github/workflows/performance.yaml
```diff
@@ -16,8 +20,8 @@ jobs:
       - uses: actions/checkout@v3
```
vadyvas: maybe better to use `actions/checkout@v4`

.github/workflows/release.yaml
```diff
@@ -170,6 +176,82 @@ jobs:
...
+
+          ⚠️ Note: This is a development build and may contain unstable features."
+
+          gh pr comment $PR_NUMBER --body "$COMMENT"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+
+      - name: Install dependencies
+        run: npm install
```
vadyvas, line 239: `npm ci` ?

tatomyr: won't work because we're actually updating package-lock.json

.github/workflows/performance.yaml
tatomyr: Updated, please re-review

PR #2384, opened 2025-10-23
fix: error with invalid type when using exclusiveMininimum and exclusiveMaximum in OpenApi 3.0

### LINKED ISSUE #1610: Lint rules for examples require exclusiveMinimum to be number in openapi 3.0.x
**Describe the bug**

The linter is returning the error `Example validation errored: exclusiveMinimum value must be ["number"]`, however in openapi 3.0.x exclusiveMinimum must be of type boolean, not number. 

**To Reproduce**

1. Given the following openapi file:

```yaml
openapi: 3.0.2
info:
  title: Example API v2
  description: |-
    An  example API.
  contact: {}
  version: 2.0.0

paths:
  /myPath:
    get:
      summary: my custom path
      operationId: getMyPath
      parameters:
        - $ref: '#/components/parameters/a'
      responses:
        '200':
          description: Response description
          content:
            application/json:
              schema: {}

components:
  parameters:
    a:
      name: a
      in: header
      description: Example 
      schema:
        type: integer
        minimum: 10
        exclusiveMinimum: true
        example: 11
```

2. And the lint configuration:
```yaml
rules:
  no-invalid-schema-examples: error
```
3. Run this command 
`npx @redocly/cli lint openapi-example.yml --config openapi-example.conf.yml --format=stylish`

```
(node:53722) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
validating /.../openapi-example.yml:
  30:9  error    no-invalid-schema-examples  Example validation errored: exclusiveMinimum value must be ["number"].

openapi-example.yml: validated in 12ms

run `redocly lint --generate-ignore-file` to add all problems to the ignore file.
```

**Expected behavior**

The validation pass, because in openapi 3.0.x `exclusiveMinimum` MUST be of type boolean and is defined correctly.
This is different in openapi 3.1.0, but the specification provided is for 3.0.2.

**Redocly Version(s)**

1.17.1

**`Node.js` Version(s)**

22.2.0

**Additional context**

Same issue is present using `no-invalid-parameter-examples` rule.
See also changes between 3.0.x and 3.1.0: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

AlbinaBlazhko17:
## What/Why/How?
The ajv validator, which we are using to validate JSON Schemas didn't supports `exclusiveMininimum` and `exclusiveMaximum` properties as boolean type. That's why i need to use old `draft-04` version to have a proper validation for these properties. Also, this old `draft-04` version supports only old refs with uri, that's why we need to create own fork and publish own package to npm called `@redocly/ajv-draft-04`.

## Reference
Resolves #1610

## Testing
Ran tests new and created.

## Check yourself

## Security

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,8 +80,5 @@
     "packages/**/*.ts": [
       "eslint --fix"
     ]
-  },
-  "dependencies": {
-    "ajv-formats": "^3.0.1"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,6 @@
       "workspaces": [
         "packages/*"
       ],
-      "dependencies": {
-        "ajv-formats": "^3.0.1"
-      },
       "devDependencies": {
         "@changesets/cli": "^2.26.2",
         "@types/mark.js": "^8.11.5",
@@ -4939,9 +4936,9 @@
       }
     },
     "node_modules/ajv-formats": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/ajv-formats/-/ajv-formats-3.0.1.tgz",
-      "integrity": "sha512-8iUql50EUR+uUcdRQ3HDqa6EVyo3docL8g5WJ3FNcWmu62IbkGUue/pEyLBW8VGKKucTPgqeks4fIU1DA4yowQ==",
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/ajv-formats/-/ajv-formats-2.1.1.tgz",
+      "integrity": "sha512-Wx0Kx52hxE7C18hkMEggYlEifqWZtYaRgouJor+WMdPnQyEK13vgEWyVNup7SoeeoLMsr4kf5h6dOW11I15MUA==",
       "license": "MIT",
       "dependencies": {
         "ajv": "^8.0.0"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -80,5 +80,8 @@
     "packages/**/*.ts": [
       "eslint --fix"
     ]
+  },
+  "dependencies": {
+    "ajv-formats": "^3.0.1"
```
tatomyr: Why do we need it in the root dependencies?

packages/core/src/rules/ajv.ts
```diff
@@ -52,15 +75,32 @@ function getAjvValidator(
...
   schemaLoc: Location,
   instancePath: string,
   resolve: ResolveFn,
-  allowAdditionalProperties: boolean
+  allowAdditionalProperties: boolean,
+  specVersion: string = 'oas3_1'
 ): { valid: boolean; errors: (ErrorObject & { suggest?: string[] })[] } {
-  const validate = getAjvValidator(schema, schemaLoc, resolve, allowAdditionalProperties);
+  const validate =
+    specVersion === 'oas3_0' || specVersion === 'oas2'
```
tatomyr, line 109: Coud you add a comment with the link to a document which clarifies the JSON Schema drafts usage for these OAS versions?

tatomyr:
Could you check if we need to adjust types in the `no-invalid-schema-examples` rule itself? (See this [comment](https://github.com/Redocly/redocly-cli/issues/1610#issuecomment-2211120429).)

package.json
AlbinaBlazhko17: Thanks for the observation, i removed it, but i think, we need to upgrade ajv-formats package to 3.0.1 in core, because old version causes errors.

packages/core/src/rules/ajv.ts
AlbinaBlazhko17: Draft-04 handles only validation of `exclusiveMinimum` and `exclusiveMaximum` for boolean, and all other validations it handles using types/functions/library/vocabulary from basic AJV.

vadyvas:
resolved in https://github.com/Redocly/redocly-cli/pull/2564

PR #2391, opened 2025-10-28
fix: add validation to mount-path

kanoru3101:
## What/Why/How?
`mount-path` was not checked, which led to errors when used with an empty path or a path identical (`/`) to the project path.
## Testing
Error when using `--mount-path` or `--mp` with an empty path or root path.

```
Mount path cannot be empty or root path. Please use --mount-path option with a valid path.
```

## Screenshots (optional)
Example of usage
Screenshot 2025-10-28 at 15 16 32

## Check yourself

## Security


packages/cli/src/reunite/commands/push.ts
```diff
@@ -54,6 +54,12 @@ export async function handlePush({
     );
   }
 
+  if (!mountPath || mountPath === '/') {
```
tatomyr: Could you check if it could be checked inside the `yargs` layer? Looks like it's more appropriate place for that.

kanoru3101: The yargs layer looks better. Moved

.changeset/violet-jobs-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": patch
+---
+
+Fixed an issue with validation the `mount-path` option.
```
JLekawa:
```suggestion
Fixed an issue where the `mount-path` option was not validated, leading to errors when used with an empty path or a path identical to the project path.
```

tatomyr:
Have you considered updating the corresponding documentation?
image

kanoru3101:
> Have you considered updating the corresponding documentation? image

Updated information about the path in the `push` command for `mount-path`.

JLekawa (approved):

docs/@v1/commands/push.md
```diff
@@ -37,7 +37,7 @@ REDOCLY_AUTHORIZATION=<api-key> redocly push <files> --organization <organizatio
 | files                 | [string] | **REQUIRED.** List of folders and/or files to upload.                                                                                                                                  |
 | --organization, -o    |  string  | **REQUIRED.** Organization slug.                                                                                                                                                       |
 | --project, -p         |  string  | **REQUIRED.** Project slug.                                                                                                                                                            |
-| --mount-path, -mp     |  string  | **REQUIRED.** The path where the files are mounted in the project.                                                                                                                     |
+| --mount-path, -mp     |  string  | **REQUIRED.** The path where the files are mounted in the project. Cannot be empty or identical to the project path.                                                                   |
```
mark-redocly, line 40: I not sure we need add this message. Because  project path always `/`

packages/cli/src/utils/__tests__/validate-mount-path.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+describe('validateMountPath', () => {
+  it('should accept valid mount path', () => {
+    expect(validateMountPath('/docs')).toBe('/docs');
+    expect(validateMountPath('/api/v1')).toBe('/api/v1');
+    expect(validateMountPath('/my-path')).toBe('/my-path');
+  });
+
+  it('should reject empty mount path', () => {
+    expect(() => validateMountPath('')).toThrow(
```
mark-redocly, line 12: Should be empty validated by `required` option?

kanoru3101: We have `required` option for `mount-path`. But it doesn't cover the case when `--mount-path=''`

docs/@v1/commands/push.md
kanoru3101: Yes, but it is better to leave a message so that other clients do not try to use `--mount-path='/'`, as this is also an incorrect path.

mark-redocly (approved):

Issue #1371, opened 2023-12-28
Redocly CLI incorrectly resolves $refs in example


tatomyr:
**Describe the bug**

**To Reproduce**

1. Given this `redocly.yaml` file
2. And this OpenAPI file(s)
3. Run this command with these arguments... `redocly ...`

**Expected behavior**

**Logs**

**OpenAPI description**

**Redocly Version(s)**

**`Node.js` Version(s)**

**Additional context**

PR #2392, opened 2025-10-28
fix: do not resolve $ref in examples.value

vadyvas:
## What/Why/How?

What: Stop resolving $ref inside Examples.value and dataValue (oas3.2) during bundling. Only resolve when the Example itself is a Reference Object, not when a $ref appears inside the literal value.
Why: Per OpenAPI spec and Redocly refs guide, Examples.value must be treated literally; resolving $ref there is incorrect.

```yaml
openapi: 3.2.0
info:
  title: Test API
  version: 1.0.0
paths:
  /test:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  id:
                    type: integer
                example:
                  $ref: example.json # should resolve
              examples:
                Test:
                  value:
                    $ref: example.json # should NOT resolve

```

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/1371

## Testing

Added new unit and e2e tests

## Check yourself

## Security

### DIFF
--- a/tests/e2e/bundle/bundle-example-value-with-ref/example.json
+++ b/tests/e2e/bundle/bundle-example-value-with-ref/example.json
@@ -1,4 +1,0 @@
-{
-  "id": 123,
-  "name": "Test User"
-}
--- a/tests/e2e/bundle/bundle-example-value-with-ref/openapi.yaml
+++ b/tests/e2e/bundle/bundle-example-value-with-ref/openapi.yaml
@@ -12,21 +12,17 @@ paths:
               schema:
                 type: object
                 properties:
-                  id:
-                    type: integer
-                  name:
+                  $ref:
                     type: string
-                  active:
-                    type: boolean
                 example:
-                  $ref: example.json
+                  $ref: example.json # should NOT be resolved
               examples:
                 Test:
                   summary: Example with $ref in value (should NOT resolve)
                   value:
-                    $ref: example.json
+                    $ref: example.json # should NOT be resolved
 
                 TestDataValue:
                   summary: Example with $ref in dataValue (should NOT resolve)
                   dataValue:
-                    $ref: example.json
+                    $ref: example.json # should NOT be resolved
--- a/tests/e2e/bundle/bundle-example-value-with-ref/snapshot.txt
+++ b/tests/e2e/bundle/bundle-example-value-with-ref/snapshot.txt
@@ -12,12 +12,8 @@ paths:
               schema:
                 type: object
                 properties:
-                  id:
-                    type: integer
-                  name:
+                  $ref:
                     type: string
-                  active:
-                    type: boolean
                 example:
                   $ref: example.json
               examples:


.changeset/loud-mails-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Do not resolve `$ref` inside `Examples.value` during bundling and linting
```
JLekawa:
```suggestion
Fixed an issue where the content of `$ref`s inside `Examples.value` was erroneously resolved during bundling and linting.
```

__tests__/bundle/bundle-example-value-with-ref/openapi.yaml
```diff
@@ -0,0 +1,92 @@
...
+                    object:
+                      nested: true
+
+                TestMultipleRefs:
+                  summary: Example with multiple $ref in value
+                  value:
+                    $ref: example.json
+                    fallback:
+                      $ref: example.json
+                    static: 'constant'
```
tatomyr: What do you need those cases for? Do they cover anything that was not covered already?
Please try keeping tests concise.

packages/core/src/types/oas2.ts
```diff
@@ -429,7 +429,7 @@ const SecurityRequirement: NodeType = {
 
 const Example: NodeType = {
   properties: {
-    value: { isExample: true },
+    value: { isExample: true, resolvable: false },
```
tatomyr: Wait, there's something wrong with this. There's no `value` or anything like that in  OAS2 [Example Object](https://swagger.io/specification/v2/).

packages/core/src/types/oas3.ts
```diff
@@ -233,7 +233,7 @@ const MediaType: NodeType = {
 
 const Example: NodeType = {
```
tatomyr, line 234: And the example/examples are different in OAS 3: https://spec.openapis.org/oas/v3.2.0.html#media-type-object.
Maybe we have to adjust the node types to reflect that first?

packages/core/src/types/oas2.ts
vadyvas: This example is not from the oas specification; it’s specifically for our custom x-examples extension, so I’ve renamed it to `XExample` to avoid confusion

packages/core/src/types/oas3.ts
vadyvas: There is no separate type for a single "any" `Example` in the spec – it only defines ExampleObject, which corresponds to the Example type we have here. For a single any example we use an inline shape with `{ isExample: true }` instead (see the existing usage here: https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/types/oas3.ts#L227
).

If we split this into separate `Example` and `ExampleObject` types, it adds complexity for lint rules and visitors because they would need to handle two different types.

Given that, do you think it’s still worth introducing a separate `Example` type here?

.changeset/loud-mails-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed an issue where the content of `$ref`s inside `Examples.value` was erroneously resolved during bundling and linting.
```
tatomyr:
```suggestion
Fixed an issue where the content of `$ref`s inside example values was erroneously resolved during bundling and linting.
```

.changeset/loud-mails-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
```
tatomyr, line 2: It also affects the CLI behaviour, so please mention it here as well.

docs/@v2/configuration/reference/resolve.md
```diff
@@ -21,7 +21,7 @@ One HTTP header is supported for each URL resolved.
 
 - doNotResolveExamples
 - boolean
-- When running `lint`, set this option to `true` to avoid resolving `$ref` fields in examples. Resolving `$ref`s in other parts of the API is unaffected.
+- Set this option to `true` to prevent resolving `$ref` fields in `Example`. This affects both `lint` and `bundle` commands. Resolving `$ref`s in other parts of the API description is unaffected.
```
tatomyr: It doesn't affect the `Example` type, only singular `example` properties.

docs/@v2/configuration/resolve.yaml
```diff
@@ -12,7 +12,8 @@ properties:
   doNotResolveExamples:
     type: boolean
     description: >-
-      You can stop `lint` from resolving `$refs` in examples by setting this configuration option to `true`.
+      Set this option to `true` to prevent resolving `$ref` fields in `Example`.
```
tatomyr: Same here.

tests/e2e/bundle/bundle-example-value-with-ref/openapi.yaml
```diff
@@ -0,0 +1,32 @@
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  id:
+                    type: integer
+                  name:
+                    type: string
+                  active:
+                    type: boolean
```
tatomyr:
```suggestion
                  $ref:
                     type: string
```

tests/e2e/bundle/bundle-example-value-with-ref/example.json
```diff
@@ -0,0 +1,4 @@
+{
```
tatomyr: Remove plz.

PR #2396, opened 2025-10-30
chore: require Redocly config update before release

tatomyr:
## What/Why/How?

Added a check on Release PR that will fail if Redocly config package version is outdated.

## Reference

Intrernal discussion: https://redoc-ly.slack.com/archives/C02UHE7EL3E/p1761816060204409?thread_ts=1761785533.570059&cid=C02UHE7EL3E

## Testing

Should be tested on a Release PR.

## Security

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -23,7 +23,6 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 24
-          cache: npm
       - name: Check Redocly Config Version
         run: |
           LATEST_VERSION=$(npm view @redocly/config version)


.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,42 @@
...
+  REDOCLY_TELEMETRY: off
+
+jobs:
+  checks:
+    # Run only on release branch (changeset-release/main):
+    if: github.head_ref == 'changeset-release/main'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
```
vadyvas, line 23: can be upgraded to a newer one

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          node-version: 24
+          cache: npm
```
vadyvas: Is cache needed?

tatomyr: removed

.github/workflows/pre-release.yaml
tatomyr: Needs to be updated in all workflows. Would be better to do it in a separate PR.

PR #2399, opened 2025-10-30
chore: update typescript to v5.9.3 

kanoru3101:
## What/Why/How?
Updated TS to the latest version 
## Check yourself

## Security

### DIFF
--- a/.changeset/lucky-chicken-switch.md
+++ b/.changeset/lucky-chicken-switch.md
@@ -1,7 +1,0 @@
----
-"@redocly/respect-core": patch
-"@redocly/openapi-core": patch
-"@redocly/cli": patch
----
-
-Update typescript to v5.9.3


packages/cli/src/reunite/api/api-client.ts
```diff
@@ -190,9 +190,9 @@ class RemotesApi {
     payload.commit.createdAt && formData.append('commit[createdAt]', payload.commit.createdAt);
 
     for (const file of files) {
-      const blob = Buffer.isBuffer(file.stream)
-        ? new Blob([file.stream])
-        : new Blob([await streamToBuffer(file.stream)]);
+      const buffer = Buffer.isBuffer(file.stream) ? file.stream : await streamToBuffer(file.stream);
```
tatomyr: Could you check how it's implemented in the VS Code extension? They should be the same.

.changeset/lucky-chicken-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@redocly/respect-core": patch
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Update typescript to v5.9.3
```
kanoru3101: Do we need changeset for the updated version?

packages/cli/src/reunite/api/api-client.ts
kanoru3101: I couldn't find it in the `vs-code-extention`. But `btoa` is used in `open-api-language-server`.
In some other packages, it is simply added as `as BlobPart`. I did the same and tested it with the `push` command, and  it worked.

.changeset/lucky-chicken-switch.md
tatomyr: No, it's a chore. Please remove.

packages/cli/src/reunite/api/api-client.ts
```diff
@@ -191,8 +191,8 @@ class RemotesApi {
 
     for (const file of files) {
       const blob = Buffer.isBuffer(file.stream)
-        ? new Blob([file.stream])
-        : new Blob([await streamToBuffer(file.stream)]);
+        ? new Blob([file.stream as BlobPart])
```
kanoru3101, line 194: @tatomyr Tested these changes, looks like it works the same

tatomyr (approved):

Issue #2410, opened 2025-11-06
`no-invalid-media-type-examples` not triggering warning for unevaluated properties


Breed43214:
**Describe the bug**

From v2.8.0 onwards, no-invalid-media-type-examples rule does not warn on invalid media type examples if any component schema contains an example when using the recommended ruleset.

**To Reproduce**

Take the example YAML below:

```
openapi: 3.0.3
info:
  title: Test API
  version: 1.0.0
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
  description: |
    Test API for Bug Report
servers:
  - url: https://test.com/v1

security:
  - SAFTokenAuth: []
  - OAuth2: []

paths:
  /test1:
    get:
      summary: Test Request
      operationId: getTest1
      responses:
        200:
          description: OK
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/testSchemaWithExamples"
              examples:
                scenario1:
                  value:
                    testArray:
                      - name: John
                        id: 1
                      - name: James
                        id: 2
                      - name: Joe
                        id: 3
                      - names: Jill
                        id: 4
        404:
          description: Not Found
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/errorResponse"
              examples:
                scenario1:
                  value:
                    code: 60
                    message: Not Found
                    reasons:
                      - errorReason: The requested resource could not be found and does not exist.
                        recommendedAction: Please correct the request and try again.
                        property: null
components:
  schemas:
    testSchemaWithExamples:
      type: object
      properties:
        testArray:
          type: array
          items:
            type: object
            properties:
              name:
                type: string
                #example: Joe
              id:
                type: integer  
    errorResponse:
      type: object
      properties:
        code:
          type: integer
        message:
          type: string
        reasons:
          type: array
          items:
            type: object
            properties:
              errorReason:
                type: string
              recommendedAction:
                type: string
              property:
                type: string
                nullable: true
  securitySchemes:
    SAFTokenAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
    OAuth2:
      type: oauth2
      flows:
        clientCredentials:
          tokenUrl: /oauth/token
          scopes: {}
```

[...]

PR #2411, opened 2025-11-06
fix: allowAdditionalProperties option in example validation rules

vadyvas:
## What/Why/How?

Align example validation with the `allowAdditionalProperties` option across example-related rules.

Previously:

- `validateExample` for a single example always passed `allowAdditionalProperties: true`.
- `no-invalid-parameter-examples` used `allowAdditionalProperties: true` for multi examples, but respected `opts.allowAdditionalProperties` only for single examples.

## Reference

- Resolves https://github.com/Redocly/redocly-cli/issues/2410

## Testing

- Updated/added unit tests for example-related rules.
- Updated e2e tests for

## Check yourself

## Security


DmitryAnansky (approved):

JLekawa (approved):

packages/core/src/rules/common/no-invalid-schema-examples.ts
```diff
@@ -31,7 +31,13 @@ export const NoInvalidSchemaExamples: Oas3Rule | Oas2Rule = (opts: any) => {
             return;
           }
 
-          validateExample(schema.example, schema, ctx.location.child('example'), ctx, true);
+          validateExample(
+            schema.example,
+            schema,
+            ctx.location.child('example'),
+            ctx,
+            !!opts.allowAdditionalProperties
```
tatomyr, line 39: This fixes the issue only partially. If the configuration of the rules is different, there will be a mismatch since the first rule that runs instantiates the AJV instance with its config and ignores the others.

PR #2412, opened 2025-11-07
fix: no-invalid-media-type-examples now respects readOnly and writeOnly properties

### LINKED ISSUE #1416: validation of sample : readOnly / writeOnly  and required statement interaction
**Describe the bug**

given an API specification mixing required keyword and readOnly , writeOnly redocly generate a warning wheras it should not

**To Reproduce**
given the minimal files attached  [small.json](https://github.com/Redocly/redocly-cli/files/14097153/small.json)
generate a warning 

**Expected behavior**

the object contains Id , name , password ,  both are required ... in differents scenario 

id is set to ReadOnly 
password as WriteOnly 

i would expect to do a single representation of my object  , for input / output  and exemple matching 

this is a quite common use case (especially for id that is generated by server so mandatory in return only ... so readOnly) 
i extended the use case to writeOnly and password just for the record. 

The Json Schema specification is not 100 % clear if there is a precedence between required statement and the readOnly and writeOnly context  . 

**Logs**
```
Example value must conform to the schema: must have required property 'id'.

80 | "inputExample" : {
81 |     "description" : "sampleOfInput",
82 |     "value" : {
83 |         "name" : "Doe",
84 |         "password" : "securityFirst"
85 |     }
86 | },
87 | "outputExample" : {

referenced from small.json:23:45 at #/paths/~1test/post/requestBody/content/application~1json

Warning was generated by the no-invalid-media-type-examples rule.


[2] small.json:89:27 at #/components/examples/outputExample/value

Example value must conform to the schema: must have required property 'password'.

87 |     "outputExample" : {
88 |         "description" : "sampleOfOutPut",
89 |         "value" : { 
90 |             "id" : "goodId",
91 |             "name" : "Doe"
92 |         }
93 |     }   
94 | }

referenced from small.json:39:49 at #/paths/~1test/post/responses/200/content/application~1json

Warning was generated by the no-invalid-media-type-examples rule. 
``` 

**Redocly Version(s)**

redocly 1.8.1

**Additional context**

thx for your attention , do not hesitate to close it if you consider it as too corner case

sakurai-ryo:
## What/Why/How?

Fixed an issue where the `no-invalid-media-type-examples` rule incorrectly reported validation errors/warnings when example/examples properly omitted readOnly properties in request bodies or writeOnly properties in response bodies.

ajv does not support `readOnly` and `writeOnly` properties, and the request and response context is specific to OpenAPI.
So this must be handled in the lint rule rather than in ajv.
https://github.com/ajv-validator/ajv/issues/2097

### How

This PR introduces context-aware validation.
It modifies the object schema based on whether mediaType examples are request or response context.

Check the `properties` field of `mediaType.schema` and remove properties that should not be required in the request or response context from the `properties` field and the `required` field.

This PR skips validating cases where mediaType.schema contains complex elements like allOf.
If there are any utility functions that support this case, please let us know.

## Reference

Fixes #1416

## Testing

Added unit tests.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
+++ b/packages/core/src/rules/oas3/__tests__/no-invalid-media-type-examples.test.ts
@@ -967,7 +967,7 @@ describe('no-invalid-media-type-examples', () => {
       expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
     });
 
-    it.only('should report if examples include contextually invalid values', async () => {
+    it('should report if examples include contextually invalid values', async () => {
       const document = parseYamlToDocument(
         outdent`
         openapi: 3.1.0


packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -50,3 +57,95 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
...
+  return (
+    mediaTypeLocation.pointer.includes('/components/requestBodies') ||
+    /paths\/[^/]+\/[^/]+\/requestBody/.test(mediaTypeLocation.pointer)
+  );
+}
+
+function isResponseExample(mediaTypeLocation: Location): boolean {
+  return (
+    mediaTypeLocation.pointer.includes('/components/responses') ||
+    /paths\/[^/]+\/[^/]+\/responses/.test(mediaTypeLocation.pointer)
```
tatomyr, line 114: This is a bit fragile and hard to read. I suggest using `Responses` and `RequestBody` visitors to narrow down the context:

```ts
Responses: {
    MediaType: {
      leave(mediaType, ctx: UserContext) {
         // Filter out `writeOnly` props...
```

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -13,35 +13,42 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
...
             if (!resolved.location) return;
-            location = isMultiple ? resolved.location.child('value') : resolved.location;
+            exampleLocation = isMultiple ? resolved.location.child('value') : resolved.location;
             example = resolved.node;
           }
           if (isMultiple && typeof example?.value === 'undefined') {
             return;
           }
+
+          const resolvedSchema = ensureSchemaIsResolved(mediaType.schema!, ctx);
```
tatomyr, line 44: I guess it will still be easier to fix that on the AJV side. We use our own AJV fork, and we can try passing down a new parameter, something like `ignoreWriteOnly`. Let me think this through.

tatomyr (commented):
Thank you for spending your time on this @sakurai-ryo! Left a couple of suggestion/concerns. 
Also, I'm not sure how this solution will work with circular references and nested objects (for instance, objects in arrays). Could you check that?

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
AlbinaBlazhko17: @tatomyr you are absolutely right. In AJV it looks better. I try to create a quick fix directly in AJV and it works. I created draft PR: https://github.com/Redocly/ajv/pull/35.

vadyvas:
Fixed in https://github.com/Redocly/redocly-cli/pull/2521
@sakurai-ryo thanks for your time

PR #2424, opened 2025-11-24
chore: improve telemetry collecting using generated SDK

AlbinaBlazhko17:
## What/Why/How?
Updated CLI telemetry to fit current otel implementation and added properties to get better analytics.

**Changes:**
- Installed SDK package, that contains types for `cloudEvent`.
- Moved creation of the nodeTracerProvider to constructor to reuse it among the class and for better code readability.
- Added `session_id` to `ResourceAttributes`.
- Added `attributes` to fit the current otel setup. This change resolves previous issue with the removing analytics, because attribute didn't passed.
- Added functionality to store in tmp directory `anonymous_id` which is generated using `ulid` package.
- Added to analytics `execution_time`.
- Minor improvements and code refactoring.

## Testing
Locally with DB.

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/utils/telemetry.ts
+++ b/packages/cli/src/utils/telemetry.ts
@@ -181,10 +181,6 @@ function cleanString(value: string): string {
     return 'folder';
   }
 
-  if ((value.includes('/') || value.includes('\\')) && /\.[^./\\]+$/.test(value)) {
-    return value.replace(/.+\.([^.]+)$/, 'invalid-file');
-  }
-
   return value;
 }
 
--- a/packages/cli/src/__tests__/utils.test.ts
+++ b/packages/cli/src/__tests__/utils.test.ts
@@ -715,21 +715,6 @@ describe('cleanArgs', () => {
       })
     );
   });
-
-  it('should clean non-existent file paths that look like files', () => {
-    const rawInput = ['redocly', 'lint', '/Users/username/test/openapi.yam'];
-    const parsedArgs = {
-      apis: ['/Users/username/test/openapi.yam'],
-    };
-    const result = cleanArgs(parsedArgs, rawInput);
-
-    expect(result.raw_input).toEqual('redocly lint invalid-file');
-    expect(result.arguments).toEqual(
-      JSON.stringify({
-        apis: ['invalid-file'],
-      })
-    );
-  });
 });
 
 describe('validateFileExtension', () => {


packages/cli/src/utils/otel.ts
```diff
@@ -4,19 +4,19 @@ import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http';
 import { ATTR_SERVICE_NAME, ATTR_SERVICE_VERSION } from '@opentelemetry/semantic-conventions';
 import { version } from './package.js';
 import { OTEL_TRACES_URL, DEFAULT_FETCH_TIMEOUT } from './constants.js';
+import { isDefined } from '../../../core/src/utils/is-defined.js';
```
tatomyr: The path is wrong. You should import it from `@redocly/openapi-core`.

AlbinaBlazhko17: Yeah, i tried to use it, but `@redocly/openapi-core` doesn't have an exported member `isDefined`, should i add it to the export?

tatomyr: Let's just compare to `undefined`.

packages/cli/src/utils/telemetry.ts
```diff
@@ -188,12 +196,12 @@ function cleanObject(obj: any, keysToClean: string[]): any {
...
   }
 
-  const cleaned: any = {};
+  const cleaned: Record<string, unknown> = {};
 
   for (const [key, value] of Object.entries(obj)) {
     if (keysToClean.includes(key)) {
       cleaned[key] = SECRET_REPLACEMENT;
-    } else if (typeof value === 'object' && value !== null) {
+    } else if (isPlainObject(value)) {
```
tatomyr, line 205: Could this also be an array?

AlbinaBlazhko17: Currently working on these two functions. I will check

AlbinaBlazhko17: I refactored this function. This function uses only to clean args inside mtls argument, which was added a month ago. We discussed with Dmytro, that inside mtls we cannot use array, so i changed this logic inside the function and added proper typing.

packages/cli/src/utils/telemetry.ts
```diff
@@ -68,23 +54,27 @@ export async function sendTelemetry({
...
     const { RedoclyOAuthClient } = await import('../auth/oauth-client.js');
     const oauthClient = new RedoclyOAuthClient();
     const reuniteUrl = getReuniteUrl(config, args.residency);
     const logged_in = await oauthClient.isAuthorized(reuniteUrl);
-    const data: Analytics = {
-      event: 'cli_command',
-      event_time,
+    let anonymous_id = getCachedAnonymousId();
+    if (!anonymous_id) {
+      anonymous_id = `ann_${ulid()}`;
```
DmitryAnansky, line 63: should it start with `anon_` ?

packages/cli/src/utils/telemetry.ts
```diff
@@ -97,8 +87,35 @@ export async function sendTelemetry({
...
 
+    const cloudEvent: CloudEvents.CommandRanMessage = {
+      id: `evt_${ulid()}`,
+      time: new Date().toISOString(),
+      type: 'command.ran',
+      object: 'event',
+      specversion: '1.0',
+      datacontenttype: 'application/json',
+      source: 'com.redocly.cli',
+      origin: 'ui',
```
DmitryAnansky: I am wondering why origin is `ui` for CLI commands?

packages/cli/src/utils/telemetry.ts
```diff
@@ -257,3 +270,36 @@ export function cleanArgs(parsedArgs: CommandArgv, rawArgv: string[]) {
...
+  const isCI = !!process.env.CI;
+  if (isCI) {
+    return;
+  }
+
+  try {
+    const anonymousIdFile = join(tmpdir(), ANONYMOUS_ID_CACHE_FILE);
+
+    if (!existsSync(anonymousIdFile)) {
+      return;
```
DmitryAnansky, line 298: Should we just generate a new anonymousID as a fallback in case there is no access to stored file?

packages/cli/src/utils/telemetry.ts
```diff
@@ -257,3 +270,36 @@ export function cleanArgs(parsedArgs: CommandArgv, rawArgv: string[]) {
...
+    }
+
+    return readFileSync(anonymousIdFile).toString().trim();
+  } catch (e) {
+    return;
```
DmitryAnansky, line 303: Please consider dynamic anonymousID generation as a fallback.

docs/@v2/usage-data.md
```diff
@@ -19,6 +19,8 @@ When a command is run, the following data is collected:
 - whether the `redocly.yaml` configuration file exists
 - API specification version
 - Platform (Linux, macOS, Windows)
+- Anonymous id (a randomly generated identifier that doesn't contain personal information)
```
JLekawa:
```suggestion
- Anonymous ID (a randomly generated identifier that doesn't contain personal information)
```

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: I asked @SerhiiOselskyi and he said, that we always send `ann_${ulid()}`.

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: Yeah, you are right, but property `origin` can be equal to `api` or `ui`, because it is enum.

packages/cli/src/utils/telemetry.ts
```diff
@@ -97,8 +87,35 @@ export async function sendTelemetry({
...
+          uri: '',
+        },
+      ],
+      environment: process.env.REDOCLY_ENVIRONMENT,
+      sourceDetails: {
+        id: anonymous_id,
+        object: 'user',
+        uri: '',
+      },
+      data: eventData,
```
kanoru3101, line 114: What about `project.id`, `project.slug`, `organization.id`, `organization.slug`? Will they be `''` in ClickHouse?

AlbinaBlazhko17: Inside AsyncAPI definition of the event i didn't declared properties, that you provided, because we don't have them in CLI at all

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: I will add fallbacks, but what about CI, do we need to return an anonymousID or `Anonymous` string?

AlbinaBlazhko17: Wait, i don't think, that i need to add fallback inside this function, because if anonymousID doesn't exist, i cached it and then passed generated anonymousID to the telemetry.
https://github.com/Redocly/redocly-cli/pull/2424/files#diff-6d078e3ba2cd199687fb2ab9da466d821a9556161dcdb00b6cf25d0002b562cfR63

DmitryAnansky: Just another ulid
```
I will add fallbacks, but what about CI, do we need to return an anonymousID or Anonymous string?
```

packages/cli/src/utils/telemetry.ts
DmitryAnansky: But should there be some additional enum value, like `cli`.
Do you know why it is restricted only to this two values?
Is it SDK missing some additional config?

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: Thanks for the suggestion, added these properties

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: Thanks for the suggestion, i updated `origin` in SDK and in CLI

packages/cli/src/utils/telemetry.ts
```diff
@@ -97,8 +87,35 @@ export async function sendTelemetry({
           : undefined,
     };
 
+    const cloudEvent: CloudEvents.CommandRanMessage = {
+      id: `evt_${ulid()}`,
+      time: new Date().toISOString(),
+      type: 'command.ran',
+      object: 'event',
+      specversion: '1.0',
```
DmitryAnansky, line 95: Is '1.0' is ok for specversion initial value?

DmitryAnansky (approved):

PR #2432, opened 2025-11-28
feat: add functionality to change default unevaluated properties on demand

### LINKED ISSUE #2410: `no-invalid-media-type-examples` not triggering warning for unevaluated properties

AlbinaBlazhko17:
## What/Why/How?
- Added using new setter method to set on demand default unevaluated properties.
- Refactored `no-invalid-media-type-examples`, `no-invalid-parameter-examples`, `no-invalid-schema-examples` rules.
- Set to validateExamples function default property of defaultUnevaluatedProperties to true due to OpenApi spec. OpenApi spec by default sets `additionalProperties` for objects to true.

## Reference
Resolves #2410 

## Testing
Ran tests and added new, which cover specific case.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/common/no-invalid-parameter-examples.ts
+++ b/packages/core/src/rules/common/no-invalid-parameter-examples.ts
@@ -8,17 +8,13 @@ export const NoInvalidParameterExamples: any = (opts: any) => {
   return {
     Parameter: {
       leave(parameter: Oas3Parameter, ctx: UserContext) {
-        const allowAdditionalProperties = isDefined(opts.allowAdditionalProperties)
-          ? opts.allowAdditionalProperties
-          : false;
-
         if (isDefined(parameter.example)) {
           validateExample(
             parameter.example,
             parameter.schema!,
             ctx.location.child('example'),
             ctx,
-            allowAdditionalProperties
+            !!opts.allowAdditionalProperties
           );
         }
 
@@ -30,7 +26,7 @@ export const NoInvalidParameterExamples: any = (opts: any) => {
                 parameter.schema!,
                 ctx.location.child(['examples', key]),
                 ctx,
-                allowAdditionalProperties
+                !!opts.allowAdditionalProperties
               );
             }
           }
--- a/packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
+++ b/packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
@@ -11,13 +11,10 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
   return {
     MediaType: {
       skip(mediaType) {
-        return !isDefined(mediaType.schema);
+        return mediaType.schema === undefined;
       },
       leave(mediaType, ctx: UserContext) {
         const { location, resolve } = ctx;
-        const allowAdditionalProperties = isDefined(opts.allowAdditionalProperties)
-          ? opts.allowAdditionalProperties
-          : false;
 
         if (isDefined(mediaType.example)) {
           resolveAndValidateExample(mediaType.example, location.child('example'));
@@ -50,7 +47,7 @@ export const ValidContentExamples: Oas3Rule = (opts) => {
             mediaType.schema!,
             location,
             ctx,
-            allowAdditionalProperties
+            !!opts.allowAdditionalProperties
           );
         }
       },
--- a/packages/core/src/rules/common/no-invalid-schema-examples.ts
+++ b/packages/core/src/rules/common/no-invalid-schema-examples.ts
@@ -10,9 +10,6 @@ export const NoInvalidSchemaExamples: Oas3Rule | Oas2Rule = (opts: any) => {
     Schema: {
       leave(schema: Oas3_1Schema | Oas3Schema, ctx: UserContext) {
         const examples = (schema as Oas3_1Schema).examples;
-        const allowAdditionalProperties = isDefined(opts.allowAdditionalProperties)
-          ? opts.allowAdditionalProperties
-          : false;
 
         if (examples) {
           for (const example of examples) {
@@ -21,7 +18,7 @@ export const NoInvalidSchemaExamples: Oas3Rule | Oas2Rule = (opts: any) => {
               schema,
               ctx.location.child(['examples', examples.indexOf(example)]),
               ctx,
-              allowAdditionalProperties
+              !!opts.allowAdditionalProperties
             );
           }
         }
@@ -41,7 +38,7 @@ export const NoInvalidSchemaExamples: Oas3Rule | Oas2Rule = (opts: any) => {
             schema,
             ctx.location.child('example'),
             ctx,
-            allowAdditionalProperties
+            !!opts.allowAdditionalProperties
           );
         }
       },


.changeset/shaggy-jokes-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fix example validation for `additionalProperties` in invalid-example rules.
```
JLekawa:
```suggestion
Fixed an issue where the `no-invalid-media-type-examples`, `no-invalid-parameter-examples`, and `no-invalid-schema-examples` would not trigger warnings when an example defined in a schema.
```

packages/core/src/rules/common/__tests__/no-invalid-schema-examples.test.ts
```diff
@@ -84,4 +84,115 @@ describe('no-invalid-schema-examples', () => {
...
+          "message": "Example value must conform to the schema: must NOT have unevaluated properties \`extraProperty\`.",
+          "ruleId": "no-invalid-schema-examples",
+          "severity": "error",
+          "suggest": [],
+        },
+      ]
+    `);
+  });
+
+  it('should report on invalid examples with additional properties', async () => {
```
tatomyr, line 88: No need to cover both example/examples in each test. Let's keep tests simple and minimalistic.

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -9,10 +10,16 @@ import type { UserContext } from '../../walk.js';
...
+      skip(mediaType) {
+        return !isDefined(mediaType.schema);
+      },
       leave(mediaType, ctx: UserContext) {
         const { location, resolve } = ctx;
-        if (!mediaType.schema) return;
-        if (mediaType.example !== undefined) {
+        const allowAdditionalProperties = isDefined(opts.allowAdditionalProperties)
+          ? opts.allowAdditionalProperties
+          : true;
```
tatomyr: Why added the `true` fallback? The intended default must be  `false`, as it currently is.

tests/e2e/lint/no-invalid-examples-allow-additional-properties-in-config/redocly.yaml
```diff
@@ -0,0 +1,14 @@
...
+  test:
+    root: ./openapi.yaml
+
+rules:
+  no-invalid-schema-examples:
+    severity: error
+    allowAdditionalProperties: true
+  no-invalid-parameter-examples:
+    severity: error
+    allowAdditionalProperties: false
```
tatomyr: Let's keep one default for testing purposes.

tatomyr: It should still be highlighted.

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
tatomyr: Same for other rules.

tatomyr (changes requested):

packages/core/src/rules/common/__tests__/no-invalid-schema-examples.test.ts
vadyvas: removed the duplicate

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
vadyvas: fixed 👍

tests/e2e/lint/no-invalid-examples-allow-additional-properties-in-config/redocly.yaml
vadyvas: updated

packages/core/src/rules/common/no-invalid-parameter-examples.ts
```diff
@@ -7,13 +8,17 @@ export const NoInvalidParameterExamples: any = (opts: any) => {
...
+          : false;
+
+        if (isDefined(parameter.example)) {
           validateExample(
             parameter.example,
             parameter.schema!,
             ctx.location.child('example'),
             ctx,
-            !!opts.allowAdditionalProperties
+            allowAdditionalProperties
```
tatomyr:
```suggestion
            !!opts.allowAdditionalProperties
```

Isn't it the same?

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
```diff
@@ -9,10 +10,16 @@ import type { UserContext } from '../../walk.js';
 export const ValidContentExamples: Oas3Rule = (opts) => {
   return {
     MediaType: {
+      skip(mediaType) {
+        return !isDefined(mediaType.schema);
```
tatomyr:
```suggestion
        return mediaType.schema === 'undefined'
```

 Not need to use additional wrapper. And it's better to avoid negations.

Anyway, why it's needed? Looks like it was working without it before.

packages/core/src/rules/common/no-invalid-parameter-examples.ts
tatomyr: Same in other rules.

tests/e2e/lint/no-invalid-schema-examples/openapi.yaml
```diff
@@ -66,7 +66,7 @@ components:
     Dog:
       type: object
       example:
-        status: 400
+        my_list: test
```
tatomyr, line 69: Why this has been changed?

tests/smoke/basic/pre-built/redoc.html
```diff
@@ -12,284 +12,13 @@
...
-data-styled.g135[id="sc-elAWhK"]{content:"dVAafl,"}/*!sc*/
-.kPfnYC{padding:5px 0;}/*!sc*/
-data-styled.g136[id="sc-kiPvrT"]{content:"kPfnYC,"}/*!sc*/
-.kMyMjD{width:calc(100% - 40px);box-sizing:border-box;margin:0 20px;padding:5px 10px 5px 20px;border:0;border-bottom:1px solid #e1e1e1;font-family:Roboto,sans-serif;font-weight:bold;font-size:13px;color:#333333;background-color:transparent;outline:none;}/*!sc*/
-data-styled.g137[id="sc-cLNono"]{content:"kMyMjD,"}/*!sc*/
-.cDWGFa{position:absolute;left:20px;height:1.8em;width:0.9em;}/*!sc*/
-.cDWGFa path{fill:#333333;}/*!sc*/
-data-styled.g138[id="sc-iJbNxx"]{content:"cDWGFa,"}/*!sc*/
-</style>
+  <script src="https://cdn.redocly.com/redoc/v2.5.1/bundles/redoc.standalone.js"></script>
```
tatomyr, line 15: Is it still working as before?

vadyvas: seems yes

image

tests/e2e/lint/no-invalid-schema-examples/openapi.yaml
vadyvas: noticed during an `e2e` failure investigation that the examples are incorrect

tests/smoke/basic/pre-built/redoc.html
tatomyr: Are the buttons clickable? If yes, then most likely everything is ok.

tests/e2e/lint/no-invalid-schema-examples/openapi.yaml
tatomyr: Didn't they supposed to be incorrect for testing purposes?

tests/smoke/basic/pre-built/redoc.html
vadyvas: clickable, everything expands and collapses, search and download works, switching by response works

tests/e2e/lint/no-invalid-schema-examples/openapi.yaml
vadyvas: snapshot expects a successful execution, so I don’t think they were intended to be incorrect

https://github.com/Redocly/redocly-cli/blob/main/tests/e2e/lint/no-invalid-schema-examples/snapshot.txt

packages/core/src/rules/oas3/no-invalid-media-type-examples.ts
AlbinaBlazhko17: I just want to reuse our utility function, but anyway we can use just comparison to undefined.
I wanted to refactor this flow and it is better to use skip instead of these if statements inside leave function:
```
if (!mediaType.schema) return;
        if (mediaType.example !== undefined) {
```

vadyvas: @tatomyr left `skip`, but removed `isDefined`

packages/core/src/rules/common/no-invalid-parameter-examples.ts
vadyvas: returned `!!opts.allowAdditionalProperties` usage

PR #2437, opened 2025-12-03
fix(project): add empty path check, throw error on unresolved path

aim4ik11:
## What/Why/How?
Currently extends property parser in config accepts not existing refs which causes builder to attempt to resolve those and throw hard error which could be hard to debug for user. I suggest adding more descriptive error message so that builder displays it

## Reference
[Issue](https://github.com/orgs/Redocly/projects/81/views/1?pane=issue&itemId=141017093&issue=Redocly%7Credocly%7C19412)

## Check yourself

## Security


packages/core/src/config/bundle-extends.ts
```diff
@@ -21,16 +21,28 @@ export function bundleExtends({
   }
 
   const resolvedExtends = (node.extends || [])
-    .map((presetItem: string) => {
+    .map((presetItem, index) => {
+      if (typeof presetItem !== 'string' || !presetItem.trim()) {
```
RomanHotsiy: it's missing file path of where it happens

packages/core/src/config/bundle-extends.ts
```diff
@@ -21,16 +21,46 @@ export function bundleExtends({
   }
 
   const resolvedExtends = (node.extends || [])
-    .map((presetItem: string) => {
+    .map((presetItem, index) => {
```
AlbinaBlazhko17: Let's investigate if we can use `reportUnresolvedRef` in some place to resolve this issue.
https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/common/no-unresolved-refs.ts

packages/core/src/rules/config/config-no-unresolved-refs.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function ConfigNoUnresolvedRefs() {
+  const base = NoUnresolvedRefs({});
+
+  function validateConfigExtends(node: unknown, ctx: UserContext) {
+    if (!isPlainObject(node)) return;
+    const exts = (node as any).extends;
+    if (!Array.isArray(exts)) return;
+
+    exts.forEach((item, index) => {
```
RomanHotsiy: let's use regular for or for of

packages/core/src/types/redocly-yaml.ts
```diff
@@ -204,20 +202,23 @@ const configGovernanceProperties: Record<
...
-      if (typeof node === 'string' && !isAbsoluteUrl(node) && !path.extname(node)) {
-        return { type: 'string' };
-      }
-      return {
-        ...ConfigGovernance,
-        directResolveAs: { name: 'ConfigGovernance', ...ConfigGovernance },
-      } as PropType;
-    },
-  } as PropType,
+  // extends: {
```
RomanHotsiy: why is this commented out?

packages/core/src/types/redocly-yaml.ts
```diff
@@ -248,12 +247,39 @@ const configGovernanceProperties: Record<
...
   arazzo1Decorators: { type: 'object' },
   overlay1Decorators: { type: 'object' },
+  openrpc1Decorators: { type: 'object' },
 };
 
 const ConfigGovernance: NodeType = {
   properties: configGovernanceProperties,
 };
 
+const ConfigGovernanceList: NodeType = {
```
tatomyr, line 273: FYI, this name will likely be changed in https://github.com/Redocly/redocly-cli/pull/2488.

Issue #2436, opened 2025-12-02
path-parameters-defined rule false positive: validates sibling operation's path parameters against callback URL runtime expressions


ZyzychO:
**Describe the bug**

The path-parameters-defined rule incorrectly produces an error when:
A path has multiple operations (e.g., post and patch)
One operation defines a callbacks section with a runtime expression URL (e.g., '{$request.body#/callbackUrl}')
A sibling operation defines its path parameters at the operation level
The rule incorrectly validates the sibling operation's path parameters against the callback URL from the other operation, treating the runtime expression as if it were the operation's path.

**To Reproduce**

1. Given this `openapi.yaml` file

```yaml
openapi: 3.1.1
info:
  title: Bug Reproduction - path-parameters-defined false positive
  version: 1.0.0

paths:
  /projects/{projectId}:
    post:
      operationId: createProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      callbacks:
        onEvent:
          '{$request.body#/callbackUrl}':
            post:
              summary: Callback endpoint
              responses:
                '200':
                  description: OK
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                callbackUrl:
                  type: string
      responses:
        '201':
          description: Created

    patch:
      operationId: updateProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
```
2. Run redocly lint openapi.yaml
4. Observe the error on the patch operation

**Expected behavior**

No error should be reported. The projectId parameter in the patch operation is correctly used in the path /projects/{projectId}.

**Redocly version(s)**

2.12.1

DmitryAnansky:
NOTE: Originally reported in `@redocly/cli`  v`1.34.2`.

PR #2438, opened 2025-12-03
fix: path-parameters-defined rule false positive

DmitryAnansky:
## What/Why/How?
NOTE: SIMILAR FIX SHOULD BE APPLIED TO OLDER REDOCLY/CLI VERSION.

Added additional `path-parameters-defined` rule context during validation to track parameters for Path and Path that comes from `callback`. 

Having this part of description as an example:
```yaml
paths:
  /projects/{projectId}:
    post:
      operationId: createProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      callbacks:
        onEvent:
          '{$request.body#/callbackUrl}':
            post:
              responses:
                '200':
                  description: OK
                '400':
                  description: Bad Request
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                callbackUrl:
                  type: string
      responses:
        '201':
          description: Created

    patch:
      operationId: updateProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
```

Both `/projects/{projectId}` and `{$request.body#/callbackUrl}` validated by this rule as a separate `PathItem`.
The issue was that without context tracking we were handling `patch` operation as if it a part of `{$request.body#/callbackUrl}` path. 
The rule logic is more complicated now, but we are able to restore correct `path` and context for `patch` operation.

## Alternative solution
This time instead of having additional context and stack we are going to narrow down visitor access, handling separately:

Paths -> PathItem + ignoring PathItem for callbacks
Callback -> PathItem

https://github.com/Redocly/redocly-cli/pull/2442

## Reference
https://github.com/Redocly/redocly-cli/issues/2436

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/common/path-params-defined.ts
+++ b/packages/core/src/rules/common/path-params-defined.ts
@@ -1,24 +1,121 @@
 import type { Oas3Rule, Oas2Rule } from '../../visitors.js';
-import type { Oas2Parameter } from '../../typings/swagger.js';
-import type { Oas3Parameter } from '../../typings/openapi.js';
+import type { Oas2Operation, Oas2Parameter } from '../../typings/swagger.js';
+import type { Oas3Operation, Oas3Parameter } from '../../typings/openapi.js';
 import type { UserContext } from '../../walk.js';
 
 const pathRegex = /\{([a-zA-Z0-9_.-]+)\}+/g;
 
+type PathContext = {
+  path: string;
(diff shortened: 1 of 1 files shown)


DmitryAnansky:
Alternative solution => https://github.com/Redocly/redocly-cli/pull/2442

PR #2442, opened 2025-12-04
fix: path-parameters-defined rule false positive

DmitryAnansky:
## What/Why/How?
NOTE: SIMILAR FIX SHOULD BE APPLIED TO OLDER REDOCLY/CLI VERSION.

Added additional `path-parameters-defined` rule context during validation to track parameters for Path and Path that comes from `callback`. 

Having this part of description as an example:
```yaml
paths:
  /projects/{projectId}:
    post:
      operationId: createProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      callbacks:
        onEvent:
          '{$request.body#/callbackUrl}':
            post:
              responses:
                '200':
                  description: OK
                '400':
                  description: Bad Request
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                callbackUrl:
                  type: string
      responses:
        '201':
          description: Created

    patch:
      operationId: updateProject
      parameters:
        - name: projectId
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: OK
```

Both `/projects/{projectId}` and `{$request.body#/callbackUrl}` validated by this rule as a separate `PathItem`.
The issue was that without context tracking we were handling `patch` operation as if it a part of `{$request.body#/callbackUrl}` path. 

This time instead of having additional context and stack we are going to narrow down visitor access, handling separately `Callback` inside `PathItem`.

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/2436

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/common/path-params-defined.ts
+++ b/packages/core/src/rules/common/path-params-defined.ts
@@ -128,9 +128,6 @@ const createOperationHandlers = (
     },
     Parameter(parameter: Oas2Parameter | Oas3Parameter, { report, location }: UserContext) {
       if (parameter.in === 'path' && parameter.name && pathContext.current) {
-        if (!currentOperationParams) {
-          currentOperationParams = new Set();
-        }
         currentOperationParams.add(parameter.name);
         validatePathParameter(
           parameter.name,


packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -1,181 +1,197 @@
...
-
-    return {
-      path: contextToUse.path,
-      templateParams: new Set(contextToUse.templateParams),
-      definedParams: new Set(definedPathParams),
-      definedOperationParams: new Set(),
-    };
-  };
+  let currentPath: string;
+  let pathTemplateParams: Set<string>;
```
AlbinaBlazhko17: Is it possible to remove these variables completely and keep them inside the `rootPathContext`?

AlbinaBlazhko17:
Have you tested this implementation with the different scenarios? e.g. adding unused parameters in the `patch` and in the callback operation?

DmitryAnansky:
> Have you tested this implementation with the different scenarios? e.g. adding unused parameters in the `patch` and in the callback operation?

Yes, its running the test from this branch https://github.com/Redocly/redocly-cli/pull/2438

packages/core/src/rules/common/path-params-defined.ts
DmitryAnansky: I still need to track some state between different pathItems.
Changed it a bit, please take a look.

.changeset/rotten-news-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed false positive reports for path parameters in callback operations by the `path-parameters-defined` rule.
```
JLekawa:
```suggestion
Fixed false positive reports for path parameters in callback operations reported by the `path-parameters-defined` rule.
```

JLekawa (approved):

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,232 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
-          if (parameter.in === 'path' && parameter.name) {
-            definedOperationParams.add(parameter.name);
-            if (!pathTemplateParams.has(parameter.name)) {
-              report({
-                message: `Path parameter \`${parameter.name}\` is not used in the path \`${currentPath}\`.`,
-                location: location.child(['name']),
-              });
-            }
-          }
+        leave: () => {},
```
AlbinaBlazhko17: Why we need here empty visitors? Just to match the type?

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,232 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
+        Parameter: () => {},
+        Operation: {
+          enter: () => {},
+          leave: () => {},
+          Parameter: () => {},
+          Callback: {},
         },
       },
     },
+  } as unknown as OperationHandlers;
```
tatomyr: Please avoid double type casting.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,232 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
 import type { UserContext } from '../../walk.js';
 
 const pathRegex = /\{([a-zA-Z0-9_.-]+)\}+/g;
+const MAX_DEPTH = 4;
```
tatomyr: I don't believe there's a reason of going that deep. It's unlikely that anyone is using nested callbacks. I'd start from 1 (or from 2 if I'm getting it right) and then expand the depth if needed.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,232 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
+type PathContext = {
+  path: string;
+  templateParams: Set<string>;
+  definedParams: Set<string>;
+};
+
+type OperationHandlers = {
+  enter: () => void;
+  leave: (op: unknown, ctx: UserContext) => void;
+  Parameter: (parameter: Oas2Parameter | Oas3Parameter, ctx: UserContext) => void;
```
tatomyr: All those types are just corresponding visitors, why are you declaring them again?

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,232 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
+        pathContext.current.definedParams,
+        pathContext.current.path,
+        report,
+        location
+      );
+    },
+    Parameter(parameter: Oas2Parameter | Oas3Parameter, { report, location }: UserContext) {
+      if (parameter.in === 'path' && parameter.name && pathContext.current) {
+        if (!operationParams.current) {
+          operationParams.current = new Set();
```
tatomyr: Why not just set the initial `current` to an empty Set?

packages/core/src/rules/common/path-params-defined.ts
DmitryAnansky: In short - yes, but I tried to simplify it, please take a look.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,195 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
+        currentOperationParams,
+        pathContext.current.definedParams,
+        pathContext.current.path,
+        report,
+        location
+      );
+    },
+    Parameter(parameter: Oas2Parameter | Oas3Parameter, { report, location }: UserContext) {
+      if (parameter.in === 'path' && parameter.name && pathContext.current) {
+        if (!currentOperationParams) {
```
tatomyr: It seems it cannot have a falsy value as you're setting it it an empty Set above.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -4,62 +4,195 @@ import type { Oas3Parameter } from '../../typings/openapi.js';
...
+    });
+  };
+  if (depth >= MAX_DEPTH) {
+    return {
+      enter: () => {},
+      leave: (_op: unknown, { report, location }: UserContext) => {
+        reportMaxDepthWarning(report, location, depth);
+      },
+      Parameter: () => {},
+      Callback: undefined,
```
tatomyr, line 87: Have you tried using the `skip` visitor instead? Should be much cleaner.

DmitryAnansky: I want not only skip but also report warning message and stop building the nested visitor.

Issue #2457, opened 2025-12-10
$ref to parameters causes error for lint command


floyd-resimate:
**Describe the bug**

When running the newest version (2.12.5), I get the following error with no further description:
```bash
floyd@[redacted]/api_documentation$ npx @redocly/cli@2.12.5 lint
Need to install the following packages:
@redocly/cli@2.12.5
Ok to proceed? (y) 

validating openapi.yaml using lint rules for api 'resimate@0.1.0'...
Something went wrong when processing [redacted]/api_documentation/openapi.yaml:

  - op?.parameters?.forEach is not a function

```

When running with version 2.11.1, I don't have any issues:

```bash
floyd@[redacted]/api_documentation$ npx @redocly/cli@2.11.1 lint

    ╔═══════════════════════════════════════════════════════╗
    ║                                                       ║
    ║  A new version of Redocly CLI (2.12.5) is available.  ║
    ║  Update now: `npm i -g @redocly/cli@latest`.          ║
    ║  Changelog: https://redocly.com/docs/cli/changelog/   ║
    ║                                                       ║
    ╚═══════════════════════════════════════════════════════╝

validating openapi.yaml using lint rules for api 'resimate@0.1.0'...
openapi.yaml: validated in 200ms
```

Given the error message, the only addition I have is that we do have numerous endpoints with the `op` query parameter defined. Unfortunately, I can't determine which singular one, if any, is the offending one. Here's an example, which should be the same definition of all of them:

```yaml
# Parameters file tied in by $ref from paths file
# ...
- name: op
  in: query
  required: false
  schema:
    default: list
    enum:
    - count
    - list
    - countAndList
    type: string
  description: 'The operation to be conducted: either counting, listing, or both.'
# ...
```

**To Reproduce**

Ok, after some trials, here's the repdroduction I can provide. 

The correctly functioning in-file definition of the parameter, which I called `test_infile.yaml`
```yaml
openapi: 3.1.0
info:
  title: Test API
  description: Test documentation to reproduce bug
  version: 1.0.0
  license:
    name: n/a
    url: n/a
components:
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
security:
  - BearerAuth: []
servers:
  - url: https://api.test.com
    description: Test server
paths:
  /test-endpoint:
    get:
      operationId: testEndpoint
      tags:
        - Test
      summary: Test endpoint to reproduce bug
      parameters:
        - name: op
[...]

PR #2458, opened 2025-12-10
fix: parameters defined as a reference in the path-params-defined rule

DmitryAnansky:
## What/Why/How?

Fixed `path-params-defined` rule to correctly skip parameters defined via `$ref`.

## Reference
Closes https://github.com/Redocly/redocly-cli/issues/2457

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/common/path-params-defined.ts
+++ b/packages/core/src/rules/common/path-params-defined.ts
@@ -155,8 +155,7 @@ const collectPathParamsFromOperation = (
   targetSet: Set<string>
 ): void => {
   if (parameter && typeof parameter === 'object' && 'in' in parameter && 'name' in parameter) {
-    const p = parameter as { in?: string; name?: string };
-    if (p.in === 'path' && p.name) {
+    if (parameter.in === 'path' && parameter.name) {
       targetSet.add(parameter.name);
     }
   }


.changeset/evil-states-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed `path-params-defined` rule to correctly skip parameters defined via `$ref`.
```
JLekawa:
```suggestion
Fixed `path-params-defined` rule to correctly skip parameters defined through `$ref`.
```

JLekawa (approved):

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -151,12 +151,19 @@ const extractTemplateParams = (path: string): Set<string> => {
 };
 
 const collectPathParamsFromOperation = (operation: unknown, targetSet: Set<string>): void => {
```
AlbinaBlazhko17: Let's create at the top of the file general type of the operation, then you can cast to this type `_op` on the 122 line and on the current line. In this case you can remove casting to the `unknown` type on 154 line, remove if statement on the 159 line and remove casting on the 160 line.

```
type Operation = {
  parameters?: Array<Oas2Parameter | Oas3Parameter> | Oas2Parameter | Oas3Parameter;
};
```

DmitryAnansky: this does not cover ref scenario.

AlbinaBlazhko17: Maybe in that way?

```
type Operation = Oas2Operation | Oas3Operation<Oas3Schema | Oas3_1Schema>;

leave(op: Operation, { report, location }: UserContext) 

const collectPathParamsFromOperation = (operation: Operation, targetSet: Set<string>): void => {
  const params = operation?.parameters;

  if (Array.isArray(params)) {
    for (const param of params) {
      if ('$ref' in param) continue; // Skip OasRef types
      if (param.in === 'path' && param.name) {
        targetSet.add(param.name);
      }
    }
  }
};
```

DmitryAnansky: Talked with @tatomyr and changed / simplify parameters collection.
Now we collect only resolved parameters.
Please take a look.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -150,13 +150,16 @@ const extractTemplateParams = (path: string): Set<string> => {
...
-  const op = operation as { parameters?: Array<{ in?: string; name?: string }> };
-  op?.parameters?.forEach((param) => {
-    if (param?.in === 'path' && param?.name) {
-      targetSet.add(param.name);
+const collectPathParamsFromOperation = (
+  parameter: Oas2Parameter | Oas3Parameter,
+  targetSet: Set<string>
+): void => {
+  if (parameter && typeof parameter === 'object' && 'in' in parameter && 'name' in parameter) {
+    const p = parameter as { in?: string; name?: string };
```
AlbinaBlazhko17: You can remove this line and use just `parameter`.

DmitryAnansky: Thx, missed this one

PR #2461, opened 2025-12-12
fix: added encoding on absolutePointer before decodeURIComponent

DmitryAnansky:
## What/Why/How?

Fixed an issue where JSON Pointers containing special characters (like `%`) were not properly URI-encoded when used as URI identifiers, causing validation errors with properties containing percent signs or other special characters.

Additionally improved performance benchmarking by unifying the pipelines, making the run faster by linting only one API, and adding more versions to compare.

## Reference

Related change => https://github.com/Redocly/redocly-cli/pull/170

## Testing

Tested with Redoc/Realm/Reunite   => https://github.com/Redocly/redocly/pull/19833

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/ajv.ts
+++ b/packages/core/src/rules/ajv.ts
@@ -1,7 +1,6 @@
 import addFormats from 'ajv-formats';
 import Ajv from '@redocly/ajv/dist/2020.js';
 import { escapePointerFragment } from '../ref-utils.js';
-import { slash } from '../utils/slash.js';
 
 import type { Location } from '../ref-utils.js';
 import type { ValidateFunction, ErrorObject } from '@redocly/ajv/dist/2020.js';
@@ -27,10 +26,14 @@ function getAjv(resolve: ResolveFn, allowAdditionalProperties: boolean) {
       validateFormats: true,
       defaultUnevaluatedProperties: allowAdditionalProperties,
       loadSchemaSync(base: string, $ref: string, $id: string) {
-        const resolvedRef = resolve({ $ref }, slash(base.split('#')[0]));
+        const decodedBase = decodeURI(base.split('#')[0]);
+        const resolvedRef = resolve({ $ref }, decodedBase);
         if (!resolvedRef || !resolvedRef.location) return false;
-        const normalizedAbsoluteRef = slash(resolvedRef.location.source.absoluteRef);
-        return { $id: normalizedAbsoluteRef + '#' + $id, ...resolvedRef.node };
+
+        return {
+          $id: encodeURI(resolvedRef.location.source.absoluteRef) + '#' + $id,
+          ...resolvedRef.node,
+        };
       },
       logger: false,
     });
@@ -46,7 +49,7 @@ function getAjvValidator(
   allowAdditionalProperties: boolean
 ): ValidateFunction | undefined {
   const ajv = getAjv(resolve, allowAdditionalProperties);
-  const $id = encodeURI(slash(loc.absolutePointer));
+  const $id = encodeURI(loc.absolutePointer);
 
   if (!ajv.getSchema($id)) {
     ajv.addSchema({ $id, ...schema }, $id);


.changeset/solid-lands-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed an issue where JSON Pointers containing special characters (like `%`) were not properly URI-encoded when used as URI identifiers, causing validation errors with properties containing percent signs or other special characters.
```
JLekawa:
```suggestion
Fixed an issue where JSON Pointers containing special characters (like `%`) were not properly URI-encoded.
When these pointers were used as URI identifiers, they caused validation errors with properties containing percent signs or other special characters.
```

packages/core/src/format/codeframes.ts
```diff
@@ -181,7 +176,7 @@ function positionsToLoc(
 }
 
 export function getAstNodeByPointer(root: YAMLNode, pointer: string, reportOnKey: boolean) {
-  const pointerSegments = parsePointer(pointer);
+  const [, ...pointerSegments] = parsePointer(pointer);
```
kanoru3101: Can there be cases where the first element is not `#`?

RomanHotsiy: This may impact perf a bit

packages/core/src/ref-utils.ts
```diff
@@ -36,12 +36,16 @@ export class Location {
...
+    return this.source.absoluteRef + (this.pointer === '#/' ? '' : encodeURI(this.pointer));
   }
 }
 
 export function unescapePointer(fragment: string): string {
-  return decodeURIComponent(fragment.replace(/~1/g, '/').replace(/~0/g, '~'));
+  return fragment.replace(/~1/g, '/').replace(/~0/g, '~');
+}
+
+export function unescapePointerFromURI(fragment: string): string {
```
RomanHotsiy:
```suggestion
export function unescapePointerFromURIFragment(fragment: string): string {
```

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Oas3 spec-components-invalid-map-name', () => {
         {
           "location": [
             {
-              "pointer": "#/components/parameters/my Param",
+              "pointer": "#/components/parameters/my%20Param",
```
RomanHotsiy: we should ensure the old way doesn't break (there may be people using it )

packages/core/src/ref-utils.ts
```diff
@@ -41,12 +41,23 @@ export class Location {
 }
 
 export function unescapePointer(fragment: string): string {
-  return decodeURIComponent(fragment.replace(/~1/g, '/').replace(/~0/g, '~'));
+  return decodeURIComponent(fragment).replaceAll('~1', '/').replaceAll('~0', '~');
 }
 
 export function escapePointer<T extends string | number>(fragment: T): T {
```
RomanHotsiy: let's rename those to escape/unescapePointerFragment or it's misleading otherwise

packages/core/src/__tests__/ref-utils.test.ts
```diff
@@ -139,4 +139,33 @@ describe('ref-utils', () => {
       expect(refBaseName('abcdefg')).toStrictEqual('abcdefg');
     });
   });
+
+  describe('escapePointer', () => {
+    it('should escape a pointer correctly', () => {
```
RomanHotsiy: Let's update everywhere
```suggestion
    it('should escape a pointer fragment correctly', () => {
```

packages/core/src/__tests__/ref-utils.test.ts
```diff
@@ -35,7 +35,7 @@ describe('ref-utils', () => {
   });
 
   it(`should unescape complex urlencoded paths`, () => {
-    const referene = 'somefile.yaml#/components/schemas/scope%2Fcomplex~name';
+    const referene = 'somefile.yaml#/components/schemas/scope~1complex~0name';
```
RomanHotsiy: we need to ensure the old format also works (people may be using it still)

DmitryAnansky: But it was encoded wrong previously.

tatomyr: @DmitryAnansky  It will still work anyway 🙂.

packages/core/src/rules/oas3/__tests__/spec-components-invalid-map-name.test.ts
tatomyr: Added a test for that: https://github.com/Redocly/redocly-cli/pull/2461/files#diff-6c16fd995e69873e49edd979fdd4e1c9768661b4dfc71cd03f2d59d40f549005R57

packages/core/src/__tests__/ref-utils.test.ts
tatomyr: Added the old test back.

packages/core/src/ref-utils.ts
```diff
@@ -40,13 +40,22 @@ export class Location {
...
 
-export function escapePointer<T extends string | number>(fragment: T): T {
+export function escapePointerFragment<T extends string | number>(fragment: T): T {
   if (typeof fragment === 'number') return fragment;
-  return (fragment as string).replace(/~/g, '~0').replace(/\//g, '~1') as T;
+  let result: string = fragment;
+  // add only % escape if % is present
+  if (result.includes('%')) {
+    result = result.replaceAll('%', '%25');
+  }
```
RomanHotsiy: we don't need this, we encode here not for URL, if someone needs to use this in the URL later they can encode

DmitryAnansky: We need this to show better error location for the example.
```
Example value must conform to the schema: type must be number.

267 |           activity_level_%:
268 |             type: number
269 |             example: ваіва
270 |
271 | TicketType:
```

instead of 

```
Example validation errored: URI malformed.

266 |             example: 32
267 |           activity_level_%:
268 |             type: number
269 |             example: ваіва
270 |
271 | TicketType:
```

PR #2462, opened 2025-12-15
feat: basic support for openrpc spec

RomanHotsiy:
## What/Why/How?

Added support for Open-RPC spec: https://spec.open-rpc.org/

## Reference
JSON-RPC is used in [A2A Protocol](https://a2a-protocol.org/latest/) so it may gain popularity.

## Testing
Added unit and e2e test.

## Check yourself

## Security


RomanHotsiy:
No idea why the smoke tests on node 22 yarn are failing. cc @tatomyr

tatomyr:
> the smoke tests on node 22 yarn are failing

@RomanHotsiy sometimes yarn fails to resolve ajv-formats (haven't figured why). Should pass after re-running smokes.

packages/core/src/config/minimal.ts
```diff
@@ -277,6 +277,13 @@ const minimal: RawGovernanceConfig<'built-in'> = {
   overlay1Rules: {
     'info-contact': 'off',
   },
+  openrpc1Rules: {
```
tatomyr, line 280: Please also add to `config/all.ts`.

RomanHotsiy: Done

Issue #1634, opened 2024-07-16
root level paths `/` are not split correctly


LukeHagar:
**Describe the bug**

When running the `redocly split` command the openapi path at the API root: `/` is not properly moved to the paths folder, and the references in the document are not properly converted to file based references.

**To Reproduce**

Minimum repro repo:
https://github.com/LukeHagar/redocly-split-repro

1. Open the repo
2. Run `make repro`
3. See the file here:
https://github.com/LukeHagar/redocly-split-repro/blob/main/openapi/paths.yaml

**Expected behavior**

I would expect this path to be handled like all the others

**Redocly Version(s)**

1.18.0

**`Node.js` Version(s)**

v20.15.0

**Additional context**

You rock! I love the `bundle` and `split` implementations

tatomyr:
I confirm this is a bug. The issue is that the tool attempts to create a file named after the path item, and since it's empty (the slash doesn't count here), it fails to resolve it. The solution I propose is to give a default name for the root path. @LukeHagar, what name would you expect here?

NB: A temporary workaround would be to rename the `/` path item in your OpenAPI file to an arbitrary unique value (e.g., `/root`), then split it, and then rename it back in the generated files (leaving the file name intact).

PR #1638, opened 2024-07-18
fix: enable path parsing to account for root level paths

LukeHagar:
## What/Why/How?

Fixes #1634

## Check yourself

- [ ] Code changed? - Tested with redoc/reference-docs/workflows (internal)
- [ ] All new/updated code is covered with tests
- [ ] New package installed? - Tested in different environments (browser/node)

## Security

- [ ] Security impact of change has been considered
- [ ] Code follows company security practices and guidelines

PR #2485, opened 2025-12-31
fix!: split root-level paths correctly

tatomyr:
## What/Why/How?

Fixed the `split` command to properly handle root-level paths.
Previously, the root path `/` was converted to an empty string as a filename, leading to incorrect file structure and broken links.
Now, it correctly maps to the specified path separator.

## Reference

Opened as a replacement for https://github.com/Redocly/redocly-cli/pull/1638

Fixes https://github.com/Redocly/redocly-cli/issues/1634

## Check yourself

## Security

### DIFF
--- a/docs/@v2/commands/split.md
+++ b/docs/@v2/commands/split.md
@@ -24,15 +24,15 @@ redocly split --version
 
 ## Options
 
-| Option        | Type    | Description                                                                                                                                                                              |
-| ------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| api           | string  | **REQUIRED.** Path to the API description file that you want to split into a multi-file structure.                                                                                       |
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/miscellaneous.ts
```diff
@@ -121,11 +121,15 @@ export function printExecutionTime(commandName: string, startedAt: number, api:
...
+  }
+
   return path
-    .replace(/~1/g, '/')
-    .replace(/~0/g, '~')
+    .replaceAll('~1', '/')
+    .replaceAll('~0', '~')
     .replace(/^\//, '')
-    .replace(/\//g, pathSeparator);
+    .replaceAll('/', pathSeparator);
```
tatomyr, line 132: No behaviour changes here, simply improving readability.

PR #2489, opened 2026-01-07
fix(core): fix bug loading of `.redocly.lint-ignore.yaml` in browser environments

kanoru3101:
## What/Why/How?
Fixed loading of `.redocly.lint-ignore.yaml` in browser environments (Reunite).
## Reference
[#86](https://github.com/Redocly/redocly-vs-code/issues/86)
[Related PR](https://github.com/Redocly/redocly/pull/19923)
## Check yourself

## Security

### DIFF
--- a/packages/core/src/config/config.ts
+++ b/packages/core/src/config/config.ts
@@ -31,6 +31,8 @@ import type {
   ResolvedConfig,
   RuleConfig,
   RuleSettings,
+  IgnoreFile,
+  ResolvedIgnore,
 } from './types.js';
 
 export class Config {
@@ -42,7 +44,7 @@ export class Config {
   _alias?: string;
 
   plugins: Plugin[];
-  ignore: Record<string, Record<string, Set<string>>> = {};
+  ignore: ResolvedIgnore = {};
   doNotResolveExamples: boolean;
   rules: Record<SpecVersion, Record<string, RuleConfig>>;
   preprocessors: Record<SpecVersion, Record<string, PreprocessorConfig>>;
@@ -59,8 +61,8 @@ export class Config {
       resolvedRefMap?: ResolvedRefMap;
       alias?: string;
       plugins?: Plugin[];
-      ignoreFile?: { content: Record<string, Record<string, string[]>>; dir: string };
-      ignore?: Record<string, Record<string, Set<string>>>;
+      ignoreFile?: IgnoreFile;
+      ignore?: ResolvedIgnore;
     } = {}
   ) {
     this.resolvedConfig = resolvedConfig;
@@ -146,21 +148,15 @@ export class Config {
     this.ignore = opts.ignore ?? (opts.ignoreFile ? this.resolveIgnore(opts.ignoreFile) : {});
   }
 
-  private resolveIgnore({
-    content,
-    dir,
-  }: {
-    content: Record<string, Record<string, string[]>>;
-    dir: string;
-  }): Record<string, Record<string, Set<string>>> {
-    const ignore = Object.create(null) as Record<string, Record<string, Set<string>>>;
+  private resolveIgnore({ content, dir }: IgnoreFile): ResolvedIgnore {
+    const ignore: ResolvedIgnore = Object.create(null);
 
     for (const fileName of Object.keys(content)) {
       const fileIgnore = content[fileName];
 
       const resolvedFileName = isAbsoluteUrl(fileName) ? fileName : resolvePath(dir, fileName);
 
-      ignore[resolvedFileName] = Object.create(null) as Record<string, Set<string>>;
+      ignore[resolvedFileName] = Object.create(null);
 
       for (const ruleId of Object.keys(fileIgnore)) {
         ignore[resolvedFileName][ruleId] = new Set(fileIgnore[ruleId]);
--- a/packages/core/src/config/load.ts
+++ b/packages/core/src/config/load.ts
@@ -12,26 +12,30 @@ import { CONFIG_FILE_NAME, IGNORE_FILE } from './constants.js';
 import { isAbsoluteUrl, getDir, resolvePath } from '../ref-utils.js';
 import { isBrowser } from '../env.js';
 
-import type { RawUniversalConfig } from './types.js';
+import type { RawUniversalConfig, IgnoreFile } from './types.js';
 
 export async function loadIgnoreFile(
   configPath: string | undefined,
   resolver: BaseResolver
-): Promise<{ content: Record<string, Record<string, string[]>>; dir: string } | undefined> {
+): Promise<IgnoreFile | undefined> {
   const configDir = configPath ? getDir(configPath) : isBrowser ? '' : process.cwd();
   const ignorePath = configDir ? resolvePath(configDir, IGNORE_FILE) : IGNORE_FILE;
   if (fs?.existsSync && !isAbsoluteUrl(ignorePath) && !fs.existsSync(ignorePath)) {
     return undefined;
   }
 
-  const ignoreDocument = await resolver.resolveDocument(null, ignorePath, true);
+  const ignoreDocument = await resolver.resolveDocument<IgnoreFile['content']>(
+    null,
+    ignorePath,
+    true
+  );
 
   if (ignoreDocument instanceof Error || !ignoreDocument.parsed) {
     return undefined;
   }
 
   return {
-    content: (ignoreDocument.parsed || {}) as Record<string, Record<string, string[]>>,
+    content: ignoreDocument.parsed || {},
     dir: configDir,
   };
 }
@@ -85,7 +89,7 @@ type CreateConfigOptions = {
   configPath?: string;
   externalRefResolver?: BaseResolver;
   resolvedRefMap?: ResolvedRefMap;
-  ignoreFile?: { content: Record<string, Record<string, string[]>>; dir: string };
+  ignoreFile?: IgnoreFile;
 };
 
 export async function createConfig(
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -263,3 +263,6 @@ export type ResolvedConfig = Omit<RawUniversalConfig, 'apis' | 'plugins'> &
     apis?: Record<string, ResolvedApiConfig>;
     plugins?: string[];
   };
+
+export type ResolvedIgnore = Record<string, Record<string, Set<string>>>;
+export type IgnoreFile = { content: Record<string, Record<string, string[]>>; dir: string };


packages/core/src/config/__tests__/fixtures/ignore-file/redocly.yaml
```diff
@@ -0,0 +1,2 @@
+rules:
+  operation-operationId: error
```
tatomyr, line 2: Could you add yet another rule (but don't ignore it) so we can be sure it ignores only what's supposed to be ignored?

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,69 @@ import {
   type Document,
   type ResolvedRefMap,
 } from '../resolve.js';
-import { CONFIG_FILE_NAME } from './constants.js';
+import { CONFIG_FILE_NAME, IGNORE_FILE } from './constants.js';
 
 import type { RawUniversalConfig } from './types.js';
 
+function isUrl(ref: string): boolean {
```
tatomyr: I believe we have something similar to this (isAbsolutePath? Maybe it needs to be adjusted.) 
 Could you check the codebase?

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,69 @@ import {
...
+
+  const ignore = (ignoreDocument.parsed || {}) as Record<string, Record<string, Set<string>>>;
+
+  for (const fileName of Object.keys(ignore)) {
+    const resolvedFileName = isUrl(fileName) ? fileName : resolvePath(configDir, fileName);
+
+    ignore[resolvedFileName] = ignore[fileName];
+
+    for (const ruleId of Object.keys(ignore[fileName])) {
+      ignore[fileName][ruleId] = new Set(ignore[fileName][ruleId]);
```
tatomyr: Why do we need this mapper? Cannot we just use whatever we've resolved?

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,69 @@ import {
...
+
+function getConfigDir(configPath: string): string {
+  if (!path.extname(configPath)) {
+    return configPath;
+  }
+
+  return isUrl(configPath)
+    ? configPath.substring(0, configPath.lastIndexOf('/'))
+    : path.dirname(configPath);
+}
```
tatomyr: It's just a general function. Could you rename it to simply `getDir` and move to utils?

packages/core/src/config/load.ts
kanoru3101: `isAbsoluteUrl` is used in many places. I created another function based on it, so as not to break anything. What do you think about this? Or is it better to edit `isAbsoluteUrl`?

packages/core/src/config/load.ts
kanoru3101: moved to ref-utils

packages/core/src/config/__tests__/fixtures/ignore-file/redocly.yaml
kanoru3101: Tested on Reunite and in VSCE. It works

packages/core/src/config/load.ts
```diff
@@ -79,11 +130,16 @@ export async function createConfig(
     configPath,
     externalRefResolver,
   });
+
+  const resolver = externalRefResolver ?? new BaseResolver();
```
tatomyr: Could you check if moving this part to `resolveConfig` makes it nicer? Looks like you have to declare the resolver 2 times now.

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,58 @@ import {
   type Document,
   type ResolvedRefMap,
 } from '../resolve.js';
-import { CONFIG_FILE_NAME } from './constants.js';
+import { CONFIG_FILE_NAME, IGNORE_FILE } from './constants.js';
+import { isAbsoluteUrlOrFileUrl, getDir } from '../ref-utils.js';
 
 import type { RawUniversalConfig } from './types.js';
 
+function resolvePath(base: string, relative: string): string {
```
tatomyr: This function also looks generic. Maybe we can move it somewhere else too?

.changeset/sour-wings-follow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed loading of `.redocly.lint-ignore.yaml` in browser environments.
```
JLekawa:
```suggestion
Fixed an issue where `.redocly.lint-ignore.yaml` was not loaded in browser environments.
```

JLekawa (approved):

packages/core/src/config/load.ts
kanoru3101: moved to ref-utils too

packages/core/src/config/load.ts
```diff
@@ -35,14 +76,17 @@ export async function loadConfig(
     rawConfigDocument: rawConfigDocument ? cloneConfigDocument(rawConfigDocument) : undefined,
     customExtends,
     configPath,
-    externalRefResolver,
+    externalRefResolver: resolver,
```
tatomyr: Please revert this back.

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,51 @@ import {
...
 import type { RawUniversalConfig } from './types.js';
 
+async function loadIgnoreFile(
+  configPath: string | undefined,
+  resolver: BaseResolver
+): Promise<Record<string, Record<string, Set<string>>> | undefined> {
+  if (!configPath) return undefined;
+
+  const configDir = getDir(configPath);
+  const ignorePath = resolvePath(configDir, IGNORE_FILE);
```
tatomyr: If there's no config file, then resolve the ignore file against CWD.

packages/core/src/ref-utils.ts
```diff
@@ -85,6 +86,27 @@ export function isAbsoluteUrl(ref: string) {
   return ref.startsWith('http://') || ref.startsWith('https://');
 }
 
+export function isAbsoluteUrlOrFileUrl(ref: string) {
```
tatomyr: Let's reuse this function in the language-server/vs-code.

kanoru3101: Added at this [commit](https://github.com/Redocly/redocly/pull/19923/commits/86a5a14e78908b26e6b1211ea0f5d841cca1f8b3)

packages/core/src/config/load.ts
kanoru3101: updated. Please check again

packages/core/src/config/load.ts
```diff
@@ -74,16 +116,22 @@ export async function createConfig(
     rawConfigDocument.parsed = config;
   }
 
+  const resolver = externalRefResolver ?? new BaseResolver();
+
   const { resolvedConfig, resolvedRefMap, plugins } = await resolveConfig({
     rawConfigDocument: cloneConfigDocument(rawConfigDocument),
     configPath,
-    externalRefResolver,
+    externalRefResolver: resolver,
```
tatomyr:
```suggestion
    externalRefResolver,
```

Let's not change the resolver here.

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,49 @@ import {
...
+import { CONFIG_FILE_NAME, IGNORE_FILE } from './constants.js';
+import { isAbsoluteUrlOrFileUrl, getDir, resolvePath } from '../ref-utils.js';
 
 import type { RawUniversalConfig } from './types.js';
 
+async function loadIgnoreFile(
+  configPath: string | undefined,
+  resolver: BaseResolver
+): Promise<Record<string, Record<string, Set<string>>> | undefined> {
+  const configDir = configPath ? getDir(configPath) : process.cwd();
```
tatomyr: Have you tested if it works in the browser env? Is there maybe a fallback needed?

kanoru3101: Found a bug. Without redocly.yaml, language-server does not work. Not sure if this is related to changes in the core, or perhaps it could be in language-server. I am investigating this.

kanoru3101: [Fixed bug](https://github.com/Redocly/redocly/pull/19923/changes/bb1395a93d53001b1ffe068dead0abab48d48282#diff-2aea901cc2e4164b95d06de8179f420db58137fe499a79390fc5a332b6867dca) on language-server side

packages/core/src/config/load.ts
```diff
@@ -74,16 +121,22 @@ export async function createConfig(
...
 
+  const resolver = externalRefResolver ?? new BaseResolver();
+
   const { resolvedConfig, resolvedRefMap, plugins } = await resolveConfig({
     rawConfigDocument: cloneConfigDocument(rawConfigDocument),
     configPath,
     externalRefResolver,
   });
+
+  const ignore = await loadIgnoreFile(configPath, resolver);
```
tatomyr:
```suggestion
const { resolvedConfig, resolvedRefMap, plugins } = await resolveConfig({
rawConfigDocument: cloneConfigDocument(rawConfigDocument),
configPath,
externalRefResolver,
});

  const ignore = await loadIgnoreFile(configPath, externalRefResolver ?? new BaseResolver());
```

tatomyr: TBH, I'm not sure if we need to read the ignore file here at all. Maybe we should provide the resolved file (maybe just a json). Then, [this adapter](https://github.com/Redocly/redocly-cli/pull/2489/changes#diff-91233afeb25fea2c651ad22452d823830a0c6e8378deba81a1109b57cdae3fb6R38-R54) should probably go to the `Config`.

packages/core/src/config/config.ts
```diff
@@ -59,6 +59,8 @@ export class Config {
       resolvedRefMap?: ResolvedRefMap;
       alias?: string;
       plugins?: Plugin[];
+      rawIgnore?: Record<string, Record<string, string[]>>;
+      ignorePath?: string;
```
tatomyr: Just pass `ignoreFile` (or, alternatively, `ignoreDocument`) instead of rawIgnore and ignorePath.

packages/core/src/config/config.ts
```diff
@@ -5,7 +5,7 @@ import { slash } from '../utils/slash.js';
 import { isPlainObject } from '../utils/is-plain-object.js';
 import { specVersions } from '../detect-spec.js';
 import { getResolveConfig } from './get-resolve-config.js';
-import { isAbsoluteUrl } from '../ref-utils.js';
+import { isAbsoluteUrl, isAbsoluteUrlOrFileUrl, resolvePath } from '../ref-utils.js';
```
tatomyr: Do we need both isAbsoluteUrl and  isAbsoluteUrlOrFileUrl? 
Could you check if one of them is redundant?

packages/core/src/config/config.ts
```diff
@@ -142,7 +144,36 @@ export class Config {
...
+      (opts.rawIgnore && opts.ignorePath
+        ? this.resolveIgnore(opts.rawIgnore, opts.ignorePath)
+        : {});
+  }
+
+  private resolveIgnore(
+    ignore: Record<string, Record<string, string[]>>,
+    ignorePath: string
+  ): Record<string, Record<string, Set<string>>> {
+    const adapted = Object.create(null) as Record<string, Record<string, Set<string>>>;
```
tatomyr: Why not just empty object literal?

packages/core/src/config/config.ts
```diff
@@ -71,6 +59,9 @@ export class Config {
       resolvedRefMap?: ResolvedRefMap;
       alias?: string;
       plugins?: Plugin[];
+      rawIgnore?: Record<string, Record<string, string[]>>;
+      ignorePath?: string;
```
tatomyr:
```suggestion
      ignoreFile: { content; path }
```

packages/core/src/config/config.ts
```diff
@@ -153,7 +144,36 @@ export class Config {
...
+      (opts.rawIgnore && opts.ignorePath
+        ? this.resolveIgnore(opts.rawIgnore, opts.ignorePath)
+        : {});
+  }
+
+  private resolveIgnore(
+    ignore: Record<string, Record<string, string[]>>,
+    ignorePath: string
+  ): Record<string, Record<string, Set<string>>> {
+    const adapted = Object.create(null) as Record<string, Record<string, Set<string>>>;
```
tatomyr:
```suggestion
    { content, path }
  ): Record<string, Record<string, Set<string>>> {
    const ignore = Object.create(null) as Record<string, Record<string, Set<string>>>;
```

Better to use meaningful names.

packages/core/src/config/load.ts
```diff
@@ -38,11 +66,15 @@ export async function loadConfig(
...
 
+  const ignoreResult = await loadIgnoreFile(configPath, resolver);
+
   const config = new Config(resolvedConfig, {
     configPath,
     document: rawConfigDocument,
     resolvedRefMap: resolvedRefMap,
     plugins,
+    rawIgnore: ignoreResult?.content,
+    ignorePath: ignoreResult?.path,
```
tatomyr:
```suggestion
  const ignoreFile = await loadIgnoreFile(configPath, resolver);

const config = new Config(resolvedConfig, {
configPath,
document: rawConfigDocument,
resolvedRefMap: resolvedRefMap,
plugins,
    ignoreFile,
```

packages/core/src/config/load.ts
```diff
@@ -58,11 +90,13 @@ type CreateConfigOptions = {
...
   externalRefResolver?: BaseResolver;
   resolvedRefMap?: ResolvedRefMap;
+  rawIgnore?: Record<string, Record<string, string[]>>;
+  ignorePath?: string;
 };
 
 export async function createConfig(
   config?: string | RawUniversalConfig,
-  { configPath, externalRefResolver }: CreateConfigOptions = {}
+  { configPath, externalRefResolver, rawIgnore, ignorePath }: CreateConfigOptions = {}
```
tatomyr:
```suggestion
  { configPath, externalRefResolver, ignoreFile }: CreateConfigOptions = {}
```

packages/core/src/config/config.ts
kanoru3101: `github-advanced-security` [says](https://github.com/Redocly/redocly-cli/security/code-scanning/136) that is `Prototype-polluting assignment` and better use for this case `Object.create(null)` instead `{}`

packages/core/src/config/config.ts
kanoru3101: Used ignoreFile

packages/core/src/config/config.ts
kanoru3101: I moved the logic from isAbsoluteUrlOrFileUrl to isAbsoluteUrl, and it seems to work fine. Tested on Reunite, VSCE (also Windows), and ran tests on CLI

packages/core/src/index.ts
```diff
@@ -26,6 +26,7 @@ export { normalizeTypes, type NormalizedNodeType, type NodeType } from './types/
 export { Stats } from './rules/other/stats.js';
 export {
   loadConfig,
+  loadIgnoreFile,
```
tatomyr, line 29: Do we reuse it somewhere else?

packages/core/src/config/load.ts
```diff
@@ -8,10 +8,38 @@ import {
...
+    return undefined;
+  }
+
+  const ignoreDocument = await resolver.resolveDocument(null, ignorePath, true);
+
+  if (ignoreDocument instanceof Error || !ignoreDocument.parsed) {
+    return undefined;
+  }
+
+  const resolvedIgnorePath =
```
tatomyr: Looks like it's not a file path, but a directory? So either rename `path` to `dir` or calculate the full path (not sure what'd be better).

tatomyr: Moreover, it looks like this directory is always the same as the configDir itself (even for browser). Not sure why do we need all those calculations.

packages/core/src/config/config.ts
```diff
@@ -153,7 +143,35 @@ export class Config {
...
+    path: string;
+  }): Record<string, Record<string, Set<string>>> {
+    const ignore = Object.create(null) as Record<string, Record<string, Set<string>>>;
+
+    for (const fileName of Object.keys(content)) {
+      const fileIgnore = content[fileName];
+
+      const resolvedFileName = isAbsoluteUrl(fileName)
+        ? fileName
+        : path
```
tatomyr: Is it possible that there's no path?

packages/core/src/index.ts
kanoru3101: In the language server, it should be added that ignoring works when there is no redocly.yaml.

tatomyr: Strange, I thought it should just work automatically when using `loadCogfig`. Let's discuss that.

kanoru3101: It is used in `language-server` to create a default configuration when we don't have redocly.yaml. There, we use `createConfig` and must send the result of `loadIgnoreFile` to `createConfig`; if we don't, the ignore file will not work.

packages/core/src/config/config.ts
kanoru3101: No, simplified

packages/core/src/config/load.ts
kanoru3101: Yes you are right, I checked `configDir` it was the same as `resolvedIgnorePath`. Used `configDir` instead `resolvedIgnorePath`

packages/core/src/config/config.ts
```diff
@@ -153,7 +143,31 @@ export class Config {
...
 
-    this.resolveIgnore(getIgnoreFilePath(opts.configPath));
+    this.ignore = opts.ignore ?? (opts.ignoreFile ? this.resolveIgnore(opts.ignoreFile) : {});
+  }
+
+  private resolveIgnore({
+    content,
+    dir,
+  }: {
+    content: Record<string, Record<string, string[]>>;
```
tatomyr: Reuse the type.

tatomyr (approved):
Left a minor comment. Otherwise looks good.

PR #2495, opened 2026-01-12
feat: implemented new linting of entities to support new scorecards

AlbinaBlazhko17:
## What/Why/How?
- Implemented `lintEntityWithScorecardLevel`, which will be used in `portal` to lint new entities with new scorecard functionality.
- Implemented `lintSchema` for linting specific schema. If entity has `data-schema` type we need to lint only provided schema.
- Added support for `Assertions` (custom rules) in `lintEntityFile`.
- The main goal of helper functions are to transform assertion subject to type, which is provided in `@redocly/config`.
- Changed `json-schema-adapter.ts` to handle `nodeTypeName` if provided. Added returning of `discriminatorFunc` from adapter to get proper rootType. Added functionality to `discriminatorFunc` for handling mapping the default naming and provided `nodeTypeName`.
- Fixed tests to match new structure.

### Opened questions:
- [ ] How to handle entities with the type of `api-operations`.
- [ ] Adding support of custom rules for arrays. Currently, we don't have any Assertions (e.g. `lt`, `lte`, `gte`, etc.) for arrays.

## Testing
Ran tests in `CLI` and `Realm`.

## Check yourself

## Security

### DIFF
--- a/packages/core/src/types/json-schema-adapter.ts
+++ b/packages/core/src/types/json-schema-adapter.ts
@@ -184,16 +184,11 @@ function extractNodeToContext(
 
   let additionalProperties;
   if (isPlainObject(schema.additionalProperties)) {
-    const additionalPropertiesNodeTypeName = (
-      schema.additionalProperties as JSONSchema & { nodeTypeName?: string }
-    )['nodeTypeName'];
-    additionalProperties = additionalPropertiesNodeTypeName
-      ? extractNodeToContext(additionalPropertiesNodeTypeName, schema.additionalProperties, ctx)
-      : transformJSONSchemaToNodeType(
-          nodeTypeName + '_additionalProperties',
-          schema.additionalProperties,
-          ctx
-        );
+    additionalProperties = transformJSONSchemaToNodeType(
+      propertyName + '_additionalProperties',
+      schema.additionalProperties,
+      ctx
+    );
   }
   if (schema.additionalProperties === true) {
     additionalProperties = {};
@@ -206,12 +201,7 @@ function extractNodeToContext(
       isPlainObject(schema.items.additionalProperties) ||
       schema.items.oneOf) // exclude scalar array types
   ) {
-    const itemsNodeTypeName = (schema.items as JSONSchema & { nodeTypeName?: string })[
-      'nodeTypeName'
-    ];
-    items = itemsNodeTypeName
-      ? extractNodeToContext(itemsNodeTypeName, schema.items, ctx)
-      : transformJSONSchemaToNodeType(nodeTypeName + '_items', schema.items, ctx);
+    items = transformJSONSchemaToNodeType(propertyName + '_items', schema.items, ctx);
   }
 
   let required = schema.required as NodeType['required'];


packages/core/src/__tests__/entity-yaml.test.ts
```diff
@@ -30,11 +30,11 @@ describe('entity-yaml', () => {
     ) as NormalizedNodeType;
 
     expect(resolvedType).toBeTruthy();
-    expect(resolvedType.name).toBe('EntityFileDefault');
+    expect(resolvedType.name).toBe('UserEntity'); // Falls back to first type when discriminator fails
```
tatomyr: Maybe it's better to put `Entity` first, so it becomes the default fallback?

packages/core/src/__tests__/entity-yaml.test.ts
```diff
@@ -50,22 +50,61 @@ describe('entity-yaml', () => {
...
     expect(defaultNode.required).toContain('type');
     expect(defaultNode.required).toContain('key');
     expect(defaultNode.required).toContain('title');
   });
+
+  it('should correctly discriminate between different entity types in an array', async () => {
+    const { lintEntityFile } = await import('../lint.js');
+    const { makeDocumentFromString, BaseResolver } = await import('../resolve.js');
+
+    const entities = `- type: user
```
tatomyr: Could you use `outdent`?

packages/core/src/__tests__/entity.test.ts
```diff
@@ -4,21 +4,21 @@ import { NormalizedNodeType, normalizeTypes, ResolveTypeFn } from '../types/inde
...
+
+    expect(entityTypes).toHaveProperty('UserEntity');
+    expect(entityTypes).toHaveProperty('ApiOperationEntity');
+    expect(entityTypes).toHaveProperty('DataSchemaEntity');
+    expect(entityTypes).toHaveProperty('ApiDescriptionEntity');
+    expect(entityTypes).toHaveProperty('ServiceEntity');
+    expect(entityTypes).toHaveProperty('DomainEntity');
+    expect(entityTypes).toHaveProperty('TeamEntity');
+    expect(entityTypes).toHaveProperty('Entity');
     expect(entityTypes).toHaveProperty('EntityFileArray');
```
tatomyr, line 18: WHy don't we expect EntityFile/Default anymore?

tatomyr: I see we use them in `createEntityTypes`.

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -109,15 +113,27 @@ function transformJSONSchemaToNodeType(
...
+
+        return actualTypeName;
       });
 
       return (value: unknown, key: string) => {
         if (isPlainObject(value)) {
-          const discriminatedTypeName = value[discriminatedPropertyName];
-          if (typeof discriminatedTypeName === 'string' && ctx[discriminatedTypeName]) {
-            return discriminatedTypeName;
+          const discriminatorValue = value[discriminatedPropertyName];
```
tatomyr: Why changed discriminatedTypeName -> discriminatorValue? It's still a property name, not a value. Or am I missing something?

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -109,15 +113,27 @@ function transformJSONSchemaToNodeType(
...
           );
         }
-        const name = discriminatedProperty.const as string;
-        return transformJSONSchemaToNodeType(name, option, ctx);
+        const discriminatorValue = discriminatedProperty.const as string;
+        const actualTypeName = transformJSONSchemaToNodeType(discriminatorValue, option, ctx);
+
+        // Store mapping from discriminator value to actual type name
+        if (typeof actualTypeName === 'string') {
+          discriminatorMapping[discriminatorValue] = actualTypeName;
```
tatomyr, line 122: Not sure why do we need to mutate it here. Don't we receive the discriminator mapping in the needed form already?

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -156,15 +172,19 @@ function extractNodeToContext(
     );
   }
 
+  // Use nodeTypeName from schema if provided, otherwise use propertyName
+  const nodeTypeName =
+    (schema as JSONSchema & { nodeTypeName?: string })['nodeTypeName'] || propertyName;
```
tatomyr: We can also populate `description` and `documentationLink` fields from the schema.

tatomyr:
```suggestion
    (schema as JSONSchema & { nodeTypeName?: string })['nodeTypeName'] ?? propertyName;
```

tatomyr: And also, I'm not sure regarding the property name itself... Let's discuss it offline.

tatomyr: This change will also affect creation of other types, isn't it? (At least, types for redocly.yaml.)

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -203,15 +223,26 @@ function extractNodeToContext(
...
+  const discriminatorResolver = transformJSONSchemaToNodeType(schemaName, entrySchema, ctx);
+  if (discriminatorResolver && typeof discriminatorResolver === 'function') {
+    return {
+      ctx,
+      discriminatorResolver,
+    };
+  }
+  return {
+    ctx,
+  };
```
tatomyr:
```suggestion
    return {
      ctx,
      discriminatorResolver: typeof discriminatorResolver === 'function' ? discriminatorResolver : undefined,
    };
```

It's better to always return the same structure, but filter inside the property itself (if needed).

packages/core/src/utils/scorecards.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  return assertionConfig;
+}
+
+export function categorizeAssertions(assertionConfig: AssertionConfig): AssertionsByTarget {
+  const entityRules: Record<string, Assertion> = {};
+  const apiRules: Record<string, Assertion | RuleConfig> = {};
+  for (const [ruleKey, ruleValue] of Object.entries(assertionConfig)) {
```
tatomyr, line 57: Why not just this?

```suggestion
  for (const ruleKey of assertionConfig) {
```

packages/core/src/utils/scorecards.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  } catch {
+    return null;
+  }
+
+  return null;
+}
+
+function isAssertionRule(ruleKey: string, ruleValue: unknown): ruleValue is RawAssertion {
+  return ruleKey.startsWith('rule/') && typeof ruleValue === 'object' && ruleValue !== null;
```
tatomyr:
```suggestion
  return ruleKey.startsWith('rule/') && isPlainObject(ruleValue);
```

BTW, don't we have this check already in other places?

packages/core/src/utils/scorecards.ts
```diff
@@ -0,0 +1,148 @@
...
+    ...(transformedWhere && { where: transformedWhere }),
+    ...(rawAssertion.message && { message: rawAssertion.message }),
+    ...(rawAssertion.severity && { severity: rawAssertion.severity }),
+  };
+}
+
+function hasComponents(parsed: unknown): parsed is { [key: string]: unknown } {
+  return (
+    typeof parsed === 'object' &&
+    parsed !== null &&
```
tatomyr:
```suggestion
    isPlainObject(parsed)
```

packages/core/src/lint.ts
```diff
@@ -211,33 +228,64 @@ export async function lintEntityFile(opts: {
   entityDefaultSchema: JSONSchema;
   severity?: ProblemSeverity;
   externalRefResolver?: BaseResolver;
+  assertionConfig?: Record<string, Assertion>;
```
tatomyr: Assertions should be converted a separate array, at least it's so for regular lint. Let's discuss this offline.

packages/core/src/lint.ts
```diff
@@ -275,3 +325,146 @@ export async function lintEntityFile(opts: {
...
+  entity: EntityFileSchema | EntityBaseFileSchema,
+  scorecardLevel: NonNullable<ScorecardConfig['levels']>[number],
+  document?: Document
+): Promise<NormalizedProblem[]> {
+  if (!scorecardLevel.rules) {
+    throw new Error(`Scorecard level "${scorecardLevel.name}" has no rules defined.`);
+  }
+
+  const externalRefResolver = new BaseResolver();
+  const entityDocument = makeDocumentFromString(JSON.stringify(entity, null, 2), 'entity.yaml');
```
tatomyr: Why do we need to create a fake entity file on the fly? We should be able to resolve the real files.

packages/core/src/lint.ts
```diff
@@ -275,3 +325,146 @@ export async function lintEntityFile(opts: {
...
+    const apiProblems = await lintDocument({
+      document,
+      externalRefResolver,
+      config: await createConfig({
+        rules: apiRules,
+      }),
+    });
+
+    return [...entityProblems, ...apiProblems];
+  } else if (Object.keys(apiRules).length !== 0) {
```
tatomyr: Please use a utility function if possible  (isEmptyObject or something like that).

packages/core/src/utils/scorecards.ts
```diff
@@ -0,0 +1,148 @@
...
+  const specificType = capitalizedEntityType + subjectType;
+
+  if ((Object.values(NODE_TYPE_NAMES) as string[]).includes(specificType)) {
+    return specificType;
+  }
+
+  return subjectType;
+}
+
+export function transformScorecardRulesToAssertions(
```
tatomyr, line 31: Why not use the existing assertion/rules conversion mechanism?

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: I've added here this logic to return proper nodeName, so i can use the discriminator function without any additional logic. Without it i will pass e.g. 'domain' and this function will return 'domain' instead of 'DomainEntity'. I can remove this logic from the discriminator function, but in this case for me this function does not make sense. WDYT?

packages/core/src/utils/scorecards.ts
AlbinaBlazhko17: Because i need both: `ruleKey` and `ruleValue`

packages/core/src/__tests__/entity-yaml.test.ts
AlbinaBlazhko17: Added

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: You are right, fixed naming

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: Thanks, fixed

packages/core/src/utils/scorecards.ts
AlbinaBlazhko17: Fixed

packages/core/src/lint.ts
AlbinaBlazhko17: Added

packages/core/src/utils/scorecards.ts
AlbinaBlazhko17: Yeah, we have similar check here, but it doesn't fit for my implementation: https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/config/group-assertion-rules.ts#L22

packages/core/src/utils/scorecards.ts
AlbinaBlazhko17: Because i am using `buildAssertionWithNormalizedTypes`, which converts `subject` and `where` properties to appropriate node types. e.g. subject.type: EntityMetadata, entity.type: 'user', so in this case we need to have subject like UserEntityMetadata.

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: I don't think, that it will affect smth, because currently we use `propertyName` and I've added `nodeTypeName`, which has a priority over `propertyName`.

packages/core/src/__tests__/entity-yaml.test.ts
AlbinaBlazhko17: Thanks, i fixed a huge issue

packages/core/src/utils/scorecards.ts
```diff
@@ -0,0 +1,148 @@
...
+    ...(rawAssertion.message && { message: rawAssertion.message }),
+    ...(rawAssertion.severity && { severity: rawAssertion.severity }),
+  };
+}
+
+function hasComponents(parsed: unknown): parsed is { [key: string]: unknown } {
+  return (
+    isPlainObject(parsed) &&
+    'components' in parsed &&
+    typeof (parsed as Record<string, unknown>).components === 'object'
```
tatomyr:
```suggestion
    isPlainObject(parsed.components)
```

tatomyr: Maybe we don't need a dedicated function for this? Will maybe this work instead?

```ts
isPlainObject(parsed?.components)
```

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -156,15 +173,19 @@ function extractNodeToContext(
...
   for (const [name, property] of Object.entries(schema.properties || {})) {
-    properties[name] = transformJSONSchemaToNodeType(propertyName + '.' + name, property, ctx);
+    properties[name] = transformJSONSchemaToNodeType(nodeTypeName + '.' + name, property, ctx);
   }
 
   let additionalProperties;
   if (isPlainObject(schema.additionalProperties)) {
     additionalProperties = transformJSONSchemaToNodeType(
-      propertyName + '_additionalProperties',
+      nodeTypeName + '_additionalProperties',
```
tatomyr: If it's a node type, we don't want to add the additional notations. It's already a desired type name.

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -180,7 +201,7 @@ function extractNodeToContext(
       isPlainObject(schema.items.additionalProperties) ||
       schema.items.oneOf) // exclude scalar array types
   ) {
-    items = transformJSONSchemaToNodeType(propertyName + '_items', schema.items, ctx);
+    items = transformJSONSchemaToNodeType(nodeTypeName + '_items', schema.items, ctx);
```
tatomyr: Same here.

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -203,15 +224,22 @@ function extractNodeToContext(
...
+  ctx: Record<string, NodeType>;
+  discriminatorResolver?: ResolveTypeFn;
+} {
   const ctx: Record<string, NodeType> = {};
-  transformJSONSchemaToNodeType(schemaName, entrySchema, ctx);
-  return ctx;
+  const discriminatorResolver = transformJSONSchemaToNodeType(schemaName, entrySchema, ctx);
+  return {
+    ctx,
+    discriminatorResolver:
```
tatomyr, line 243: This will only return one discriminatorResolver for all types. Instead, it should return discriminator resolvers in all NodeTypes that are returned in `ctx`. Please leave a TODO to improve that later.

packages/core/src/utils/scorecards.ts
AlbinaBlazhko17: Thanks, fixed

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: Fixed

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: Fixed

packages/core/src/types/json-schema-adapter.ts
AlbinaBlazhko17: Added comment

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -180,7 +206,12 @@ function extractNodeToContext(
       isPlainObject(schema.items.additionalProperties) ||
       schema.items.oneOf) // exclude scalar array types
   ) {
-    items = transformJSONSchemaToNodeType(propertyName + '_items', schema.items, ctx);
+    const itemsNodeTypeName = (schema.items as JSONSchema & { nodeTypeName?: string })[
```
tatomyr: Let's revert this.

packages/core/src/types/json-schema-adapter.ts
```diff
@@ -156,18 +173,27 @@ function extractNodeToContext(
...
   }
 
   let additionalProperties;
   if (isPlainObject(schema.additionalProperties)) {
-    additionalProperties = transformJSONSchemaToNodeType(
-      propertyName + '_additionalProperties',
-      schema.additionalProperties,
-      ctx
-    );
+    const additionalPropertiesNodeTypeName = (
```
tatomyr: Let's revert this.

PR #2524, opened 2026-01-30
feat: add support for mediaTypes in OAS 3.2

n0rahh:
## What/Why/How?
Added bundling support for OpenAPI 3.2 `components/mediaTypes` references used in request and response content.

## Reference
[OpenAPI 3.2 support](https://github.com/Redocly/redocly/issues/15509)
[openapi-docs changes](https://github.com/Redocly/redocly/pull/20646)

## Check yourself

## Security


packages/core/src/rules/oas3/no-unused-components.ts
```diff
@@ -15,7 +15,16 @@ export const NoUnusedComponents: Oas3Rule = () => {
...
       if (
-        ['Schema', 'Header', 'Parameter', 'Response', 'Example', 'RequestBody'].includes(type.name)
+        [
+          'Schema',
+          'Header',
+          'Parameter',
+          'Response',
+          'Example',
+          'RequestBody',
+          'MediaType',
```
tatomyr: Why do we need to check media types here?

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -8,23 +9,25 @@ import type {
   Oas3_2Definition,
   Oas3Components,
   Oas3_1Components,
+  Oas3_2Components,
 } from '../../typings/openapi.js';
 
 type AnyOas3Definition = Oas3Definition | Oas3_1Definition | Oas3_2Definition;
+type AnyOas3Component = keyof Oas3Components | keyof Oas3_1Components | keyof Oas3_2Components;
```
tatomyr: Technically, it's `AnyOas3ComponentKey`. Or you can use `keyof` where you need it directly.

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -34,6 +34,9 @@ export function mapTypeToComponent(typeName: string, version: SpecMajorVersion)
           return 'links';
         case 'Callback':
           return 'callbacks';
+        case 'MediaTypesMap':
+        case 'MediaType':
+          return 'mediaTypes';
```
tatomyr, line 38: Is your intention to populate the `components` with `mediaTypes` when bundling? Is there a test for that?

n0rahh: You’re right — MediaType isn’t needed, it was added defensively and has now been removed

packages/core/src/rules/oas3/no-unused-components.ts
n0rahh: I added MediaType initially out of caution, my bad, I've removed it now

packages/core/src/bundle/bundle-visitor.ts
tatomyr: Again, could you add a test to cover the case when an external media type link transforms to a component?

.changeset/lucky-bees-say.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
```
tatomyr: It's a feature, so this should be minor rather than patch.

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -8,23 +9,25 @@ import type {
   Oas3_2Definition,
   Oas3Components,
   Oas3_1Components,
+  Oas3_2Components,
 } from '../../typings/openapi.js';
 
 type AnyOas3Definition = Oas3Definition | Oas3_1Definition | Oas3_2Definition;
+type AnyOas3ComponentKey = keyof Oas3Components | keyof Oas3_1Components | keyof Oas3_2Components;
```
tatomyr:
```suggestion
type AnyOas3ComponentsKey = keyof Oas3Components | keyof Oas3_1Components | keyof Oas3_2Components;
```

tatomyr (commented):
Left a couple of minor comments.

packages/core/src/__tests__/bundle.test.ts
```diff
@@ -144,6 +144,99 @@ describe('bundle', () => {
...
+              id: 1,
+              name: 'John',
+            },
+          },
+        },
+      },
+    });
+    expect(parsed.paths?.['/pets']?.get?.responses?.['200']?.content).toEqual({
+      $ref: '#/components/mediaTypes/JsonPets',
+    });
```
tatomyr: It's really challenging to read those expects. Could you use inline snapshots instead? Same in other places.

packages/core/src/__tests__/fixtures/refs/media-types.yaml
```diff
@@ -0,0 +1,17 @@
...
+          properties:
+            id:
+              type: integer
+            name:
+              type: string
+        examples:
+          example1:
+            value:
+              id: 1
+              name: John
```
tatomyr: You don't have to wrap it into OpenAPI, it could be literally what you need:

```yaml
      application/json:
        schema:
          type: object
          properties:
            id:
              type: integer
        examples:
          Test:
            value:
              id: 1
```

packages/core/src/__tests__/fixtures/refs/external-media-types.yaml
```diff
@@ -0,0 +1,20 @@
...
+  /pets:
+    get:
+      summary: List all pets
+      operationId: listPets
+      responses:
+        '200':
+          description: OK
+          content:
+            $ref: ./media-types.yaml#/components/mediaTypes/JsonPets
+components:
```
tatomyr: What do you need `components` for? Please keep your tests minimal so it's easier to read them and understand the purpose. Avoid unnecessary fields (like summary and operationId since they are not needed for this particular test), and don't use strange naming like 'pets', just stick to foo, bar, test, user, etc., or some other general names. 
This also applies to other test samples.

PR #2525, opened 2026-01-30
fix: do not print bundle warnings into stdout

tatomyr:
## What/Why/How?

Fixed an issue where schema name collision warnings appeared in bundle output when using stdout.

## Reference

Addresses https://github.com/Redocly/redocly-cli/issues/661#issuecomment-3823004190

## Check yourself

## Security

### DIFF
--- a/.changeset/yummy-pandas-retire.md
+++ b/.changeset/yummy-pandas-retire.md
@@ -3,4 +3,4 @@
 "@redocly/cli": patch
 ---
 
-Fixed an issue where schema name collision warnings appeared in bundle output when using stdout.
+Fixed an issue where schema name collision warnings appeared in bundle output when using `stdout`.


.changeset/yummy-pandas-retire.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed an issue where schema name collision warnings appeared in bundle output when using stdout.
```
JLekawa:
```suggestion
Fixed an issue where schema name collision warnings appeared in bundle output when using `stdout`.
```

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -289,11 +290,9 @@ export function makeBundleVisitor(
     }
 
     if (!componentsGroup[name]) {
-      ctx.report({
-        message: `Two schemas are referenced with the same name but different content. Renamed ${prevName} to ${name}.`,
-        location: ctx.location,
```
DmitryAnansky, line 294: Is there some reason for changing the report?
No location is needed anymore ?

DmitryAnansky: Or even no report enymore?

tatomyr: The original file names are absolutely correct, and we are right to rename components during bundling too. So it doesn’t look like a bundler problem to me. 

The only thing is that it won't show a specific location anymore, but it will still produce the warning in Terminal with the component names, so it's not a big deal to figure it out.

tatomyr:
Closing this one in favour of this: https://github.com/Redocly/redocly-cli/pull/2541.

Issue #2464, opened 2025-12-17
Decorator to include only public items


sanscontext:
**Is your feature request related to a problem? Please describe.**

I'm working on an OAS file for a repo with a _lot_ of private routes. In probably about 30% or less of these routes are actually exposed to the "public" partners, and the rest are used internally.  

I got this working "well enough" by tagging all the routes with `x-internal` and then bundling with the `remove-x-internal` decorator. But then I had a (totally reasonable) request from one of the engineers to make this "fail closed", which is to say only include routes that are explicitly tagged, rather than relying on a dev knowing they need to tag new work and any imported routes with `x-internal`.

**Describe the solution you'd like**

I would love a built-in decorator that's something like `include-only-x-public` that would do sort of the inverse of `remove-x-internal` and keep ONLY the items explicitly tagged.  

**Describe alternatives you've considered**

I did try to get this working with `filter-in`, but that can't filter on the `absence` of a tag, so having to have my devs remember to tag everything with `x-public: false` isn't any better than having them need to tag them as `x-internal: true`.

**Additional context**

From [a WritetheDocs slack thread](https://writethedocs.slack.com/archives/C0YH9K2JY/p1765931418522369). :)

tatomyr:
Hi @sanscontext, would the [filter-in](https://redocly.com/docs/cli/decorators/filter-in#examples) decorator serve your purpose? Something like the following:

```yaml
    decorators:
      filter-in:
        property: operationId
        value: [createSpecialEvent, listSpecialEvents]
```

or 

```yaml
    decorators:
      filter-in:
        property: x-public
        value: true
```

or 

```yaml
    decorators:
      filter-in:
        property: x-visibility
        value: public
```

tatomyr:
If that doesn't work, you can use a [custom decorator](https://redocly.com/docs/cli/custom-plugins) like the following one:

```js
export default () => ({
  id: 'plugin',
  decorators: {
    oas3: {
      'filter-operations': (opts) => {
        let usedOperations = [];
        let unusedOperations = [];
        let unusedPathItems = [];
        return {
          Paths: {
            enter() {
              unusedPathItems = [];
            },
            leave(paths) {
              unusedPathItems.forEach((pathItem) => {
                delete paths[pathItem];
              });
            },
            PathItem: {
              enter() {
                usedOperations = [];
                unusedOperations = [];
              },
              leave(pathItem, ctx) {
                unusedOperations.forEach((operation) => {
[...]

PR #2536, opened 2026-02-04
feat(filter-in): add applyTo option to limit filtering to specific node types

balvig:
## What/Why/How?

When using `filter-in` with `property: tags` to filter operations by tag, the decorator filters _any_ schema properties that happen to be named "tags", which can case undesirable results in cases such as this:

```yaml
# configuration
filter-in:
  property: tags
  value: App

# schema
openapi: 3.0.0
paths:
  /events:
    get:
      tags:
        - App
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Event'
components:
  schemas:
    Event:
      type: object
      properties:
        id:
          type: string
        tags:  # <-- This property gets incorrectly filtered out!
          type: array
          items:
            type: string
          description: User-defined tags for the event
```

This PR adds an `applyTo` option that lets users specify which node type the filter applies to (e.g., `PathItem`). 
This prevents unintended filtering of schema properties while still allowing filtering at the operation level:

```yaml
filter-in:
  property: tags
  value: App
  applyTo: PathItem
```

## Testing

Added a test case that verifies:
- Operations with the correct tag are included
- Operations with wrong tags are filtered out
- Schema properties named "tags" are preserved (not filtered)

## Check yourself

- [x] Have you added a changeset for your changes?
- [x] Have you added tests?

### DIFF
--- a/.changeset/filter-in-applyto.md
+++ b/.changeset/filter-in-applyto.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added `applyTo` option to `filter-in` decorator to limit filtering to specific node types. This allows users to avoid accidentally filtering schema properties when filtering operations by tags.
--- a/packages/core/src/decorators/__tests__/filter-in.test.ts
+++ b/packages/core/src/decorators/__tests__/filter-in.test.ts
@@ -250,6 +250,92 @@ describe('oas3 filter-in', () => {
 
     `);
   });
+
+  it('should support applyTo option to limit filtering to specific node types', async () => {
+    // Using applyTo: PathItem prevents filtering schema properties that happen
+    // to have the same name as the filter property (e.g., a schema with a "tags"
+    // property definition should not be filtered when filtering operations by tags).
+    const testDoc = parseYamlToDocument(
+      outdent`
+        openapi: 3.0.0
+        paths:
+          /events:
+            get:
+              tags:
+                - App
+              summary: List events
+              responses:
+                '200':
+                  description: OK
+                  content:
+                    application/json:
+                      schema:
+                        $ref: '#/components/schemas/Event'
+          /users:
+            get:
+              tags:
+                - Public
+              summary: List users
+        components:
+          schemas:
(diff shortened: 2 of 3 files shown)


tatomyr:
This is a good contribution, @balvig, thank you!

However, I'd like to point to a couple of things here. 
First, if adding `applyTo` to `filter-in`, it's worth adding it to `filter-out` as well (and reflect that in the corresponding docs). 
Second, it looks odd that we have to specify `PathItem` when we're actually targeting the `Operation` level.

Currently I'm trying to revise the logic at all in scope of [this ticket](https://github.com/Redocly/redocly-cli/issues/2464). I'm leaning towards creating a separate  decorator (something like `filter-in-operations`) that will target specifically operations as those are the atomic values of OpenAPI.  I might end up introducing the `applyTo` configuration property there too though. 

What do you think? Wouldn't that new decorator serve your purpose better?

balvig:
> Wouldn't that new decorator serve your purpose better?

@tatomyr it probably would! 

All I really need is a way to filter in/out operations based on tags, without affecting other nodes that happen to have a `tags` property. 🙂

tatomyr:
@balvig I'll try to file a draft PR later today or early next week.

tatomyr:
@balvig could you check if the tests are covering your case? https://github.com/Redocly/redocly-cli/pull/2555

balvig:
Thank you @tatomyr, I think so!

The particular issue I had was with a child element _also_ containing a `tags` property and getting unintentionally removed, so not sure if worth covering that edge case in the tests, but otherwise looks good!

tatomyr:
@balvig, after a lot of discussions we've eventually decided to go with the interface you proposed 🙂. Released in v2.21.0. Here are the related docs: https://redocly.com/docs/cli/decorators/filter-in. Could you verify it works for you?

Closing this as the idea was implemented in https://github.com/Redocly/redocly-cli/pull/2570.

balvig:
Thank you @tatomyr! 
This seems to work for me. Thank you for the effort, appreciate it! 🙇

PR #2538, opened 2026-02-04
chore(perf): migrate from prettier+eslint to oxfmt+oxlint

### LINKED ISSUE #2535: suggestion: migrate from prettier+eslint to oxfmt+oxlint
## Why

[oxfmt+oxlint](https://oxc.rs) is much faster than prettier+eslint

## How

I'll submit a PR

gameroman:
## What/Why/How?

oxfmt+oxlint is much faster than prettier+eslint

## Reference

Closes #2535 

## Testing

Formatting should stay the same

## Check yourself

## Security

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -10,8 +10,15 @@
   },
   "ignorePatterns": ["**/__tests__/**", "packages/*/lib/**", "*.js", "*.cjs"],
   "rules": {
-    "no-array-constructor": "error",
-    "no-unused-vars": [
+    "eslint/no-case-declarations": "error",
+    "eslint/no-empty": "error",
+    "eslint/no-redeclare": "error",
+    "eslint/no-regex-spaces": "error",
+    "eslint/no-unexpected-multiline": "error",
+    "eslint/no-unreachable": "error",
+    "eslint/no-array-constructor": "error",
+
+    "eslint/no-unused-vars": [
       "error",
       {
         "argsIgnorePattern": "^_",
@@ -19,7 +26,18 @@
         "caughtErrors": "none"
       }
     ],
+
     "typescript/consistent-type-imports": ["error", { "prefer": "type-imports" }],
+
+    "typescript/ban-ts-comment": "error",
+    "typescript/no-empty-object-type": "error",
+    "typescript/no-namespace": "error",
+    "typescript/no-require-imports": "error",
+    "typescript/no-unnecessary-type-constraint": "error",
+    "typescript/no-unsafe-function-type": "error",
+
+    "typescript/no-explicit-any": "warn",
+
     "import/no-duplicates": "error"
   }
 }


gameroman:
There are some formatting changes, mainly because of import sorting and due to the fact that we had `prettier` `v2.7.1` but `oxfmt` is compatible with `v3.8`

.oxlintrc.json
```diff
@@ -0,0 +1,906 @@
...
+        "@typescript-eslint/consistent-type-imports": [
+          "error",
+          {
+            "prefer": "type-imports"
+          }
+        ],
+        "import/no-duplicates": "error",
+        "no-empty-function": "off"
+      },
+      "globals": {
```
RomanHotsiy: do we need all of those?

RomanHotsiy: we didn't have it in our eslint

packages/cli/src/__tests__/commands/build-docs.test.ts
RomanHotsiy, line 1: we did not have import order before? Or we had and it didn't work? why so many fies changes

RomanHotsiy:
@gameroman any more details on time to format/lint before vs now?

.oxlintrc.json
gameroman: This is from running [`npx @oxlint/migrate`](https://oxc.rs/docs/guide/usage/linter/migrate-from-eslint.html#migrating-from-an-eslint-flat-config)

packages/cli/src/__tests__/commands/build-docs.test.ts
gameroman: Import order was checked by `eslint` previously, now it is automatically formatted by `oxfmt`

gameroman:
> Any more details on time to format/lint before vs now?

For linting, `oxlint` is much faster than `eslint`, it takes less than a second now

```
Finished in 359ms on 377 files with 59 rules using 4 threads.
```

For formatting, it takes about 25 seconds on my weak laptop, which is more than I'd expect

```
Finished in 22028ms on 1278 files using 4 threads.
```

gameroman:
It seems like `oxfmt` takes a lot of time formatting [`rebilly-description.yaml`](https://github.com/gameroman/redocly-cli/blob/migrate-to-oxfmt-oxlint/tests/smoke/rebilly/rebilly-description.yaml) file

gameroman:
If I understand correctly `oxfmt` fallbacks to `prettier` for formatting `yaml` files

gameroman:
I think we can exclude those files from being checked by `oxfmt` as it is a smoke test file and shouldn't be modified anyway

.oxlintrc.json
gameroman: Most of them are default values I guess and can be removed probably, I will double check

packages/cli/src/__tests__/commands/build-docs.test.ts
gameroman: Most of the changes are due to `"newlinesBetween": true` option. We can set it to `false` and then there will be no empty new lines between imports

.oxlintrc.json
gameroman: done

gameroman:
Hi, @RomanHotsiy is there anything else I need to address?

RomanHotsiy:
> For linting, oxlint is much faster than eslint, it takes less than a second now

Yes, I would like to see "before / after" comparison.

> For formatting, it takes about 25 seconds on my weak laptop, which is more than I'd expect

Same, how long it took with prettier?

> There are some formatting changes, mainly because of import sorting and due to the fact that we had prettier v2.7.1 but oxfmt is compatible with v3.8

can we disable import sorting and raise a separate PR for it. It's hard to review  both tooling setup and so many changes.

@gameroman

gameroman:
> Yes, I would like to see "before / after" comparison.

On my laptop

|           | Before            | After           |
|-----------|-------------------|-----------------|
| Linter    | 23384ms - 37783ms | 251ms - 525ms   |
| Formatter | 25633ms - 29072ms | 7102ms - 9083ms |


> can we disable import sorting and raise a separate PR for it. It's hard to review both tooling setup and so many changes.

Sure, disabled it for now

@RomanHotsiy

gameroman:
Okay, let me just resolve the conflicts then

packages/cli/src/auth/device-flow.ts
```diff
@@ -52,8 +52,8 @@ export class RedoclyOAuthDeviceFlow {
         process.platform === 'win32'
           ? `start ${url}`
           : process.platform === 'darwin'
-          ? `open ${url}`
-          : `xdg-open ${url}`;
+            ? `open ${url}`
```
tatomyr, line 55: I like the previous formatting more. Can it be configured back?

packages/core/src/config/utils.ts
```diff
@@ -29,7 +29,7 @@ export function prefixRules<
     | Async2RuleSet
     | Arazzo1RuleSet
     | Overlay1RuleSet
-    | OpenRpc1RuleSet
+    | OpenRpc1RuleSet,
```
tatomyr, line 32: What's the point? Can it be configured back?

packages/respect-core/src/modules/flow-runner/prepare-request.ts
```diff
@@ -97,10 +97,7 @@ export async function prepareRequest(
     replacements,
   } = await parseRequestBody(step['requestBody'], ctx);
 
-  const requestBody =
-    stepRequestBodyPayload !== undefined
-      ? stepRequestBodyPayload
-      : requestDataFromOpenAPI?.requestBody;
+  const requestBody = stepRequestBodyPayload || requestDataFromOpenAPI?.requestBody;
```
tatomyr: This is wrong. It'll work incorrectly when the requestBody equals to `null`, `false`, empty string, &c..
I'd only suggest to swap it to be more readable: 


```suggestion
  const requestBody = stepRequestBodyPayload === undefined
    ? requestDataFromOpenAPI?.requestBody
    : stepRequestBodyPayload;
```

package.json
```diff
@@ -14,9 +14,10 @@
     "unit": "VITEST_SUITE=unit vitest run",
     "e2e": "VITEST_SUITE=e2e vitest run",
     "smoke:rebilly": "VITEST_SUITE=smoke-rebilly vitest run",
-    "prettier": "npx prettier --write \"**/*.{ts,js,yaml,yml,json,md}\"",
-    "prettier:check": "npx prettier --check \"**/*.{ts,js,yaml,yml,json,md}\"",
-    "eslint": "eslint packages/**",
+    "format": "oxfmt",
+    "check": "oxfmt --check",
```
tatomyr:
```suggestion
    "format:check": "oxfmt --check",
```

packages/respect-core/src/modules/flow-runner/prepare-request.ts
tatomyr: The tests are failing due to this change, I believe.

.oxlintrc.jsonc
```diff
@@ -0,0 +1,129 @@
...
+          {
+            "prefer": "type-imports",
+          },
+        ],
+        "import/no-duplicates": "error",
+        "no-empty-function": "off",
+      },
+      "env": {
+        "es2026": true,
+        "commonjs": true,
```
tatomyr: We don't use CommonJS.

.oxlintrc.jsonc
```diff
@@ -0,0 +1,129 @@
...
+        "no-unused-expressions": "error",
+        "no-unused-vars": [
+          "error",
+          {
+            "argsIgnorePattern": "^_",
+            "varsIgnorePattern": "^_",
+            "caughtErrors": "none",
+          },
+        ],
+        "@typescript-eslint/no-wrapper-object-types": "error",
```
tatomyr: Is it still valid? I mean the eslint plugin rule?

.github/workflows/tests.yaml
```diff
@@ -45,5 +45,7 @@ jobs:
           cache: npm
       - name: Install dependencies
         run: npm ci
-      - run: npm run eslint
-      - run: npm run prettier:check
+      - name: Lint
+        run: npm run lint
+      - name: Check formatting
+        run: npm run check
```
tatomyr:
```suggestion
        run: npm run format:check
```

package.json
```diff
@@ -53,30 +54,27 @@
...
     "vitest": "^3.0.8"
   },
   "lint-staged": {
     "**/*.{ts,js,yaml,yml,json,md}": [
-      "prettier --write"
+      "oxfmt"
     ],
     "packages/**/*.ts": [
-      "eslint --fix"
+      "oxlint --fix"
```
tatomyr: Why not this?


```suggestion
      "lint:fix"
```

.oxlintrc.jsonc
```diff
@@ -0,0 +1,128 @@
+{
```
vadyvas: I’d like to propose a simplified `oxlint` config for this PR.

### Why simplify?
- Enable `oxc` recommended baseline: turning on the `correctness` category (as `error`) gives us a solid default rule set without manually curating a long allow/deny list.
- ts-first setup: since the majority of the repo is `ts`, it’s cleaner to keep the `ts` config at the root and only add overrides for exceptional cases later (if/when needed).
- less duplication: many explicit rules become redundant once we rely on `categories` + `plugins` (and `extends: ["eslint"]`).
- use JSON instead of JSONC: I don’t think we need comments in the linter config; a strict JSON file is simpler and more tooling-friendly.

### Suggested config
```json
{
  "$schema": "./node_modules/oxlint/configuration_schema.json",
  "extends": ["eslint"],
  "categories": {
    "correctness": "error"
  },
  "env": {
    "builtin": true,
    "node": true,
    "es2026": true,
  },
  "ignorePatterns": [
    "**/__tests__/**",
    "packages/*/lib/**",
    "*.js",
    "*.cjs",
    "packages/respect-core/src/modules/runtime-expressions/abnf-parser.js"
  ],
  "plugins": ["typescript", "import"],
  "rules": {
    "no-thenable": "off",
    "no-undef": "allow",
    "no-prototype-builtins": "allow",
    "no-useless-escape": "deny",
    "no-unused-vars": [
      "deny",
      {
        "argsIgnorePattern": "^_",
        "varsIgnorePattern": "^_",
        "caughtErrors": "none"
      }
    ],
    "typescript/no-explicit-any": "warn",
    "typescript/no-var-requires": "allow",
    "typescript/no-empty-function": "allow",
    "typescript/no-inferrable-types": "allow",
    "typescript/consistent-type-imports": [
      "deny",
      {
        "prefer": "type-imports"
      }
    ],
    "import/no-duplicates": "deny",
    "import/order": [
      "deny",
      {
        "groups": [
          ["builtin", "external", "parent", "sibling", "index", "object"],
          ["type"]
        ],
        "newlines-between": "always"
      }
    ]
  }
}
```

packages/cli/src/auth/device-flow.ts
gameroman: It's a legacy prettier behaviour and this oxc behaviour matches new prettier behaviour, so it [can't be enabled](https://oxc.rs/docs/guide/usage/formatter/unsupported-features)

packages/core/src/config/utils.ts
gameroman: This matches latest prettier version behaviour and is enabled when `"trailingComma": "es5"`

.oxlintrc.jsonc
gameroman: Yeah, that should make sense

gameroman: Like this should be good I think
```json
{
  "$schema": "./node_modules/oxlint/configuration_schema.json",
  "plugins": ["typescript", "import"],
  "categories": {
    "correctness": "error"
  },
  "env": {
    "builtin": true,
    "es2026": true
  },
  "ignorePatterns": ["**/__tests__/**", "packages/*/lib/**", "*.js", "*.cjs"],
  "rules": {
    "no-array-constructor": "error",
    "no-unused-vars": [
      "error",
      {
        "argsIgnorePattern": "^_",
        "varsIgnorePattern": "^_",
        "caughtErrors": "none"
      }
    ],
    "typescript/consistent-type-imports": [
      "error",
      { "prefer": "type-imports" }
    ],
    "import/no-duplicates": "error"
  }
}
```

.oxlintrc.jsonc
gameroman: enabled by default in oxlint

packages/core/src/config/utils.ts
tatomyr: Ok.

packages/cli/src/auth/device-flow.ts
tatomyr: Ok, thanks.

.oxlintrc.jsonc
vadyvas: Yep, that looks good to me.

I did a quick pass over what we get from `categories.correctness` and found a few rules we’re currently missing (or that I think we should keep explicit for parity / consistency). Could we add these as overrides?

```json
"eslint/no-case-declarations": "error"
"eslint/no-empty": "error"
"eslint/no-redeclare": "error"
"eslint/no-regex-spaces": "error"
"eslint/no-unexpected-multiline": "error"
"eslint/no-unreachable": "error"
"eslint/no-array-constructor": "error"

"typescript/ban-ts-comment": "error",
"typescript/no-empty-object-type": "error",
"typescript/no-namespace": "error",
"typescript/no-require-imports": "error",
"typescript/no-unnecessary-type-constraint": "error",
"typescript/no-unsafe-function-type": "error",

"typescript/no-explicit-any": "warn",

"import/order": [
  "error",
  {
    "groups": [
      ["builtin", "external", "parent", "sibling", "index", "object"],
      ["type"]
    ],
    "newlines-between": "always"
  }
]
```

@gameroman With these overrides in place, the config should be complete

gameroman: I don't think `import/order` exists in `oxlint`

tatomyr: I believe we can handle import order separately to speed up this PR. I was thinking about changing that rule anyway.

PR #2546, opened 2026-02-06
chore(deps): remove `ts-loader` package 

kanoru3101:
## What/Why/How?
`ts-loader` is not used in the project, as is `webpack`.
## Reference
close https://github.com/Redocly/redocly-cli/pull/2542
## Check yourself

## Security

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,6 @@
   "license": "MIT",
   "devDependencies": {
     "@changesets/cli": "^2.26.2",
-    "@types/mark.js": "^8.11.5",
-    "@types/marked": "^4.0.3",
     "@types/node": "^22.15.3",
     "@typescript-eslint/eslint-plugin": "^8.54.0",
     "@typescript-eslint/parser": "^8.54.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -50,6 +50,9 @@
     "Roman Hotsiy <roman@redocly.com> (https://redocly.com/)"
   ],
   "license": "MIT",
+  "overrides": {
+    "webpack": "5.105.0"
```
kanoru3101: Wepack uses `ts-loader`, but `ts-loader` has `“webpack”: "*"`, so updating `ts-loader` won't help. I decided to use a override for this. What do you think about it?

package.json
```diff
@@ -69,7 +67,6 @@
     "pegjs": "0.11.0-master.b7b87ea",
     "prettier": "^2.1.2",
     "slackify-markdown": "^4.3.1",
-    "ts-loader": "^8.0.2",
```
DmitryAnansky, line 72: @kanoru3101 do we know why it was added before?

kanoru3101: It was added 5 years ago, when we used webpack for bundling

PR #2551, opened 2026-02-10
feat: add rule to validate query and querystring parameters usage

n0rahh:
## What/Why/How?
A new rule has been added because if any parameters are in `in: query`, you cannot use `in: querystring`, and vice versa.

## Reference
https://github.com/Redocly/redocly/issues/15509

## Check yourself

## Security

### DIFF
--- a/.changeset/wicked-towns-rest.md
+++ b/.changeset/wicked-towns-rest.md
@@ -1,6 +1,0 @@
----
-"@redocly/openapi-core": patch
-"@redocly/cli": patch
----
-
-Fixed false positive errors in example validation for OpenAPI 3.0.x and OpenAPI 2.x.
--- a/docs/@v2/changelog.md
+++ b/docs/@v2/changelog.md
@@ -7,6 +7,13 @@ toc:
 
 <!-- do-not-remove -->
 
+## 2.18.2 (2026-02-16)
+
+### Patch Changes
+
+- Fixed false positive errors in example validation for OpenAPI 3.0.x and OpenAPI 2.x.
+- Updated @redocly/openapi-core to v2.18.2.
+
 ## 2.18.1 (2026-02-13)
 
 ### Patch Changes
--- a/package-lock.json
+++ b/package-lock.json
@@ -4160,9 +4160,9 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/fast-xml-parser": {
-      "version": "4.5.3",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-4.5.3.tgz",
-      "integrity": "sha512-RKihhV+SHsIUGXObeVy9AXiBbFwkVk7Syp8XgwN5U3JV416+Gwp/GO9i0JYKmikykgz/UHRrrV4ROuZEo/T0ig==",
+      "version": "5.3.6",
+      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.3.6.tgz",
+      "integrity": "sha512-QNI3sAvSvaOiaMl8FYU4trnEzCwiRr8XMWgAHzlrWpTSj+QaCSvOf1h82OEP1s4hiAXhnbXSyFWCf4ldZzZRVA==",
       "funding": [
         {
           "type": "github",
@@ -4171,7 +4171,7 @@
       ],
       "license": "MIT",
       "dependencies": {
-        "strnum": "^1.1.1"
+        "strnum": "^2.1.2"
       },
       "bin": {
         "fxparser": "src/cli/cli.js"
@@ -6181,13 +6181,13 @@
       }
     },
     "node_modules/openapi-sampler": {
-      "version": "1.6.2",
-      "resolved": "https://registry.npmjs.org/openapi-sampler/-/openapi-sampler-1.6.2.tgz",
-      "integrity": "sha512-NyKGiFKfSWAZr4srD/5WDhInOWDhfml32h/FKUqLpEwKJt0kG0LGUU0MdyNkKrVGuJnw6DuPWq/sHCwAMpiRxg==",
+      "version": "1.7.0",
+      "resolved": "https://registry.npmjs.org/openapi-sampler/-/openapi-sampler-1.7.0.tgz",
+      "integrity": "sha512-fWq32F5vqGpgRJYIarC/9Y1wC9tKnRDcCOjsDJ7MIcSv2HsE7kNifcXIZ8FVtNStBUWxYrEk/MKqVF0SwZ5gog==",
       "license": "MIT",
       "dependencies": {
         "@types/json-schema": "^7.0.7",
-        "fast-xml-parser": "^4.5.0",
+        "fast-xml-parser": "^5.3.4",
         "json-pointer": "0.6.2"
       }
     },
@@ -7406,9 +7406,9 @@
       }
     },
     "node_modules/strnum": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/strnum/-/strnum-1.1.2.tgz",
-      "integrity": "sha512-vrN+B7DBIoTTZjnPNewwhx6cBA/H+IS7rfW68n7XxC1y7uoiGQBxaKzqucGUgavX15dJgiGztLJ8vxuEzwqBdA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/strnum/-/strnum-2.1.2.tgz",
+      "integrity": "sha512-l63NF9y/cLROq/yqKXSLtcMeeyOfnSQlfMSlzFt/K73oIaD8DGaQWd7Z34X9GPiKqP5rbSh84Hl4bOlLcjiSrQ==",
       "funding": [
         {
           "type": "github",
@@ -8535,15 +8535,15 @@
     },
     "packages/cli": {
       "name": "@redocly/cli",
-      "version": "2.18.1",
+      "version": "2.18.2",
       "license": "MIT",
       "dependencies": {
         "@opentelemetry/exporter-trace-otlp-http": "0.202.0",
         "@opentelemetry/resources": "2.0.1",
         "@opentelemetry/sdk-trace-node": "2.0.1",
         "@opentelemetry/semantic-conventions": "1.34.0",
-        "@redocly/openapi-core": "2.18.1",
-        "@redocly/respect-core": "2.18.1",
+        "@redocly/openapi-core": "2.18.2",
+        "@redocly/respect-core": "2.18.2",
         "abort-controller": "^3.0.0",
         "ajv": "npm:@redocly/ajv@8.17.4",
         "ajv-formats": "^3.0.1",
@@ -8673,7 +8673,7 @@
     },
     "packages/core": {
       "name": "@redocly/openapi-core",
-      "version": "2.18.1",
+      "version": "2.18.2",
       "license": "MIT",
       "dependencies": {
         "@redocly/ajv": "^8.17.4",
@@ -8731,19 +8731,19 @@
     },
     "packages/respect-core": {
       "name": "@redocly/respect-core",
-      "version": "2.18.1",
+      "version": "2.18.2",
       "license": "MIT",
       "dependencies": {
         "@faker-js/faker": "^7.6.0",
         "@noble/hashes": "^1.8.0",
         "@redocly/ajv": "8.17.4",
-        "@redocly/openapi-core": "2.18.1",
+        "@redocly/openapi-core": "2.18.2",
         "ajv": "npm:@redocly/ajv@8.17.4",
         "better-ajv-errors": "^1.2.0",
         "colorette": "^2.0.20",
         "json-pointer": "^0.6.2",
         "jsonpath-rfc9535": "1.3.0",
-        "openapi-sampler": "^1.6.2",
+        "openapi-sampler": "^1.7.0",
         "outdent": "^0.8.0",
         "picomatch": "^4.0.3"
       },
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @redocly/cli
 
+## 2.18.2
+
+### Patch Changes
+
+- Fixed false positive errors in example validation for OpenAPI 3.0.x and OpenAPI 2.x.
+- Updated @redocly/openapi-core to v2.18.2.
+
 ## 2.18.1
 
 ### Patch Changes
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@redocly/cli",
-  "version": "2.18.1",
+  "version": "2.18.2",
   "description": "",
   "license": "MIT",
   "bin": {
@@ -42,8 +42,8 @@
     "@opentelemetry/resources": "2.0.1",
     "@opentelemetry/sdk-trace-node": "2.0.1",
     "@opentelemetry/semantic-conventions": "1.34.0",
-    "@redocly/openapi-core": "2.18.1",
-    "@redocly/respect-core": "2.18.1",
+    "@redocly/openapi-core": "2.18.2",
+    "@redocly/respect-core": "2.18.2",
     "abort-controller": "^3.0.0",
     "ajv": "npm:@redocly/ajv@8.17.4",
     "ajv-formats": "^3.0.1",
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @redocly/openapi-core
 
+## 2.18.2
+
+### Patch Changes
+
+- Fixed false positive errors in example validation for OpenAPI 3.0.x and OpenAPI 2.x.
+
 ## 2.18.1
 
 ### Patch Changes
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@redocly/openapi-core",
-  "version": "2.18.1",
+  "version": "2.18.2",
   "description": "",
   "type": "module",
   "types": "lib/index.d.ts",
--- a/packages/respect-core/CHANGELOG.md
+++ b/packages/respect-core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @redocly/respect-core
 
+## 2.18.2
+
+### Patch Changes
+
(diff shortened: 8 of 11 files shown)


packages/core/src/config/spec.ts
```diff
@@ -220,6 +222,7 @@ const spec: RawGovernanceConfig<'built-in'> = {
     'spec-components-invalid-map-name': 'error',
     'spec-discriminator-defaultMapping': 'error',
     'spec-example-values': 'error',
+    'spec-no-mixed-query-and-querystring-parameters': 'error',
```
tatomyr: Please put it in alphabetical order.

packages/core/src/rules/oas3/spec-no-mixed-query-and-querystring-parameters.ts
```diff
@@ -0,0 +1,72 @@
...
+        });
+      }
+      operationQueryStringLocation ??= parameterLocation;
+    }
+  }
+
+  return {
+    PathItem: {
+      enter(_pathItem, ctx: UserContext) {
+        isOas3_2 = ctx.specVersion === 'oas3_2';
```
tatomyr: You don't have to use this as `detectSpec` already detects the spec flavour and version and separates what rules to run. You can bypass the rule in the rulesets you don't want the rule to be ran. You'll have to do a type tweek though (see how the OAS3.2 `spec-no-invalid-tag-parents` rule is configured).

.changeset/young-ghosts-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added `spec-no-mixed-query-and-querystring-parameters` rule to validate `query` and `querystring` parameter usage in OpenAPI 3.2.
```
tatomyr: Maybe it's also worth checking that it's the only one `querystring` declared on the Operation/PathItem to follow the other part of the spec restriction (https://spec.openapis.org/oas/v3.2.0.html#parameter-locations)? I'd suggest renaming the rule to something more general then, i.e. to `spec-querystring-parameters`.

packages/core/src/rules/oas3/spec-querystring-parameters.ts
```diff
@@ -0,0 +1,72 @@
...
+        operationQueryLocation = pathQueryLocation;
+        operationQueryStringLocation = pathQueryStringLocation;
+
+        const parameterLocation = ctx.parentLocations.PathItem.child(['parameters', ctx.key]);
+        checkParameter(parameter, parameterLocation, ctx);
+
+        pathQueryLocation = operationQueryLocation;
+        pathQueryStringLocation = operationQueryStringLocation;
+      },
+      Operation: {
```
tatomyr, line 78: 1. Will the rule allow one operation containing `query` and another containing `querystring`?
2. Will the rule allow an operation-level parameter redefining the path-level one? Say, you have a path-level parameter `foo` with `querystring` defined, and the operation-level parameter `foo` overriding the path-level one with `query` + other operation-level parameters also using `query`. So, will this trigger the error?

packages/core/src/config/minimal.ts
```diff
@@ -220,6 +222,7 @@ const minimal: RawGovernanceConfig<'built-in'> = {
     'spec-components-invalid-map-name': 'warn',
     'spec-discriminator-defaultMapping': 'off',
     'spec-example-values': 'off',
+    'spec-no-mixed-query-and-querystring-parameters': 'error',
```
DmitryAnansky: Should it be error in minimal config?
Maybe `warn`?
cc: @tatomyr

packages/core/src/rules/oas3/__tests__/spec-no-mixed-query-and-querystring-parameters.test.ts
```diff
@@ -0,0 +1,153 @@
+import { outdent } from 'outdent';
+import { lintDocument } from '../../../lint.js';
+import { parseYamlToDocument, replaceSourceWithRef } from '../../../../__tests__/utils.js';
+import { BaseResolver } from '../../../resolve.js';
+import { createConfig } from '../../../config/index.js';
+
+describe('OAS3 spec-no-mixed-query-and-querystring-parameters (OAS 3.2)', () => {
+  it('reports when query and querystring are mixed within operation parameters', async () => {
```
DmitryAnansky:
```suggestion
  it('should report error when query and querystring are mixed within operation parameters', async () => {
```

packages/core/src/config/minimal.ts
n0rahh: Documentation is clear that this is invalid and can cause incorrect behavior, so I’d leave it as error, what do you think @tatomyr ?

packages/core/src/rules/oas3/spec-querystring-parameters.ts
n0rahh: Yes, I've added one more test and extended current test, to make it clear

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
```diff
@@ -0,0 +1,324 @@
...
+          struct: 'error',
+          'spec-querystring-parameters': 'error',
+        },
+      }),
+    });
+
+    expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
+  });
+
+  it('should report error when an operation-level parameter redefines a path-level parameter (querystring → query)', async () => {
```
tatomyr, line 104: Should not, perhaps? Since the operation itself doesn't contain the querystring parameter (for the path-level get's redefined with the op-level one)?

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
```diff
@@ -0,0 +1,324 @@
...
+  it('should report error when double querystring is used in the same path parameter set', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.2.0
+        info:
+          title: Test
+          version: 1.0.0
+        paths:
+          /test:
+            parameters:
```
tatomyr, line 278: Do you have a test for one querysting in path params and one in operation params?

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
n0rahh: Since `querystring` and `query` are different `in` values, wouldn’t both parameters still apply to the operation?

tatomyr: They both are `foo`. I'm not sure how the spec handles the inheritance of `querystring` and `query`. I was under impression that the operation `query` should override `querystring`, but now I'm not sure. Is it explained in the spec?

n0rahh: From what I can see in the spec, parameters are only considered redefinitions when both `name` and `in` match (see the [Parameter Object](https://spec.openapis.org/oas/v3.2.0.html#parameter-object)), but at the same time OAS 3.2 explicitly says `in: query` and `in: querystring` MUST NOT appear together in the same operation or its path-item (see the [[Parameter Locations](https://spec.openapis.org/oas/v3.2.0.html#parameter-locations)]), so they shouldn’t override each other at all because mixing them is already invalid imo.

packages/core/src/config/minimal.ts
tatomyr: Yes, could be error.

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
tatomyr: Makes sense. Ok. I'll re-review this tomorrow.

packages/core/src/rules/oas3/spec-querystring-parameters.ts
```diff
@@ -0,0 +1,94 @@
...
+      enter(pathItem, ctx: UserContext) {
+        pathState = {};
+        operationState = {};
+        pathQuerystringCount = 0;
+
+        reportIfMultipleQuerystring(countQuerystring(pathItem.parameters || []), ctx);
+      },
+
+      Parameter(parameter: Oas3Parameter, ctx: UserContext) {
+        const location = ctx.parentLocations.PathItem.child(['parameters', ctx.key]);
```
tatomyr: This looks a bit too complicated. Cannot we simply use the `ctx.location` or something similar instead?

packages/core/src/rules/oas3/spec-querystring-parameters.ts
```diff
@@ -0,0 +1,94 @@
...
+type QueryState = {
+  queryLocation?: Location;
+  querystringLocation?: Location;
+};
+
+function countQuerystring(parameters: ReadonlyArray<Referenced<Oas3Parameter>>): number {
+  return parameters.filter((p) => !('$ref' in p) && p.in === 'querystring').length;
+}
+
+const QUERYSTRING_ONCE_MESSAGE =
```
tatomyr: Doesn't make sense to declare a const. Simply use it in place.

packages/core/src/rules/oas3/spec-querystring-parameters.ts
```diff
@@ -0,0 +1,94 @@
...
+  'Parameters with `in: querystring` should be defined only once per path/operation parameter set (OpenAPI 3.2).';
```
tatomyr:
```suggestion
  'Parameters with `in: querystring` should be defined only once per path/operation parameter set.';
```

I don't think it's needed.

.changeset/young-ghosts-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added `spec-querystring-parameters` rule (OpenAPI 3.2): enforces that `query` and `querystring` are not mixed in the same operation/path parameter set, and that at most one `querystring` parameter is declared per operation or path.
```
JLekawa:
```suggestion
Added the `spec-querystring-parameters` rule (OpenAPI 3.2) which enforces that `query` and `querystring` are not mixed in the same operation/path parameter set, and that at most one `querystring` parameter is declared per operation or path.
```

JLekawa:
@tatomyr: should this rule be documented in [Built-in rules](https://redocly.com/docs/cli/rules/built-in-rules)?

tatomyr:
@JLekawa certainly. And not only there but in some other places perhaps too (sidebars, rulesets, the rule's documentation page itself, etc.). @n0rahh please add the corresponding docs as well and mark the corresponding point:

image

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
```diff
@@ -0,0 +1,324 @@
+import { outdent } from 'outdent';
+import { lintDocument } from '../../../lint.js';
+import { parseYamlToDocument, replaceSourceWithRef } from '../../../../__tests__/utils.js';
+import { BaseResolver } from '../../../resolve.js';
+import { createConfig } from '../../../config/index.js';
+
+describe('OAS3 spec-querystring-parameters (OAS 3.2)', () => {
```
DmitryAnansky, line 7: Although `querystring` was introduced in 3.2, this rule also included into other that 3.2 specs, do we need to mention `(OAS 3.2)` in describe?

packages/core/src/config/spec.ts
```diff
@@ -163,6 +164,7 @@ const spec: RawGovernanceConfig<'built-in'> = {
     'security-defined': 'off',
     'spec-components-invalid-map-name': 'error',
     'spec-example-values': 'off',
+    'spec-querystring-parameters': 'off',
```
tatomyr: You don't have to add it here (nor in 3.0). Please have a look at how the `spec-no-invalid-tag-parents` is declared (it's also scoped to the OAS 3.2).

tatomyr:
Please disregard the failing smokes. They are related to a separate issue.

packages/core/src/rules/oas3/__tests__/spec-querystring-parameters.test.ts
tatomyr: No, it won't appear on other spec flavours.

tatomyr: Oh, you mean in the describe's message? Let it remain there for the sake of clarity.

n0rahh: I think now it makes sense to keep it, I did the changes that @tatomyr suggested below

.changeset/young-ghosts-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added the `spec-querystring-parameters` rule (OpenAPI 3.2) which enforces that `query` and `querystring` are not mixed in the same operation/path parameter set, and that at most one `querystring` parameter is declared per operation or path.
```
JLekawa:
```suggestion
Added the `spec-querystring-parameters` rule (OpenAPI 3.2).
This rule enforces that `query` and `querystring` are not mixed in the same operation/path parameter set, and that at most one `querystring` parameter is declared per operation or path.
```

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,112 @@
+---
+slug: /docs/cli/rules/oas/spec-querystring-parameters
+---
+
+# spec-querystring-parameters
+
+Enforce valid use of parameters with `in: querystring` (OpenAPI 3.2).
```
tatomyr:
```suggestion
Enforces valid use of querystring parameters.
```

Please follow the style of other documents.

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+OpenAPI 3.2 introduces the `querystring` parameter location for representing the full query string as a single schema (e.g. `application/x-www-form-urlencoded`). This rule ensures:
+
+1. **At most one querystring parameter** — Parameters with `in: querystring` may be defined only once per path/operation parameter set.
+2. **No mixing with query** — Parameters with `in: query` cannot be used together with `in: querystring` in the same operation/path parameter set.
+
+## Configuration
+
+| Option   | Type   | Description                                                                                       |
+| -------- | ------ | ------------------------------------------------------------------------------------------------- |
+| severity | string | Possible values: `off`, `warn`, `error`. Default `off` (3.0/3.1), `error` in `recommended` (3.2). |
```
tatomyr:
```suggestion
| severity | string | Possible values: `off`, `warn`, `error`. Default `error` (in `recommended` configuration).  |
```

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+| 3.0 | ❌            |
+| 3.1 | ❌            |
+| 3.2 | ✅            |
+
+## API design principles
+
+OpenAPI 3.2 introduces the `querystring` parameter location for representing the full query string as a single schema (e.g. `application/x-www-form-urlencoded`). This rule ensures:
+
+1. **At most one querystring parameter** — Parameters with `in: querystring` may be defined only once per path/operation parameter set.
+2. **No mixing with query** — Parameters with `in: query` cannot be used together with `in: querystring` in the same operation/path parameter set.
```
JLekawa:
```suggestion
OpenAPI 3.2 introduces the `querystring` parameter location for representing the full query string as a single schema (e.g. `application/x-www-form-urlencoded`). 

This rule ensures that:

- There is at most one `querystring` parameter.
  Parameters with `in: querystring` may be defined only once per path/operation parameter set.
- No mixing `querystring` with `query`.
  Parameters with `in: query` cannot be used together with `in: querystring` in the same operation/path parameter set.
```

JLekawa (approved):

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+Example of **incorrect** use (multiple `querystring` parameters):
+
+```yaml
+paths:
+  /search:
+    get:
+      parameters:
+        - name: qs1
```
tatomyr: Coudl you come up with something meaningful? This is the documentation that our users will face, and as a user I'd love to see a well-thought example.

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+                properties:
+                  q: { type: string }
+                  page: { type: integer }
+```
+
+## Related rules
+
+- [path-not-include-query](./path-not-include-query.md)
+- [operation-parameters-unique](./operation-parameters-unique.md)
+- [configurable rules](../configurable-rules.md)
```
tatomyr:
```suggestion
- [path-not-include-query](./path-not-include-query.md) # -- not sure about this one
- [operation-parameters-unique](./operation-parameters-unique.md)
```

docs/@v2/rules/built-in-rules.md
```diff
@@ -37,6 +37,7 @@ The rules list is split into sections.
 - [security-defined](./oas/security-defined.md): Security rules must be defined, either globally or per-operation
 - [struct](./common/struct.md): Conform to the declared OpenAPI specification version
 - [spec-components-invalid-map-name](./oas/spec-components-invalid-map-name.md): Use only alphanumeric and basic punctuation as key names in the components section
+- [spec-querystring-parameters](./oas/spec-querystring-parameters.md): Enforce valid use of `in: querystring` (OpenAPI 3.2): at most one per path/operation, and not mixed with `in: query`
```
tatomyr, line 40: Please also check if we need to add the rule reference to a sidebar.

tatomyr:
Left comments mostly related to the docs.
cc @JLekawa

docs/@v2/rules/oas/spec-querystring-parameters.md
```diff
@@ -0,0 +1,131 @@
+---
+slug: /docs/cli/rules/oas/spec-querystring-parameters
+---
+
+# spec-querystring-parameters
+
+Enforces valid use of querystring parameters.
```
JLekawa:
```suggestion
Enforces valid use of `querystring` parameters.
```

PR #2555, opened 2026-02-10
feat: add filter-operations decorator

tatomyr:
## What/Why/How?

- Added `filter-operations` decorator that allows filtering operation which should be included in the resulting bundle.
- Adjusted configuration for the `remove-unused-components` decorator to always be executed last.
## Reference

Closes https://github.com/Redocly/redocly-cli/issues/2464

Related contribution: https://github.com/Redocly/redocly-cli/pull/2536

## Check yourself

- [ ] Code changed? - Tested with Redoc/Realm/Reunite (internal)
- [ ] All new/updated code is covered by tests
- [ ] New package installed? - Tested in different environments (browser/node)
- [ ] Documentation update considered

## Security

- [ ] The security impact of the change has been considered
- [ ] Code follows company security practices and guidelines

PR #2570, opened 2026-02-17
feat: add the applyTo option to the filter-in and filter-out decorators

tatomyr:
## What/Why/How?

Added new options to the `filter-in` and `filter-out` decorators, allowing to filter through a specific target nodes (`PathItem` or `Operation`).

This solves an issue where users want to filter in/out specific nodes (i.e. Operations) and leave other nodes that might contain the `property` untouched (for instance, in Schemas). However, with the current implementation of filter-in/out we cannot scope where to look by nodes, only by specifying the `property`, so the decorator walks through all nodes looking for the specified property.

~**A note for reviewers:** I'm still not convinced that we need the `noPropertyStrategy` option in `filter-in`. For me it seems natural that we don't keep a node that doesn't match the specified property when filtering in.~

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/2464

Related contribution: https://github.com/Redocly/redocly-cli/pull/2536

Alters https://github.com/Redocly/redocly-cli/pull/2555.

## Check yourself

## Security

### DIFF
--- a/docs/@v2/decorators/filter-in.md
+++ b/docs/@v2/decorators/filter-in.md
@@ -20,7 +20,7 @@ Giant monolithic API docs can be overwhelming. By filtering what is most relevan
 
 ## Examples
 
-### Filter operations by operationId
+### Filter operations by `operationId`
 
 Using the [Museum API](https://github.com/Redocly/museum-openapi-example) (v1.0.0), use the stats command to get a summary of its contents:
 


packages/core/src/decorators/common/filters/filter-in.ts
```diff
@@ -1,11 +1,89 @@
...
+export const FilterIn: Oas3Decorator | Oas2Decorator = ({
+  property,
+  value,
+  matchStrategy,
+  target,
+  noPropertyStrategy,
+}) => {
   const strategy = matchStrategy || DEFAULT_STRATEGY;
+
+  if (target !== undefined) {
```
tatomyr: Added a separate implemetation for the `target` mode to avoid clashing with the previous implementation.

docs/@v2/custom-plugins/custom-decorators.md
```diff
@@ -163,6 +163,8 @@ Here are the key points to understand:
   1. The order of plugins as listed in the `plugins` array in `redocly.yaml` configuration file.
   1. The order of decorators as defined in the `decorators` object of each plugin.
 
+Note, that the built-in decorators are considered to be part of a special default plugin which is always executed last.
```
JLekawa:
```suggestion
Note that the built-in decorators are considered to be part of a special default plugin which is always executed last.
```

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,33 @@
...
+| Option             | Type     | Description                                                                                                                                                             |
+| ------------------ | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| property           | string   | **REQUIRED.** The property name used for evaluation. It attempts to match the values.                                                                                   |
+| value              | [string] | **REQUIRED.** List of values used for the matching.                                                                                                                     |
+| matchStrategy      | string   | Possible values: `all`, `any`. If `all` it needs to match all of the values supplied. If `any` it needs to match only one of the values supplied. Default value: `any`. |
```
JLekawa:
```suggestion
| matchStrategy      | string   | Possible values: `all`, `any`. When `all`, must match all of the values supplied. When `any`, must match only one of the values supplied. Default value: `any`. |
```

.changeset/social-groups-switch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added a new option to the `filter-out` decorator, allowing to filter through a specific target nodes (`PathItem` or `Operation`).
```
JLekawa:
```suggestion
Added a new option to the `filter-out` decorator, allowing to filter through specific target nodes (`PathItem` or `Operation`).
```

.changeset/clear-apples-trade.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": minor
+"@redocly/cli": minor
+---
+
+Added new options to the `filter-in` decorator, allowing to filter through a specific target nodes (`PathItem` or `Operation`).
```
JLekawa:
```suggestion
Added new options to the `filter-in` decorator, allowing to filter through specific target nodes (`PathItem` or `Operation`).
```

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,33 @@
...
-| property      | string   | **REQUIRED.** The property name used for evaluation. It attempts to match the values.                                                                                   |
-| value         | [string] | **REQUIRED.** List of values used for the matching.                                                                                                                     |
-| matchStrategy | string   | Possible values: `all`, `any`. If `all` it needs to match all of the values supplied. If `any` it needs to match only one of the values supplied. Default value: `any`. |
+| Option             | Type     | Description                                                                                                                                                     |
+| ------------------ | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| property           | string   | **REQUIRED.** The property name used for evaluation. It attempts to match the values.                                                                           |
```
JLekawa:
```suggestion
| property           | string   | **REQUIRED.** The property name used for evaluation. Attempts to match the values.                                                                           |
```

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,33 @@
...
+| target             | string   | Possible values: `PathItem`, `Operation`. When set, filtering is scoped to the specified target.                                                                |
+| noPropertyStrategy | string   | Possible values: `keep`, `remove` (default value: `keep`). Decides whether to keep nodes without the specified property. Useful with `target`.                  |
+
+### Explicit vs. implicit target behavior
+
+When `target` is explicitly set, the decorator walks through all nodes of that type and keeps those where the `property` matches the specified values.
+Target nodes without the specified property are either kept (when `noPropertyStrategy` is set to `keep`) or removed (when `noPropertyStrategy` is set to `remove`).
+
+If there's no explicit `target`, the decorator evaluates every node in the API description that has the `property` and keeps those where the property matches the specified values.
+Nodes without the property are left untouched.
```
JLekawa:
```suggestion
Nodes without the property are left unchanged.
```

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,33 @@
...
+| ------------------ | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| property           | string   | **REQUIRED.** The property name used for evaluation. Attempts to match the values.                                                                           |
+| value              | [string] | **REQUIRED.** List of values used for the matching.                                                                                                             |
+| matchStrategy      | string   | Possible values: `all`, `any`. When `all`, must match all of the values supplied. When `any`, must match only one of the values supplied. Default value: `any`. |
+| target             | string   | Possible values: `PathItem`, `Operation`. When set, filtering is scoped to the specified target.                                                                |
+| noPropertyStrategy | string   | Possible values: `keep`, `remove` (default value: `keep`). Decides whether to keep nodes without the specified property. Useful with `target`.                  |
```
tatomyr: This option still looks wrong to me (at least the default value).

To me, removing the properties that don't match with the value, aligns better with the filter-out logic. What do you think?

adamaltman: I don't understand this property name. I think that "filter-in" means keep and "filter-out" means remove. So how can filter-in also have remove?

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,33 @@
...
+
+### Explicit vs. implicit target behavior
+
+When `target` is explicitly set, the decorator walks through all nodes of that type and keeps those where the `property` matches the specified values.
+Target nodes without the specified property are either kept (when `noPropertyStrategy` is set to `keep`) or removed (when `noPropertyStrategy` is set to `remove`).
+
+If there's no explicit `target`, the decorator evaluates every node in the API description that has the `property` and keeps those where the property matches the specified values.
+Nodes without the property are left unchanged.
```
adamaltman: Why do we need `noPropertyStrategy`?  Isn't the decorator name strategy enough?

docs/@v2/decorators/filter-in.md
RomanHotsiy: The filter in applies only to nodes with the `property` defined. All the other properties are skipped at all.

So if you do `filter-in` by `x-public: true` in the following case:


```
/path:
  get:
    x-public: true
    # ...
  post:
    x-public: false
    # ...
  delete:
    # ....
```

It will keep the `get` and `delete`.

docs/@v2/decorators/filter-in.md
tatomyr: @adamaltman We discussed that offline with @RomanHotsiy  and decided that it's needed because of the ambiguity when there's no `property` defined on the node: if you don't specify target, it won't remove those nodes without the property, so the users might get confused with the changed behaviour when using `target`. 
But after another look (especially after comparing it with filter-out), it also seems needless to me. That's why I'm asking if it wouldn't be better to remove.

docs/@v2/decorators/filter-in.md
RomanHotsiy: That's why we need to have a option which answers the question what to do with those that don't have property defined.

We can't change default behavior because filter-in will remove everything without a tag (schemas, properties, everything deep).

And I don't want to change behavior based on the presence of `target`. It would be confusing I think.
If only we come up with a better name for `target` that implies both targeting and change in remove behavior.

tatomyr: It's because of how the original implementation works. Really, it's not filtering-in nodes where the `property` matches, but removes nodes where the `property` is declared and doesn't match the `value`. 
The `target` is actually here to allow really filter in, but without changing the existing implementation.
I didn't find a better name, and it also sounds confusing to me.

tatomyr: > come up with a better name for target that implies both targeting and change in remove behavior.

I can only think of something like `requireOn` or `exclusiveTarget` or `strictTarget`:

```yaml
decorators:
  filter-in:
    property: x-public
    value: true
    strictTarget: Operation
```

adamaltman: > it's not filtering-in nodes where the property matches, but removes nodes where the property is declared and doesn't match the value.

This sentence would have made the docs so much clearer.

--- 

What if someone had multiple targets like they put it at both the operation and tag declaration levels? Should it be a string or a list of strings? 

What if we don't really require a subject type and have something like this:

```
filterIn:
  property: x-audience
  value: public
  requireProperty: true
```

For docs:

> `filter-in` applies to any node type where the property is declared (for example Operations, schema properties, or Tags). Nodes are removed only when the property exists and does not match the specified **value**. Nodes missing the **property** are kept unless `requireProperty: true` is set.

tatomyr: > What if someone had multiple targets like they put it at both the operation and tag declaration levels? Should it be a string or a list of strings?

Currently, we don't support multiple targets (if it's really needed, I'd suggest applying the decorator multiple times sequentially). 

> What if we don't really require a subject type

You mean the target node? But the whole point is to allow targeting a specific node, because we already can target multiple nodes, and people were asking to scope the filtering to a specific target.

> requireProperty: true

@adamaltman, do you suggest replacing `noPropertyStrategy` with `requireProperty` and apply it to the current behavior as well?

tatomyr: Would this work?

Image

adamaltman: Does that mean the default value for target is `any`?

tatomyr: @adamaltman technically there's no default value. But we can potentially treat it as such.

RomanHotsiy: > ```
> filterIn:
>   property: x-audience
>   value: public
>   requireProperty: true
> ```

@adamaltman  this will technically remove everything where x-audience is not defined so you have to have it **everywhere**: in info section, in each tag, in each operation, in each schema, in each schema property.

adamaltman: That sounds good. But I think target needs to support multiple values, not a single value in that case?

tatomyr: @adamaltman I'd start with a single target value for simplicity, and we can enhance that later if we receive a request. All usage of filter-in that I'm aware of tries to target the operation level only (except for Rebilly).

tatomyr: Updated the docs and the code. Please take a look.

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,27 @@
...
+| --------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| property        | string   | **REQUIRED.** The property name used for evaluation. Attempts to match the values.                                                                              |
+| value           | [string] | **REQUIRED.** List of values used for the matching.                                                                                                             |
+| matchStrategy   | string   | Possible values: `all`, `any`. When `all`, must match all of the values supplied. When `any`, must match only one of the values supplied. Default value: `any`. |
+| target          | string   | Possible values: `PathItem`, `Operation`. When set, filtering is scoped to the specified target.                                                                |
+| requireProperty | boolean  | When set to `true`, nodes without the specified property are removed. Default value: `false`.                                                                   |
```
Marshevskyy: When requireProperty is true, target is required, right?

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,27 @@
 # filter-in
 
-Preserves nodes that have specific `property` set to the specific `value` and removes others. Nodes that don't have the `property` defined are not impacted.
+Applies to any node type where the property is declared (for example Operations, schema properties, or Tags).
```
Marshevskyy: how about adding `When target is set, filtering is limited to that node type.`

docs/@v2/decorators/filter-in.md
Marshevskyy: LGTM overall
1. filter-out has `target` but not `requireProperty`. Should we add it for consistency?
2. Since `target` and `requireProperty` are coupled, maybe we should consider slightly different naming? For example, `filterInNodeTypes` or `applyTo`. Or do we already have some names in the existing config that would fit here?

docs/@v2/decorators/filter-in.md
tatomyr: `target` is required for `requireProperty` to take effect (we can change that, and allow stripping all nodes without the `property` even without `target` being set, but it hardly makes any sense as it will efectively remove all the description unless we have the `property` declared on the root).

docs/@v2/decorators/filter-in.md
tatomyr: Good catch. Adjusting the wording.

docs/@v2/decorators/filter-in.md
tatomyr: @Marshevskyy 

1. No because filter-out naturally can work only one way.
2. I actually like `applyTo`, but would like to here other thoughts.

docs/@v2/decorators/filter-in.md
Marshevskyy: what would happen now if you add requireProperty without target, will cli fail?

tatomyr: No, for now it'll just ignore it and work as before.

tatomyr: At least, it shouldn't fail with the current implementation.

adamaltman: I don't like having conditionally require properties. I would veto this change until that is adjusted.

docs/@v2/decorators/filter-in.md
adamaltman: I don't think that we should have conditionally required properties. This is a design problem.

tatomyr: @adamaltman ok, I'll make `requireProperty` to work without `target` too.

tatomyr: Updated the code and added the test for [removing nodes missing the property without target explicitly set when requireProperty is true](https://github.com/Redocly/redocly-cli/pull/2570/changes#diff-c3657e57a9972353c42158a3f79b87b6bda3e6160107dcc0fbe18880bbc2e5f5R259-R260).

Basically, it requires the `x-audience` property being set on every level to be preserved:

```yaml
openapi: 3.0.0
paths:
  x-audience: Public
  /keep:
    x-audience: Public
    get:
      x-audience: Public
      summary: Should be kept
      responses: # will be removed
        '200':
          description: Will be removed because it's missing the x-audience property
  /remove: # will be removed
    get:
      x-audience: Public
      summary: Will be removed because parent node is missing the x-audience property
```

This behaviour hardly makes any sense to me, so that's why I was doubting if we need the additional `requireProperty`  option at all in filter-in only as a complementary one to `target` (especially taking into account that filter-out doesn't need it at all).

Could we maybe come up with a better name for `target` so it will clearly reflect that the target nodes without the `property` will be removed and drop that option (as Roman suggested [above](https://github.com/Redocly/redocly-cli/pull/2570#discussion_r2841480339))?

tatomyr: I tried to visualize how I see the configuration. We have two possible scenarios: with the `requiredProperty` and without:



image


From a user's perspective, having an additional option for filter-in makes it orthogonal to filter-out. The options I've marked in red are illogical to me, because users won't expect that filter-in leaves the nodes without the `property` by default, nor they might want to strip off all the nodes without the `property` when the target is not defined. 

Thus I vote for removing `requiredProperty` and clarifying the difference between using and not using the `target` option in the docs. This will leave the current behaviour as it is, and allow users   to scope the filtering to certain nodes with the same option for filter-in and filter-out.

tatomyr: Discussed with @RomanHotsiy  and decided to go without the additional property in filter-in (to be on par with filter-out). @adamaltman hopefully, this should resolve your [vetoing comment](https://github.com/Redocly/redocly-cli/pull/2570#discussion_r2879312079). 

Also, I renamed `target` -> `applyTo` to be more specific with the intention of the option, but I'm not entirely convinced it sounds better this way. You can check how the wording changes in [this commit](https://github.com/Redocly/redocly-cli/pull/2570/changes/67a301fd48d9e18107b733746daececdaac63b06?show-viewed-files=true&file-filters%5B%5D=.md).

docs/@v2/decorators/filter-in.md
```diff
@@ -1,21 +1,27 @@
...
+| ------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| property      | string   | **REQUIRED.** The property name used for evaluation. Attempts to match the values.                                                                              |
+| value         | [string] | **REQUIRED.** List of values used for the matching.                                                                                                             |
+| matchStrategy | string   | Possible values: `all`, `any`. When `all`, must match all of the values supplied. When `any`, must match only one of the values supplied. Default value: `any`. |
+| applyTo       | string   | Possible values: `PathItem`, `Operation`. When set, filtering is scoped to the specified target.                                                                |
 
 ## Examples
 
+### Filter operations by operationId
```
adamaltman:
```suggestion
### Filter operations by `operationId`
```

PR #2571, opened 2026-02-17
fix: issue assigning list response to the output

DmitryAnansky:
## What/Why/How?

- [x] fix: issue assigning object list response to the output
```yaml
outputs:
  products: $response.body#/items
```

- [x] Change AJV setting for Respect schema check
```
validateFormats: true
```

## Testing

tested internally => https://github.com/Redocly/redocly/actions/runs/22134093462/job/63981335759?pr=21051 

## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/modules/__tests__/runtime-expressions/evaluate.test.ts
+++ b/packages/respect-core/src/modules/__tests__/runtime-expressions/evaluate.test.ts
@@ -96,6 +96,16 @@ const runtimeExpressionContext = {
                 'Join us as we review and classify a rare collection of 20 thingamabobs, gadgets, gizmos, whoosits, and whatsits, kindly donated by Ariel.',
               dates: ['2023-12-15', '2023-12-22'],
               price: 0,
+              items: [
+                {
+                  name: 'item1',
+                  description: 'item1 description',
+                },
+                {
+                  name: 'item2',
+                  description: 'item2 description',
+                },
+              ],
             },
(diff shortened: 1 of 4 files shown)


packages/respect-core/src/modules/__tests__/runtime-expressions/evaluate.test.ts
```diff
@@ -639,4 +667,18 @@ describe('evaluateRuntimeExpression', () => {
     expect(evaluateRuntimeExpression(expression2, runtimeExpressionContext, logger)).toEqual(true);
     expect(evaluateRuntimeExpression(expression3, runtimeExpressionContext, logger)).toEqual(false);
   });
+
+  it('should evaluate list runtime expression value', () => {
+    const expression = '$response.body#/items';
```
vadyvas, line 672: can you also add test for `{$response.body#/items}` to check:
> A runtime expression wrapper has the form {$variable...}

DmitryAnansky: The use of curly braces `{$smth}` indicates that the value will be stringified.

Imho this does not make sense for a list of objects, as it would prevent proper usage in other runtime expressions. I have checked the [spec](https://spec.openapis.org/arazzo/latest.html) and it seems like this case is not described well.

I will confirm this scenario with the Arazzo community and address it separately in a follow-up PR.

PR #2575, opened 2026-02-18
chore: fix flaky smoke test

AlbinaBlazhko17:
## What/Why/How?
Fixed flaky tests (yarn 20 + windows and node-yarn-24)

Tried different approach, but locally the issue is still reproducible: https://github.com/Redocly/redocly/pull/21083

## Reference
https://github.com/Redocly/redocly-cli/actions/runs/21981360702/job/63504577156?pr=2555
https://github.com/Redocly/redocly-cli/actions/runs/22145856147/job/64022774477

## Check yourself

## Security

### DIFF
--- a/.changeset/young-ghosts-drop.md
+++ b/.changeset/young-ghosts-drop.md
@@ -1,7 +1,0 @@
----
-"@redocly/openapi-core": minor
-"@redocly/cli": minor
----
-
-Added the `spec-querystring-parameters` rule (OpenAPI 3.2).
-This rule enforces that `query` and `querystring` are not mixed in the same operation/path parameter set, and that at most one `querystring` parameter is declared per operation or path.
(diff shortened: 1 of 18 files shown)


packages/core/src/utils/scorecards.ts
```diff
@@ -1,4 +1,8 @@
-import { NODE_TYPE_NAMES } from '@redocly/config';
+// Fix issue on yarn and windows with dynamic import of @redocly/config
```
vadyvas, line 1: not sure if it's worth leaving a comment here, but the choice is yours

AlbinaBlazhko17: I will leave it, to avoid any misunderstandings, as these functionalities may be changed by other teams.

PR #2579, opened 2026-02-18
fix: respect display issue for multipart/form-data encoded requestBody

DmitryAnansky:
## What/Why/How?
Fixed an issue in Respect where only the last value for repeated keys in `multipart/form-data` request bodies was shown.
Arazzo example:
arazzo-example
Before:
Before
After (slightly different Arazzo file version):
After

## Check yourself

## Security

### DIFF
--- a/packages/respect-core/src/modules/logger-output/display-checks.ts
+++ b/packages/respect-core/src/modules/logger-output/display-checks.ts
@@ -90,7 +90,7 @@ function displayVerboseLogs({
     const formDataObject: Record<string, any> = {};
     for (const [key, value] of body.entries()) {
       const displayValue = value instanceof File ? `[File: ${value.name}]` : value;
-      // FormData allows multiple values per key (e.g., for lists), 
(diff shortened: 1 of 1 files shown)


packages/respect-core/src/modules/logger-output/display-checks.ts
```diff
@@ -89,10 +89,14 @@ function displayVerboseLogs({
     // Convert FormData to a simple object for display
     const formDataObject: Record<string, any> = {};
     for (const [key, value] of body.entries()) {
-      if (value instanceof File) {
-        formDataObject[key] = `[File: ${value.name}]`;
+      const displayValue = value instanceof File ? `[File: ${value.name}]` : value;
+      if (key in formDataObject) {
+        const existing = formDataObject[key];
+        formDataObject[key] = Array.isArray(existing)
```
tatomyr, line 97: Please leave a short comment explaining this behaviour.

PR #2597, opened 2026-02-23
chore: update analytics data format

kanoru3101:
## What/Why/How?
This PR updates usage analytics to align with the new data format.

Changes include:
- [x] adopting new data types
- [x] using the updated mapper logic
- [x] update dasboards

## Reference
Internal issue
## Check yourself

## Security

### DIFF
--- a/.changeset/witty-taxes-serve.md
+++ b/.changeset/witty-taxes-serve.md
@@ -1,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed an issue where invalid discriminator mapping values could cause linting to fail.
--- a/packages/core/src/__tests__/ref-utils.test.ts
+++ b/packages/core/src/__tests__/ref-utils.test.ts
@@ -9,6 +9,7 @@ import {
   refBaseName,
   unescapePointerFragment,
   isAbsoluteUrl,
+  isMappingRef,
   getDir,
   resolvePath,
 } from '../ref-utils.js';
@@ -205,6 +206,17 @@ describe('ref-utils', () => {
       expect(isAbsoluteUrl('./api.yaml')).toBe(false);
       expect(isAbsoluteUrl('../api.yaml')).toBe(false);
       expect(isAbsoluteUrl('/Users/test/api.yaml')).toBe(false);
+    });
+  });
+
+  describe('isMappingRef', () => {
+    it('should return true for a valid mapping ref', () => {
+      expect(isMappingRef('#/components/schemas/Foo')).toBe(true);
+      expect(isMappingRef('./test.yaml')).toBe(true);
+    });
+
+    it('should return false if the ref is not a string', () => {
+      expect(isMappingRef(null as any)).toBe(false);
     });
   });
 
--- a/packages/core/src/ref-utils.ts
+++ b/packages/core/src/ref-utils.ts
@@ -114,12 +114,13 @@ export function resolvePath(base: string, relative: string): string {
 export function isMappingRef(mapping: string) {
   // TODO: proper detection of mapping refs
   return (
-    mapping.startsWith('#') ||
-    mapping.startsWith('https://') ||
-    mapping.startsWith('http://') ||
-    mapping.startsWith('./') ||
-    mapping.startsWith('../') ||
-    mapping.indexOf('/') > -1
+    typeof mapping === 'string' &&
+    (mapping.startsWith('#') ||
+      mapping.startsWith('https://') ||
+      mapping.startsWith('http://') ||
+      mapping.startsWith('./') ||
+      mapping.startsWith('../') ||
+      mapping.indexOf('/') > -1)
   );
 }
 


packages/cli/src/utils/otel-attributes.ts
```diff
@@ -0,0 +1,111 @@
+import type { Attributes } from '@opentelemetry/api';
```
kanoru3101: Copied part of the logic from the mapper in the SDK

packages/cli/src/utils/otel.ts
```diff
@@ -30,39 +31,16 @@ export class OtelServerTelemetry {
...
-    };
+    const spanName = `event.${cloudEvent.type}`;
+    const attributes = processCloudEventAttributes(cloudEvent, time);
 
-    for (const [key, value] of Object.entries(cloudEvent.data)) {
-      const keySnakeCase = key.replace(/([A-Z])/g, '_$1').toLowerCase();
-      if (value !== undefined) {
-        attributes[`cloudevents.event_data.${keySnakeCase}`] = value;
-      }
+    if (!process.env.REDOCLY_ENVIRONMENT || process.env.REDOCLY_ENVIRONMENT == 'local') {
```
kanoru3101: Is it possible that `REDOCLY_ENVIRONMENT` may not exist outside of local development?

SVell:
I really do not like this solution with chuck of mapper's code just inserted here. I think we should look for a solution to reuse the mapper here...

packages/cli/src/utils/otel.ts
tatomyr: Absolutely.

kanoru3101: Removed local logs

SVell:
https://github.com/Redocly/redocly-cli/pull/2606

packages/cli/src/utils/__tests__/otel.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      command: 'lint',
+      exit_code: 0,
+      execution_time: 1500,
+    },
+  ],
+};
+
+describe('OtelServerTelemetry.send', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
```
tatomyr: Why do you need to clear mocks? Don't the default setup work?

packages/cli/src/utils/__tests__/otel.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    mockSpanEnd.mockReturnValue(undefined);
+    mockStartSpan.mockReturnValue({ end: mockSpanEnd });
+    mockGetTracer.mockReturnValue({ startSpan: mockStartSpan });
+    mockProcessCloudEventAttributes.mockReturnValue({
+      'cloudevents.event_id': 'evt_test_001',
+    });
+  });
+
+  it('should process cloud event into OTEL span', () => {
```
tatomyr: This test has more mocks than the actual testing. I don't think we need it in this form.

packages/cli/src/utils/telemetry.ts
```diff
@@ -64,55 +69,46 @@ export async function sendTelemetry({
...
+        environment: process.env.REDOCLY_ENVIRONMENT,
+        has_config: typeof config?.document?.parsed === 'undefined' ? 'no' : 'yes',
+        spec_version,
+        spec_keyword,
+        spec_full_version,
+        respect_x_security_auth_types:
+          spec_version === 'arazzo1' && respect_x_security_auth_types?.length
+            ? JSON.stringify(respect_x_security_auth_types)
+            : undefined,
+      } as EventPayload<EventType>[0],
```
tatomyr: Why are you casting the type? Is the initial type incorrect here?

packages/cli/src/utils/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,107 @@
+import * as process from 'node:process';
```
tatomyr: Again, this test looks like you're testing a library, not the actual implementation. I don't see the purpose of these tests.

packages/cli/src/utils/telemetry.ts
kanoru3101: Yes, we missed add `environment`to scheme. Is it better to update `cli-otel` now and remove this casting type?

tatomyr: Yes. But this time please test the package in CLI before releasing it.

packages/cli/src/utils/__tests__/telemetry.test.ts
kanoru3101: Removed it and used `vi.stubEnv` for set up envs

vitest.config.ts
```diff
@@ -47,6 +47,7 @@ export default mergeConfig(
       globals: true,
       restoreMocks: true,
       mockReset: true,
+      unstubEnvs: true,
```
kanoru3101: Allows you to add environments using `vi.stubEnv()`

packages/cli/src/utils/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,36 @@
...
+vi.mock('../otel.js', () => ({ otelTelemetry: { send: mockOtelSend } }));
+vi.mock('../network-check.js');
+vi.mock('../../auth/oauth-client.js');
+vi.mock('../../reunite/api/index.js');
+
+it('sendTelemetry calls all telemetry functions', async () => {
+  vi.mocked(respondWithinMs).mockResolvedValue(true);
+
+  await sendTelemetry({
+    config: { document: { parsed: {} } } as any,
```
kanoru3101: The `any` type was passed to avoid passing all config and command props

tatomyr: You don't have to mock config, just use `createConfig({})` to create an empty one like in other tests.

kanoru3101: `createConfig` was used, but it seems that `any` should be left for `argv`

PR #2598, opened 2026-02-23
chore: update vitest and @vitest/coverage-istanbul dependencies

vadyvas:
## What/Why/How?

Upgrade `vitest` and `@vitest/coverage-istanbul` to `4.0.18` and fix tests that broke after the upgrade.

Tests relying on previous behavior started failing:

* date spies no longer behave the same way (switched to fake timers instead of spying on `Date`).
* mock call counts are no longer reset automatically between tests (added `vi.clearAllMocks()` / `vi.resetAllMocks()` in `beforeEach` where tests assert call counts, so counts don’t leak between tests).
* constructor mocks must use real functions/classes (replaced arrow functions with function/class declarations).

## Reference

- https://github.com/Redocly/redocly-cli/pull/2584
- https://vitest.dev/guide/migration.html 

## Check yourself

## Security

### DIFF
--- a/packages/cli/src/auth/__tests__/oauth-client.test.ts
+++ b/packages/cli/src/auth/__tests__/oauth-client.test.ts
@@ -30,8 +30,6 @@ describe('RedoclyOAuthClient', () => {
         return mockDeviceFlow;
       });
 
-      console.log(typeof RedoclyOAuthDeviceFlow);
-
       await client.login(mockBaseUrl);
 
       expect(mockDeviceFlow.run).toHaveBeenCalled();


packages/cli/src/__tests__/wrapper.test.ts
```diff
@@ -10,6 +10,7 @@ const originalFetch = global.fetch;
 
 describe('commandWrapper', () => {
   beforeEach(() => {
+    vi.clearAllMocks();
```
tatomyr: Doesn't `resetMocks: true` in the config already do the same?

vadyvas: updated the config and reverted the changes

packages/cli/src/auth/__tests__/oauth-client.test.ts
```diff
@@ -26,7 +26,9 @@ describe('RedoclyOAuthClient', () => {
       const mockDeviceFlow = {
         run: vi.fn().mockResolvedValue(mockToken),
       };
-      vi.mocked(RedoclyOAuthDeviceFlow).mockImplementation(() => mockDeviceFlow as any);
+      vi.mocked(RedoclyOAuthDeviceFlow).mockImplementation(function () {
```
tatomyr, line 29: What's the reason of changing this?

packages/cli/src/auth/__tests__/oauth-client.test.ts
```diff
@@ -26,7 +26,9 @@ describe('RedoclyOAuthClient', () => {
...
+        return mockDeviceFlow as any;
```
tatomyr: Is there still a need for casting this?

packages/cli/src/auth/__tests__/oauth-client.test.ts
vadyvas: Image

https://vitest.dev/api/vi#vi-spyon

packages/cli/src/auth/__tests__/oauth-client.test.ts
vadyvas: removed✅

packages/cli/src/auth/__tests__/oauth-client.test.ts
tatomyr: Aha. Okay.

tatomyr (approved):

PR #2608, opened 2026-02-26
chore: update stats command documentation about async support

DmitryAnansky:
## What/Why/How?

Update `stats` command documentation about async support.

## Check yourself

## Security

### DIFF
--- a/docs/@v2/commands/stats.md
+++ b/docs/@v2/commands/stats.md
@@ -191,40 +191,7 @@ The following is an example JSON output for an OpenAPI description:
 }
 </pre>
 
-For AsyncAPI descriptions, the output uses `channels` instead of `links`, `pathItems`, and `webhooks`:
-
-<pre>
-{
-  "refs": {
-    "metric": "🚗 References",
-    "total": 2
-  },
-  "externalDocs": {
-    "metric": "📦 External Documents",
-    "total": 1
-  },
-  "schemas": {
-    "metric": "📈 Schemas",
-    "total": 1
-  },
-  "parameters": {
-    "metric": "👉 Parameters",
-    "total": 0
-  },
-  "channels": {
-    "metric": "📡 Channels",
-    "total": 1
-  },
-  "operations": {
-    "metric": "👷 Operations",
-    "total": 1
-  },
-  "tags": {
-    "metric": "🔖 Tags",
-    "total": 2
-  }
-}
-</pre>
+For AsyncAPI descriptions, the output uses `channels` instead of `links`, `pathItems`, and `webhooks`.
 
 The JSON format output is suitable when you want to use the stats data in another program.
 


packages/cli/src/commands/stats.ts
```diff
@@ -115,23 +93,42 @@ export async function handleStats({ argv, config, collectSpecData }: CommandArgs
...
-  const statsAccumulatorAsync2 = createAsyncAPIStatsAccumulator();
-  const statsAccumulatorAsync3 = createAsyncAPIStatsAccumulator();
-
-  const statsVisitor =
-    specVersion === 'async2'
-      ? StatsAsync2(statsAccumulatorAsync2)
-      : specVersion === 'async3'
-        ? StatsAsync3(statsAccumulatorAsync3)
-        : StatsOAS(statsAccumulatorOAS);
+  const statsAccumulatorOAS: OASStatsAccumulator = {
```
vadyvas: how about creating a factory to avoid duplication?

DmitryAnansky: I change it actually to switch case to make it more obvious witch one is used.
What do you think?

vadyvas (approved):

docs/@v2/commands/stats.md
```diff
@@ -150,14 +191,49 @@ The JSON format output is shown in the following example:
...
+  "operations": {
+    "metric": "👷 Operations",
+    "total": 1
+  },
+  "tags": {
+    "metric": "🔖 Tags",
+    "total": 2
+  }
+}
+</pre>
```
tatomyr:
```suggestion
For AsyncAPI descriptions, the output uses `channels` instead of `links`, `pathItems`, and `webhooks`.
```

No need for example of each spec output, I believe.

packages/cli/src/commands/stats/index.ts
```diff
@@ -21,31 +18,10 @@ import type {
...
-  schemas: { metric: '📈 Schemas', total: 0, color: 'white' },
-  parameters: { metric: '👉 Parameters', total: 0, color: 'yellow', items: new Set() },
-  channels: { metric: '📡 Channels', total: 0, color: 'green' },
-  operations: { metric: '👷 Operations', total: 0, color: 'yellow' },
-  tags: { metric: '🔖 Tags', total: 0, color: 'white', items: new Set() },
-});
+import type { VerifyConfigOptions } from '../../types.js';
+import { getFallbackApisOrExit, printExecutionTime } from '../../utils/miscellaneous.js';
+import type { CommandArgs } from '../../wrapper.js';
+import { resolveStatsVisitorAndAccumulator } from './resolve-visitor-and-accumulator.js';
```
tatomyr: It's usually a bad idea to use verbs in filenames.

packages/cli/src/commands/stats/visitor-and-accumulator-resolver.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  type SpecVersion,
+  type OASStatsAccumulator,
+  type AsyncAPIStatsAccumulator,
+  StatsAsync2,
+  StatsAsync3,
+  StatsOAS,
```
tatomyr, line 7: Doesn't this introduce circular referencing because of [this improt](https://github.com/Redocly/redocly-cli/pull/2608/changes#diff-dcd43e9085709b5a89aea11fb3f885193cf2d7e12aa7a049588a690c9af4cc31R24)?

tatomyr: Maybe let's move accumulators here but leave the combination logic where it was? Or just move each visitor to a separate file?

DmitryAnansky: I don't see the circular ref, could you elaborate on this?

docs/@v2/commands/stats.md
DmitryAnansky: we have it for openAPI, not sure why you against it, but updated.

tatomyr: The point of the doc is to demonstrate how it would look like in general and would it fit user's needs, not how it will work in every case. I think it's suffice to look at the OAS example to surmise what the Async output would look like.

Issue #1241, opened 2023-08-22
Reported Error location does not point at correct node for Parameters


jeremyfiel:
**Describe the bug**

When an error is produced for a required parameter in the uri, the error location points at the component schema rather than the actual error location in the `parameters` collection, this makes it difficult to trace which line has the actual error. In this example, the `$ref` on line 89 should be indicated. 

**To Reproduce**

1. Given this `redocly.yaml` file

```yaml
extends:
  - recommended
```

2. And this OpenAPI file(s)

```json
{
    "openapi": "3.1.0",
    "info": {
        "title": "freeCodeCamp Classroom",
        "version": "0.0.1",
        "contact": {
            "name": "Classroom Team",
            "email": ""
        },
        "license": {
            "name": "",
            "url": ""
        }
    },
    "servers": [
        {
            "url": "https://api.freecodecamp.org"
        }
    ],
    "security": [],
    "tags": [
        {
            "name": "Classroom user data",
            "description": ""
        }
    ],
    "paths": {
        "/classroom/userData": {
            "get": {
                "summary": "",
                "description": "",
                "operationId": "getUserData",
                "tags": [
                    "Classroom user data"
                ],
                "parameters": [
                    {
                        "$ref": "#/components/parameters/header_accept"
                    }
                ],
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/userData"
                                }
                            }
                        }
                    },
                    "400": {
                        "$ref": "#/components/responses/400"
                    },
                    "401": {
                        "$ref": "#/components/responses/401"
                    },
                    "403": {
                        "$ref": "#/components/responses/403"
                    },
                    "404": {
                        "$ref": "#/components/responses/404"
                    },
                    "500": {
                        "$ref": "#/components/responses/500"
                    },
                    "503": {
                        "$ref": "#/components/responses/503"
                    }
                }
            },
            "post": {
                "summary": "",
                "description": "",
                "operationId": "queryUserData",
                "tags": [
                    "Classroom user data"
                ],
                "parameters": [
                    {
                        "$ref": "#/components/parameters/header_contentType"
                    },
                    {
                        "$ref": "#/components/parameters/path_userId"
                    }
                ],
                "requestBody": {
                    "$ref": "#/components/requestBodies/userDataRequest"
                },
                "responses": {
                    "200": {
                        "description": "OK",
                        "content": {
[...]

PR #2626, opened 2026-03-02
fix: reported error location at correct node for parameters

kanoru3101:
## What/Why/How?
For parameters referenced via `$ref`, the error location now points at the `$ref` usage instead of the component definition.

## Reference
 #1241
## Testing
```
openapi: 3.1.0
servers:
  - url: https://test-example.com
info:
  title: Repro issue
  version: 0.0.1
  license:
    name: MIT
    url: https://opensource.org/licenses/MIT
security: []
paths:
  /users/data:
    post:
      summary: Query user data
      operationId: queryUserData
      parameters:
        - $ref: "#/components/parameters/header_contentType"
        - $ref: "#/components/parameters/path_userId"
      responses:
        "200":
          description: OK
        "400":
          description: Bad Request

components:
  parameters:
    header_contentType:
      name: content-type
      in: header
      schema:
        type: string
    path_userId:
      name: user-id
      in: path
      required: true
      schema:
        type: string

```

Screenshot 2026-03-02 at 18 24 18

## Check yourself

## Security

### DIFF
--- a/docs/@v2/guides/migrate-to-v2.md
+++ b/docs/@v2/guides/migrate-to-v2.md
@@ -118,6 +118,40 @@ rules:
       defined: true
 ```
 
+### Changes in the `join` command server handling (since v2.0.5)
+
+The `join` command no longer merges `servers` into the root level.
+Each path item now retains the `servers` from its source description, so that joined descriptions remain unmodified.
+This fixes incorrect behavior but may break workflows that relied on root-level server inheritance.
+
+To restore root-level `servers`, use a custom [decorator](../custom-plugins/custom-decorators.md) like the following:
+
+```js
+export default function plugin() {
+  return {
+    id: 'my-plugin',
+    decorators: {
+      oas3: {
+        'add-servers': () => ({
+          Root: {
+            leave: (root) => {
+              root.servers = [{ url: 'https://your-server-url.com' }];
+            },
+          },
+        }),
+      },
+    },
+  };
+}
+```
+
+Register the plugin in `redocly.yaml` and run `join` followed by `bundle` to apply it:
+
+```bash
+redocly join foo.yaml bar.yaml -o tmp.yaml
+redocly bundle tmp.yaml -o joined.yaml
+```
+
 ### Platform changes
 
 - **Legacy Registry**: Support for the legacy Redocly API Registry has been removed in favor of [Reunite](https://app.cloud.redocly.com/).
@@ -174,6 +208,7 @@ paths:
 1. **Replace `spec` rule** with `struct`.
 1. **Update configurable rules** to use `rule/` prefix instead of `assert/`.
 1. **Replace `undefined` assertions** with `defined: true`.
+1. **Revise `join` command usage** if your workflow relies on root-level `servers` in joined output.
 1. **Update configuration structure**:
    - Replace `apiDefinitions` with `apis`
    - Move `features.openapi.*` to `openapi.*`
--- a/packages/core/src/types/redocly-yaml.ts
+++ b/packages/core/src/types/redocly-yaml.ts
@@ -230,8 +230,7 @@ const configGovernanceProperties: Record<
         directResolveAs: { name: 'ConfigGovernance', ...ConfigGovernance },
       } as PropType;
     },
-    description:
-      'The extends configuration entry allows your project configuration to extend an existing configuration set.',
+    description: 'Use extends to inherit rules and their configurations from other rulesets.',
     documentationLink: 'https://redocly.com/docs/cli/configuration/reference/extends#extends',
   } as PropType,
   plugins: { type: 'array', items: { type: 'string' } },


.changeset/sad-mice-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fix `path-parameters-defined` rule to report error at `$ref` location instead of component definition when path parameter is referenced via `$ref` but not used in path
```
tatomyr:
```suggestion
Fixed the `path-parameters-defined` rule to report an error at the `$ref` location instead of at the component definition when a path parameter is referenced via `$ref` but not used in the path.
```

packages/core/src/rules/common/__tests__/path-params-defined.test.ts
```diff
@@ -429,4 +429,68 @@ describe('Oas3 path-params-defined', () => {
 
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
   });
+
+  it('should report error at $ref location when path parameter via $ref is not used in path (issue #1241)', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.1.0
```
tatomyr: Please  use OAS 3.2, and strip off all the nodes that aren't related to the test case if possible.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -25,8 +25,11 @@ export const PathParamsDefined: Oas3Rule | Oas2Rule = () => {
       leave() {
         pathItemLeave(pathContext);
       },
-      Parameter(parameter: Oas2Parameter | Oas3Parameter, { report, location }: UserContext) {
-        createPathItemParameterHandler(parameter, pathContext, report, location);
+      Parameter(
+        parameter: Oas2Parameter | Oas3Parameter,
+        { report, location, rawLocation }: UserContext
+      ) {
+        createPathItemParameterHandler(parameter, pathContext, report, location, rawLocation);
```
tatomyr: Does it make sense to pass parameters as an object instead?

```suggestion
        createPathItemParameterHandler({ parameter, pathContext, report, location, rawLocation });
```

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -167,13 +179,17 @@ const validatePathParameter = (
...
+  rawLocation: UserContext['rawLocation']
 ): void => {
   if (!templateParams.has(paramName)) {
-    report({
-      message: `Path parameter \`${paramName}\` is not used in the path \`${path}\`.`,
-      location: location.child(['name']),
-    });
+    const message = `Path parameter \`${paramName}\` is not used in the path \`${path}\`.`;
+
+    if (rawLocation !== location) {
```
tatomyr: Why do we need this check? Are other rules also make it?

packages/core/src/rules/common/__tests__/path-params-defined.test.ts
kanoru3101: used

packages/core/src/rules/common/path-params-defined.ts
kanoru3101: I rewrote on
```ts
const refLocation = rawLocation !== location ? rawLocation : undefined;
report({ message, location: location.child(['name']), from: refLocation });
```

For example, the `struct` rule, used in the same way

.changeset/sad-mice-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed the `path-parameters-defined` rule to report an error at the `$ref` location instead of at the component definition when a path parameter is referenced via `$ref` but not used in the path.
```
tatomyr: Something is wrong here. It should point at the actual error, not the $ref. What we want is to provide an additional context via the `from` field.

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -45,20 +56,22 @@ const pathItemLeave = (pathContext: { current: PathContext | null }) => {
...
+const createPathItemParameterHandler = ({
+  parameter,
+  pathContext,
+  report,
+  location,
+  rawLocation,
+}: PathItemParameterHandlerParams) => {
   if (parameter.in === 'path' && parameter.name && pathContext.current) {
     pathContext.current.definedParams.add(parameter.name);
     validatePathParameter(
```
tatomyr: Does it make sense to convert the parameters into an object too?

packages/core/src/rules/common/path-params-defined.ts
```diff
@@ -167,13 +187,14 @@ const validatePathParameter = (
...
+  location: UserContext['location'],
+  rawLocation: UserContext['rawLocation']
 ): void => {
   if (!templateParams.has(paramName)) {
-    report({
-      message: `Path parameter \`${paramName}\` is not used in the path \`${path}\`.`,
-      location: location.child(['name']),
-    });
+    const message = `Path parameter \`${paramName}\` is not used in the path \`${path}\`.`;
+    const refLocation = rawLocation !== location ? rawLocation : undefined;
```
tatomyr:
```suggestion
    const from = rawLocation === location ?  undefined : rawLocation;
```

Positive  boolean expressions are  just easier to read. Also, why introduce another name if we can use the name of the existing field?

tatomyr (commented):
Almost there. Please also update the PR description and the test name.

packages/core/src/rules/common/path-params-defined.ts
kanoru3101: Yes, updated for `validateRequiredPathParamsParams` too

.changeset/sad-mice-tease.md
kanoru3101: Updated, please check again

.changeset/sad-mice-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+"@redocly/openapi-core": patch
+"@redocly/cli": patch
+---
+
+Fixed the `path-parameters-defined` rule to report an error at the actual error location (the parameter definition) and provide additional context via the `from` field when a path parameter is referenced via `$ref` but not used in the path.
```
tatomyr: > to report an error at the actual error location 

It was the same before, wasn't it?

tatomyr: > ... provide additional context via the `from` 

This is a technical detail. Changelogs are for the users' benefit, so let's focus on what this contribution brings to users.

kanoru3101: @tatomyr what about this?
```
Improved error messaging when a path parameter referenced via `$ref` is not used in the path.
```

tatomyr: @kanoru3101 LGTM.

PR #2636, opened 2026-03-04
chore: expand exported paths

vital4u4ok:
## What/Why/How?
This pull request adds new entry points to the `packages/respect-core/package.json` exports, making it easier to import specific modules directly.

New module exports:

* Added export for `./runtime-expressions`, allowing direct import of runtime expressions module and its types.
* Added export for `./run`, enabling direct import of the `run` module and its types.

This change helps significantly reduce Replay's bundle size in `openapi-docs`.

## Check yourself

## Security


packages/respect-core/package.json
```diff
@@ -9,6 +9,14 @@
     ".": {
       "import": "./lib/index.js",
       "types": "./lib/index.d.ts"
+    },
+    "./runtime-expressions": {
```
tatomyr, line 13: I'd suggest propagating the exports to the root in the code instead of creating new entrypoints. This should be a safer approach in terms of maintenance.

tatomyr: Unless it's for smaller bundle size. If this is the case, please update the PR's description.

vital4u4ok: Yes, this is for bundle size reduction

Issue #2641, opened 2026-03-10
$NO_PROXY environment variable is ignored when using respect


capkr011:
**Describe the bug**

When executing an Arazzo workflow with `redocly respect` inside a Docker container, requests to domains specified in the `NO_PROXY` environment variable (e.g., `localhost`) are still routed through the HTTP(S) proxy instead of being accessed directly. This leads to connection failures for local endpoints.

We are using the docker image like this:

`docker run -e HTTP_PROXY=$HTTP_PROXY -e HTTPS_PROXY=$HTTPS_PROXY -e http_proxy=$HTTP_PROXY -e https_proxy=$HTTPS_PROXY -e no_proxy=$NO_PROXY -e NO_PROXY=$NO_PROXY  respect test.arazzo.yaml --verbose... 
`
The variable $NO_PROXY contains something like this:  "localhost,.domain.com"

The request to `http://localhost:8000/endpoint` fails.

Errors we get:
`
```
Request URL: http://localhost:8000/endpoint
    Request Headers:
      content-type: text/plain
      accept: application/json
    Request Body: ...
 
    ✗ failed network request
 
  Failed tests info:
  Workflow name: ...
 
    stepId - ... 
    ✗ failed network request 
      fetch failed
```

Using  `strace` , we confirmed that `redocly respect` attempts to `CONNECT` to `localhost:8000` via the proxy:

```
write(19, "CONNECT localhost:8000 HTTP/1.1\r\nhost: localhost:8000\r\nconnection: close\r\n\r\n", 76)
```
When all proxy-related environment variables are unset, the `localhost` endpoint is accessed correctly, confirming the issue is related to `NO_PROXY` not being honored (endpoints behind proxy will fail).

**To Reproduce**

Set proxy URL in  $HTTP(S)_PROXY environment variables and add domains you do not want to be forwarded through proxy in the  $NO_PROXY variable.

**Expected behavior**
redocly respect should not route network calls through proxy for domains listed in the $NO_PROXY (or $no_proxy) environment variable

**Logs**
See above

**OpenAPI description**

OpenAPI 3.0.3 and Arazzo 1.0.1

**Redocly version(s)**

2.20.4

**Node.js version(s)**

nodejs: v24.14.0

npm: 11.9.0

**OS, environment**

Linux, Ubuntu 24 LTS, running in a docker container

**Additional context**

This issue may be related to ticket 
https://github.com/Redocly/redocly-cli/issues/2345

[...]

PR #2643, opened 2026-03-10
feat: add support for NO_PROXY environment variable

DmitryAnansky:
## What/Why/How?

Added support for `NO_PROXY` environment variable.
Users will be able to add domains you do not want to be forwarded through proxy.

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/2641

## Check yourself

## Security

### DIFF
--- a/.changeset/solid-jeans-report.md
+++ b/.changeset/solid-jeans-report.md
@@ -2,4 +2,4 @@
 "@redocly/cli": minor
 ---
 
-Added support for the NO_PROXY environment variable.
+Added support for the `NO_PROXY` environment variable.
--- a/docs/@v2/installation.md
+++ b/docs/@v2/installation.md
@@ -115,22 +115,6 @@ The value is a comma-separated list of hostnames or domain suffixes:
 - `example.com` — matches `example.com` and any subdomain such as `api.example.com`.
 - `*` — bypasses the proxy for all requests.
 
-```bash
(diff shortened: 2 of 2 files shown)


JLekawa (approved):

.changeset/solid-jeans-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Added support for the NO_PROXY environment variable.
```
kanoru3101:
```suggestion
Added support for the `NO_PROXY `environment variable.
```

packages/cli/src/utils/proxy-agent.ts
```diff
@@ -1,10 +1,41 @@
...
+  );
 }
 
 export function getProxyAgent() {
   const proxy = getProxyUrl();
   return proxy ? new HttpsProxyAgent(proxy) : undefined;
 }
+
+export function shouldBypassProxy(url: string): boolean {
+  const noProxy = process.env.NO_PROXY || process.env.no_proxy;
```
kanoru3101, line 18: Is it possible to create a helper tool that will handle environments with upper and lower case letters?  What you do think? 
Does that make sense?

DmitryAnansky: There is a way to do this on yargs level, but I would avoid all env vars matching, just support those 3 as it is a common practice.

PR #2648, opened 2026-03-12
feat: add score command

adamaltman:
## What/Why/How?

**What:** Adds a new score command to Redocly CLI that analyzes OpenAPI 3.x descriptions and produces two composite scores: Integration Simplicity (0-100, how easy is this API to integrate) and Agent Readiness (0-100, how usable is this API by AI agents/LLM tooling).

**Why**: API producers currently lack a quick, deterministic way to assess how developer-friendly or AI-agent-friendly their API descriptions are. The existing stats command counts structural elements but doesn't evaluate quality signals like documentation coverage, example presence, schema complexity, or error response structure. This command fills that gap with actionable, explainable scores.

**How**: The implementation follows the same pattern as the `stats` command (`bundle` + `analyze`), with a clean separation between metric collection and score calculation:
- **Metric collection (`collectors/`)**: Walks the bundled document, resolving internal $refs, to gather per-operation raw metrics (parameter counts, schema depth, polymorphism, description/constraint/example coverage, structured error responses, workflow dependency depth via shared schema refs).
- **Scoring (`scoring.ts`)**: Pure functions that normalize raw metrics into subscores and compute weighted composite scores. Thresholds and weights are configurable constants. anyOf is penalized more heavily than oneOf/allOf; discriminator presence improves the agent readiness polymorphism clarity subscore.
- **Hotspots (`hotspots.ts`)**: Identifies the operations with the most issues, sorted by number of reasons, with human-readable explanations.
- **Output**: `--format=stylish` (default, with color bar charts) and `--format=json` (machine-readable for CI/dashboards).

## Reference

Related to API governance and developer experience tooling. No existing issue -- this is a new feature.

## Testing

- 43 unit tests across 3 test files covering schema depth calculation, `$ref` resolution, polymorphism counting (`oneOf/anyOf/allOf`), constraint detection (including const), example coverage scoring, `anyOf` penalty multiplier, `discriminator` impact on agent readiness, deterministic output, and score range validation.
- Manually tested against three real OpenAPI descriptions (Redocly Cafe: 12 operations, Reunite Main: 299 operations, Rebilly: 606 operations) to verify scores are reasonable and hotspot reasoning is actionable.
- TypeScript compiles cleanly (`tsc --noEmit`), all existing tests continue to pass.

## Screenshots (optional)

Stylish output for Redocly Cafe:

```
  Scores

  Integration Simplicity:  85.3/100
  Agent Readiness:         94.4/100

  Integration Simplicity Subscores

  Parameter Simplicity     [█████████████████░░░] 83%
  Schema Simplicity        [████████████░░░░░░░░] 62%
  Documentation Quality    [███████████████████░] 97%
  Constraint Clarity       [███████████████████░] 96%
  Example Coverage         [██████████████████░░] 92%
  Error Clarity            [████████████████████] 100%
  Workflow Clarity         [█████████████████░░░] 83%

  Top 4 Hotspot Operations

  GET /orders (listOrders)
    Integration Simplicity: 69.1  Agent Readiness: 93.9
    - High parameter count (6)
    - Deep schema nesting (depth 5)

  PATCH /orders/{orderId} (updateOrder)
    Integration Simplicity: 77.6  Agent Readiness: 85.3
    - Missing request body examples
```

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it introduces a new CLI surface area and substantial new scoring/metric-collection logic (schema walking, ref handling, output formatting), plus a small core export change (`isMappingRef`, `Oas3MediaType`) that could affect downstream consumers.
> 
> **Overview**
> Adds a new experimental `score` command to the CLI that bundles an OpenAPI 3.x description, collects per-operation complexity/documentation/example/error metrics (including shared `$ref` dependency depth), computes normalized subscores and a composite **Agent Readiness** score, and reports top “hotspot” operations with human-readable reasons.
> 
> Implements both `--format=stylish` (terminal summary + optional per-operation table + `--debug-operation-id` schema breakdown) and `--format=json` outputs, with extensive unit coverage plus new e2e snapshot fixtures; docs/navigation and a changeset are updated accordingly, and `@redocly/openapi-core` now re-exports `isMappingRef` and `Oas3MediaType` to support the new metric collection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c506c42dfef0126c120bf17dae3685e2b99e7147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/add-score-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': minor
+---
+
+Added new `score` command that analyzes OpenAPI 3.x descriptions and produces integration simplicity and agent readiness scores (0-100). Reports normalized subscores, raw per-operation metrics, and top hotspot operations with human-readable explanations. Supports `--format=stylish` (default) and `--format=json` output.
```
RomanHotsiy: Should we use AI word for more clarity?

\```suggestion
Added new `score` command that analyzes OpenAPI 3.x descriptions and produces integration simplicity and AI agent readiness scores (0-100). Reports normalized subscores, raw per-operation metrics, and top hotspot operations with human-readable explanations. Supports `--format=stylish` (default) and `--format=json` output.
```

File: docs/@v2/commands/index.md [hunk 2]
```diff
@@ -13,10 +13,11 @@ Documentation commands:
 
 API management commands:
 
-- [`stats`](stats.md) Gather statistics for a document.
 - [`bundle`](bundle.md) Bundle API description.
-- [`split`](split.md) Split API description into a multi-file structure.
 - [`join`](join.md) Join API descriptions [experimental feature].
+- [`score`](score.md) Score an API for integration simplicity and agent readiness.
```
RomanHotsiy: same here
```suggestion
- [`score`](score.md) Score an API for integration simplicity and AI agent readiness.
```

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,152 @@
...
+The `score` command supports OpenAPI 3.0, 3.1, and 3.2 descriptions. OpenAPI 2.0 (Swagger) and AsyncAPI are not currently supported.
+{% /admonition %}
+
+### Metrics
+
+The following raw metrics are collected per operation and aggregated across the document:
+
+| Metric                             | Description                                                                                                                          |
+| ---------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| Parameter count                    | Total parameters (path, query, header, cookie) per operation.                                                                        |
```
RomanHotsiy: should we have security schemas documentation as a metric?

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,152 @@
...
+redocly score --config=./another/directory/config.yaml
+```
+
+### Specify output format
+
+#### Stylish output (default)
+
+The default output format shows a human-readable summary in your terminal:
+
+<pre>
```
RomanHotsiy: use backtricks

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,261 @@
...
+    let current: any = document;
+    for (const part of parts) {
+      if (!current || typeof current !== 'object') return undefined;
+      current = current[decodeURIComponent(part.replace(/~1/g, '/').replace(/~0/g, '~'))];
+    }
+    return current && typeof current === 'object' ? current : undefined;
+  };
+}
+
+function resolveIfRef(
```
RomanHotsiy: this is not needed, just call resolveRef - it will return original if it's not ref

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,261 @@
...
+      if (!pathObj) continue;
+
+      const pathLevelParams: any[] = Array.isArray(pathObj.parameters) ? pathObj.parameters : [];
+
+      for (const method of HTTP_METHODS) {
+        const operation = pathObj[method];
+        if (!operation || typeof operation !== 'object') continue;
+
+        const opKey = operation.operationId ?? `${method.toUpperCase()} ${pathStr}`;
+        const metrics = collectOperationMetrics(
```
RomanHotsiy: oh i think it's bad. I think instead it should use visitor approach like stats command instead of going manually down the tree

RomanHotsiy: it should allow to support other spec formats more easily too. I think we should support asyncapi and other specs we have (probably except the overlays and redocly.yaml)

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,153 @@
...
+
+The `score` command analyzes an OpenAPI description and produces two composite scores:
+
+- **Integration Simplicity** (0–100): How easy is this API to integrate? Higher is better.
+- **Agent Readiness** (0–100): How usable is this API by AI agents and LLM-based tooling? Higher is better.
+
+In addition to the top-level scores, the command reports normalized subscores, raw metrics for every operation, and a list of **hotspot operations** — the endpoints most likely to cause integration friction — along with human-readable explanations.
+
+{% admonition type="warning" name="OpenAPI 3.x only" %}
+The `score` command supports OpenAPI 3.0, 3.1, and 3.2 descriptions. OpenAPI 2.0 (Swagger) and AsyncAPI are not currently supported.
```
JLekawa:
```suggestion
The `score` command supports OpenAPI 3.0, 3.1, and 3.2 descriptions.
OpenAPI 2.0 (Swagger) and AsyncAPI are not currently supported.
```

.changeset/add-score-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': minor
+---
+
+Added new `score` command that analyzes OpenAPI 3.x descriptions and produces integration simplicity and AI agent readiness scores (0-100). Reports normalized subscores, raw per-operation metrics, and top hotspot operations with human-readable explanations. Supports `--format=stylish` (default) and `--format=json` output.
```
JLekawa:
```suggestion
Added new `score` command that analyzes OpenAPI 3.x descriptions and produces integration simplicity and AI agent readiness scores (0-100).
Reports normalized subscores, raw per-operation metrics, and top hotspot operations with human-readable explanations. Supports `--format=stylish` (default) and `--format=json` output.
```

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,154 @@
...
+
+```yaml
+apis:
+  core@v1:
+    root: ./openapi/api-description.json
+```
+
+```bash
+redocly score core@v1
+```
```
tatomyr: I don't think we need this extra explanation. It follows our general `apis` approach and don't need a separate example. I'd suggest write docs as concisely as possible so users won't skip reading them.

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Use alternative configuration file
```
tatomyr: Same here. It's described here: https://redocly.com/docs/cli/commands#config-file.

packages/cli/src/commands/score/__tests__/document-metrics.test.ts
```diff
@@ -0,0 +1,415 @@
...
+
+describe('collectDocumentMetrics', () => {
+  it('should return zero operations for empty paths', async () => {
+    const result = await collectDocumentMetrics(makeDoc({}));
+    expect(result.operationCount).toBe(0);
+    expect(result.operations.size).toBe(0);
+  });
+
+  it('should count operations across methods', async () => {
+    const doc = makeDoc({
```
tatomyr: Why use `makeDoc`? We gain a few lines and lose readability. I'd suggest following the current approach and providing the actual API in yaml using `outdent`, like [here](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/__tests__/bundle.test.ts#L14).

tatomyr: Same for other tests. We don't necessarily need that many tests, but they should be easily understandable, so the review makes sense. I tried to go through them, but I quickly got lost. I'd prefer less tests but more eloquent ones.

packages/cli/src/commands/score/__tests__/formatters.test.ts
```diff
@@ -0,0 +1,221 @@
...
+import type { ScoreResult } from '../types.js';
+
+vi.mock('@redocly/openapi-core', () => ({
+  logger: { output: vi.fn() },
+}));
+
+import { logger } from '@redocly/openapi-core';
+const mockOutput = vi.mocked(logger.output);
+
+function makeResult(overrides: Partial<ScoreResult> = {}): ScoreResult {
```
tatomyr: Again, helpers like this don't really help, but make the tests more complicated than they need to be. I think it would be easier to go with snapshots instead, or check only specific fields in expects.

packages/cli/src/commands/score/__tests__/formatters.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    );
+    const output = mockOutput.mock.calls.map(([s]: [string]) => s).join('');
+    expect(output).toContain('Total operations: 0');
+    expect(output).not.toContain('Avg parameters');
+  });
+
+  it('appends newline to every output call', () => {
+    printScoreStylish(makeResult());
+    for (const call of mockOutput.mock.calls) {
+      expect(call[0]).toMatch(/\n$/);
```
tatomyr: I'm not sure what this test does.

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,154 @@
...
+| Parameter count                    | Total parameters (path, query, header, cookie) per operation.                                                                        |
+| Required parameter count           | How many of those parameters are required.                                                                                           |
+| Request body presence              | Whether the operation defines a request body.                                                                                        |
+| Top-level writable field count     | Number of non-`readOnly` top-level properties in request schemas.                                                                    |
+| Max request/response schema depth  | Deepest nesting level in request and response schemas.                                                                               |
+| Polymorphism count                 | Number of `oneOf`, `anyOf`, and `allOf` usages. `anyOf` is penalized more heavily.                                                   |
```
tatomyr: Why is `anyOf` different to `oneOf` in terms of scoring them?

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+  const externalRefResolver = new BaseResolver(config.resolve);
+  const { bundle: document } = await bundle({ config, ref: path });
+  collectSpecData?.(document.parsed);
+
+  const specVersion = detectSpec(document.parsed);
+  if (!specVersion.startsWith('oas3')) {
+    logger.error(
+      colors.red(`The score command currently supports only OpenAPI 3.x. Detected: ${specVersion}`)
+    );
+    process.exitCode = 1;
```
tatomyr:
```suggestion
  if (getMajorSpecVersion(specVersion) !== 'oas3') {
    return exitWithError(`The score command currently supports only OpenAPI 3.x. Detected: ${specVersion}`)
    );
```

packages/cli/src/commands/score/collectors/dependency-graph.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  for (const opKey of operations.keys()) {
+    const depth = bfsMaxDepth(opKey, adjacency);
+    depths.set(opKey, depth);
+  }
+
+  return depths;
+}
+
+function bfsMaxDepth(start: string, adjacency: Map<string, Set<string>>): number {
```
tatomyr: What `bfs` stands for?

packages/cli/src/index.ts
```diff
@@ -73,6 +74,30 @@ yargs(hideBin(process.argv))
...
+        .option({
+          config: { description: 'Path to the config file.', type: 'string' },
+          'lint-config': {
+            description: 'Severity level for config file linting.',
+            choices: ['warn', 'error', 'off'] as ReadonlyArray<RuleSeverity>,
+            default: 'warn' as RuleSeverity,
+          },
+          format: {
+            description: 'Use a specific output format.',
+            choices: ['stylish', 'json', 'markdown'] as ReadonlyArray<OutputFormat>,
```
tatomyr:
```suggestion
            choices: ['stylish', 'json'] as ReadonlyArray<OutputFormat>,
```

No markdown yet.

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    workflowDepths,
+  };
+
+  logger.info(`Document: ${colors.magenta(path)} score:\n`);
+
+  switch (argv.format) {
+    case 'json':
+      printScoreJson(result);
+      break;
+    default:
```
tatomyr, line 114:
```suggestion
    case 'stylish':
    default:
```

Just for the sake of clarity.

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,154 @@
...
+| Ambiguous identifier count         | Parameters with generic names (e.g. `id`, `name`, `type`) and no description.                                                        |
+| Constraint coverage                | Count of constraining keywords (`enum`, `format`, `pattern`, `minimum`, `maximum`, `minLength`, `maxLength`, `discriminator`, etc.). |
+| Request/response example coverage  | Whether request and response media types include `example` or `examples`.                                                            |
+| Structured error response coverage | How many 4xx/5xx responses include a content schema or meaningful description.                                                       |
+| Security scheme coverage           | Whether operations reference documented security schemes with descriptions.                                                          |
+| Workflow dependency depth          | Inferred from shared `$ref` usage across operations; deeper shared-schema graphs indicate more tightly coupled workflows.            |
```
tatomyr: What does the word workflow mean in this context?

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+  const accumulator = createScoreAccumulator();
+  const scoreVisitor = createScoreVisitor(accumulator);
+
+  const resolvedRefMap = await resolveDocument({
+    rootDocument: document,
+    rootType: types.Root,
+    externalRefResolver,
+  });
+
+  const walkCtx: WalkContext = {
```
tatomyr:
```suggestion
  const ctx: WalkContext = {
```

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    [{ severity: 'warn', ruleId: 'score', visitor: scoreVisitor }],
+    types
+  );
+
+  walkDocument({
+    document,
+    rootType: types.Root,
+    normalizedVisitors,
+    resolvedRefMap,
+    ctx: walkCtx,
```
tatomyr:
```suggestion
    ctx,
```

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+      break;
+    default:
+      printScoreStylish(result);
+      break;
+  }
+
+  printExecutionTime('score', startedAt, path);
+}
+
+function averageIntegrationSubscores(
```
tatomyr: I suggest keeping all the calculation logic somewhere else, e.g. in `scoring.ts` or a separate file.

packages/cli/src/commands/score/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    agentSubscores: avgAgentSubscores,
+    rawMetrics,
+    hotspots,
+    operationScores,
+    workflowDepths,
+  };
+
+  logger.info(`Document: ${colors.magenta(path)} score:\n`);
+
+  switch (argv.format) {
```
tatomyr: Also, please move these lines into a separate function, e.g. `printStats`, like [here](https://github.com/Redocly/redocly-cli/blob/main/packages/cli/src/commands/stats/index.ts#L129), to keep things similar.

tatomyr (commented):
Left a couple of comments. I haven'd fully reviewed the scoring and collectors though as it takes time.

packages/cli/src/commands/score/__tests__/collect-metrics.ts
```diff
@@ -0,0 +1,82 @@
+import {
```
tatomyr: Is it code or tests? It code, then it should be outside the `__tests__` folder.

tatomyr: This code doesn't seems to be used anywhere except in tests. Please consider removing it and the tests.

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,503 @@
...
+    hasPropertyExamples: false,
+    writableTopLevelFields: 0,
+    refsUsed: [],
+    debugEntries: null,
+    pendingRef: null,
+  };
+}
+
+export function resetSchemaWalkState(s: SchemaWalkState): void {
+  s.depth = -1;
```
tatomyr: What does the negative depth mean?

packages/cli/src/commands/score/__tests__/document-metrics.test.ts
```diff
@@ -0,0 +1,244 @@
+import { parseYaml } from '@redocly/openapi-core';
+import { outdent } from 'outdent';
+
+import { collectDocumentMetrics } from './collect-metrics.js';
+
+function yaml(source: string): Record<string, unknown> {
+  return parseYaml(source) as Record<string, unknown>;
+}
```
tatomyr:
```suggestion
import { parseYaml as yaml } from '@redocly/openapi-core';
import { outdent } from 'outdent';

import { collectDocumentMetrics } from './collect-metrics.js';
```

No need in this wrapper.

packages/cli/src/commands/score/__tests__/formatters.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      integrationSimplicityScore: 65,
+      agentReadinessScore: 55,
+      reasons: ['Missing operation description'],
+    },
+  ],
+  operationScores: new Map(),
+  workflowDepths: new Map(),
+};
+
+function getStylishOutput(result: ScoreResult = RESULT): string {
```
tatomyr: It's a bad pattern to use default parameters in tests. They should be as explicit as possible.

tatomyr (changes requested):
I found some dead code. Please check if it's needed and remove if not. Also, I'm not sure what tests to review as many appear to only test that dead code, so let's handle that first.

packages/cli/src/commands/score/__tests__/document-metrics.test.ts
```diff
@@ -0,0 +1,321 @@
...
+  resolveDocument,
+  BaseResolver,
+  Source,
+  parseYaml,
+  type Document,
+  type WalkContext,
+} from '@redocly/openapi-core';
+import { outdent } from 'outdent';
+
+const yaml = (source: string) => parseYaml(source) as Record<string, unknown>;
```
adamaltman: Restored a `yaml()` cast wrapper since `parseYaml` returns unknown but the function expects `Record<string, unknown>`. Otherwise checks fail.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    if (ref) {
+      if (!debug) {
+        const cached = schemaCache.get(ref);
+        if (cached) return cached;
+      }
+      resolved = resolveJsonPointer(document.parsed, ref) ?? schemaNode;
+    }
+
+    if (resolved && typeof resolved === 'object' && computing.has(resolved)) {
```
tatomyr: This can pass arrays in too. It's better to use a utility function instead:

```suggestion
    if (isPlainObject(resolved) && computing.has(resolved)) {
```

packages/cli/src/commands/score/__tests__/formatters.test.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+function getStylishOutput(result: ScoreResult): string {
+  mockOutput.mockClear();
+  printScoreStylish(result);
+  return mockOutput.mock.calls.map(([s]: [string]) => s).join('');
+}
+
+describe('printScoreStylish', () => {
+  it('outputs scores, subscores, metrics, and hotspots', () => {
```
tatomyr: Instead of mocking and presetting, it's better to simply create e2e tests for that cases.

packages/cli/src/commands/score/__tests__/hotspots.test.ts
```diff
@@ -0,0 +1,304 @@
...
+    totalErrorResponses: 0,
+    ambiguousIdentifierCount: 0,
+    refsUsed: new Set(),
+    ...overrides,
+  };
+}
+
+function makeScores(overrides: Partial<OperationScores> = {}): OperationScores {
+  return {
+    integrationSimplicity: 80,
```
tatomyr: What does this object mean? Is this kind of a default score?

adamaltman: Left a comment in the code above: https://github.com/Redocly/redocly-cli/pull/2648/changes/d6175e2bc99b869602cf24a37d037dbd35f07d1c#diff-32ae15ab9d5aed1dd1d04aa78ef227e00866148987c233b12b2da12fe9657a94R34

packages/cli/src/commands/score/__tests__/formatters.test.ts
adamaltman: Did this, but:

> ERROR: Coverage for branches (70.99%) does not meet global threshold (71%)
> Error: Process completed with exit code 1.

tatomyr: That's not a problem. You can lower the threshold in **vitest.config.ts**. It's expected that we only cover with unit tests only what's reasonable.

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,136 @@
+# `score`
+
+## Introduction
+
```
tatomyr, line 4:
```suggestion

{% admonition type="warning" name="Important" %}
The `score` command is considered an experimental feature. This means it's still a work in progress and may go through major changes.

The `score` command supports OpenAPI 3.x descriptions only.
{% /admonition %}

```

I suggest adding the 'experimental' admonition similar to the `join` command.

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,136 @@
...
+- **Integration Simplicity** (0–100): How easy is this API to integrate? Higher is better.
+- **Agent Readiness** (0–100): How usable is this API by AI agents and LLM-based tooling? Higher is better.
+
+In addition to the top-level scores, the command reports normalized subscores, raw metrics for every operation, and a list of **hotspot operations** — the endpoints most likely to cause integration friction — along with human-readable explanations.
+
+{% admonition type="warning" name="OpenAPI 3.x only" %}
+The `score` command supports OpenAPI 3.0, 3.1, and 3.2 descriptions.
+OpenAPI 2.0 (Swagger) and AsyncAPI are not currently supported.
+{% /admonition %}
```
tatomyr:
```suggestion
```

docs/@v2/commands/score.md
```diff
@@ -0,0 +1,136 @@
...
+- "Polymorphism (anyOf) without discriminator"
+- "Missing request and response examples"
+- "No structured error responses (4xx/5xx)"
+- "Missing operation description"
+
+## Usage
+
+```bash
+redocly score <api>
+redocly score <api> [--format=<option>] [--config=<path>]
```
tatomyr:
```suggestion
redocly score <api> [--format=<option>]
```

I don't think we explicitly use any **redocly.yaml** configuration in this command, so for me it doesn't make too much sense to list it in the examples.

packages/cli/src/commands/score/__tests__/index.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    argv: { format: 'stylish' as const, ...overrides } as ScoreArgv,
+    config: { extendTypes: (types: any) => types, resolve: {} } as any,
+    version: '0.0.0-test',
+    collectSpecData: vi.fn(),
+  };
+}
+
+describe('handleScore', () => {
+  beforeEach(() => {
+    mockOutput.mockClear();
```
tatomyr: This is redundant as we configured `restoreMocks: true` in vitest config.
```suggestion
```

packages/cli/src/commands/score/__tests__/index.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    mockedGetFallback.mockResolvedValue([{ path: 'test.yaml' }] as any);
+    mockedBundle.mockResolvedValue({ bundle: { parsed: {} } } as any);
+    mockedDetectSpec.mockReturnValue('oas3_0');
+    mockedNormalizeTypes.mockReturnValue({ Root: {} } as any);
+    mockedResolveDocument.mockResolvedValue(new Map() as any);
+    mockedCollectMetrics.mockReturnValue({
+      metrics: makeDocumentMetrics(new Map([['listItems', makeTestMetrics()]])),
+      debugLogs: [],
+    });
+    process.exitCode = undefined;
```
tatomyr: Seems redundant. At least, I don't see where we use `exitCode` in the tests.

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,505 @@
...
+              totalProperties: stats.propertyCount,
+              totalPolymorphism: stats.polymorphismCount,
+              totalConstraints: stats.constraintCount,
+              maxDepth: stats.maxDepth,
+            });
+          }
+        }
+      },
+    },
+    Paths: {
```
tatomyr, line 427: What is this for? Could PathItems reside outside the Paths?

packages/cli/src/commands/score/formatters/json.ts
```diff
@@ -0,0 +1,61 @@
...
+    discoverability: Math.round(result.discoverability * 100),
+    integrationSubscores: result.integrationSubscores,
+    agentSubscores: result.agentSubscores,
+    rawMetrics: {
+      operationCount: result.rawMetrics.operationCount,
+      operations: Object.fromEntries(
+        Array.from(result.rawMetrics.operations.entries()).map(([key, m]) => [
+          key,
+          {
+            path: m.path,
```
tatomyr: What is this mapping useful for? Why not simply use `m` or `{ ...m }`? Are you trying to clean some properties?

packages/cli/src/commands/score/formatters/json.ts
```diff
@@ -0,0 +1,61 @@
...
+            ambiguousIdentifierCount: m.ambiguousIdentifierCount,
+          },
+        ])
+      ),
+    },
+    operationScores: Object.fromEntries(
+      Array.from(result.operationScores.entries()).map(([key, s]) => [
+        key,
+        {
+          integrationSimplicity: s.integrationSimplicity,
```
tatomyr: Same here.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+        : Array.isArray(anyOfBranches) && anyOfBranches.length > 1
+          ? 'anyOf'
+          : null;
+    const polyBranches: any[] | null =
+      polyKeyword === 'oneOf' ? oneOfBranches : polyKeyword === 'anyOf' ? anyOfBranches : null;
+
+    const hasAllOf = Array.isArray(resolved?.allOf) && resolved.allOf.length > 0;
+
+    const disc = resolved?.discriminator ?? stripped?.discriminator;
+    const discriminatorRefs =
```
tatomyr, line 150: It looks like you're trying to reimplement `isMappingRef` utility function.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+    let resolved = schemaNode;
+    const ref: string | undefined =
+      schemaNode?.$ref && typeof schemaNode.$ref === 'string' ? schemaNode.$ref : undefined;
+
+    if (ref) {
+      if (!debug) {
+        const cached = schemaCache.get(ref);
+        if (cached) return cached;
+      }
+      resolved = resolveJsonPointer(document.parsed, ref) ?? schemaNode;
```
tatomyr, line 130: I'm not 100% sure, but it looks like if you run the walker within the `UserContext`, you should be able to reuse the `ctx.resolve` method to resolve refs, so you don't have to reimplement that itself.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+    const hasAllOf = Array.isArray(resolved?.allOf) && resolved.allOf.length > 0;
+
+    const disc = resolved?.discriminator ?? stripped?.discriminator;
+    const discriminatorRefs =
+      !polyKeyword && disc?.mapping
+        ? (Object.values(disc.mapping) as string[])
+            .filter((v) => typeof v === 'string' && v.startsWith('#/'))
+            .map((v) => ({ $ref: v }))
+        : null;
+    const hasDiscriminatorBranches = discriminatorRefs != null && discriminatorRefs.length > 0;
```
tatomyr:
```suggestion
    const hasDiscriminatorBranches = Array.isArray(discriminatorRefs) && discriminatorRefs.length > 0;
```

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+        };
+      }
+
+      let discStats: SchemaStats | null = null;
+      if (hasDiscriminatorBranches) {
+        delete parentOnly.discriminator;
+
+        let maxBranch = walkSchema(discriminatorRefs![0], debug);
+        for (let i = 1; i < discriminatorRefs!.length; i++) {
+          const branchStats = walkSchema(discriminatorRefs![i], debug);
```
tatomyr:
```suggestion
        for (let i = 1; i < discriminatorRefs.length; i++) {
          const branchStats = walkSchema(discriminatorRefs[i], debug);
```

I don't think you need those asserts.

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,505 @@
...
+        for (const keyword of ['oneOf', 'anyOf', 'allOf'] as const) {
+          const list = schema[keyword];
+          if (!Array.isArray(list)) continue;
+          state.polymorphismCount += list.length;
+          if (keyword === 'anyOf') state.anyOfCount += list.length;
+          localPoly[keyword] = list.length;
+        }
+
+        const localPropertyNames: string[] = [];
+        if (schema.properties && typeof schema.properties === 'object') {
```
tatomyr:
```suggestion
        if (isPlainObject(schema.properties)) {
```

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,505 @@
...
+  for (const raw of opLevel) {
+    const p = resolveParam(raw, resolve);
+    if (p?.name && p.in) merged.set(`${p.in}:${p.name}`, p);
+  }
+  return merged;
+}
+
+function hasExample(mediaType: { example?: unknown; examples?: Record<string, unknown> }): boolean {
+  if (mediaType.example !== undefined) return true;
+  if (mediaType.examples && typeof mediaType.examples === 'object') {
```
tatomyr:
```suggestion
  if (isNotEmptyObject(mediaType.examples)) {
```

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,360 @@
...
+      debugEntries:
+        a.debugEntries || b.debugEntries
+          ? [...(a.debugEntries ?? []), ...(b.debugEntries ?? [])]
+          : undefined,
+    };
+  }
+
+  const walkSchema = (schemaNode: any, debug = false): SchemaStats => {
+    let resolved = schemaNode;
+    const ref: string | undefined =
```
tatomyr: It's better to reuse the `isRef` utility function here.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Yes, PathItems can appear outside Paths in OAS 3.1+ (e.g. in webhooks). Wrapping under `Paths` scopes the visitor to only score the main API operations, not webhook operations. This is intentional since webhook operations have different discoverability/integration characteristics.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Good catch. `isMappingRef` is not exported from the public `@redocly/openapi-core` API, so I can't import it directly here. I've added `isPlainObject(disc?.mapping)` as a guard and kept the inline `startsWith("#/")` filter. If `isMappingRef` gets exported in the future, we can switch to it.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: The `walkSchema` function in `collect-metrics.ts` runs outside a visitor callback — it's called from the accumulator's `MediaType.enter` handler, which does have `ctx.resolve`, but `walkSchema` itself recursively resolves `$ref` strings against the raw parsed document to handle composition stripping and max-branch logic. Switching to `ctx.resolve` would require threading the resolve function through all the recursive calls, which is feasible but a larger refactor. Filed as a follow-up improvement.

packages/cli/src/commands/score/collectors/document-metrics.ts
tatomyr: Indeed, I've forgotten about the webhooks.

packages/cli/src/commands/score/collect-metrics.ts
tatomyr: You can export if from the openapi-core if needed. Simply reexport it from the [index file](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/index.ts).

packages/cli/src/commands/score/collect-metrics.ts
tatomyr: That could be improved later, for sure.

packages/cli/src/commands/score/collectors/dependency-graph.ts
tatomyr: Could you use a more descriptive name here?

packages/cli/src/commands/score/__tests__/document-metrics.test.ts
```diff
@@ -0,0 +1,246 @@
+import { parseYaml } from '@redocly/openapi-core';
+import { outdent } from 'outdent';
+
+import { collectDocumentMetrics } from '../collect-metrics.js';
+
+const yaml = (source: string) => parseYaml(source) as Record<string, unknown>;
+
+async function collect(doc: Record<string, unknown>) {
+  return (await collectDocumentMetrics(doc)).metrics;
+}
```
tatomyr: This wrapper seems needless and doesn't add to the tests clarity.

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,512 @@
...
+    hasPropertyExamples: false,
+    writableTopLevelFields: 0,
+    refsUsed: [],
+    debugEntries: null,
+    pendingRef: null,
+  };
+}
+
+export function resetSchemaWalkState(s: SchemaWalkState): void {
+  s.depth = -1; // starts at -1 because the root Schema.enter increments to 0
```
tatomyr: It seems that resetting the state is the same as `createSchemaWalkState`. Could it be reused to avoid creating several source of truth?

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+  ctx,
+  debugOperationId,
+}: CollectMetricsOptions): CollectMetricsResult {
+  const removedComposition = stripCompositionKeywords(document.parsed);
+
+  try {
+    const schemaWalkState = createSchemaWalkState();
+    const schemaVisitor = createSchemaMetricVisitor(schemaWalkState);
+    const normalizedSchemaVisitors = normalizeVisitors(
+      [{ severity: 'warn', ruleId: 'score-schema', visitor: schemaVisitor as any }],
```
tatomyr: What is the type cast needed for? Is the visitor type wrong?

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+  if (!ref.startsWith('#/')) return undefined;
+  let node = root;
+  for (const segment of ref.slice(2).split('/')) {
+    const decoded = decodeURIComponent(segment).replace(/~1/g, '/').replace(/~0/g, '~');
+    node = node?.[decoded];
+  }
+  return node;
+}
+
+interface SchemaStats {
```
tatomyr: This looks similar to `SchemaStats` in [document-metrics.ts](https://github.com/Redocly/redocly-cli/pull/2648/changes#diff-9abcbba74119f9188aa71c13e66d7741e9c5423eb21ec91ece6efb4201c7da8dR169) but with a wrong type import. Are they the same?

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+  createSchemaMetricVisitor,
+  getDocumentMetrics,
+} from './collectors/document-metrics.js';
+import type { DebugMediaTypeLog, DocumentMetrics } from './types.js';
+
+function resolveJsonPointer(root: any, ref: string): any {
+  if (!ref.startsWith('#/')) return undefined;
+  let node = root;
+  for (const segment of ref.slice(2).split('/')) {
+    const decoded = decodeURIComponent(segment).replace(/~1/g, '/').replace(/~0/g, '~');
```
tatomyr: It's probably better to reuse `unescapePointerFragment` from `ref-utils.ts`.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+  } finally {
+    restoreCompositionKeywords(removedComposition);
+  }
+}
+
+/**
+ * Convenience wrapper that resolves a parsed OpenAPI document and collects metrics.
+ * Useful in tests and standalone usage where you don't already have resolved types.
+ */
+export async function collectDocumentMetrics(
```
tatomyr: Where is this function  being used?

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+      }
+
+      return result;
+    };
+
+    const accumulator = createScoreAccumulator(walkSchema, debugOperationId);
+    const scoreVisitor = createScoreVisitor(accumulator);
+
+    const normalizedVisitors = normalizeVisitors(
+      [{ severity: 'warn', ruleId: 'score', visitor: scoreVisitor as any }],
```
tatomyr: The type casting seems redundant to me.

packages/cli/src/commands/score/hotspots.ts
```diff
@@ -0,0 +1,114 @@
...
+    const aAvg = (a.integrationSimplicityScore + a.agentReadinessScore) / 2;
+    const bAvg = (b.integrationSimplicityScore + b.agentReadinessScore) / 2;
+    return aAvg - bAvg;
+  });
+
+  return entries.slice(0, constants.hotspotLimit);
+}
+
+function getHotspotReasons(
+  metrics: {
```
tatomyr:
```suggestion
  metrics: OperationMetrics
```

Isn't it the same type?

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,512 @@
...
+  }
+  return merged;
+}
+
+function hasExample(mediaType: { example?: unknown; examples?: Record<string, unknown> }): boolean {
+  if (mediaType.example !== undefined) return true;
+  if (isNotEmptyObject(mediaType.examples)) {
+    return true;
+  }
+  return false;
```
tatomyr:
```suggestion
  return (mediaType.example !== undefined)  || isNotEmptyObject(mediaType.examples)
```

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,512 @@
...
+    if (p?.name && p.in) merged.set(`${p.in}:${p.name}`, p);
+  }
+  for (const raw of opLevel) {
+    const p = resolveParam(raw, resolve);
+    if (p?.name && p.in) merged.set(`${p.in}:${p.name}`, p);
+  }
+  return merged;
+}
+
+function hasExample(mediaType: { example?: unknown; examples?: Record<string, unknown> }): boolean {
```
tatomyr:
```suggestion
function hasExample(mediaType: Oas3MediaType): boolean {
```

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,512 @@
...
+        }
+      },
+      leave() {
+        if (accumulator.current) accumulator.current.inResponse = false;
+      },
+    },
+    MediaType: {
+      enter(
+        mediaType: { schema?: any; example?: unknown; examples?: Record<string, unknown> },
+        _ctx: UserContext
```
tatomyr:
```suggestion
        mediaType
```

tatomyr: Looks like you're trying to put as much code as possible where you can perfectly get away with fewer 😄. It works correctly, but the codebase gets bloated and harder to maintain.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+        maxDepth: schemaWalkState.maxDepth,
+        polymorphismCount: schemaWalkState.polymorphismCount,
+        anyOfCount: schemaWalkState.anyOfCount,
+        hasDiscriminator: schemaWalkState.hasDiscriminator,
+        propertyCount: schemaWalkState.propertyCount,
+        totalSchemaProperties: schemaWalkState.totalSchemaProperties,
+        schemaPropertiesWithDescription: schemaWalkState.schemaPropertiesWithDescription,
+        constraintCount: schemaWalkState.constraintCount,
+        hasPropertyExamples: schemaWalkState.hasPropertyExamples,
+        writableTopLevelFields: schemaWalkState.writableTopLevelFields,
```
tatomyr:
```suggestion
       ...schemaWalkState
```

Could it be simplified like so?

packages/cli/src/commands/score/collectors/document-metrics.ts
```diff
@@ -0,0 +1,512 @@
...
+    inRequestBody: false,
+    inResponse: false,
+    currentResponseCode: '',
+    errorStructuredCounted: false,
+
+    refsUsed: new Set(),
+  };
+}
+
+function buildOperationMetrics(ctx: CurrentOperationContext): OperationMetrics {
```
tatomyr, line 297: What is the difference between those two types? It looks like they are either the same or could be reused, so this function could be avoided or at least shortened.

packages/cli/src/commands/score/collect-metrics.ts
```diff
@@ -0,0 +1,367 @@
...
+  oneOf?: any[];
+  anyOf?: any[];
+}
+
+function stripCompositionKeywords(parsed: any): Map<object, StrippedComposition> {
+  const removed = new Map<object, StrippedComposition>();
+  const seen = new WeakSet<object>();
+
+  function walk(node: any): void {
+    if (!node || typeof node !== 'object' || seen.has(node)) return;
```
tatomyr:
```suggestion
    if (!isPlainObject(node) || seen.has(node)) return;
```

Was this the intention? Or you also want to traverse arrays?

Another question--could nodes be $refs? If yes, you have to resolve them. On the other hand, you already have `walkSchemaRaw` function which walks through schemas, maybe you can reuse it, but with a different visitor?

tatomyr (commented):
These metrics look like a duplication:

Image

Is it intended?

adamaltman:
Yes, I think there is too much overlap between the two scores so I'm going to consolidate it into one. It will make docs easier, result understanding easier, etc... Addressing other comments too.

packages/cli/src/commands/score/__tests__/document-metrics.test.ts
adamaltman: Done — the collect() wrapper is removed. Tests now destructure { metrics } directly from collectDocumentMetrics().

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Done — resetSchemaWalkState now uses Object.assign(s, createSchemaWalkState()) to reuse the single source of truth.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Yes, the as any cast is needed because normalizeVisitors expects an internal type (RuleConfig[]) that doesn't structurally match Oas3Visitor. The visitor types aren't perfectly aligned for this use case — the same pattern exists for the outer document walker on line 238. We can't avoid it without changing openapi-core's exported types.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Done — consolidated SchemaStats into types.ts and both files now import from there.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Done — now using unescapePointerFragment from @redocly/openapi-core.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: This is collectDocumentMetrics — it's used by tests as a convenience wrapper around the lower-level collectMetrics. It resolves the document, sets up types, and calls collectMetrics so tests don't have to repeat that boilerplate. Kept it since it's used across multiple test files.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Done — simplified to use spread with destructuring to exclude the extra SchemaWalkState fields (depth, pendingRef, debugEntries).

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: CurrentOperationContext has visitor tracking fields (inRequestBody, inResponse, currentResponseCode, errorStructuredCounted, refsUsed as Set) that aren't part of OperationMetrics. Also field names differ (e.g. writableTopLevelFieldCount vs topLevelWritableFieldCount). So we can't use spread or avoid the explicit mapping.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: This entire stripping mechanism has been removed in the latest commit. The parentOnly shallow copy already handles the top-level composition, and nested schemas now retain their composition keywords naturally — which was actually a bug (Bugbot caught it: nested oneOf/anyOf were silently lost by the global stripping).

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Done — simplified to a single return expression.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Oas3MediaType isn't currently exported from the public API of @redocly/openapi-core, so we kept the inline type. Simplified the signature to just (mediaType: any) to avoid the verbose inline type.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Done — simplified to (mediaType: any), removed the unused _ctx parameter.

packages/cli/src/commands/score/hotspots.ts
adamaltman: Done — using OperationMetrics type directly.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Good point — addressed by consolidating both scores into a single Agent Readiness score. The overlap is gone now.

adamaltman:
Addressed all remaining review comments. The two scores have been consolidated into a single **Agent Readiness** score with all 9 subscores. Also removed the global `stripCompositionKeywords` mechanism (Bugbot's nested-composition bug), unused `maxPropertiesGood` threshold, and deduplicated the `median` function into a shared utility.

packages/cli/src/commands/score/collectors/document-metrics.ts
tatomyr: The `any` type is not needed here. The correct type is inferred from the visitor by default.

packages/cli/src/commands/score/collect-metrics.ts
tatomyr: If it's only used in tests, it should reside there. Having test code alongside the regular one is misleading.
Better yet, let's name it something like `coolectMetricsTestWrapper` to reflect the intention.

packages/cli/src/commands/score/collectors/document-metrics.ts
tatomyr: It could be reexported via openapi-core's index file and reused here (similar to `isMappingRef`). It's always safer to use the correct type instead of creating it's simulations.

packages/cli/src/commands/score/collectors/document-metrics.ts
tatomyr: Then I'd suggest using the same name and consolidate the types, so you don't have to do the transformation and this function could be simply removed.

packages/cli/src/commands/score/collect-metrics.ts
tatomyr: Sorry, I think you are wrong. I've run it locally without the casts and the code compiles without warnings.

adamaltman: Done — removed both `as any` casts. Confirmed that `tsc --noEmit --skipLibCheck` compiles cleanly without them.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Done — renamed `writableTopLevelFieldCount` to `topLevelWritableFieldCount` in `CurrentOperationContext` to match `OperationMetrics`, and simplified `buildOperationMetrics` to a destructuring spread that strips only the four tracking-only fields.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Good call — exported `Oas3MediaType` from `@redocly/openapi-core` and now using it for the `hasExample` parameter type.

packages/cli/src/commands/score/collect-metrics.ts
adamaltman: Done — moved `collectDocumentMetrics` to `__tests__/collect-metrics-helper.ts` since it's only used in tests. Both `document-metrics.test.ts` and `example-coverage.test.ts` now import from there.

packages/cli/src/commands/score/collectors/document-metrics.ts
adamaltman: Done — removed the `any` type annotation. The visitor framework infers the correct type by default.

adamaltman:
Addressed all review comments in fdb8c94:

1. **Removed `as any` casts** — both `normalizeVisitors` calls compile cleanly without them.
2. **Consolidated `CurrentOperationContext` → `OperationMetrics`** — renamed `writableTopLevelFieldCount` to `topLevelWritableFieldCount`, simplified `buildOperationMetrics` to a destructuring spread.
3. **Exported `Oas3MediaType`** from `@redocly/openapi-core` and use it in `hasExample`.
4. **Removed `any` type** from `MediaType.enter` — visitor infers the type by default.
5. **Moved `collectDocumentMetrics`** to `__tests__/collect-metrics-helper.ts` since it's test-only.
6. **Removed duplicate `propertyCount`** (Bugbot) — `totalSchemaProperties` is used everywhere now.
7. **Fixed `constraintCount` media type mixing** (Bugbot) — `constraintCount`, `totalSchemaProperties`, and `schemaPropertiesWithDescription` now all come from the same schema walk.

tatomyr (approved):
LGTM. There are a few bugbot comments left -- please review whether they make sense.

JLekawa (approved):

PR #2681, opened 2026-03-23
fix: prevent failing when discriminator mapping value is null

tatomyr:
## What/Why/How?

Fixed an issue where invalid discriminator mapping values could cause linting to fail, for instance like `foo: null` in the following example:

```yaml
openapi: 3.2.0
components:
  schemas:
    Test:
      type: object
      properties:
        base:
          type: string
      discriminator:
        propertyName: test
        mapping:
          foo: null # This cause failure

```

 

## Check yourself

## Security

### DIFF
--- a/packages/core/src/ref-utils.ts
+++ b/packages/core/src/ref-utils.ts
@@ -87,11 +87,10 @@ export function refBaseName(ref: string) {
 
 export function isAbsoluteUrl(ref: string) {
   return (
-    typeof ref === 'string' &&
-    (ref.startsWith('http://') ||
-      ref.startsWith('https://') ||
-      ref.startsWith('file://') ||
-      ref.startsWith('data:'))
+    ref.startsWith('http://') ||
+    ref.startsWith('https://') ||
+    ref.startsWith('file://') ||
+    ref.startsWith('data:')
   );
 }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/ref-utils.ts
```diff
@@ -87,10 +87,11 @@ export function refBaseName(ref: string) {
 
 export function isAbsoluteUrl(ref: string) {
   return (
-    ref.startsWith('http://') ||
-    ref.startsWith('https://') ||
-    ref.startsWith('file://') ||
-    ref.startsWith('data:')
+    typeof ref === 'string' &&
```
AlbinaBlazhko17: I was thinking about this type guard, but in many places where we use this function, the type of ref is a string. This gives us confidence in the type, and if we receive an error, it means that we have forgotten to handle a specific behaviour. 
In other words, if the reference is not a valid reference, `isAbsoluteUrl` will return `undefined`, and we will bypass it during execution, which will cause other errors or inconvenient behavior.

tatomyr: If it's not a string, it will crash the app which is even worse. This happens mostly in openapi-language-server when we're typing, but until we actually write something, the value is considered `null` and the server crashes.

AlbinaBlazhko17: Yeah, i totally agree, that we need to avoid the crash. But there is cases when we throw a meaningful error and we don't proceed to next steps and if we receive an error like `Cannot read properties of null (reading 'startsWith')` it means, that we forget to add guard or throw a meaningful error.

Issue #2702, opened 2026-03-30
`lint` crashes with OAS3.2 `defaultMapping` property


je-l:
**Describe the bug**

Runtime crash `Cannot read properties of undefined (reading 'includes')` when using the new OAS3.2 `defaultMapping` field and referencing it in the path definitions.

**To Reproduce**

1. Given this OpenAPI file
2. Run `redocly lint schema.yaml`
3. Observe crash

**Expected behavior**

Should print linter warnings and errors as usual.

**Logs**

```
$ redocly lint schema.yaml
No configurations were provided -- using built in recommended configuration by default.

validating schema.yaml...
Something went wrong when processing schema.yaml:

  - Cannot read properties of undefined (reading 'includes')

```

**OpenAPI description**

```yaml
openapi: 3.2.0
paths:
  /x:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/A'
components:
  schemas:
    A:
      oneOf:
        - $ref: '#/components/schemas/B'
      discriminator:
        propertyName: t
        defaultMapping: C

    B:
      type: object
    C:
      type: object
```

**Redocly version(s)**

2.25.2

**Node.js version(s)**

v24.0.2

**OS, environment**

Fedora 42

tatomyr:
Thanks for reporting. Looking into it.

tatomyr:
@je-l  should be fixed in v2.25.3. Please let me know if it works for you.

je-l:
I updated the CLI and it's working now, thank you.

PR #2703, opened 2026-03-30
fix: discriminator defaultMapping rule

tatomyr:
## What/Why/How?

Fixed the `spec-discriminator-defaultMapping` rule:
- prevent crashing by collecting components.schemas names from root to ensure correct order
- traverse composite schemas to find the required property
- treat defaultMapping as a ref to a schema
- add location to resolve refs correctly from other files
- handle cyclic dependencies

## Reference

Fixes https://github.com/Redocly/redocly-cli/issues/2702
 

## Check yourself

## Security

### DIFF
--- a/packages/core/src/rules/oas3/__tests__/spec-discriminator-defaultMapping.test.ts
+++ b/packages/core/src/rules/oas3/__tests__/spec-discriminator-defaultMapping.test.ts
@@ -197,6 +197,45 @@ describe('spec-discriminator-defaultMapping', () => {
     `);
   });
 
+  it('should pass when propertyName is optional in a descendant oneOf schema but defaultMapping is provided', async () => {
+    const document = parseYamlToDocument(
+      outdent`
+        openapi: 3.2.0
+        components:
+          schemas:
+            Used:
+              discriminator:
+                propertyName: test
+                defaultMapping: Foo
+              oneOf:
+                - $ref: '#/components/schemas/Foo'
+                - $ref: '#/components/schemas/Bar'
+            Foo:
+              type: object
+              properties:
+                test: # test is not required here
+                  type: string
+                  const: foo
+            Bar:
+              type: object
+              properties:
+                test:
+                  type: string
+                  const: bar
+              required: [test]
+      `,
+      'foobar.yaml'
+    );
+
+    const results = await lintDocument({
+      externalRefResolver: new BaseResolver(),
+      document,
+      config: await createConfig({ rules: { 'spec-discriminator-defaultMapping': 'error' } }),
+    });
+
+    expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
+  });
+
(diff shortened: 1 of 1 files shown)


packages/cli/src/auth/__tests__/oauth-client.test.ts
```diff
@@ -5,16 +5,17 @@ import * as path from 'node:path';
 import { RedoclyOAuthDeviceFlow } from '../device-flow.js';
 import { RedoclyOAuthClient } from '../oauth-client.js';
 
+vi.mock('node:fs');
```
tatomyr, line 8: Just moved the mocks up as suggested by the linter. Nothing changed.

packages/core/src/types/oas3_2.ts
```diff
@@ -241,10 +242,15 @@ const Discriminator: NodeType = {
   ...Oas3_1Types.Discriminator,
   properties: {
     ...Oas3_1Types.Discriminator.properties,
-    defaultMapping: {
-      type: 'string',
-      description:
-        'The schema name or URI reference to a schema that is expected to validate the structure of the model when the discriminating property is not present in the payload or contains a value for which there is no explicit or implicit mapping.',
+    defaultMapping: (value) => {
```
tatomyr, line 245: This allows `defaultMapping` to be resolved as schema (similar to how `mapping` is resolved).

.oxlintrc.json
```diff
@@ -12,6 +12,7 @@
   "rules": {
     "eslint/no-array-constructor": "error",
     "eslint/no-case-declarations": "error",
+    "eslint/no-console": "error",
```
tatomyr, line 15: Adjusted linting rules.

packages/core/src/rules/utils.ts
```diff
@@ -11,9 +11,27 @@ import type {
...
+  resolveFrom?: string
+): {
+  schema: T | undefined;
+  location: string | undefined;
+} => {
+  if (isRef(schemaOrRef)) {
+    const resolved = ctx.resolve<T>(schemaOrRef, resolveFrom);
+    return resolved
+      ? { schema: resolved.node, location: resolved.location?.source.absoluteRef }
+      : { schema: undefined, location: resolveFrom };
```
AlbinaBlazhko17, line 29:
```suggestion
      : { schema: undefined, location: undefined };
```
Is it better to return `undefined` if we cannot resolve schema? The same as we do inside `walk.ts` resolve function?

tatomyr: This is not exactly the case here since we know the location -- it's the current file. So no reason for making it undefined.

AlbinaBlazhko17: Make sense, but you are returning execution when schema is undefined:
`if (!resolved.schema || visited.has(resolved.schema))`

So, it looks like returning location is not neccessary,

tatomyr: You are correct, it's not. But if that changes, we'll get `undefined` where we should not get it.

PR #2705, opened 2026-03-31
fix: no-required-schema-properties rule to report when required properties are not defined in all branches

tatomyr:
## What/Why/How?
 

- Fixed the `no-required-schema-properties-undefined` rule to report when a required property is not defined in every `oneOf`/`anyOf` branch.
- Refactored the `no-required-schema-properties` rule code to avoid the 'grandparents' concept 
- Refactored tests: make the descriptions clearer and removed duplications.

Discovered internally.
 
## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Updates a core lint rule’s schema-walking logic, which may change reported violations (and potentially introduce false positives/negatives) across OpenAPI/AsyncAPI/Arazzo documents.
> 
> **Overview**
> Fixes `no-required-schema-properties-undefined` so a `required` property must be present in **every** `anyOf`/`oneOf` branch (while still honoring `allOf` and `$ref` chains), using a new `hasProperty` traversal with cycle protection and a simplified composition-root lookup.
> 
> Updates and expands the test suite to cover the new branch-aware behavior, `$ref`/`allOf` nesting cases, and adjusts diagnostics (message text from *"is undefined"* to *"is not defined"*) and expectations accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 954825287c318d011215cc14890af62b425fb059. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/rules/common/no-required-schema-properties-undefined.ts
+++ b/packages/core/src/rules/common/no-required-schema-properties-undefined.ts
@@ -1,6 +1,7 @@
 import type { Oas3Schema, Oas3_1Schema } from '../../typings/openapi.js';
 import type { Oas2Schema } from '../../typings/swagger.js';
 import { getOwn } from '../../utils/get-own.js';
+import { isNotEmptyArray } from '../../utils/is-not-empty-array.js';
 import type { Async2Rule, Async3Rule, Arazzo1Rule, Oas2Rule, Oas3Rule } from '../../visitors.js';
 import type { UserContext } from '../../walk.js';
 import { resolveSchema } from '../utils.js';
@@ -45,13 +46,15 @@ export const NoRequiredSchemaPropertiesUndefined:
           }
 
           if (
-            schema.anyOf?.every((s) => hasProperty(s, propertyName, new Set(visited), location))
+            isNotEmptyArray<AnySchema>(schema.anyOf) &&
+            schema.anyOf.every((s) => hasProperty(s, propertyName, new Set(visited), location))
           ) {
             return true;
           }
 
           if (
-            schema.oneOf?.every((s) => hasProperty(s, propertyName, new Set(visited), location))
+            isNotEmptyArray<AnySchema>(schema.oneOf) &&
+            schema.oneOf.every((s) => hasProperty(s, propertyName, new Set(visited), location))
           ) {
             return true;
           }


packages/core/src/rules/common/__tests__/no-required-schema-properties-undefined.test.ts
```diff
@@ -728,24 +610,24 @@ describe('no-required-schema-properties-undefined', () => {
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
   });
 
-  it('should NOT report if one or more of the required properties are defined when used in schema with anyOf keyword', async () => {
+  it('should report when required properties are not declared in all anyOf branches', async () => {
```
tatomyr, line 613: This is the actual test case added.

packages/core/src/rules/common/__tests__/no-required-schema-properties-undefined.test.ts
```diff
@@ -281,108 +295,6 @@ describe('no-required-schema-properties-undefined', () => {
     `);
   });
 
-  it('should not report required properties are present after resolving $refs when used in schema with allOf keyword', async () => {
```
tatomyr, line 284: Removed as duplicated.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -19,87 +22,70 @@ export const NoRequiredSchemaPropertiesUndefined:
...
-          schema: AnySchema,
-          from?: string,
-          includeFirstLevelExclusiveSchemas: boolean = true
-        ): Record<string, AnySchema> => {
-          // Check if the schema has been visited before processing it
-          if (visitedSchemas.has(schema)) {
-            return {};
-          }
-          visitedSchemas.add(schema);
+        const hasProperty = (
```
tatomyr, line 30: Refactored similarly to the [spec-discriminator-defaultMapping](https://github.com/Redocly/redocly-cli/blob/main/packages/core/src/rules/oas3/spec-discriminator-defaultMapping.ts#L27) rule.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -1,11 +1,14 @@
...
 import { getOwn } from '../../utils/get-own.js';
 import type { Async2Rule, Async3Rule, Arazzo1Rule, Oas2Rule, Oas3Rule } from '../../visitors.js';
 import type { UserContext } from '../../walk.js';
+import { resolveSchema } from '../utils.js';
 
-type AnySchema = Oas3Schema | Oas3_1Schema | Oas2Schema;
+type AnySchema =
+  | Oas3Schema
+  | Oas3_1Schema
+  | (Oas2Schema & { anyOf?: undefined; oneOf?: undefined });
```
DmitryAnansky, line 12: Why are we mixing OAS2 with anyOf and oneOf?
As far as I understand, these constructs are not supported by the specification.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -19,87 +22,70 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          if (!resolved.schema || visited.has(resolved.schema)) return false;
+          visited.add(resolved.schema);
 
-          if (isRef(schema)) {
-            const resolved = resolve<AnySchema>(schema, from);
-            if (!resolved.node) {
-              return {};
-            }
-            return elevateProperties(resolved.node, resolved.location?.source.absoluteRef);
+          if (
```
DmitryAnansky: Maybe we could make it more readable, like

```
const { schema, location } = resolved;

const hasDirectProperty =
  schema.properties &&
  getOwn(schema.properties, propertyName) !== undefined;

const checkSome = (schemas?: any[]) =>
  schemas?.some((s) => hasProperty(s, propertyName, visited, location));

const checkEvery = (schemas?: any[]) =>
  schemas?.every((s) => hasProperty(s, propertyName, visited, location));

if (
  hasDirectProperty ||
  checkSome(schema.allOf) ||
  checkEvery(schema.anyOf) ||
  checkEvery(schema.oneOf)
) {
  return true;
}
```

DmitryAnansky: or similar.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
tatomyr: We are not mixing, I've added both of them as `undefined` which technically doesn't change anything. However, it  allows the rule to cover all those specs at once.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
tatomyr: The point was to eagerly return once we've found the required property, but you're right--it might be not the most readable solution. I'll give it another try.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -19,91 +23,72 @@ export const NoRequiredSchemaPropertiesUndefined:
...
-        const allProperties = elevateProperties(schema);
-        const parentProperties = recursivelyGetParentProperties(location.pointer.split('/'));
+        const compositionRoot = findCompositionRoot(parents.length - 2, currentSchema);
 
-        for (const [i, requiredProperty] of schema.required.entries()) {
+        for (const [i, requiredProperty] of currentSchema.required.entries()) {
           if (
-            (!allProperties || getOwn(allProperties, requiredProperty) === undefined) &&
-            (!parentProperties || getOwn(parentProperties, requiredProperty) === undefined)
+            !hasProperty(currentSchema, requiredProperty, new Set()) &&
```
DmitryAnansky, line 86: Could we extract some variables to make the conditions in this if statement clearer and easier to understand?

tatomyr: Yes, but that will have performance impact. Moreover,  I'm not convinced it will add much to readability, as 
`!hasProperty(currentSchema, requiredProperty, new Set())` will become just `currentSchemaDoesNotContainRequiredProperty` or something alike.

DmitryAnansky (approved):

Issue #2652, opened 2026-03-12
CLI `lint` command `--generate-ignore-file` option override previously generated ignore file


DmitryAnansky:
**Is your feature request related to a problem? Please describe.**

Currently, the CLI `lint` command with the `--generate-ignore-file` option overwrites the previously generated `.redocly.lint-ignore.yaml` file.

That leads to a number of issues:
- Some lint ignore records that were added manually or generated with a different configuration get deleted.
- Command is executed against all OAS description.

The another unsolved question is to decide how `scorecard-classic` rules should be taken into consideration or handled separatelly.

**Describe the solution you'd like**

- When the `lint` command with the `--generate-ignore-file` option is executed against a specific description, only the records related to that file should be updated or appended.
- [UNSOLVED MYSTERY] - there should be some way to get into consideration scorecard level individual rules and allow apply ignore per API description.

**Describe alternatives you've considered**

Keep manually updating ignore file after generation.

**Additional context**

harshit078:
Hi @DmitryAnansky , should we close this issue ?

tatomyr:
Yes, thanks @harshit078.

PR #2716, opened 2026-04-03
fix: enhance linting process to preserve existing files

harshit078:
## What/Why/How?

- `redocly lint api1.yaml --generate-ignore-file` updates only entries related to api1.yaml and its referenced files

## Reference

- #2652

## Check yourself

- [x] Code follows the [contributing guide](https://github.com/Redocly/redocly-cli/blob/main/CONTRIBUTING.md#pull-request-guidelines) 

## Security

### DIFF
--- a/.changeset/rotten-wombats-cry.md
+++ b/.changeset/rotten-wombats-cry.md
@@ -1,5 +1,5 @@
 ---
-"@redocly/cli": major
+"@redocly/cli": minor
 ---
 
 redocly lint api1.yaml --generate-ignore-file updates only entries related to api1.yaml and its referenced files
--- a/.changeset/rotten-wombats-cry.md
+++ b/.changeset/rotten-wombats-cry.md
@@ -1,0 +1,5 @@
+---
+"@redocly/cli": major
+---
+
+redocly lint api1.yaml --generate-ignore-file updates only entries related to api1.yaml and its referenced files
--- a/.changeset/rotten-wombats-cry.md
+++ b/.changeset/rotten-wombats-cry.md
@@ -2,4 +2,4 @@
 "@redocly/cli": minor
 ---
 
-redocly lint api1.yaml --generate-ignore-file updates only entries related to api1.yaml and its referenced files
+Changed lint behavior with the `--generate-ignore-file` option — now it updates only the entries related to the file being linted, leaving other files' entries untouched.
--- a/.changeset/rotten-wombats-cry.md
+++ b/.changeset/rotten-wombats-cry.md
@@ -2,4 +2,6 @@
 "@redocly/cli": minor
 ---
 
-Changed lint behavior with the `--generate-ignore-file` option — now it updates only the entries related to the file being linted, leaving other files' entries untouched.
+Changed `lint` behavior with the `--generate-ignore-file` option.
+Now `lint` updates only the entries related to the file being linted.
+Other files' entries are unchanged.
--- a/docs/@v2/commands/lint.md
+++ b/docs/@v2/commands/lint.md
@@ -372,6 +372,12 @@ The errors in the ignore file `.redocly.lint-ignore.yaml` are ignored when the `
 
 If you run:
 
+```bash
+redocly lint api1.yaml --generate-ignore-file
+```
+
+Only the ignore entries related to api1.yaml are updated.
+Ignore entries for other API descriptions remain unchanged.
 
 To generate an ignore file for multiple API descriptions, pass them as arguments:
 
--- a/docs/@v2/commands/lint.md
+++ b/docs/@v2/commands/lint.md
@@ -352,7 +352,8 @@ This option is useful when you have an API design standard, but have some except
 It allows for highly granular control.
 
 {% admonition type="warning" %}
-This command updates ignore entries only for the API descriptions passed to the command. Existing entries for other API descriptions are preserved.
+This command updates ignore entries only for the API descriptions passed to the command.
+Existing entries for other API descriptions are preserved.
 {% /admonition %}
 
 The following command runs `lint` and adds all the errors to an ignore file:
@@ -369,13 +370,8 @@ Generated ignore file with 3 problems.
 
 The errors in the ignore file `.redocly.lint-ignore.yaml` are ignored when the `lint` command is run.
 
-if you run:
+If you run:
 
-```bash
-redocly lint api1.yaml --generate-ignore-file
-```
-
-only ignore entries related to `api1.yaml` are updated. Ignore entries for other API descriptions remain unchanged.
 
 To generate an ignore file for multiple API descriptions, pass them as arguments:
 
--- a/docs/@v2/commands/lint.md
+++ b/docs/@v2/commands/lint.md
@@ -352,7 +352,7 @@ This option is useful when you have an API design standard, but have some except
 It allows for highly granular control.
 
 {% admonition type="warning" %}
-This command overwrites an existing ignore file.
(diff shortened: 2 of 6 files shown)


tatomyr:
Thank you for your contribution, @harshit078!
Will this clean up other ignored entries related to that API file? 
Could you add a test with ignored rules for two files, and then re-generate the entries only for one file, assuming that one of the problems in this file has been resolved, so when ignoring, it only adds entries that are relevant to the new version of the linted file.

Say, the starting ignore file:

```yaml
ignore-bar.yaml:
  no-empty-servers:
    - '#/servers'
  operation-summary:
    - '#/paths/~1items/get/summary'
ignore-foo.yaml:
  no-empty-servers:
    - '#/servers'
```

Then you fix the `operation-summary` rule in ignore-bar.yaml and re-generate the ignore file only for it.
What you should get is this (the ignore entry for foo is preserved, and bar is updated):

```yaml
ignore-bar.yaml:
  no-empty-servers:
    - '#/servers'
ignore-foo.yaml:
  no-empty-servers:
    - '#/servers'
```

harshit078:
@tatomyr Yes, it will clean up entries for the linted API file but only for files that still have problems in current run. Got it, I'll add the test

packages/cli/src/commands/lint.ts
```diff
@@ -110,13 +113,25 @@ export async function handleLint({
...
+    }
+    for (const absRef of lintedAbsRefs) {
+      delete config.ignore[absRef];
+    }
+    for (const m of ignoreFileResults) {
+      config.addIgnore(m);
+    }
     config.saveIgnore();
     logger.info(
       `Generated ignore file with ${totalIgnored} ${pluralize('problem', totalIgnored)}.\n\n`
```
tatomyr: The message is somewhat misleading, as now the existing ignore file could be enhanced. What if we write it like so?

```suggestion
      `Explicitly ignored ${totalIgnored} ${pluralize('problem', totalIgnored)}.\n\n`
```

cc @Redocly/technical-writers

packages/cli/src/commands/lint.ts
```diff
@@ -110,13 +113,25 @@ export async function handleLint({
...
-      handleError(e, path);
+      handleError(e, apiPath);
     }
   }
 
   if (argv['generate-ignore-file']) {
+    for (const m of ignoreFileResults) {
+      const loc = m.location?.[0];
+      if (loc?.pointer !== undefined) {
+        lintedAbsRefs.add(loc.source.absoluteRef);
```
tatomyr: I don't get why do you need to add `lintedAbsRefs` two times. Moreover, you already have the linted file `apiPath` and you can use it to cleanup the config earlier. The logic becomes scattered  across the file, and I'd prefer to consilidate it in two locations (if possible): 
- cleanup the file we're linting in the ignore file
- use the `.addIgnore` as before

tatomyr (commented):
Left a couple of suggestions for improvement. Could you also check if the corresponding docs need to be updated?

packages/cli/src/commands/lint.ts
harshit078: I agree to your suggestion.

packages/cli/src/commands/lint.ts
harshit078: hmm, thanks for pointing it out. I agree, we can definitely simplify the logic. I'll make the changes !

.changeset/rotten-wombats-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": major
```
tatomyr: Should be minor.

harshit078: ok !

packages/cli/src/commands/lint.ts
```diff
@@ -51,14 +53,11 @@ export async function handleLint({
...
 
-  if (argv['generate-ignore-file']) {
-    config.ignore = {}; // clear ignore
-  }
   const totals: Totals = { errors: 0, warnings: 0, ignored: 0 };
   let totalIgnored = 0;
 
   // TODO: use shared externalRef resolver, blocked by preprocessors now as they can mutate documents
-  for (const { path, alias } of apis) {
+  for (const { path: apiPath, alias } of apis) {
```
tatomyr: What's the need for renaming `path`? Could you revert that so there's fewer changes?

packages/cli/src/commands/lint.ts
```diff
@@ -95,6 +94,8 @@ export async function handleLint({
       totals.ignored += fileTotals.ignored;
 
       if (argv['generate-ignore-file']) {
+        const apiAbsRef = isAbsoluteUrl(apiPath) ? apiPath : resolvePath(apiPath);
```
tatomyr: This check seems incorrect. The ref should align with `mappedDefinitionName` from `Config.saveIgnore`. I believe it would be cleaner to move this and the following lines into a new cleanup method next to `saveIgnore` and use it here.

.changeset/rotten-wombats-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+redocly lint api1.yaml --generate-ignore-file updates only entries related to api1.yaml and its referenced files
```
tatomyr:
```suggestion
Changed lint behavior with the `--generate-ignore-file` option — now it updates only the entries related to the file being linted, leaving other files' entries untouched.
```

tatomyr (commented):
Please do consider updating the corresponding docs in **docs/@v2/commands/lint.md**.

.changeset/rotten-wombats-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/cli": minor
+---
+
+Changed lint behavior with the `--generate-ignore-file` option — now it updates only the entries related to the file being linted, leaving other files' entries untouched.
```
JLekawa:
```suggestion
Changed `lint` behavior with the `--generate-ignore-file` option.
Now `lint` updates only the entries related to the file being linted.
Other files' entries are unchanged.
```

docs/@v2/commands/lint.md
```diff
@@ -352,7 +352,7 @@ This option is useful when you have an API design standard, but have some except
 It allows for highly granular control.
 
 {% admonition type="warning" %}
-This command overwrites an existing ignore file.
+This command updates ignore entries only for the API descriptions passed to the command. Existing entries for other API descriptions are preserved.
```
JLekawa:
```suggestion
This command updates ignore entries only for the API descriptions passed to the command.
Existing entries for other API descriptions are preserved.
```

docs/@v2/commands/lint.md
```diff
@@ -369,6 +369,14 @@ Generated ignore file with 3 problems.
 
 The errors in the ignore file `.redocly.lint-ignore.yaml` are ignored when the `lint` command is run.
 
+if you run:
+
+```bash
+redocly lint api1.yaml --generate-ignore-file
+```
+
+only ignore entries related to `api1.yaml` are updated. Ignore entries for other API descriptions remain unchanged.
```
JLekawa:
```suggestion
If you run:

```bash
redocly lint api1.yaml --generate-ignore-file
```

Only the ignore entries related to `api1.yaml` are updated.
Ignore entries for other API descriptions remain unchanged.
```

JLekawa (approved):

tatomyr:
@harshit078, your tests are failing. Could you fix that?

harshit078:
@tatomyr pushed a fix for the failing test. thanks !

tatomyr (approved):
LGTM.

Issue #2639, opened 2026-03-06
Respect doesn't resolve complex constraint pattern


AlbinaBlazhko17:
**Describe the bug**
AJV validation error when using a discriminator with a schema property that uses allOf + not pattern instead of explicit enum values. The discriminator expects the mapped schema's discriminator property to have const or enum, but the AlternativeInstrument schema uses a more complex constraint pattern.

‼️ Error: Ajv error: discriminator: "properties/method" must have "const" or "enum"

**To Reproduce**
1. Given the `openapi.yaml` file with the AlternativeInstrument schema that uses allOf + not:
```yaml
AlternativeInstrument:
  properties:
    method:
      allOf:
        - $ref: '#/components/schemas/AlternativePaymentMethods'
        - not:
            enum:
              - payment-card
              - paypal
              - ach
```
2. And the `openapi.arazzo.yaml` workflow file
3. Run: `redocly respect openapi.arazzo.yaml`
4. See the error: `Ajv error: discriminator: "properties/method" must have "const" or "enum"`

**Expected behavior**
The schema should br validated successfully.

**Logs**
```
Workflow name: test-cash-payment
 
  stepId - create-cash-instrument 
  ✗ schema check 
     Ajv error: discriminator: "properties/method" must have "const" or "enum"
```

**OpenAPI description**

**Redocly version(s)**
2.20.4

**Node.js version(s)**
v24

**OS, environment**
macOs

**Additional context**
If we use directly enum without `allOf` + `not` pattern everything works as expected.

[complex-schema-repro.zip](https://github.com/user-attachments/files/25788884/complex-schema-repro.zip)

tatomyr:
Note: The behaviour should be the same for both lint and respect commands.

tatomyr:
Related: https://github.com/Redocly/redocly-cli/issues/1232#issuecomment-4223517962. So the behavior might be actually correct.

PR #2720, opened 2026-04-06
fix(core): fix Ajv error discriminator 

harshit078:
## What/Why/How?

- This PR addresses issue #2639
- added a utility function  `remove-discriminators-from-schema` which just strips discriminator keywords from schemas
- added this logic before AJV validation 

## Check yourself

- [ ] Documentation update has been considered 

## Security

---

> [!NOTE]
> **Medium Risk**
> Touches core response schema validation behavior; while the change is targeted, normalizing schemas could affect edge-case discriminator matching or validation outcomes if the promoted `const`/`enum` is ambiguous.
> 
> **Overview**
> Fixes AJV discriminator validation crashes by preprocessing dereferenced schemas so discriminator tag properties expose `const`/`enum` at the top level even when they’re nested inside `allOf` (while keeping the original `allOf` constraints intact).
> 
> `checkSchema` now runs schemas through the new `normalizeDiscriminatorSchemas` utility before AJV validation, and tests were added/updated to cover `oneOf`/`anyOf` discriminators with `allOf + not` patterns and to ensure failures are reported as normal validation errors (not `Ajv error`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ce2215961b2fd1555c067557af1d58a36609d9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


DmitryAnansky:
@harshit078 
Could you please elaborate on why you believe removing discriminators is a good fix for this issue?
Have you considered any alternative approaches?

.changeset/sixty-falcons-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/respect-core": major
```
DmitryAnansky: Please use `patch` for fixes.

DmitryAnansky (changes requested):

harshit078:
@DmitryAnansky sure, 
my intent/logic behind creating a function which removes discriminators is because when I researched about OpenAPI discriminator schema I found out that it serves mainly one purpose which is its optimises routing and act as routing shortcut.
The alternatives I considered before submitting PR were - 

- Selective stripping: this was more complex than intented and had to have many edge cases
- making discriminator as false : it would have affected globally rather than per schema
- brute force with simple Catch error and retry: same as first point

This is my POV of issue. If you think approach can be better, I am ready to implement it. Thanks :)

.changeset/sixty-falcons-glow.md
harshit078: sure

packages/respect-core/src/utils/remove-discriminators-from-schema.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * removes `discriminator` keywords from a fully-dereferenced schema.
+ */
+export function removeDiscriminatorsFromSchema(schema: any): any {
+  if (!schema || typeof schema !== 'object') return schema;
+
+  const schemaCopy = JSON.parse(JSON.stringify(schema));
```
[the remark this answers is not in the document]
harshit078: thanks for this mistake ! I agree

packages/respect-core/src/modules/flow-runner/schema/schema-checker.ts
AlbinaBlazhko17, line 24: Thanks @harshit078 for the contribution. You don't need to remove discriminator manually, you can just drop this setting from the `Ajv` and it will not cause the error. The issue looks more deeply. Could you please investigate the root cause?

harshit078: sure, I'll look into it

packages/respect-core/src/modules/flow-runner/schema/schema-checker.ts
```diff
@@ -20,7 +20,7 @@ const ajvStrict = new Ajv({
   strictSchema: false,
   inlineRefs: false,
   validateSchema: false,
-  discriminator: true,
+  discriminator: false,
```
DmitryAnansky: I’d suggest taking a deeper look into this issue instead of disabling discriminator in Ajv options. It would be worthwhile to review the Ajv implementation and verify how it handles this particular case.

DmitryAnansky:
Hi, @harshit078 
Thank you for the time you’ve invested in investigating this issue — it’s greatly appreciated.
It appears to be a more complex problem that may require deeper investigation or even changes in external libraries (such as Ajv). We’ll take it forward and handle it internally from here.
Closing this PR for now.

harshit078:
Hey @DmitryAnansky ,
No worries, while resolving the issue I eventually realised that fixing the issue from the root ajv will solve this problem. On digging deep into ajv repo, I found out that we can recursively search for const/enum terms in `addMappings` function in index.ts under discriminator. This would resolve the problem once for all.

https://github.com/ajv-validator/ajv/blob/master/lib/vocabularies/discriminator/index.ts

PR #2732, opened 2026-04-09
feat: add cleaning cache for plugin files

kanoru3101:
## What/Why/How?
In VSCE, we have a problem with clean cache for plugin files. This PR adds cleaning the general cache for plugin files, cjs files, and esm files.

* Moved all logic with plugin caches to separate files and covered by tests
* Cleaning cache for ESM and СJS files, cover cases with nested imports.
* Removed `__webpack_require__`

## Reference
Issue https://github.com/Redocly/redocly-vs-code/issues/10
VSCE Part https://github.com/Redocly/redocly/pull/22293
## Testing
Tested in [VSCE part](https://github.com/Redocly/redocly/pull/22293)
## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Touches plugin loading/caching and introduces cache-busting behavior for both CJS and ESM (with Node-version-dependent hooks), which could affect plugin reload semantics and module singletons.
> 
> **Overview**
> Adds a dedicated `plugins-cache` module and wires `resolvePlugins` to use it, enabling explicit cache management instead of the previous in-function map and removing the special `__webpack_require__` path.
> 
> Introduces `clearPluginsCache()` (CJS `require.cache` eviction for local deps + ESM cache-busting via `?v=` versioning and a Node `registerHooks` resolver that propagates the version to nested `file:` imports while skipping `node_modules`). Exposes `clearPluginsCache`/`getPluginCacheVersion` via config/core exports and adds tests/fixtures to cover reload behavior and dependency eviction, plus a changeset bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28eb234275dedcea7ef0afc01dc93b5cc528a34a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


JLekawa (approved):

packages/core/src/config/__tests__/fixtures/resolve-config/local-config-with-plugin-init.yaml
```diff
@@ -1,2 +1,2 @@
 plugins:
-  - plugin-with-init-logic.js
+  - plugin-with-init-logic.cjs
```
DmitryAnansky, line 2: why this become .cjs?

kanoru3101: This file uses `cjs` (`require`, `module.exports`). `packages/core` is `"type": "module"`, so `.js` would be parsed as `ESM` and fail. `.cjs` forces Node to use the `CommonJS` loader.

DmitryAnansky (approved):

packages/core/src/config/__tests__/fixtures/resolve-config/local-config-with-plugin-init.yaml
tatomyr: Can we use one cjs and one mjs to cover ESM setup a well? Or we have such tests already?

packages/core/src/config/config-resolvers.ts
```diff
@@ -40,6 +40,23 @@ import {
 
 // Cache instantiated plugins during a single execution
 const pluginsCache: Map<string, Plugin[]> = new Map();
+let pluginsCacheVersion = 0;
+
+export const clearPluginsCache = (): void => {
+  for (const pluginPath of pluginsCache.keys()) {
+    if (pluginPath.endsWith('.cjs')) {
```
tatomyr: Plugins could be in other formats too. Basically anything that could be interpreted as  a JavaScript file. Is that covered?

kanoru3101: Yes, by default, if we have a `.js` file, it must be an `ESM` type. But for the `ESM` type, we can’t clear the cache (this is blocked at the engine level), so we use a different method (we add a query parameter—this is exactly how they solve this problem in Vite). And for `.cjs` files, we can clear the cache.

My only concern is whether we'll be able to better determine whether a file has `CommonJS` or `ESM` type. Because the method based on the file extension doesn't look very good.

packages/core/src/config/__tests__/fixtures/resolve-config/local-config-with-plugin-init.yaml
kanoru3101: @tatomyr, added test with ESM file

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+import * as url from 'node:url';
+
+import type { Plugin } from './types.js';
+
+// Cache instantiated plugins during a single execution
+export const pluginsCache: Map<string, Plugin[]> = new Map();
+let pluginsCacheVersion = 0;
+let isEsmCacheBustHookRegistered = false;
+
+const ESM_CACHE_BUST_HOOK_SOURCE = `
```
kanoru3101: ESM resolver hook (registered via module.register) that propagates the ?v=N cache-busting query parameter from a parent module's URL to its nested imports, so clearing the cache reloads the entire plugin subgraph instead of just the entry file

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,68 @@
+import module from 'node:module';
+import * as path from 'node:path';
+import * as url from 'node:url';
+
+import type { Plugin } from './types.js';
+
+// Cache instantiated plugins during a single execution
+export const pluginsCache: Map<string, Plugin[]> = new Map();
```
tatomyr: I wouldn't export it as is. In my opinion, it's safer to export a getter instead, so noone will unintentionally modify it.

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+  try {
+    module.register(
+      `data:text/javascript,${encodeURIComponent(ESM_CACHE_BUST_HOOK_SOURCE)}`,
+      import.meta.url
+    );
+    isEsmCacheBustHookRegistered = true;
+  } catch {
+    // silently fail; without the hook only the entry plugin is cache-busted.
+    // The flag stays false so a later `clearPluginsCache()` will retry.
+  }
```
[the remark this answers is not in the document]
kanoru3101: module.registerHooks() is not supported for node 20

packages/core/src/config/plugins-cache.ts
kanoru3101: Created functions for working with pluginsCache: `hasCachedPlugin`, `getCachedPlugins` and `setCachedPlugins`

packages/core/src/config/__tests__/plugins-cache.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      const second = await loadPluginModule(cjsPluginPath);
+
+      expect(second).toBe(first);
+    });
+
+    it('should return the same module on subsequent imports of the same esm plugin', async () => {
+      const first = await loadPluginModule(esmPluginPath);
+      const second = await loadPluginModule(esmPluginPath);
+
+      expect(second).toBe(first);
```
kanoru3101: Think about these two test. Check in vitest. `===`

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+const PLUGIN_VERSION_PARAM = 'v';
+
+// Cleared in `clearPluginsCache()` so the next load re-walks the directory; if any
+// file changed on disk, the new `?v=` busts Node's ESM cache.
+let cacheVersion = 0;
+
+let isEsmCacheBustHookRegistered = false;
+
+function registerEsmCacheBustHook(): void {
+  module.registerHooks({
```
kanoru3101, line 18: Read docs about registerHooks and props

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+    resolve(specifier, context, nextResolve) {
+      const result = nextResolve(specifier, context);
+      if (!result.url.startsWith('file:')) return result;
+      if (!context.parentURL) return result;
```
kanoru3101, line 27: add comment what to do

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+      if (!parentVersion) return result;
+
+      const childURL = new URL(result.url);
+      if (childURL.pathname.includes('/node_modules/')) return result;
+      if (childURL.searchParams.has(PLUGIN_VERSION_PARAM)) return result;
+
+      // Propagate parent's version down the whole plugin graph so every nested
+      // URL is busted together with the entry.
+      childURL.searchParams.set(PLUGIN_VERSION_PARAM, parentVersion);
+      return { ...result, url: childURL.href, shortCircuit: true };
```
kanoru3101: Check shortCircuit if it needed

packages/core/src/config/plugins-cache.ts
```diff
@@ -0,0 +1,112 @@
...
+// Marks plugin-scoped URLs and carries the cache-bust version.
+const PLUGIN_VERSION_PARAM = 'v';
+
+// Cleared in `clearPluginsCache()` so the next load re-walks the directory; if any
+// file changed on disk, the new `?v=` busts Node's ESM cache.
+let cacheVersion = 0;
+
+let isEsmCacheBustHookRegistered = false;
+
+function registerEsmCacheBustHook(): void {
```
kanoru3101, line 17: Check if one project is opened in two different IDEs. How it works. Test with Cursor and VS-code

packages/core/src/config/__tests__/plugins-cache.test.ts
kanoru3101: Removed tests

packages/core/src/config/plugins-cache.ts
kanoru3101: Added description about these props

packages/core/src/config/plugins-cache.ts
kanoru3101: removed

packages/core/src/config/plugins-cache.ts
kanoru3101: It works. Added video to [PR](https://github.com/Redocly/redocly/pull/22293)

tatomyr (approved):

PR #2733, opened 2026-04-09
feat: setup the post-bundle decorators

### LINKED ISSUE #1783: remove-unused-components doesn't remove recursive schemas
**Describe the bug**

Using remove-unused-components decorator leave some schemas unused by operations but that are recursive. 

**To Reproduce**

```
decorators:
  remove-unused-components:
    type: remove-unused-components
    
apis:
  test:
    root: test.json
``` 

2. And this OpenAPI file(s)

``` 
{
    "openapi": "3.1.0",
    "components": {
      "schemas": {
        "iwontbedrop": {
          "properties": {
            "Myprop": {
              "items": {
                "anyOf": [
                  {
                    "$ref": "#/components/schemas/iwontbedrop"
                  }
                ]
              },
              "type": "array"
            }
          },
          "type": "object"
        },
        "iwillbedrop": {
          "properties": {
            "name": {
              "title": "Name",
              "type": "string"
            }
          },
          "type": "object"
        }
     }
  }
}
``` 
3. Run this command `redocly bundle`

**Expected behavior**

The schema `iwontbedrop` being recursive is not drop even if not used outside of it's scope. So the expectation would be to have both   `iwontbedrop` and `iwillbedrop` to be dropped

**Logs**
```
bundling .../test.json...
openapi: 3.1.0
components:
  schemas:
    iwontbedrop:
      properties:
        Myprop:
          items:
            anyOf:
              - $ref: '#/components/schemas/iwontbedrop'
          type: array
      type: object
📦 Created a bundle for .../test.json at stdout 13ms.
🧹 Removed 1 unused components.
``` 

**OpenAPI description**

Using version 3.1

**Redocly Version(s)**

1.25.8

**`Node.js` Version(s)**

22.4.0

**OS, environment**

MacOS Sonoma + Ubuntu

### LINKED ISSUE #2350: remove-unused-components removes parameters that are in use
**Describe the bug**

Adding the `remove-unused-components` decorator causes the bundle command to generate a bundled spec with no `components.parameters` included, even though the bundle still has multiple refs to `#/components/parameters/foobar`. This means that the generated bundle is invalid.

Adding the `--dereference` option causes all parameters to be pulled into the bundled spec, so I am confident that all the refs are pointing to the correct places. This is not a satisfactory workaround because I don't want a dereferenced spec.

Adding the `no-unused-components: error` lint doesn't change anything. This linting rule apparently doesn't find any unused components, but the decorator still manages to find 473 "unused" components to remove

**To Reproduce**

I haven't been able to create a minimal reproducible example, because it only seems to happen once the spec grows to a certain size, and/or a certain number/structure of refs are used. 

1. Create redocly.yaml with `remove-unused-components: on`
2. generate bundled spec with `redocly bundle foo@latest --output=bundled.yml`
3. Open bundled.yml and observe that there are no `components.parameters`

**Expected behavior**

All components that are reachable via any number of nested refs should be included in the bundle. 

The `no-unused-components` lint and the `remove-unused-components` decorator should agree on which components are unused.

**OpenAPI description**

openapi 3.0.3

**Redocly Version(s)**

2.3.1

**`Node.js` Version(s)**

24.9.0

**OS, environment**

macos 26.0.1

**Additional context**

AlbinaBlazhko17:
## What/Why/How?
Currently we run `remove-unused-components` as a regular decorator in the same walk as bundle phase. This creates a lot of side issues with refs in this decorator, which is hard to fix right in the decorator. I decided to run `remove-unused-components` after doc is bundled.

What i did:
- Added to  `bundleDocument` new phase to walk through the doc one more time and invoke `remove-unused-components` on bundled doc.
- Refactored `remove-unused-components` to track components by local pointer strings (#/components/schemas/Foo) instead of resolved Location objects. This removes the dependency on `resolve()` during ref tracking and makes "used-in" tracking simpler and more reliable when running post-bundle (where all refs are already inlined).
- Added tests to fully cover cases with `remove-unused-components`.

## Reference
Resolves #2350 
Resolves #1783 

## Check yourself

- [ ] Documentation update has been considered 

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes bundling flow to run `remove-unused-components` in a new post-bundle pass and refactors its reference-tracking logic; this can change final bundled output and component pruning behavior across OAS2/OAS3.
> 
> **Overview**
> Moves `remove-unused-components` out of the main bundle walk and into a dedicated **post-bundle** pass (with a fresh `resolveDocument`), ensuring components that become unused only after `$ref` resolution are removed.
> 
> Refactors the OAS2/OAS3 `remove-unused-components` decorators to track component usage via stable `#/...` pointer-derived keys (using `parseRef`) instead of resolved `Location` objects, and adds new e2e bundle fixtures covering parameter-to-schema refs and recursive refs. Includes a changeset bumping `@redocly/openapi-core` and `@redocly/cli` as minor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83d363af52f6eb0ccd9b041b91e721ece898fcdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlbinaBlazhko17:
@cursor review

JLekawa (approved):

packages/core/src/decorators/oas2/remove-unused-components.ts
```diff
@@ -1,78 +1,81 @@
...
+      usedIn: components.get(key)?.usedIn ?? [],
       componentType,
       name,
     });
   }
 
-  function removeUnusedComponents(root: Oas2Definition, removedPaths: string[]): number {
-    const removedLengthStart = removedPaths.length;
+  function getContainingComponentKey(pointer: string): string | undefined {
+    const parts = pointer.replace(/^#\//, '').split('/');
```
tatomyr: We should have a similar mechanism for working with pointers.

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -14,40 +14,44 @@ import type { Oas3Decorator } from '../../visitors.js';
 type AnyOas3Definition = Oas3Definition | Oas3_1Definition | Oas3_2Definition;
 type AnyOas3ComponentsKey = keyof Oas3Components | keyof Oas3_1Components | keyof Oas3_2Components;
 
+const COMPONENTS_PREFIX = '#/components/';
```
tatomyr: Should be already defined in constants.

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -14,40 +14,44 @@ import type { Oas3Decorator } from '../../visitors.js';
...
+    for (const [key, { usedIn, name, componentType }] of components) {
       const used = usedIn.some(
-        (location) =>
-          !removedPaths.some(
-            (removed) =>
-              location.absolutePointer.startsWith(removed) &&
-              (location.absolutePointer.length === removed.length ||
-                location.absolutePointer[removed.length] === '/')
-          )
+        (sourceKey) => sourceKey === undefined || !removedKeys.has(sourceKey)
```
tatomyr: This is confusing:

```
 sourceKey === undefined  
```

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -85,22 +89,17 @@ export const RemoveUnusedComponents: Oas3Decorator = () => {
...
-          const resolvedRef = resolve(ref);
-          if (!resolvedRef.location) return;
-
-          const [fileLocation, localPointer] = resolvedRef.location.absolutePointer.split('#', 2);
-          if (!localPointer) return;
-
-          const componentLevelLocalPointer = localPointer.split('/').slice(0, 4).join('/');
-          const pointer = `${fileLocation}#${componentLevelLocalPointer}`;
+          const targetPointer = getContainingComponentKey(ref.$ref);
+          if (!targetPointer) return;
```
tatomyr, line 88: It's not obvious that external refs are undefined, so we can skip registering it.

packages/core/src/bundle/bundle-document.ts
```diff
@@ -127,6 +134,14 @@ export async function bundleDocument(opts: {
     ctx,
   });
 
+  await runPostBundleDecorators({
```
tatomyr: I'd prefer writing this inline like we do with preporcessors above.

tests/e2e/bundle/bundle-remove-unused-components-from-config/oas3-parameter-ref-to-schema-with-unused-schema-opposite-ref/parameters.yaml
```diff
@@ -0,0 +1,6 @@
+Param:
+  name: param
+  in: query
+  required: true
+  schema:
+    $ref: schemas.yaml#/Schema
```
tatomyr:
```suggestion
    $ref: schemas.yaml#/Schema

```

packages/core/src/decorators/oas2/remove-unused-components.ts
AlbinaBlazhko17: Thanks, replaced with `parseRef`, which has escape functionality.

packages/core/src/decorators/oas3/remove-unused-components.ts
AlbinaBlazhko17: Changed logic

packages/core/src/bundle/bundle-document.ts
AlbinaBlazhko17: Thanks, did this

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -14,66 +14,65 @@ import type { Oas3Decorator } from '../../visitors.js';
...
     });
   }
 
-  function removeUnusedComponents(root: AnyOas3Definition, removedPaths: string[]): number {
-    const removedLengthStart = removedPaths.length;
+  function removeUnusedComponents(
+    root: AnyOas3Definition,
+    removedKeys: Set<string> = new Set()
+  ): number {
+    const countBefore = removedKeys.size;
```
tatomyr:
```suggestion
    const removedCountBefore = removedKeys.size;
```

packages/core/src/decorators/oas3/remove-unused-components.ts
```diff
@@ -85,22 +84,17 @@ export const RemoveUnusedComponents: Oas3Decorator = () => {
...
         ) {
-          const resolvedRef = resolve(ref);
-          if (!resolvedRef.location) return;
-
-          const [fileLocation, localPointer] = resolvedRef.location.absolutePointer.split('#', 2);
-          if (!localPointer) return;
-
-          const componentLevelLocalPointer = localPointer.split('/').slice(0, 4).join('/');
-          const pointer = `${fileLocation}#${componentLevelLocalPointer}`;
+          const targetPointer = getContainingComponentKey(ref.$ref);
```
tatomyr:
```suggestion
          const targetPointer = getComponentKey(ref.$ref);
```

tatomyr: Maybe use something like `isComponent` instead?

AlbinaBlazhko17: I think it is better to keep `componentKey` instead of raw `location.pointer`, because we have one unified type of targetPointers and we can easily compare them. For example, when we have a circular ref, we can easily compare it. In this case we will have two equal refs: `schemas/RecursiveRef`. In case of just local pointers we will have `#/components/schemas/RecursiveRef` and `#/components/schemas/RecursiveRef/properties/prop/items/anyOf/0` and it will be hard to compare.
```yaml
components: 
  schemas: 
    RecursiveRef:
      type: object
      properties: 
        prop:
          type: array
          items: 
            anyOf: 
              - $ref: '#/components/schemas/RecursiveRef'
```

packages/core/src/decorators/oas3/remove-unused-components.ts
AlbinaBlazhko17: Bundled doc should have only internal refs, so if we have some external ref the error occurred and how we can identify with what we can compare this external ref to delete.

packages/core/src/decorators/oas2/index.ts
```diff
@@ -16,5 +16,5 @@ export const decorators = {
   'remove-x-internal': RemoveXInternal as Oas2Decorator,
   'filter-in': FilterIn as Oas2Decorator,
   'filter-out': FilterOut as Oas2Decorator,
-  'remove-unused-components': RemoveUnusedComponents, // always the last one
+  'remove-unused-components': RemoveUnusedComponents,
```
tatomyr, line 19: Do we still need it here?

.changeset/tame-spoons-show.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
```
tatomyr, line 2: It also affects how the CLI behaves and I think it would be useful to specify it directly in it's changelog.

tatomyr (approved):
Left minor comments. Otherwise looks good to me.

.changeset/tame-spoons-show.md
AlbinaBlazhko17: Thanks, will add it.

packages/core/src/decorators/oas2/index.ts
AlbinaBlazhko17: Yes, we are initing rules and decorators using `initRules` and pass to it spec version to get config for decorator. Without this it will break `remove-unused-components` in config.

tatomyr: Indeed.

.changeset/tame-spoons-show.md
```diff
@@ -0,0 +1,6 @@
...
+"@redocly/cli": minor
+---
+
+Moved the `remove-unused-components` decorator to a post-bundle phase so that components that become unused only after `$ref` resolution are correctly removed.
```
JLekawa:
```suggestion
Moved the `remove-unused-components` decorator to the post-bundle phase so that components that become unused only after `$ref` resolution are correctly removed.
```

JLekawa (approved):

Issue #2667, opened 2026-03-17
Add `asyncapi-operation-security-defined` lint rule — AsyncAPI operations can reference undefined security schemes with no warning


saismrutiranjan18:
### Problem
The AsyncAPI rule set has only structural rules (`channels-kebab-case`, `no-channel-trailing-slash`). There is no rule validating that security scheme names in AsyncAPI operation/server `security` arrays are defined in `components/securitySchemes`. Name mismatches are silently invalid.

### Proposed Rule
```yaml
rules:
  asyncapi-operation-security-defined: error
```

Analogous to the missing `operation-security-defined` rule for OpenAPI — traverses AsyncAPI operation and server security arrays and validates each referenced scheme exists in `components/securitySchemes`.

### Labels
`Type: Enhancement`, `p3`

DmitryAnansky:
@saismrutiranjan18 
Should be available in v`2.44.0`.
Thank you.

PR #2759, opened 2026-04-21
feat(core): add asyncapi security-defined lint rule

harshit078:
## What/Why/How?

- Added `asyncapi-operation-security-defined` rule for AsyncAPI 2.x and 3.x. both which reports when a security scheme referenced from an operation or server `security` array is not defined in ``components.securitySchemes`.`

## Reference

https://github.com/Redocly/redocly-cli/issues/2667

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> New default `error` in recommended rulesets can break CI for existing AsyncAPI specs; security lint logic is complex (multi-file refs, server applicability) though heavily tested.
> 
> **Overview**
> Adds a **`security-defined`** built-in rule for **AsyncAPI 2.x and 3.x**, wired into rule registries, preset configs (`minimal` at warn, `recommended`/`all` at error), and docs/changeset.
> 
> **AsyncAPI 2.x** validates named schemes on operation/server `security` against `components.securitySchemes`, and requires each operation to have security unless every applicable root server already does (channel server bindings, traits, and component-only definitions are handled in tests).
> 
> **AsyncAPI 3.x** validates `$ref` targets (including external files and map-level `servers`/`operations` refs), rejects refs outside `components.securitySchemes`, and applies the same operation-vs-server security coverage rules.
> 
> Shared helpers **`hasSecurityRequirements`** / **`isAsyncOperationSecured`** were added in `rules/utils.ts`; AsyncAPI typings and visitor hooks were extended to support the new checks. CLI AsyncAPI **split** code got a small **`AnyAsyncApiComponents`** typing refactor only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bce0a3207fd352f6c86ac30fb198671dc662448. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/core/src/rules/async2/asyncapi-operation-security-defined.ts
```diff
@@ -0,0 +1,16 @@
+import type { Async2Rule } from '../../visitors.js';
+import type { UserContext } from '../../walk.js';
+import { createSecuritySchemeReferencesChecker } from '../common/security-scheme-references.js';
+
+export const AsyncApiOperationSecurityDefined: Async2Rule = () => {
```
vadyvas: I think it would be better to keep this logic inside the AsyncAPI rule instead of sharing it with the OAS rule

packages/core/src/rules/common/security-scheme-references.ts
```diff
@@ -0,0 +1,38 @@
+import type { Location } from '../../ref-utils.js';
+import type { UserContext } from '../../walk.js';
+
+type ReferencedSchemes = Map<string, { defined?: boolean; from: Location[] }>;
+
+export function createSecuritySchemeReferencesChecker() {
```
vadyvas: I am not sure we need a shared helper here

For this PR, I think it may be better to avoid changing the existing OAS implementation and keep the AsyncAPI version separate, even if some code is duplicated

packages/core/src/rules/async2/index.ts
```diff
@@ -23,6 +24,7 @@ export const rules: Async2RuleSet<'built-in'> = {
   'info-contact': InfoContact as Async2Rule,
   'info-license-strict': InfoLicenseStrict as Async2Rule,
   'operation-operationId': OperationOperationId as Async2Rule,
+  'asyncapi-operation-security-defined': AsyncApiOperationSecurityDefined,
```
vadyvas: Could we use the same rule name `security-defined` for AsyncAPI as well?

Then the AsyncAPI ruleset can register its own implementation under the same name. I think this would make the rule naming more consistent for users.

docs/@v1/guides/lint-asyncapi.md
```diff
@@ -53,6 +53,7 @@ some of the built-in rules. The currently-supported rules are:
 
 - `info-contact`: the `Info` section must contain a valid `Contact` field.
 - `operation-operationId`: every operation must have a valid `operationId`.
+- `asyncapi-operation-security-defined`: every scheme referenced from an operation or server `security` array must be defined in `components.securitySchemes`.
```
vadyvas: I think we can skip the v1 docs in this PR.

vadyvas (changes requested):
I would suggest a slightly different approach:

1. keep the AsyncAPI logic separate and do not reuse shared logic from the OAS rule
2. use the same rule name, `security-defined`, for AsyncAPI as well, and register the AsyncAPI implementation in the AsyncAPI ruleset
3. do not update the v1 docs in this PR

I think this would make the change smaller, clearer, and safer.

Thank you for the contribution, overall the PR looks good

packages/core/src/rules/async2/index.ts
```diff
@@ -23,6 +24,7 @@ export const rules: Async2RuleSet<'built-in'> = {
   'info-contact': InfoContact as Async2Rule,
   'info-license-strict': InfoLicenseStrict as Async2Rule,
   'operation-operationId': OperationOperationId as Async2Rule,
+  'security-defined': SecurityDefined,
```
vadyvas, line 28: Please add support for AsyncAPI 3 as well. Right now the rule only applies to AsyncAPI2

docs/@v2/rules/async/security-defined.md
vadyvas, line 1: The code uses the rule name `security-defined`, but the docs still say `asyncapi-operation-security-defined`
Can you update related changes?

docs/@v1/rules/async/asyncapi-operation-security-defined.md
```diff
@@ -0,0 +1,83 @@
+# asyncapi-operation-security-defined
```
vadyvas: Please don’t add this rule to the v1 docs

vadyvas (changes requested):
left a few comments, could you take a look?

packages/core/src/rules/async2/index.ts
harshit078: okay sure !

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,31 @@
+import type { Async3Rule } from '../../visitors.js';
+
+const SECURITY_LOCATION = /\/security\/\d+$/;
+const COMPONENT_SCHEME_FRAGMENT = /^\/components\/securitySchemes\/([^/]+)$/;
+
+export const SecurityDefined: Async3Rule = () => {
+  return {
+    ref: {
+      leave(node, { location, report }, resolved) {
+        if (!SECURITY_LOCATION.test(location.pointer)) return;
```
AlbinaBlazhko17: Please, avoid using regex pattern to check the location. For the rules we have visitors pattern, so you can use `SecurityScheme` visitor here.

docs/@v1/rules/ruleset-templates.md
```diff
@@ -438,6 +438,7 @@ rules:
   info-contact: off
   info-license-strict: warn
   operation-operationId: warn
+  asyncapi-operation-security-defined: error
```
AlbinaBlazhko17: It looks like you forgot to remove the new rule from the v1 docs. Please, remove this line

packages/core/src/rules/async3/security-defined.ts
harshit078: sure thanks for the help. I'll push a fix for it

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,30 @@
+import type { Async3Rule } from '../../visitors.js';
+
+const COMPONENT_SCHEME_PREFIX = '#/components/securitySchemes/';
+
+export const SecurityDefined: Async3Rule = () => {
+  return {
+    ref: {
+      leave(node, { type, location, report }, resolved) {
+        if (type.name !== 'SecurityScheme') return;
```
AlbinaBlazhko17: Please, follow our visitors pattern as you did in rule for async2

.changeset/stupid-dryers-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+Added `asyncapi-operation-security-defined` rule for AsyncAPI 2.x that reports when a security scheme referenced from an operation or server `security` array is not defined in `components.securitySchemes`.
```
AlbinaBlazhko17: Also, please, change naming of the rule here as well

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,43 @@
...
+  >();
+
+  return {
+    Root: {
+      leave(_root: unknown, { report }: UserContext) {
+        for (const [name, scheme] of referencedSchemes.entries()) {
+          if (scheme.defined) continue;
+          for (const reportedFromLocation of scheme.from) {
+            report({
+              message: `There is no \`${name}\` security scheme defined.`,
```
AlbinaBlazhko17, line 35: I think we should align these rules with the existing `OpenAPI`; in other words, we need to report if any operation doesn't have a security defined

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,90 @@
...
+    },
+    SecuritySchemeList: {
+      enter(list: unknown[] | undefined, { location }: UserContext) {
+        if (!list) return;
+        for (let i = 0; i < list.length; i++) {
+          const item = list[i];
+          if (!isRef(item)) continue;
+          const itemLocation = location.child([i]);
+          const ref = item.$ref;
+          if (!ref.startsWith(COMPONENT_SCHEME_PREFIX)) {
```
AlbinaBlazhko17: Please, avoid using this type of check, because it will cause issues in future. e.g. `other.yaml#/components/securitySchemes/jwt` will be rejected. Try to track defined schemes by their `location.absolutePointer` and in `SecuritySchemeList` resolve each ref and compare absolute pointers.

.changeset/stupid-dryers-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+Added `security-defined` rule for AsyncAPI 2.x that reports when a security scheme referenced from an operation or server `security` array is not defined in `components.securitySchemes`.
```
AlbinaBlazhko17: Please, add in changeset `AsyncAPI` 3.x support.

AlbinaBlazhko17:
Thanks @harshit078  for your patience. Almost there. My biggest concern is based on AsyncAPI v3 rule and manually checking ref locations, i left comment regarding it.

harshit078:
Hey @AlbinaBlazhko17 ,  post your comment I understood the mistake I did for that check in the `async3 security defined`. I have pushed a fix addressing it. Thanks !

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,84 @@
...
+      SecurityScheme(_scheme: unknown, { location }: UserContext) {
+        definedSchemeAbsolutePointers.add(location.absolutePointer.toString());
+      },
+    },
+    SecuritySchemeList: {
+      enter(list: unknown[] | undefined, { location, resolve }: UserContext) {
+        if (!list) return;
+        for (let i = 0; i < list.length; i++) {
+          const item = list[i];
+          if (!isRef(item)) continue;
```
AlbinaBlazhko17, line 152: Could you explain why are you skipping not refs? And the purpose of this statement `item.$ref.split('/').pop()`?

.changeset/stupid-dryers-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/openapi-core": minor
+---
+
+Added `security-defined` rule for AsyncAPI 2.x and 3.x that reports when a security scheme referenced from an operation or server `security` array is not defined in `components.securitySchemes`.
```
JLekawa:
```suggestion
Added `security-defined` rule for AsyncAPI 2.x and 3.x.
This rule reports when a security scheme referenced from an operation or server `security` array is not defined in `components.securitySchemes`.
```

docs/@v2/rules/async/security-defined.md
```diff
@@ -0,0 +1,120 @@
...
+
+| AsyncAPI | Compatibility |
+| -------- | ------------- |
+| 2.6      | ✅            |
+| 3.0      | ✅            |
+
+## API design principles
+
+In AsyncAPI 2.x, `security` entries on operations and servers are bare security scheme names that must match a key under `components.securitySchemes`.
+A typo or rename leaves the reference dangling, and the document remains structurally valid — the mismatch is only visible to clients at runtime.
```
JLekawa:
```suggestion
A typo or rename breaks the reference but the document remains structurally valid.
The key mismatch is only visible to clients at runtime.
```

packages/core/src/rules/async3/security-defined.ts
harshit078: - my intent behind the skipping not refs in the logic is that since every security item can be either be a `$ref` or an inline object. Since inline objects are complete in itself hence there's nothing to check or cross check against `components.securitySchemes` hence only $ref items can be tracked as references.

- The purpose of tthis statement is that it basically just grabs the last pointer segment purely as a display name in the "**There is no name security scheme defined."** message which matches with the async2 wording.

I hope my explanation did a justice to what I was trying to say. If not, I'm happy to provide further explanation. Thanks !

docs/@v2/rules/async/security-defined.md
```diff
@@ -0,0 +1,120 @@
...
+
+In AsyncAPI 3.0, `security` entries are `SecurityScheme` objects, typically expressed as `$ref`s into `components.securitySchemes`. The rule reports when a security `$ref` does not point into `components.securitySchemes` or when it points at a name that is not defined there.
+
+This rule catches those name mismatches at lint time.
```
JLekawa:
```suggestion
In AsyncAPI 3.0, `security` entries are `SecurityScheme` objects, typically expressed as `$ref`s into `components.securitySchemes`. 

The `security-defined` rule reports when a security `$ref` does not point into `components.securitySchemes` or when it points at a name that is not defined there.
This rule catches these name mismatches at lint time.
```

docs/@v2/rules/async/security-defined.md
```diff
@@ -0,0 +1,122 @@
...
+| 2.6      | ✅            |
+| 3.0      | ✅            |
+
+## API design principles
+
+In AsyncAPI 2.x, `security` entries on operations and servers are bare security scheme names that must match a key under `components.securitySchemes`.
+A typo or rename breaks the reference but the document remains structurally valid.
+The key mismatch is only visible to clients at runtime.
+
+In AsyncAPI 3.0, `security` entries are `SecurityScheme` objects, typically expressed as `$ref`s into `components.securitySchemes`.
```
JLekawa:
```suggestion
In AsyncAPI 3.0, `security` entries are `SecurityScheme` objects, typically expressed as `$ref`s into `components.securitySchemes`.
```

JLekawa (approved):

harshit078:
Hi @AlbinaBlazhko17 , I have addressed all comments and resolved comments left by cursor. Though I have no clue on how to resolve the failing test hence I did not fix it. Can you review the PR again ? Thanks !

harshit078:
Hey @AlbinaBlazhko17 , all comments and checks are now addressed and cleared. Can you review the PR again ? 
Thanks !

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    Server(server: { security?: unknown }, { key }: UserContext) {
+      serverHasSecurity.set(key.toString(), Boolean(server?.security));
+    },
+    Channel: {
+      enter(channel: { servers?: string[] }) {
+        currentChannelServers = Array.isArray(channel?.servers) ? channel.servers : undefined;
+      },
+      Operation(
+        operation: { security?: unknown; traits?: unknown[] },
```
AlbinaBlazhko17: I was thinking about various inline types and casts, and the best solution would be to add the appropriate interfaces to the `./typings/asyncapi.ts` and `./typings/asyncapi3.ts` and then enrich with these types with visitors in `visitors.ts` (`Async2FlatVisitor` and `Async3FlatVisitor`). This would remove all casting and make the functions easy to read. You can drop in these cases `_scheme: unknown`  unknown type as well.

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,109 @@
...
+  const resolveMaybeRef = (node: unknown, resolve: UserContext['resolve']): unknown =>
+    isRef(node) ? resolve(node).node : node;
+
+  const operationSecuredByServers = (
+    operation: { channel?: unknown },
+    resolve: UserContext['resolve']
+  ): boolean => {
+    const channelRaw = operation?.channel;
+    let channel: { servers?: unknown[] } | undefined;
+    if (isRef(channelRaw)) {
```
AlbinaBlazhko17: You have reimplemented our resolve pattern. You can remove this code and use just the `resolve` function. Please remove the `resolveMaybeRef` function because its functionality can be used directly and it is only used in one place and rename function to the `isOperationSecuredByServers`.  So, the final function will look like:
```ts
 const isOperationSecuredByServers = (
    operation: Async3Operation,
    resolve: UserContext['resolve']
  ): boolean => {
    const channelRef = operation.channel;
    const channel = isRef(channelRef) ? resolve<Channel>(channelRef).node : channelRef;
    const applicableServers = channel?.servers ?? (rootServers ? Object.values(rootServers) : []);
    return applicableServers.some((server) => {
      const serverNode = isRef(server) ? resolve<Async3Server>(server).node : server;
      return Boolean(serverNode?.security);
    });
  };
```

packages/core/src/rules/common/operation-has-security.ts
```diff
@@ -0,0 +1,15 @@
+import { isRef } from '../../ref-utils.js';
+import type { UserContext } from '../../walk.js';
+
+export function operationHasSecurity(
```
AlbinaBlazhko17: Please, rename this function to `isOperationSecured` and move it to the `utils.ts`, because `common` folder only for shared rules.

AlbinaBlazhko17:
Hi @harshit078! Thanks for your patience and work. Please, do the final refactoring with types and function names.

packages/core/src/rules/async3/security-defined.ts
harshit078: sure, I agree with your approach. I'll revamp my code. Thanks !

harshit078:
Hi @AlbinaBlazhko17 , I have addressed all the refactoring you mentioned and fixed all comments by cursor. Can you check again ? Thanks for the patience to go back and forth !

.changeset/stupid-dryers-add.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+---
+
+Added `security-defined` rule for AsyncAPI 2.x and 3.x.
+This rule reports when a security scheme referenced from an operation or server `security` array is not defined in `components.securitySchemes`.
```
AlbinaBlazhko17: Please, add warning here, because it is breaking change due to adding new rule in r`ecommended` ruleset with `error` severity.
e.g.
`**Warning**: this change may break workflows that relied on root-level server inheritance.`
And, i think, we need to remove second sentence, which explain how rule works.

packages/cli/src/commands/split/asyncapi/gather-asyncapi-component-files.ts
```diff
@@ -19,7 +19,7 @@ export function gatherAsyncApiComponentFiles({
   ext: string;
   specVersion: 'async2' | 'async3';
 }) {
-  const { components } = asyncapi;
+  const components = asyncapi.components as Record<string, Record<string, unknown>> | undefined;
```
AlbinaBlazhko17: Please, avoid using `as` casting in these three files. You can add proper type inside `split` folder in `types.ts` similar to the `AnyOas3Definition`. It will look like this: 
`export type AnyAsyncApiComponents = Partial<Record<AsyncApi2SplittableComponent | AsyncApi3SplittableComponent, Record<string, unknown>>>;`

.changeset/stupid-dryers-add.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/openapi-core': minor
+---
+
+Added `security-defined` rule for AsyncAPI 2.x and 3.x.
+
+**Warning**: this change may break workflows that relied on root-level server inheritance.
```
AlbinaBlazhko17: I provided to you just an example from another PR. Please update this warning to align with this PR.

harshit078: I just realiased that and completely missed that in hurry. Aplogies !

AlbinaBlazhko17: No worries! Please, fix the issue from bugbot and i will review PR one more time. Thanks!

harshit078: Addressed cursor bot comment. The Pr is good to review again. Thanks !

harshit078:
Hi @AlbinaBlazhko17 , I have addressed all the last comments you left and Pr is good to review again. Thanks !

packages/core/src/utils/is-operation-secured.ts
```diff
@@ -0,0 +1,20 @@
+import { isRef } from '../ref-utils.js';
+import type { Async2Operation, Async2OperationTrait } from '../typings/asyncapi.js';
+import type { Async3Operation, Async3OperationTrait } from '../typings/asyncapi3.js';
+import type { UserContext } from '../walk.js';
+
+type SecuredOperation = Async2Operation | Async3Operation;
+type SecuredTrait = Async2OperationTrait | Async3OperationTrait;
+
+export function isOperationSecured(
```
DmitryAnansky: This function seems to work only with Async2 and Async3 but located in `utils` and have general naming.
If it is only work with Async, will it make sense to change its name to `isAsyncOperationSecured` ?
Also please consider file ranaming.

packages/core/src/typings/asyncapi3.ts
```diff
@@ -42,17 +44,85 @@ export interface ExternalDoc {
...
+ * @deprecated Use `Async3Channel` instead.
+ */
+export type Channel = Async3Channel;
+
+export interface Async3OperationTrait {
+  title?: string;
+  summary?: string;
+  description?: string;
+  tags?: unknown[];
+  externalDocs?: unknown;
```
DmitryAnansky: Why this type is unknown?
Can it be described with the existing ExternalDoc type?
```
export interface ExternalDoc {
  url: string;
  description?: string;
}
```

packages/core/src/typings/asyncapi3.ts
```diff
@@ -42,17 +44,85 @@ export interface ExternalDoc {
...
+/**
+ * @deprecated Use `Async3Channel` instead.
+ */
+export type Channel = Async3Channel;
+
+export interface Async3OperationTrait {
+  title?: string;
+  summary?: string;
+  description?: string;
+  tags?: unknown[];
```
DmitryAnansky: Could you please check other unknown types here, maybe some can be defined with the existing types.

packages/core/src/typings/asyncapi3.ts
```diff
@@ -42,17 +44,85 @@ export interface ExternalDoc {
...
+  parameters?: Record<string, unknown>;
+  correlationIds?: Record<string, unknown>;
+  operationTraits?: Record<string, Async3OperationTrait>;
+  messageTraits?: Record<string, unknown>;
+  serverBindings?: Record<string, unknown>;
+  channelBindings?: Record<string, unknown>;
+  operationBindings?: Record<string, unknown>;
+  messageBindings?: Record<string, unknown>;
+  channels?: Record<string, Async3Channel>;
+  servers?: Record<string, Async3Server>;
```
DmitryAnansky: @harshit078 can servers be also Reference here?

packages/core/src/utils/is-operation-secured.ts
harshit078: Hi @DmitryAnansky , I renamed and shifted this function to utils for a shared logic as guided by - 

https://github.com/Redocly/redocly-cli/pull/2759#discussion_r3361611704

packages/core/src/typings/asyncapi3.ts
harshit078: Yes, I agree. I forgot while implementing it.

packages/core/src/typings/asyncapi3.ts
harshit078: yes, I'll make changes

harshit078:
Hey @DmitryAnansky , I have fixed all comments and failing e2e tests and PR is good to review again. Thanks !

packages/cli/src/commands/split/asyncapi/gather-asyncapi-component-files.ts
```diff
@@ -19,7 +23,7 @@ export function gatherAsyncApiComponentFiles({
   ext: string;
   specVersion: 'async2' | 'async3';
 }) {
-  const { components } = asyncapi;
+  const components: AnyAsyncApiComponents | undefined = asyncapi.components;
```
Daryna-del: Why do we need to rewrite this and add type?

packages/respect-core/src/modules/__tests__/flow-runner/runner/resolve-workflow-context.test.ts
```diff
@@ -379,6 +379,30 @@ describe('resolveWorkflowContext', async () => {
     );
   });
 
+  it('should call createTestContext for arazzo type using the spec form without `workflows` segment', async () => {
```
Daryna-del: Why do we need those changes?

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,74 @@
...
+        const allServerNames = [...serverHasSecurity.keys()];
+        for (const { location, channelServers } of operationsWithoutSecurity) {
+          const applicableServers = channelServers ?? allServerNames;
+          const securedByServer = applicableServers.some((name) => serverHasSecurity.get(name));
+          if (!securedByServer) {
+            report({
+              message: `Every operation should have security defined on it.`,
+              location: location.key(),
+            });
+          }
```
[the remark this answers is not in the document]
Daryna-del: Could you please take a look to this issue found by bugbot?

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,80 @@
...
+    Server(server: Async2Server, { key }: UserContext) {
+      serverHasSecurity.set(key.toString(), Boolean(server?.security));
+    },
+    Channel: {
+      enter(channel: Async2Channel) {
+        currentChannelServers = Array.isArray(channel?.servers) ? channel.servers : undefined;
+      },
+      Operation(operation: Async2Operation, { location, resolve }: UserContext) {
+        if (isOperationSecured(operation, resolve)) return;
+        operationsWithoutSecurity.push({ location, channelServers: currentChannelServers });
```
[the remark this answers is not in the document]
Daryna-del: Let's add skip visitor to skip visiting channels under components. You have already tests for the server, please, do the same for the channels.

packages/core/src/utils/is-operation-secured.ts
Daryna-del: Could you move this function under the file `packages/core/src/rules/utils.ts`. Also, please take a look to the @DmitryAnansky 's comment regarding the name of the function.

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,117 @@
...
+    operation: Async3Operation,
+    resolve: UserContext['resolve']
+  ): boolean => {
+    const channelRef = operation.channel;
+    const channel: Async3Channel | undefined = isRef(channelRef)
+      ? resolve<Async3Channel>(channelRef).node
+      : channelRef;
+    const applicableServers: Array<Referenced<Async3Server>> =
+      channel?.servers ?? (rootServers ? Object.values(rootServers) : []);
+    if (applicableServers.length === 0) return false;
```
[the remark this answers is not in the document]
harshit078: expected. Solving it

packages/respect-core/src/modules/__tests__/flow-runner/runner/resolve-workflow-context.test.ts
harshit078: I missed the existing check which was already present hence I agree with you and I should remove this.

harshit078:
Hey @Daryna-del , I have addressed all the comments left by you and cursor and also resolved the failing e2e tests. Can you review the recent changes again ? Thanks !

docs/@v2/rules/async/security-defined.md
```diff
@@ -0,0 +1,120 @@
+# security-defined
+
+Verifies that every security scheme referenced from an operation or server `security` array is defined in `components.securitySchemes`.
+
+| AsyncAPI | Compatibility |
+| -------- | ------------- |
+| 2.6      | ✅            |
+| 3.0      | ✅            |
```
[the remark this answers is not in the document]
DmitryAnansky: @harshit078 please update PR description.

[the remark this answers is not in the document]
harshit078: Hi @DmitryAnansky updated PR description

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,122 @@
...
+          if (!isRef(item)) continue;
+          const itemLocation = location.child([i]);
+          const resolved = resolve<Async3SecurityScheme>(item);
+          const refPointer = getRefPointer(item.$ref);
+          const name = refPointer.split('/').pop() ?? item.$ref;
+          references.push({
+            location: itemLocation,
+            name,
+            refPointer,
+            resolved: resolved.node !== undefined,
```
DmitryAnansky: This check treats a null-valued scheme entry as defined, am I right?
Seems that a scheme key with no value (half-finished YAML) silently passes.

packages/core/src/rules/utils.ts
```diff
@@ -44,6 +46,22 @@ export function oasTypeOf(value: unknown) {
...
 }
 
+type SecuredOperation = Async2Operation | Async3Operation;
+type SecuredTrait = Async2OperationTrait | Async3OperationTrait;
+
+export function isAsyncOperationSecured(
+  operation: SecuredOperation | undefined,
+  resolve: UserContext['resolve']
+): boolean {
+  if (operation?.security) return true;
```
DmitryAnansky: This condition might be confusing.
Seems the function consider empty security (e.g. `security: []`) as Secured operation.
But it is actually public.

And have similar check that will result in opposite verdict:
```
      serverHasSecurity.set(
        key.toString(),
        Array.isArray(server?.security) && server.security.length > 0
      );
```

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,122 @@
...
+              message: `There is no \`${reference.name}\` security scheme defined.`,
+              location: reference.location.key(),
+            });
+          }
+        }
+
+        if (rootOperations && rootLocation) {
+          const operationsLocation = rootLocation.child(['operations']);
+          for (const [opName, opRef] of Object.entries(rootOperations)) {
+            const operation = isRef(opRef) ? resolve<Async3Operation>(opRef).node : opRef;
```
DmitryAnansky: Looks like only refs against the root document will be resolved here.
Please double check, but it seems that refs recorded relative to another file never resolve, producing false 'Every operation should have security defined' errors on multi-file docs.
Please check `context.resolve` functionality, maybe you need to handle this case as well.

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,122 @@
...
+    return applicableServers.every((server) => {
+      const serverNode = isRef(server) ? resolve<Async3Server>(server).node : server;
+      return Array.isArray(serverNode?.security) && serverNode.security.length > 0;
+    });
+  };
+
+  return {
+    Root: {
+      enter(root, { location }: UserContext) {
+        rootServers = root?.servers;
```
DmitryAnansky: Could you please check if you handle the scenario with two files correctly:

```
# main.yaml
asyncapi: 3.0.0
info:
  title: Test
  version: 1.0.0
servers:
  $ref: './servers.yaml'      # ← map-level ref
channels:
  myChannel:
    address: user/signedup
operations:
  sendUserSignedUp:
    action: send
    channel:
      $ref: '#/channels/myChannel'

# servers.yaml — the server IS secured
prod:
  host: example.com
  protocol: kafka
  security:
    - type: apiKey
      in: user
```

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+function getRefPointer(ref: string): string {
+  const hashIndex = ref.indexOf('#');
+  return hashIndex === -1 ? ref : ref.slice(hashIndex);
+}
+
+function pointsToSecurityScheme(pointer: string): boolean {
+  return (
+    pointer.startsWith(SECURITY_SCHEMES_POINTER) &&
```
DmitryAnansky: Seems external file references are not handled here.

The AsyncAPI 3.0 specification defines security requirements as an array of Security Scheme Objects, and states the general rule: any time a Schema Object can be used, a Reference Object can be used in its place, allowing referencing definitions instead of defining them inline. The Reference Object itself uses JSON Reference resolution — "For this specification, reference resolution is done as defined by the JSON Reference specification and not by the JSON Schema specification." JSON Reference allows the $ref value to be any URI, which includes relative file paths and remote URLs.

Please check the example. We able to bundle it with ref resolution but this new rule will produce false failure during file lint.

```yaml
asyncapi: 3.0.0
info:
  title: Async3 External Security Ref
  version: 1.0.0
servers:
  production:
    host: broker.example.com
    protocol: mqtt
    security:
      - $ref: './apiKey.yaml'
channels:
  userSignups:
    address: user/signedup
    messages:
      userSignedUp:
        payload:
          type: object
          properties:
            name:
              type: string
operations:
  sendUserSignups:
    action: send
    channel:
      $ref: '#/channels/userSignups'
    security:
      - $ref: './apiKey.yaml'

```

apiKey.yaml => 
```yaml
type: httpApiKey
name: api_key
in: header
description: API key security scheme defined in an external file
```

packages/core/src/rules/async3/security-defined.ts
harshit078: I guess you are correct and I didnt catch it. $`ref` to a key with a null value resolves to node === null and since null !== undefined was marked as defined and silently passed.

harshit078:
Hey @DmitryAnansky , I have addressed all your comments in my recent fix and I hope they address the issue. If not I'm still ready to improve further. Can you take a look again ? Thanks !

packages/core/src/rules/async2/__tests__/security-defined.test.ts
```diff
@@ -0,0 +1,711 @@
...
+        asyncapi: '2.6.0'
+        info:
+          title: Cool API
+          version: 1.0.0
+        servers:
+          production:
+            url: kafka.example.com
+            protocol: kafka
+            security:
+              - missingScheme: []
```
DmitryAnansky, line 72: How is the `missingSchema` test different from the `undefinedSchema` test?

harshit078: `MissingSchema` has its security on operation which is `channels.some/channel.subscribe` while `undefinedSchema` has its security on `server`, `operation` and `operationTrait` in `asyncapi2`.

DmitryAnansky: Seems it is undefinedSchema in both cases just on different level.
Can you update test description.

harshit078: done !

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,164 @@
...
+          rootOperationsFrom = undefined;
+          rootOperationsLocation = location.child(['operations']);
+        }
+      },
+      leave(_root, { report, resolve }: UserContext) {
+        for (const reference of references) {
+          const checkPointer =
+            reference.local || reference.refPointer.startsWith(COMPONENTS_POINTER);
+          if (checkPointer && !pointsToSecurityScheme(reference.refPointer)) {
+            report({
```
DmitryAnansky, line 105: @harshit078 
Could you please add reference to the corresponding docs page.
Like in other rules reports, e.g. =>
```
            report({
              message: `The "${schemeName}" security requirement must list at least one scope.`,
              location: location.key(),
              reference: 'https://redocly.com/docs/cli/rules/common/security-scopes-defined',
            });
```

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,164 @@
...
+              message: `Security scheme \`$ref\` must point to \`#/components/securitySchemes\`.`,
+              location: reference.location.key(),
+            });
+            continue;
+          }
+
+          if (!reference.resolved) {
+            report({
```
DmitryAnansky, line 114: Could you please add reference to the corresponding docs page?

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,164 @@
...
+              operation = resolvedOperation.node;
+              resolveFrom = resolvedOperation.location?.source.absoluteRef;
+            } else {
+              operation = opRef;
+              resolveFrom = rootOperationsFrom;
+            }
+            if (!operation) continue;
+            if (isAsyncOperationSecured(operation, resolve, resolveFrom)) continue;
+            if (isOperationSecuredByServers(operation, resolve, resolveFrom)) continue;
+            report({
```
DmitryAnansky, line 138: Could you please add reference to the corresponding docs page?

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        const allServerNames = [...serverHasSecurity.keys()];
+        for (const { location, channelServers } of operationsWithoutSecurity) {
+          const applicableServers = channelServers ?? allServerNames;
+          const securedByServer =
+            applicableServers.length > 0 &&
+            applicableServers.every((name) => serverHasSecurity.get(name));
+          if (!securedByServer) {
+            report({
```
DmitryAnansky, line 49: Could you please add reference to the corresponding docs page?

packages/core/src/rules/async2/security-defined.ts
```diff
@@ -0,0 +1,97 @@
...
+  let currentChannelServers: string[] | undefined;
+  let inComponents = false;
+
+  return {
+    Root: {
+      leave(_root, { report }: UserContext) {
+        for (const [name, scheme] of referencedSchemes.entries()) {
+          if (scheme.defined) continue;
+          for (const reportedFromLocation of scheme.from) {
+            report({
```
DmitryAnansky, line 34: Could you please add reference to the corresponding docs page?

packages/core/src/rules/async3/security-defined.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    if (applicableServers.length === 0) return false;
+    return applicableServers.every((server) => {
+      const serverNode = isRef(server) ? resolve<Async3Server>(server, serversFrom).node : server;
+      return hasSecurityRequirements(serverNode);
+    });
+  };
+
+  return {
+    Root: {
```
DmitryAnansky, line 76: @harshit078 have you consider using Operation and Servers node instead of entering Root level?
Its not a critical, but maybe worth to check if this would be easier approach. (You can skip it, its optional)

packages/core/src/typings/asyncapi3.ts
```diff
@@ -61,18 +46,77 @@ export interface ExternalDoc {
   description?: string;
 }
 
-export type Channel = {
+export interface Async3Components {
```
DmitryAnansky: Do we use it anywhere?

harshit078: thanks for pointing it out, It was a dead code. I have removed it.

packages/core/src/rules/async3/security-defined.ts
harshit078: The reason behind I went the latter way of entering the root level is because `root.operations/root.servers` and `components.operations/components.servers` resolve to the same Operation/Server node types so plain node visitors would also fire on reusable component definitions. Also the operation check is global, an operation's validity depends on servers declared elsewhere  so we would still need `Root.leave` regardless of ordering. As a result it would require two extra visitors plus a Root.leave. Async2 keys servers by bare name whereas async3 servers are `$refs` resolved by pointer, so it would be an extra step hence operation→channel→server fallback stays equally complex either way. this is what I think so far based on my research while working on this PR. Happy to revisit if you feel strongly.

vadyvas (approved):

DmitryAnansky (approved):

DmitryAnansky:
@harshit078 
Thank you for your contribution! We’ll notify you here once the feature is released.

DmitryAnansky:
@harshit078 , The feature will be available in v`2.44.0`.

PR #2764, opened 2026-04-22
chore: fix flaky smoke tests on yarn

### LINKED ISSUE #2618: Flaky smoke tests
**Describe the bug**
Flaky smoke tests:
https://github.com/Redocly/redocly-cli/actions/runs/23151274019/job/67256114952?pr=2588
https://github.com/Redocly/redocly-cli/actions/runs/21981360702/job/63504577156?pr=2555
https://github.com/Redocly/redocly-cli/actions/runs/22145856147/job/64022774477
https://github.com/Redocly/redocly-cli/actions/runs/23187483436/job/67374669208

**To Reproduce**

**Expected behavior**

**Logs**

**OpenAPI description**

**Redocly version(s)**

**Node.js version(s)**
Node - 20
Yarn

**OS, environment**
Windows

**Additional context**
Weird behavior, because it throws and import error only in new `@redocly/config` versions.

AlbinaBlazhko17:
## What/Why/How?
I found the root cause of the failing tests. Old `redoc` dependency has a `v0.22.0` version of `@redocly/config` and yarn tries to resolve this version instead of latest.

**The root cause:** Yarn v1's flat-hoisting is non-deterministic when the install graph contains two incompatible versions of the same package.

Changes:
- Migrated to `yarn install` and use `yarn add` only for `redoc`. Because `yarn` add doesn't respect `resolutions` property in `package.json`, which overrides deps.
- Added `*.tgz` file in `package.json`.
- Added `pin-config-version.sh` script to read version of `@redocly/config` and `@redocly/ajv` from `openapi-core` and write proper version to `package.json` for smoke tests.
- Added to steps with redoc one bash script to inject it to the `package.json`.
- Added to tests missing workaround for cache issue on windows. Fixed existing workaround: removed `try catch`, because bash exiting non-zero doesn't throw — it sets $LASTEXITCODE. The old catch block would never fire, so on the final attempt the step would exit 0 and the job would be reported.

After changes we have only tiny warning, which means, that the version now resolves properly.
`warning Resolution field "@redocly/config@0.48.1" is incompatible with requested version "@redocly/config@0.22.0"`

## Reference
Resolves #2618

## Testing
Ran 10 times smoke tests on this PR 

## Screenshots (optional)
**When test failed:**
Screenshot 2026-04-22 at 15 19 01

**When test passed:**
Screenshot 2026-04-22 at 15 21 35

## Check yourself

## Security

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to CI smoke-test setup and dependency installation behavior, with no production code impact. Main risk is unintended CI breakage due to new install/pinning steps across npm/yarn/pnpm and Windows runners.
> 
> **Overview**
> Refactors the smoke-test GitHub Actions workflow to **stop installing the CLI tarball via `add`** and instead run `npm/yarn/pnpm install`, with a dedicated step to inject `redoc` only for the `--redoc` matrix jobs.
> 
> Adds a smoke-test `package.json` dependency on `@redocly/cli` via `file:./redocly-cli.tgz`, plus scripts (`pin-intersecting-deps.sh`, `add-redoc.sh`) to pin `@redocly/config`/`@redocly/ajv` via Yarn `resolutions` and to add `redoc@latest` to avoid non-deterministic Yarn v1 hoisting.
> 
> Improves Windows smoke jobs by clearing npm cache and retrying installs correctly (removing ineffective `try/catch` and exiting with the final failure code).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 92985f2bbc764f41ae247dd0fdd9abea3dd4cd8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/smoke.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version: 24
-      - run: bash ./tests/smoke/basic/run-smoke.sh "npm i redoc redocly-cli.tgz" "npm run"
+      - run: bash ./tests/smoke/basic/run-smoke.sh "npm install && npm i redoc" "npm run"
```
tatomyr: This doesn't make any difference but the intention becomes less clear.

.github/workflows/smoke.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         run: yarn cache clean
       - run: |
           for i in {1..2}; do # workaround for yarn cache issue
-            sleep 5 && bash ./tests/smoke/basic/run-smoke.sh "yarn add ./redocly-cli.tgz" "yarn" && break
+            sleep 5 && bash ./tests/smoke/basic/run-smoke.sh "yarn install --no-lockfile" "yarn" && break
```
tatomyr: What does this change? There's no lockfile before the install, and we're installing the dependencies only once during this job, so having or not having the lockfile should change nothing.

tests/smoke/basic/pin-config-version.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+REDOCLY_CONFIG_VERSION=$(node -e "const fs = require('fs'); const p = JSON.parse(fs.readFileSync('./packages/core/package.json', 'utf8')); console.log(p.dependencies['@redocly/config'].replace(/\^/, ''));")
```
tatomyr: We're emulating something that will not be the case in our users' setup, so the smokes become irrelevant. Let's try keeping them as much close to the reality as possible.

tests/smoke/basic/run-smoke.sh
```diff
@@ -8,8 +8,8 @@ echo "Directory content:"
 ls -a
 echo
 
-# Executing the command provided as the first argument 
-$1
+# Executing the command provided as the first argument
+eval "$1"
```
tatomyr: What's the difference?

.github/workflows/smoke.yaml
AlbinaBlazhko17: I wanted to use `yarn install` instead of `yarn add` to fix the issue, so i added into `package.json` dependency, which points to the `tgz` cli file. I changed all tests to use just install and be consistent across the tests.

tests/smoke/basic/run-smoke.sh
AlbinaBlazhko17: Before this change, smoke tests became unstable, because we are using `&&` operator to install `redoc` dep: `yarn install && yarn add redoc`. Using eval i can support this operator

tests/smoke/basic/pin-config-version.sh
AlbinaBlazhko17: This bash script read the version of `@redocly/config` from `openapi-core` and write this version to `package.json` of smoke tests to override this dep to be sure, that all tests will resolve proper version of `@redocly/config`. Users will resolve the same version of `@redocly/config`. My first try to fix this issue is manually adding this properties to the `package.json`, but we can forget to change it there, so i decided to create a script.

.github/workflows/smoke.yaml
AlbinaBlazhko17: You are right. Let me drop this arg and re-ran tests, because my concern is that we are running `yarn install` and then `yarn add redoc` and the second command may overwrite the lock file and `@redocly/config` as a result.

tests/smoke/basic/pin-config-version.sh
tatomyr: Let's discuss this offine.

.github/workflows/smoke.yaml
```diff
@@ -255,12 +262,14 @@ jobs:
         run: |
           yarn cache clean
           npm cache clean --force
+      - name: Pin intersecting deps
+        run: bash ./tests/smoke/basic/pin-intersecting-deps.sh
```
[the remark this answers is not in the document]
AlbinaBlazhko17: But smoke tests filed, so i added this pin

tatomyr (approved):

PR #2771, opened 2026-04-24
chore: improve types by reducing the any type usage and type casting

tatomyr:
## What/Why/How?

Improved typing by reducing the `any` type usage and type casting
 
## Testing
 
[Pipeline](https://github.com/Redocly/redocly/actions/runs/24875488741/job/72831069288?pr=22749) and [E2E](https://github.com/Redocly/redocly/actions/runs/24876583308/job/72837313410?pr=22749) passed.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Mostly type-safety improvements, but `resolve.ts` changes the `$ref` pointer-walk logic (object/array handling and anchor collection timing), which could affect resolution behavior across bundling/linting.
> 
> **Overview**
> Reduces `any` usage across core utilities/rules by switching to `unknown`, tightening option/return types, and removing several casts (e.g., JSON formatter output, filter helpers, env/platform check, AJV error formatting, stats visitors).
> 
> Hardens reference resolution in `resolve.ts` by treating traversed nodes as `unknown`, guarding walks with `isPlainObject`/`Array.isArray`, and correctly following JSON Pointer segments through both objects and arrays (including when encountering nested `$ref`s). It also adjusts when `$anchor` entries are collected and tightens rule exports/casts in the OAS2/OAS3 rule registries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e65ed0f0c4dc91a144bae4370f1f20d26c146ac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/resolve.ts
+++ b/packages/core/src/resolve.ts
@@ -17,6 +17,7 @@ import {
 import { isNamedType, SpecExtension, type NormalizedNodeType } from './types/index.js';
 import type { OasRef } from './typings/openapi.js';
 import { getOwn } from './utils/get-own.js';
+import { isPlainObject } from './utils/is-plain-object.js';
 import { makeRefId } from './utils/make-ref-id.js';
 import { nextTick } from './utils/next-tick.js';
 import { readFileFromUrl } from './utils/read-file-from-url.js';
@@ -193,17 +194,17 @@ export type ResolvedRefMap = Map<string, ResolvedRef>;
 
 type RefFrame = {
   prev: RefFrame | null;
-  node: any;
+  node: unknown;
 };
 
-function pushRef(head: RefFrame, node: any): RefFrame {
+function pushRef(head: RefFrame, node: unknown): RefFrame {
   return {
     prev: head,
     node,
   };
 }
 
-function hasRef(head: RefFrame | null, node: any): boolean {
+function hasRef(head: RefFrame | null, node: unknown): boolean {
   while (head) {
     if (head.node === node) {
       return true;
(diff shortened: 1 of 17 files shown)


packages/core/src/resolve.ts
```diff
@@ -420,33 +421,42 @@ export async function resolveDocument(opts: {
         nodePointer: '#/',
       };
 
-      let target = targetDoc.parsed as any;
+      let target = targetDoc.parsed;
 
       const segments = pointer;
       for (const segment of segments) {
-        if (typeof target !== 'object') {
```
tatomyr, line 427: Replaced the guard with explicit Array/Object detection.

packages/core/src/resolve.ts
```diff
@@ -420,33 +421,42 @@ export async function resolveDocument(opts: {
...
+          target = target[+segment];
+          resolvedRef.nodePointer = joinPointer(
+            resolvedRef.nodePointer!,
+            escapePointerFragment(segment)
+          );
         } else if (isRef(target)) {
           resolvedRef = await followRef(targetDoc, target, pushRef(refStack, target));
           targetDoc = resolvedRef.document || targetDoc;
 
-          if (typeof resolvedRef.node !== 'object') {
```
tatomyr, line 440: Moved to the `else` block below.

packages/core/src/rules/oas2/index.ts
```diff
@@ -48,7 +48,7 @@ import { ResponseMimeType } from './response-mime-type.js';
 export const rules: Oas2RuleSet<'built-in'> = {
   struct: Struct as Oas2Rule,
   'no-invalid-schema-examples': NoInvalidSchemaExamples as Oas2Rule,
-  'no-invalid-parameter-examples': NoInvalidParameterExamples,
+  'no-invalid-parameter-examples': NoInvalidParameterExamples as Oas2Rule,
```
DmitryAnansky, line 51: Will it make sense to update type inside the rule if it is a common one?

Smth, like
```
export const NoInvalidParameterExamples: Oas2Rule |  Oas3Rule = (opts: any) => {
```

We can also use another pattern, like with `ResponseContainsProperty` - we have two files, separately for each specification, but in this case there is also difference in implementation.

tatomyr: It's already [this way](https://github.com/Redocly/redocly-cli/pull/2771/changes#diff-3916123e19e29623613e96ee5ef11ad261f06d356356caad5085f83af0dfe750R8). We just have to explicitly specify the type. Sometimes we make 2 different rules for different specs, and then the cast is not needed. But for common rules this is the proper use case.

DmitryAnansky (approved):

PR #2774, opened 2026-04-24
fix: scope AJV instances per lintDocument call to fix concurrent validation

vlindhol:
## Summary

The global AJV singleton in `packages/core/src/rules/ajv.ts` captured a `resolve` closure from whichever document hit it first. When consumers like [`openapi-typescript`](https://github.com/openapi-ts/openapi-typescript) run `lintDocument` over multiple APIs concurrently via `Promise.all`, the second document's `resolve` was silently discarded, producing spurious warnings:

```
⚠ Example validation errored: can't resolve reference #/components/schemas/Foo
  from id .../spec.yaml#/paths/~1items/post/requestBody/content/application~1json/schema.
```

The warnings only appear when:
- Multiple APIs are linted concurrently
- A same-doc `$ref` has an `example`/`examples` sibling
- The spec contains a cross-file `$ref` (which is what triggers AJV's `loadSchemaSync`)

`redocly lint` itself wasn't affected because it processes APIs sequentially.

There was an earlier attempt at this fix in #2135 that went stale. This time I built a minimal repro that's checked in as a regression test.

## Fix

Refactored `rules/ajv.ts` from free functions + global singleton into an `AjvValidator` class. Each of the three example-validation rules (`no-invalid-media-type-examples`, `no-invalid-schema-examples`, `no-invalid-parameter-examples`) now creates its own `AjvValidator` in its factory closure. Since `initRules` invokes each factory once per `lintDocument` call, every call gets a fresh validator — no shared state across documents.

The walker (`walk.ts`) stays untouched; AJV is purely a JSON-schema-validation concern living inside the rules that need it.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes shared lint-time JSON Schema validation infrastructure used by multiple rules; behavior should be equivalent for sequential lint but affects ref resolution under concurrency.
> 
> **Overview**
> Fixes **false “can't resolve reference” / example validation errors** when multiple OpenAPI specs are linted in parallel (e.g. `Promise.all` over `lint`).
> 
> **AJV is no longer a process-wide singleton.** `rules/ajv.ts` is refactored to an **`AjvValidator` class** with per-instance AJV caches; the global `releaseAjvInstance()` hook and `validateJsonSchema` export are removed. **`lintDocument` no longer resets AJV** at the start of each run.
> 
> The three example rules (`no-invalid-media-type-examples`, `no-invalid-schema-examples`, `no-invalid-parameter-examples`) each construct **`new AjvValidator()` in their rule factory**, so every `initRules` / lint pass gets validators whose `loadSchemaSync` **`resolve` closure matches that document**—avoiding cross-document races when concurrent lints shared one AJV instance.
> 
> `validateExample` in `rules/utils.ts` now takes a **`validator`** and calls **`validator.validate(...)`** instead of the old module function.
> 
> **Regression coverage:** concurrent-lint fixtures (`spec-a` / `spec-b` + shared `common.yaml`) and a test that runs both specs with `no-invalid-media-type-examples` via `Promise.all` expecting zero issues. AJV unit tests target **`AjvValidator`** directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 051ff9af243de1048daeeec324fe98a217bd8a33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -3,6 +3,7 @@ import { YamlParseError } from './errors/yaml-parse-error.js';
 import type { SpecVersion } from './oas-types.js';
 import { Location, isRef } from './ref-utils.js';
 import type { ResolveError, Source, ResolvedRefMap, Document } from './resolve.js';
+import type { AjvValidator } from './rules/ajv.js';
 import { isNamedType, SpecExtension, type NormalizedNodeType } from './types/index.js';
 import type { Referenced } from './typings/openapi.js';
 import { getOwn } from './utils/get-own.js';
@@ -49,6 +50,7 @@ export type UserContext = {
   specVersion: SpecVersion;
   config?: Config;
   getVisitorData: () => Record<string, unknown>;
+  ajvValidator: AjvValidator;
 };
 
 export type Loc = {
@@ -99,6 +101,7 @@ export type WalkContext = {
   config?: Config;
   visitorsData: Record<string, Record<string, unknown>>; // custom data store that visitors can use for various purposes
   refTypes?: Map<string, NormalizedNodeType>;
+  ajvValidator: AjvValidator;
 };
 
 function collectParents(ctx: VisitorLevelContext) {
@@ -199,6 +202,7 @@ export function walkDocument<T extends BaseVisitor>(opts: {
             specVersion: ctx.specVersion,
             config: ctx.config,
             getVisitorData: () => getVisitorDataFn(ruleId),
+            ajvValidator: ctx.ajvValidator,
           },
           { node: resolvedNode, location: resolvedLocation, error }
         );
@@ -416,6 +420,7 @@ export function walkDocument<T extends BaseVisitor>(opts: {
               specVersion: ctx.specVersion,
               config: ctx.config,
               getVisitorData: () => getVisitorDataFn(ruleId),
+              ajvValidator: ctx.ajvValidator,
             },
             { node: resolvedNode, location: resolvedLocation, error }
           );
@@ -452,6 +457,7 @@ export function walkDocument<T extends BaseVisitor>(opts: {
             ignoredNodes.add(`${currentLocation.absolutePointer}${currentLocation.pointer}`);
           },
           getVisitorData: () => getVisitorDataFn(ruleId),
+          ajvValidator: ctx.ajvValidator,
         },
         collectParents(context),
         context
--- a/.changeset/fix-ajv-concurrent-lint.md
+++ b/.changeset/fix-ajv-concurrent-lint.md
@@ -1,0 +1,5 @@
+---
+"@redocly/openapi-core": patch
+---
+
+Fixed spurious "can't resolve reference" warnings when linting multiple APIs concurrently.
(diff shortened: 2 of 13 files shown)


packages/core/src/walk.ts
```diff
@@ -49,6 +50,7 @@ export type UserContext = {
   specVersion: SpecVersion;
   config?: Config;
   getVisitorData: () => Record<string, unknown>;
+  ajvValidator: AjvValidator;
```
AlbinaBlazhko17: Please, avoid adding `ajvValidator` in walker, because it is a separate concept of `JSON schema validation`. We use this validator only for specific rule and we use Ajv validators only there, meaning, that we don't need this in general pattern of walker. Additionally, that's why all tests in pipeline failed.

AlbinaBlazhko17:
Thanks @vlindhol for the contribution! I left comment. Please, feel free to ask any questions.

packages/core/src/walk.ts
vlindhol: Good point, moved `AjvValidator` out of the walker. Each of the three example-validation rules now creates its own `AjvValidator` instead.

DmitryAnansky:
@vlindhol
Could you please update the PR description with the section required by the contribution guide?
Please make sure to review it as well.

```md
## Check yourself

- [ ] This PR follows the [contributing guide](https://github.com/Redocly/redocly-cli/blob/main/CONTRIBUTING.md#pull-request-guidelines) <!-- TODO: make it required -->
- [ ] Documentation update has been considered <!-- TODO: make it required -->

## Security

- [ ] The security impact of the change has been considered <!-- required 🔒 -->
```

vlindhol:
> @vlindhol Could you please update the PR description with the section required by the contribution guide? Please make sure to review it as well.

Done! Was a bit unclear which boxes I need to check but I'm happy to be corrected!

jeremyfiel:
Does this have any impact on the single entry document concept?   This appears to be a fix for code generators but the main purpose of this package is to lint/join/bundle descriptions.  Please verify this does not introduce any other concept outside of using the entry document as the starting point for the functionality of this package.

AlbinaBlazhko17:
Hi @vlindhol! Could you please rebase main and resolve conflicts?
We introduced a new performance benchmark and we want to check stats for this PR.

vlindhol:
@jeremyfiel sorry, didn't notice your question! Each lintDocument call still operates on one entry document at a time. The fix just stops the validator from sharing state with other concurrent lintDocument calls — no new multi-document concepts are introduced.

@AlbinaBlazhko17 done!

AlbinaBlazhko17:
@vlindhol Thanks! I will test functionality manually and go back soon.

JLekawa (approved):

AlbinaBlazhko17 (approved):

AlbinaBlazhko17:
Thanks @vlindhol for the contribution! The fix released in v2.32.1.

PR #2807, opened 2026-05-11
fix: send query languages strings in request body payload

### LINKED ISSUE #2798: respect: JSONPath string in request body payload is misinterpreted as Arazzo runtime expression
## Summary

When a request body payload contains a string value that starts with `$` (e.g. a JSONPath filter expression), Redocly Respect incorrectly treats it as an Arazzo runtime expression and throws a parse error — even though the string is a plain literal inside a request body, not an expression context.

## Steps to reproduce

**Minimal Arazzo workflow:**

```yaml
arazzo: 1.0.1
info:
  title: Reproduce runtime expression false positive
  version: 1.0.0
sourceDescriptions:
  - name: myapi
    type: openapi
    url: ./openapi.yaml
    x-serverUrl: http://localhost:8080
workflows:
  - workflowId: discover
    steps:
      - stepId: search
        operationId: myapi.search
        requestBody:
          contentType: application/json
          payload:
            message:
              filters:
                type: jsonpath
                expression: "$.items[*].attributes[?(@.type=='foo')]"
```

**Run:**
```
npx @redocly/cli respect workflow.arazzo.yaml
```

**Result:**
```
✗ unexpected error
Reason: Runtime expression is not valid: $.items[*].attributes[0]
```

The `expression` field value is a JSONPath string — it is not an Arazzo runtime expression, has no business being evaluated as one, and should be passed through to the HTTP request body as-is.

## Root cause (traced in source)

In `packages/respect-core/src/modules/runtime-expressions/evaluate.ts`:

1. `isPureRuntimeExpression()` tests every string value with:
   ```ts
   const regex = /^\$[^\s{}]+(\([^)]*\))?$/;
   ```
   `$.items[*].attributes[?(@.type=='foo')]` matches (no spaces after trimming the filter).

2. Because it matches, `evaluateRuntimeExpression()` is called directly and the string is passed to the Arazzo ABNF parser.

3. The parser only recognises `$url`, `$method`, `$statusCode`, `$request.*`, `$response.*`, `$steps.*`, etc. JSONPath syntax (`[*]`, `[?(...)]`) is not valid Arazzo grammar, so it throws:
   > `Runtime expression is not valid: $.items[*].attributes[0]`

## Why this matters

**Any API that accepts a JSONPath, XPath, SPARQL, OPA, or similar query string in its request body will hit this bug** — and these are extremely common patterns:

- OpenAPI's own `DiscoverAction` (`message.intent.filters.expression`) uses JSONPath
- OPA policy engines accept Rego expressions
- GraphQL variables can contain `$variable` references
- Search/filter APIs routinely accept JSONPath or XPath queries

The Arazzo spec (v1.0.1) states that runtime expressions in string values are delimited by `{}` curly braces: _"Expressions can be embedded into string values by surrounding the expression with `{}` curly braces."_ A bare `$`-prefixed string that does not match the Arazzo expression ABNF grammar should be treated as a literal, not as a failed expression evaluation.

Three open Arazzo spec issues acknowledge the same ambiguity: [#363](https://github.com/OAI/Arazzo-Specification/issues/363), [#424](https://github.com/OAI/Arazzo-Specification/issues/424), [#425](https://github.com/OAI/Arazzo-Specification/issues/425). The spec leans toward literals being safe; the implementation disagrees.

## Expected behaviour

String values in request body payloads that do not match a valid Arazzo runtime expression token should be passed through as literals — not throw an error.

## Suggested fix

In `evaluate.ts`, catch the `lintExpression` throw inside `evaluateExpressionString` / `evaluateRuntimeExpression`. If the expression does not parse as a valid Arazzo grammar token, return it as a literal string rather than rethrowing:

```ts
try {
  lintExpression(expression);
} catch {
  return expression; // not a valid Arazzo expression — treat as literal
}
```

## Environment

- `@redocly/cli` 2.29.x – 2.30.3 (latest)
- Arazzo spec v1.0.1
- Reproduced on macOS and Linux

DmitryAnansky:
## What/Why/How?

Changed runtime-expression detection so it no longer relies on a broad regex that treats any $... string as an Arazzo runtime expression.

The new flow is:

1. For string payloads, `isPureRuntimeExpression()` first does only a light shape check.
2. It then calls the existing `lintExpression()` parser validation.
3. If the string is valid Arazzo grammar, it is evaluated normally.
4. If it looks like JSONPath/XPath/SPARQL/OPA and fails Arazzo validation, **it stays unchanged instead of throwing**.
5. Embedded replacements use the same validation guard before evaluating matches.

Added regression coverage to ensure query-language strings like $.items[*].attributes[?(@.type=='foo')] pass through as literal payload values, while normal runtime expressions still resolve.

## Reference

Closes: https://github.com/Redocly/redocly-cli/issues/2798

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Adjusts runtime-expression detection/evaluation for request-body strings, which could change when substitutions occur or when invalid expressions are left literal. Guarded by parser-based validation and expanded tests, but behavior changes may impact existing payload templating edge cases.
> 
> **Overview**
> Fixes Respect request-body handling so *query-language strings* (e.g. JSONPath/XPath/SPARQL/OPA that contain `$`) are no longer mistakenly parsed/evaluated as Arazzo runtime expressions.
> 
> Runtime-expression detection now uses `lintExpression()` validation (via a new `isValidRuntimeExpression()` guard) before evaluating both whole-string expressions (`isPureRuntimeExpression`) and embedded replacements (`evaluateExpressionsInString`), while preserving `$faker` evaluation. Tests add regression coverage for pass-through query strings and strengthen assertions for embedded `{$faker...}` replacements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb04f518627406e2d4538e9ab831190445cc6b99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/silent-words-divide.md
+++ b/.changeset/silent-words-divide.md
@@ -2,4 +2,4 @@
 "@redocly/respect-core": patch
 ---
 
-Stopped evaluating query-language strings (JSONPath, XPath, SPARQL, OPA) in Respect request bodies as runtime expressions.
+Fixed an issue where query-language strings (JSONPath, XPath, SPARQL, OPA) in Respect request bodies were incorrectly treated as runtime expressions.
--- a/.changeset/silent-words-divide.md
+++ b/.changeset/silent-words-divide.md
@@ -1,0 +1,5 @@
+---
+"@redocly/respect-core": patch
+---
+
+Stopped evaluating query-language strings (JSONPath, XPath, SPARQL, OPA) in Respect request bodies as runtime expressions.
--- a/packages/respect-core/src/modules/__tests__/runtime-expressions/evaluate.test.ts
+++ b/packages/respect-core/src/modules/__tests__/runtime-expressions/evaluate.test.ts
@@ -339,6 +339,33 @@ describe('evaluateRuntimeExpressionPayload', () => {
     ).toEqual('2023-12-15');
   });
 
+  it('should not evaluate query-language strings as runtime expressions', () => {
+    const payload = {
+      jsonpath: "$.items[*].attributes[?(@.type=='foo')]",
+      xpath: '$bookstore/book[price>35]/title',
+      sparql: 'SELECT $item WHERE { $item <http://example.com/type> "foo" }',
+      opa: 'data.items[_].type == "foo"',
+    };
+
+    expect(
+      evaluateRuntimeExpressionPayload({ payload, context: runtimeExpressionContext, logger })
+    ).toEqual(payload);
+  });
+
+  it('should evaluate {$faker.*} wrapped expression embedded in object payload', () => {
+    const payload = { x: '{$faker.number.integer({ min: 5, max: 5 })}' };
+    expect(
+      evaluateRuntimeExpressionPayload({ payload, context: runtimeExpressionContext, logger })
+    ).toEqual({ x: '5' });
+  });
+
+  it('should evaluate {$faker.*} wrapped expression mixed with surrounding text', () => {
+    const payload = '{$faker.number.integer({ min: 5, max: 5 })} suffix';
+    expect(
+      evaluateRuntimeExpressionPayload({ payload, context: runtimeExpressionContext, logger })
+    ).toEqual('5 suffix');
+  });
+
   it('should evaluate $faker runtime expression value', () => {
     const payload = '$faker.number.integer({ min: 1, max: 10 })';
     expect(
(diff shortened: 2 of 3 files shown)


.changeset/silent-words-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@redocly/respect-core": patch
+---
+
+Stopped evaluating query-language strings (JSONPath, XPath, SPARQL, OPA) in Respect request bodies as runtime expressions.
```
JLekawa:
```suggestion
Fixed an issue where query-language strings (JSONPath, XPath, SPARQL, OPA) in Respect request bodies were incorrectly treated as runtime expressions.
```

JLekawa (approved):

Daryna-del (approved):

PR #2320, opened 2025-09-22
fix: no-required-schema-properties-undefined now considers properties in any/oneOf

baywet:
## What/Why/How?

fixes #2061
fixes #2060

## Reference

Required keyword is not defined as being only valid if the properties are defined within the same schema https://json-schema.org/draft/2020-12/json-schema-validation#name-required

## Testing

Ran the existing and additional unit tests

## Check yourself

* \[x] Code changed? - Tested with Redoc/Realm/Reunite (internal)
* \[x] All new/updated code is covered by tests
* ~\[ ] New package installed? - Tested in different environments (browser/node)~
* \[x] Documentation update considered

## Security

* \[x] The security impact of the change has been considered
* \[ ] Code follows company security practices and guidelines

⦁	chore: refactoring

⦁	fix: adds oneOf properties to the properties evaluated for required validation

⦁	fix: no-required-schema-properties-undefined now considers any/oneOf

Issue #2061, opened 2025-04-17
Built in rule: no-required-schema-properties-undefined - required with sibling composition keywords


jeremyfiel:
**Describe the bug**

related #2060 

uses the same redocly.yaml file.  Error received is for `anyOf > required` keyword are undefined.

This is a different test case of similar behavior to the other issue.  This one has sibling properties: `required` and a composition keyword `allOf, oneOf, anyOf`.  It doesn't look like the code checks for sibling properties.

```yaml
openapi: 3.0.4
info:
  title: >-
    test sibling required and composition keywords
    no required-schema-definitions-undefined
  version: 0.0.1
paths:
  '/things':
    get:
      responses:
        '200':
          description: OK
          content:
            'application/json':
              schema:
                $schema: http://json-schema.org/draft-04/schema#
                title: serviceEndpoint
                description: Defines the endpoint information for an API Specification implementation
                type: object
                properties:
                  hostName:
                    description: The host name for the endpoint
                    type: string
                  ipAddress:
                    description: The IP Address for the endpoint
                    $ref: ./ipAddressType_v01.json
                  port:
                    description: The port to use
                    type: string
                  healthCheck:
                    type: object
                    properties:
                      uri:
                        description: The URI used to reach the health check.  This URI should not
                          include the host
                        $ref: ./uriPathType_v01.json
                      callInterval:
                        description: The interval in which to call the health check
                        $ref: ../../common/numberUnitTimeType_v01.json
                    required:
                      - uri
                      - callInterval
                    additionalProperties: false
                  status:
                    description: Service endpoint status indicator
                    type: string
                    enum:
                      - UP
                      - DOWN
                    default: UP
                anyOf:
                  - required:
                      - hostName
                  - required:
                      - ipAddress
                required:
                  - port


```

**To Reproduce**

1. Given this `redocly.yaml` file
2. And this OpenAPI file(s)
3. Run this command with these arguments... `redocly ...`

**Expected behavior**

I would go one step further and also test sibling composition keywords together..

```yaml
openapi: 3.0.4
info:
  title: >-
    test sibling required and composition keywords
    no required-schema-definitions-undefined
  version: 0.0.1
paths:
  '/things':
    get:
      responses:
        '200':
          description: OK
          content:
            'application/json':
              schema:
                properties:
                  test:
                    type: string
                  something:
                    type: string
                  else:
                    type: string
                  again:
                    type: string
                  here:
                    type: string
                required:
                  - test
                anyOf:
[...]

Issue #2060, opened 2025-04-17
Built in rule: no-required-schema-properties-undefined - nested allOf, oneOf, anyOf


jeremyfiel:
**Describe the bug**

This rule is failing to resolve nested allOf, oneOf, anyOf subschemas to determine if the `required` keyword has valid definitions.

```yaml
openapi: 3.0.4
info:
  title: >-
    test nested composition keywords
    no required-schema-definitions-undefined
  version: 0.0.1
paths:
  '/things':
    get:
      responses:
        '200':
          description: OK
          content:
            'application/json':
              schema:
                properties:
                  validityPeriod:
                    description: Time bound validity period of the offer
                    allOf:
                      - $schema: http://json-schema.org/draft-04/schema#
                        title: DateTimePeriodType
                        description:
                          The string representation of the date-time range used to specify
                          effective period.
                        type: object
                        properties:
                          startDateTime:
                            $ref: '#/components/schemas/DateTimeType'
                          endDateTime:
                            $ref: '#/components/schemas/DateTimeType'
                      - anyOf:
                          - required:
                              - startDateTime
                          - required:
                              - endDateTime
              examples:
                startDate:
                  value:
                    validityPeriod:
                      startDateTime: 2024-04-01T00:00:01Z
                endDate:
                  value:
                    validityPeriod:
                      endDateTime: 2024-04-01T00:00:01Z
                fullPeriod:
                  value:
                    validityPeriod:
                      startDateTime: 2024-04-01T00:00:01Z
                      endDateTime: 2024-04-01T00:00:01Z
                invalidPeriod:
                  value:
                    validityPeriod: {}
components:
  schemas:
    DateTimeType:
      $schema: http://json-schema.org/draft-04/schema#
      title: DateTimeType
      description: The string representation of the date-time value using the
        ISO-8601:2000 format
      type: string
      format: date-time

```

**To Reproduce**

1. Given this `redocly.yaml` file
2. And this OpenAPI file(s)
3. Run this command with these arguments... `redocly ...`

```yaml
# redocly.yaml
rules:
  no-required-schema-properties-undefined: error
  no-invalid-media-type-examples: error
```
**Expected behavior**

This is a valid schema definition

**Logs**

**OpenAPI description**

**Redocly Version(s)**

Redocly 1.34.1

**`Node.js` Version(s)**

node 20+

**OS, environment**

Windows 10 
**Additional context**

tatomyr:
[...]

PR #2836, opened 2026-05-22
fix(no-required-schema-properties-undefined): resolve false positives for bare-required schemas in allOf-composed types

jeremyfiel:
## Summary

Fixes false positives in `no-required-schema-properties-undefined` for bare `required` constraints used in parent-property composition contexts.

## Problem

The `no-required-schema-properties-undefined` rule could report false positives for valid JSON Schema patterns where a bare `required` list is used as a presence constraint on a property sub-schema, while the property’s actual type is defined through a parent composition context.

This affected two cases:

1. **Bare-required property sub-schema override**
```yaml
properties:
  personName:
    required: [givenName, familyName]
allOf:
- $ref: '#/components/schemas/PersonBase'
```
2. **Constraint-fragment branches**
```yaml
properties:
  communication:
    oneOf:
    - required: [landlines]
    - required: [mobiles]
allOf:
- $ref: '#/components/schemas/PersonBase'
```

Both patterns are valid under JSON Schema, where required and properties are independent keywords.

## Solution

The rule now checks the parent-property context before reporting a missing required property. When the usual local-property and composition-root checks do not find the required key, it walks up the ancestor chain to find the containing property schema and then evaluates sibling composition branches (`allOf`/`anyOf`/`oneOf`) for that property.

This refactor keeps the logic focused on the parent-property context rather than treating it as a general composition-root lookup. In particular, the new parent-context check is distinct from `findCompositionRoot` / `isCompositionChild`: the latter answers whether the current schema belongs to a composition chain, while the new logic answers whether the enclosing property owns the schema and whether a sibling compositional branch defines the required property.

This preserves the existing behavior for misspellings: if a required key is not actually defined in the resolved property type, the rule still reports it as undefined.

## Testing

Added regression coverage for:

- bare required constraints on property sub-schemas resolved through a parent `allOf` sibling
- `oneOf`-style constraint fragments
- `anyOf`-style constraint fragments
- existing misspelling behavior and other composition-related cases
- a new test ensuring the rule still reports undefined properties when the required key is not present in the parent-context sibling composition

## Documentation

Updated the V2 rule documentation to describe the supported parent-context behavior for `oneOf`/`anyOf` constraint fragments.

relates to #2320 #2060 #2061

## Check yourself

## Security

---

> [!NOTE]
> **Low Risk**
> Scoped to one lint rule’s property-resolution logic with broad test and doc updates; no auth, runtime API, or data-path changes.
> 
> **Overview**
> Fixes **false positives** in `no-required-schema-properties-undefined` when schemas use valid JSON Schema patterns: property sub-schemas with only `required` (no local `properties`), or `oneOf`/`anyOf` branches that are pure `required` fragments, while the property’s shape comes from a parent `allOf` sibling (often via `$ref`).
> 
> The rule’s `Schema` visitor now falls back to **`hasPropertyInParentContext`**: it walks ancestor schemas, finds which named property holds the current sub-schema, and checks `allOf`/`anyOf`/`oneOf` siblings for definitions of keys listed in `required`. Misspelled required keys still fail that check and are reported.
> 
> Lint messages from the main schema path use **`is undefined`** instead of **`is not defined`**. Docs add **PersonBase/Person** examples for these patterns; a changeset marks `@redocly/openapi-core` and `@redocly/cli` as patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d34a79f715e4f8484b397171ce90346fee3c0ed4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/nasty-peas-agree.md
+++ b/.changeset/nasty-peas-agree.md
@@ -1,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Updated `no-required-schema-properties-undefined` rule to cover additional edge cases.
--- a/packages/core/src/rules/common/__tests__/no-required-schema-properties-undefined.test.ts
+++ b/packages/core/src/rules/common/__tests__/no-required-schema-properties-undefined.test.ts
@@ -46,7 +46,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'test' is not defined.",
+          "message": "Required property 'test' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -60,7 +60,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'test2' is not defined.",
+          "message": "Required property 'test2' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -110,7 +110,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'test' is not defined.",
+          "message": "Required property 'test' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -234,7 +234,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'foo' is not defined.",
+          "message": "Required property 'foo' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -290,7 +290,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'surname' is not defined.",
+          "message": "Required property 'surname' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -433,7 +433,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'test' is not defined.",
+          "message": "Required property 'test' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
@@ -542,7 +542,7 @@ describe('no-required-schema-properties-undefined', () => {
               "source": "foobar.yaml",
             },
           ],
-          "message": "Required property 'name' is not defined.",
+          "message": "Required property 'name' is undefined.",
           "reference": "https://redocly.com/docs/cli/rules/common/no-required-schema-properties-undefined",
           "ruleId": "no-required-schema-properties-undefined",
           "severity": "error",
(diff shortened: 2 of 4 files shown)


jeremyfiel:
cc @baywet  @drowze  would appreciate your feedback if you are using these constructs.

docs/@v1/rules/oas/no-required-schema-properties-undefined.md
```diff
@@ -98,6 +98,86 @@ schemas:
       - Name
 ```
 
+The rule accepts bare `required` constraints on property sub-schemas when the property's type is defined in a parent `allOf` sibling. This is a valid JSON Schema pattern for adding presence constraints on top of a referenced base type:
```
tatomyr: Please do not modify V1 docs.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -81,13 +81,50 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          }
+          return false;
+        };
+
         for (const [i, requiredProperty] of currentSchema.required.entries()) {
           if (
             !hasProperty(currentSchema, requiredProperty, new Set()) &&
-            !hasProperty(compositionRoot, requiredProperty, new Set())
+            !hasProperty(compositionRoot, requiredProperty, new Set()) &&
+            !hasPropertyInParentContext(requiredProperty, compositionRoot ?? currentSchema)
```
tatomyr: This line actually seems like it should do the same as the previous one, but from a different prospective. Maybe it worth modifying `findCompositionRoot` / `isCompositionChild` instead? 
Also, the existing code has protection against circular references. Does your code have it? (I haven't found that in your tests.)

tatomyr (commented):
Finally got my hands on this. It looks mostly okay for me. However, I'd like to ask you to explore the possibility to solve this without introducing a new concept leveraging the existing logic (if possible).

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
jeremyfiel: I think line 85 is actually a different concern from `findCompositionRoot` / `isCompositionChild`.

- `isCompositionChild` only checks whether the current schema is an item in a parent `allOf` / `anyOf` / `oneOf`.
- `findCompositionRoot` then walks that composition chain upward to the outer composed ancestor.
 
The logic at L85 is not just “what composition chain this schema belongs to”; it is “which ancestor property owns this schema, and does a sibling compositional branch define the required property?”
 
So I’d keep `findCompositionRoot` for composition-root lookup, and keep the parent-property / sibling-branch logic separate.


I did update the code a bit to reuse some existing utils like getOwn.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -80,14 +81,58 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+                if (!siblingSchema?.properties) return false;
+                const propertyDef = getOwn(
+                  siblingSchema.properties as Record<string, AnySchema>,
+                  propertyKey
+                ) as AnySchema | undefined;
+                return (
+                  propertyDef !== undefined &&
+                  schemaHasProperty(propertyDef, propertyName, ctx, new Set(), location)
+                );
+              });
```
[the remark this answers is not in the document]
jeremyfiel: I feel like this _could be_ important for some very advanced schema users but, also a pretty advanced edge case.  I'll play with this a bit and see if I can get it working with nested composition but, as it stands, I think I'm ok with the current implementation, if you are.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -80,14 +81,58 @@ export const NoRequiredSchemaPropertiesUndefined:
...
           if (
-            !schemaHasProperty(currentSchema, requiredProperty, ctx) &&
-            !schemaHasProperty(compositionRoot, requiredProperty, ctx)
+            !schemaHasProperty(currentSchema, requiredProperty, ctx, new Set(), ctx.location) &&
+            !schemaHasProperty(compositionRoot, requiredProperty, ctx, new Set(), ctx.location) &&
+            !hasPropertyInParentContext(requiredProperty, propertyContainerSchema)
           ) {
             ctx.report({
-              message: `Required property '${requiredProperty}' is not defined.`,
+              message: `Required property '${requiredProperty}' is undefined.`,
```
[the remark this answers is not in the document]
jeremyfiel: i'll leave this one to you if you want to update the grammar.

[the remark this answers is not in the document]
tatomyr: I'm okay with the grammar, but please keep the wording consistent.

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -80,14 +81,58 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          }
+
+          return false;
+        };
 
         for (const [i, requiredProperty] of currentSchema.required.entries()) {
           if (
-            !schemaHasProperty(currentSchema, requiredProperty, ctx) &&
-            !schemaHasProperty(compositionRoot, requiredProperty, ctx)
+            !schemaHasProperty(currentSchema, requiredProperty, ctx, new Set(), ctx.location) &&
```
tatomyr, line 130: Could you simply take the `location` from the context (already passed as the 3rd argument) inside `schemaHasProperty`?

packages/core/src/rules/common/__tests__/no-required-schema-properties-undefined.test.ts
```diff
@@ -931,90 +987,43 @@ describe('no-required-schema-properties-undefined', () => {
...
-    const results = await lintDocument({
-      externalRefResolver: new BaseResolver(),
-      document,
-      config: await createConfig({ rules: { 'no-required-schema-properties-undefined': 'error' } }),
-    });
-
     expect(replaceSourceWithRef(results)).toMatchInlineSnapshot(`[]`);
   });
 
-  it('should not report a required property defined further down a composed $ref chain', async () => {
```
tatomyr, line 992: Why have you removed tests? If you think there's no need in a specific test anymore, please leave a PR comment explaining why it has been removed.

.changeset/nasty-peas-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Updated `no-required-schema-properties-undefined` rule to cover additional edge cases.
```
tatomyr, line 6: > to cover additional edge cases

This sounds too general. Could you be a tiny bit more specific (not too much though 🙂)?

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
tatomyr: Ah, I see. It could differ to the one from the ctx. But this makes the arguments too messy. Let's wrap them in an object then.

Issue #2797, opened 2026-05-04
Bundler does not use user defined files


hojows:
I have this file structure in my project:
src/
├── index.yaml (Root OpenAPI definition)
├── info/
│   ├── index.yaml (API Metadata: title, version)
│   ├── contact.yaml (Contact information)
│   └── license.yaml (License information)
├── servers/
│   └── index.yaml (Server configuration, default: http://localhost:8080)
├── security/
│   └── index.yaml (Global security requirements: bearerAuth)
├── tags/
│   └── index.yaml (Tag definitions)
├── components/
│   ├── index.yaml (Entry point for components)
│   ├── callbacks/
│   │   └── index.yaml
│   ├── examples/
│   │   └── index.yaml
│   ├── links/
│   │   └── index.yaml
│   ├── securitySchemes/
│   │   ├── index.yaml
│   │   └── bearerAuth.yaml (JWT scheme definition)
│   ├── headers/
│   │   ├── index.yaml
│   │   └── auth/
│   │       └── TenantHeader.yaml (Tenant header definition)
│   ├── parameters/
│   │   ├── index.yaml
│   │   ├── path/ (Path items definitions)
│   │   └── query/ (Query items definitions)
│   ├── requestBodies/
│   │   ├── index.yaml
│   │   ├── [...] (Request bodies definitions)
│   ├── responses/
│   │   ├── index.yaml 
│   │   ├── common/ (Standard errors: 400, 401, 404, 500, etc.)
│   │   ├── [...]/ (Responses definitions)
│   └── schemas/
│       ├── index.yaml
│       ├── models/ (Entities definitions)
│       └── requests/ (Payloads for commands definitions)
└── paths/
    ├── index.yaml (Endpoint mapping)
    └── [...]/ (Routes definitions)

When I run the bundle command, though, it doesn't read my "barrel files" (the index.yaml in each folder) but it creates its own object in components starting from the path reference and naming them with the file name.

1. Root index should contain 
components:
  $ref: ./components/index.yaml
and that file should contain
schemas:
  $ref: ./schemas/index.yaml
this file should have an object structure, like:
ObjectModel:
  $ref: ./ObjectName.yaml
2. Run bundle command

I would expect the bundled definition to contain 'ObjectModel' (as defined in my schema file); but it rather contains 'ObjectName' (taking it from the referenced file name).

services:
  bundle:
    image: redocly/cli:latest
    volumes:
      - ./:/spec
    working_dir: /spec
[...]

PR #2851, opened 2026-06-03
feat: add bundle `--component-names-strategy` flag

kanoru3101:
## What/Why/How?
Added an opt-in `--component-names-strategy` flag to `bundle`: inlined Schema components are named from each schema's `title` (PascalCased format) instead of the file basename.
## Reference
#2797

### With flag

Screenshot 2026-06-03 at 22 07 20

### Error `missing title`

Screenshot 2026-06-03 at 23 32 15

### Warn `title collision`

Screenshot 2026-06-10 at 11 01 39

### Warn `two title collision`

Screenshot 2026-06-10 at 11 02 47

## Check yourself

## Security

---

> [!NOTE]
> **Low Risk**
> Opt-in bundle flag with default unchanged; scope is schema naming during bundling with tests and docs, not auth or runtime API behavior.
> 
> **Overview**
> Adds **`--component-names-strategy`** to `redocly bundle` (default **`basename`**, opt-in **`title`**) so inlined **Schema** components can be named from each schema’s **`title`** (PascalCased and sanitized for valid component keys) instead of file basename / pointer.
> 
> The bundler gains title-based naming with errors for missing `title`, warnings (or errors via existing **`--component-renaming-conflicts-severity`**) for title collisions, and basename fallback when title naming fails; **parameters** and other component types keep the previous behavior. Shared **`COMPONENT_NAME_CHARS`** is used for sanitization and the **`spec-components-invalid-map-name`** rule.
> 
> CLI wiring, bundle docs, changeset, unit/e2e tests, and help snapshots are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9612a51ac0727020f3c7231216d4d4e383a4c960. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/@v2/commands/bundle.md
```diff
@@ -147,3 +148,31 @@ You can adjust how the CLI handles these naming conflicts with the `--component-
...
+# schemas/requests/Authority.yaml
+type: object
+title: Authority request
+```
+
+```bash
+redocly bundle openapi.yaml -o bundled.yaml --use-titles-for-component-names
+```
+
+The output now uses `AuthorityModel` and `AuthorityRequest`. A `title` containing `.`, `-`, or `_` keeps them — `order-item` becomes `Order-item`.
```
JLekawa:
```suggestion
The output uses `AuthorityModel` and `AuthorityRequest`.
A `title` containing `.`, `-`, or `_` keeps them — `order-item` becomes `Order-item`.
```

docs/@v2/commands/bundle.md
```diff
@@ -147,3 +148,31 @@ You can adjust how the CLI handles these naming conflicts with the `--component-
 - `off`: No warnings or errors are shown.
 - `warn` (default): Shows a warning and renames conflicting components automatically.
 - `error`: Treats conflicts as errors; the bundling process fails if a naming conflict is detected.
+
+### Use titles for component names
+
+By default, the bundler names each inlined Schema component after its `$ref` — the JSON Pointer fragment, or the file's basename when there's no fragment. Two files that share a basename — for example `schemas/models/Authority.yaml` and `schemas/requests/Authority.yaml` — collide into `Authority` and `Authority-2`. That auto-numbered suffix is brittle: an unrelated `$ref` change can renumber it.
```
JLekawa:
```suggestion
By default, the bundler names each inlined Schema component after its `$ref` — the JSON Pointer fragment, or the file's basename when there's no fragment.
Two files that share a basename — for example `schemas/models/Authority.yaml` and `schemas/requests/Authority.yaml` — collide into `Authority` and `Authority-2`.
The auto-numbered suffix is brittle: an unrelated `$ref` change can renumber it.
```

JLekawa (approved):

tatomyr:
@kanoru3101 let's check bugbot's comments. Also, please do not just resolve them, but leave meaningful comments.

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -297,9 +305,30 @@ export function makeBundleVisitor({
...
+        const title = (target.node as { title?: string }).title;
+        ctx.report({
+          message:
+            `Title "${title}" maps to component name \`${titleName}\`, ` +
+            `already used by another schema. Rename one of the titles.`,
+          location: target.location.child('title'),
+          from: titleNameLocations.get(titleName),
+          forceSeverity: 'error',
+        });
+        return target.location.absolutePointer;
```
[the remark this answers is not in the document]
kanoru3101: It's already fixed.
If user uses `--force` with `--use-titles-for-component-names`, it creates component names with a suffix.

docs/@v2/commands/bundle.md
```diff
@@ -147,3 +148,34 @@ You can adjust how the CLI handles these naming conflicts with the `--component-
...
+redocly bundle openapi.yaml -o bundled.yaml --use-titles-for-component-names
+```
+
+The output uses `AuthorityModel` and `AuthorityRequest`.
+A `title` containing `.`, `-`, or `_` keeps them — `order-item` becomes `Order-item`.
+
+Bundling fails (no output file unless you pass `--force`) when:
+
+- A referenced schema has no `title`.
+- A `title` uses anything other than ASCII letters, digits, spaces, `.`, `-`, or `_` (for example `&` or a non-ASCII title).
```
tatomyr: Could you explain your consideration for choosing this characters set?

packages/cli/src/commands/bundle.ts
```diff
@@ -39,6 +39,7 @@ export type BundleArgv = {
   'remove-unused-components'?: boolean;
   'keep-url-references'?: boolean;
   'component-renaming-conflicts-severity'?: RuleSeverity;
+  'use-titles-for-component-names'?: boolean;
```
tatomyr: Could this option conflict with other options?

packages/core/src/bundle/bundle-title-naming.ts
```diff
@@ -0,0 +1,43 @@
+import type { Location } from '../ref-utils.js';
+import type { UserContext } from '../walk.js';
+
+/** Build a PascalCase component key from a `title`; `null` if empty or has unsupported characters. */
```
tatomyr: - Use a meaningful name so we don't need comments. 
- Consider making this function general (not only for titles but for any kind of strings), maybe it can be reused.
- Null is not a valid object key, isn't any better option out there?

tatomyr (commented):
Could there be fewer changes? It's only adding a new option, so I'd suggest isolating everything related to it so the changes are not scattered too much. 
Also, please make sure all tests are meaningful and don't duplicate each other.

docs/@v2/commands/bundle.md
kanoru3101: The character it's the Components Object key pattern from the `OpenAPI` and `AsyncAPI` specs `(^[a-zA-Z0-9.\-_]+$)`, the same one the `spec-components-invalid-map-name` rule uses.

packages/cli/src/commands/bundle.ts
kanoru3101: Tested with different flags, seems to work without conflicts

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -297,9 +305,32 @@ export function makeBundleVisitor({
...
+        const title = (target.node as { title?: string }).title;
+        ctx.report({
+          message:
+            `Title "${title}" maps to component name \`${titleName}\`, ` +
+            `already used by another schema. Rename one of the titles.`,
+          location: target.location.child('title'),
+          from: titleNameLocations.get(titleName),
+          forceSeverity: 'error',
+        });
+        return baseName;
```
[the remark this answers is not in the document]
kanoru3101: added fallback

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -291,22 +299,84 @@ export function makeBundleVisitor({
...
+    target: ComponentTarget,
+    componentsGroup: Record<string, unknown>,
+    ctx: UserContext,
+    problem: { message: string; location: Location; from?: Location }
+  ) {
+    const name = componentNameFromBasename(target, componentsGroup, ctx).name;
+    if (!componentsGroup[name]) {
+      ctx.report({ ...problem, forceSeverity: 'error' });
+      titleNameLocations.set(name, target.location);
+    }
```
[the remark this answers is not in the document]
kanoru3101: The error is reported per output component, not per source file. The bundler merges schemas into one component `a/Thing.yaml` and `b/Thing.yaml` both become a single `#/components/schemas/Thing`. So two identical sources are one component and get one error. Schemas with different content become different components and are each reported.

kanoru3101:
> Could there be fewer changes? It's only adding a new option, so I'd suggest isolating everything related to it so the changes are not scattered too much. Also, please make sure all tests are meaningful and don't duplicate each other.

I rewrote the tests and shortened them. Some of the tests check error handling, account for conflicts, and cover various types of errors

packages/core/src/__tests__/fixtures/refs/title-naming-collision/schemas/a/User.yaml
```diff
@@ -0,0 +1,5 @@
+type: object
+title: User
+properties:
+  id:
```
tatomyr: Let's tie the property to the file name.

packages/core/src/__tests__/bundle.test.ts
```diff
@@ -844,3 +844,73 @@ describe('sibling $ref resolution by spec', () => {
...
+    expect(Object.keys(parsed.components.schemas).sort()).toEqual([
+      'AuthorityModel',
+      'AuthorityRequest',
+    ]);
+    expect(
+      parsed.paths['/authorities'].get.responses['200'].content['application/json'].schema
+    ).toEqual({ $ref: '#/components/schemas/AuthorityModel' });
+    expect(
+      parsed.paths['/authorities'].post.requestBody.content['application/json'].schema
+    ).toEqual({ $ref: '#/components/schemas/AuthorityRequest' });
```
tatomyr: Could you use inline snapshots instead?

packages/core/src/__tests__/utils.test.ts
```diff
@@ -139,6 +140,20 @@ describe('utils', () => {
...
+  describe('toPascalCase', () => {
+    it.each([
+      ['Authority model', 'AuthorityModel'],
+      ['authority-model', 'Authority-model'],
+      ['user_profile', 'User_profile'],
+      ['foo.bar', 'Foo.bar'],
+      ['API v2 User', 'APIV2User'],
+      ['  padded  ', 'Padded'],
+      ['', ''],
+    ])('converts %j to %j', (input, expected) => {
```
tatomyr: What is `%j`?

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -206,7 +223,7 @@ export function makeBundleVisitor({
   };
 
   if (version === 'oas3') {
-    const componentType = mapTypeToComponent('Schema', version)!;
+    const componentType = schemaComponentType!;
```
tatomyr: Why reassigning?

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -291,22 +300,88 @@ export function makeBundleVisitor({
...
+    let name = prevName;
     let serialId = 2;
     while (componentsGroup[name] && !isEqualOrEqualRef(componentsGroup[name], target, ctx)) {
       name = `${prevName}-${serialId}`;
       serialId++;
     }
+    return { name, prevName };
+  }
+
+  function componentNameFromTitle(
```
tatomyr, line 298: Let's check if we can reuse the existing logic.

tests/e2e/bundle/bundle.test.ts
```diff
@@ -200,3 +201,20 @@ describe('bundle with long description', () => {
...
+describe('bundle with option: use-titles-for-component-names', () => {
+  test('fails when two schemas share a title-based name', async () => {
+    const testPath = join(__dirname, 'bundle-use-titles-for-component-names-collision');
+    const entryPoints = getEntrypoints(testPath);
+    const args = [
+      indexEntryPoint,
+      'bundle',
+      '--use-titles-for-component-names',
+      '--component-renaming-conflicts-severity',
+      'error',
```
tatomyr, line 215: LEt's add `off` as well.

kanoru3101: Added test with `off`

packages/core/src/__tests__/utils.test.ts
kanoru3101: The names `$input` and `$expected` have been changed for clarity

packages/core/src/__tests__/bundle.test.ts
kanoru3101: Changed to `toMatchInlineSnapshot` where checking error messages and left `toMatchSnapshot` for the happy path

packages/core/src/bundle/bundle-visitor.ts
kanoru3101: Removed this reassigning

packages/core/src/bundle/bundle-visitor.ts
kanoru3101: I tried collapsing it into a single if inside `getComponentName` without the extracted helper, but it comes out worse: the title path has failure modes the basename path doesn't (missing title / unusable chars / collision) and on failure it falls back to the basename, so you either duplicate the `-N` loop or scatter `useTitle && `checks through one big function. Keeping `componentNameFromBasename` as the shared primitive avoids both while leaving the default logic untouched.

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -116,17 +122,23 @@ export function makeBundleVisitor({
...
   resolvedRefMap: ResolvedRefMap;
   keepUrlRefs: boolean;
   componentRenamingConflicts?: RuleSeverity;
+  useTitlesForComponentNames?: boolean;
 }) {
-  let components: Record<string, Record<string, unknown>>;
+  let components: Record<string, ComponentsGroup>;
   let rootLocation: Location;
 
+  const firstSchemaLocationByName = new Map<string, Location>();
```
kanoru3101, line 139: firstSchemaLocationByName exists only to populate the collision's from — the "referenced from …" link that points at the first schema which already took the name. It can't be derived the way the basename path handles a conflict, because the two collisions behave differently:
* Basename collision auto-resolves: it appends -N and reports a rename warning at the current location. It never references the other schema, so it needs no location tracking.
*  Title collision is a user-fixable error — the fix is to rename one of the two conflicting titles, so the message points from at the other (first) schema to help the user find both.

tatomyr (approved):
LGTM

drwicid:
Question — Should lint also evaluate the effective derived component name?

My understanding is that bundled inlined schemas already receive derived component names today, traditionally based on the referenced file basename or $ref pointer segment. This option adds another derivation strategy for schemas: when --use-titles-for-component-names is enabled, the component name can instead come from the schema’s title, with fallback to the existing basename/pointer logic when the title is missing or invalid.

Given that the bundle process already has to resolve an effective component name either way, would it make sense for the lint/rule phase to expose or reuse that same derived-name calculation?

Linting can certainly validate the presence or shape of title fields directly. However, that only covers one input to the naming strategy. What seems more useful is linting against the final effective name that bundle would produce, regardless of whether that name came from title or from the traditional basename / $ref pointer fallback.

That would allow rules such as component-name-unique to detect collisions before bundling, including collisions introduced by title-derived names as well as collisions that still occur through fallback naming. It would also let validity and naming/casing rules apply to the actual component key that will be emitted, with diagnostics reported against the original schema files and source locations.

This seems preferable to requiring a second lint pass against the generated bundled file, where diagnostics are tied to generated output and are less directly traceable to the originating schema.

I understand the lifecycle split: bundle runs preprocessors, rules, and decorators, while lint runs preprocessors and rules but not decorators. Is keeping effective component-name resolution bundle-only an intentional boundary, or would sharing the derived-name helper with lint be reasonable so these checks can happen before bundling?

docs/@v2/commands/bundle.md
```diff
@@ -45,6 +45,7 @@ redocly bundle --version
...
 | --skip-decorator                        | [string] | Ignore certain decorators. See the [Skip preprocessor, rule, or decorator section](#skip-preprocessor-rule-or-decorator).                                                                                                                                 |
 | --skip-preprocessor                     | [string] | Ignore certain preprocessors. See the [Skip preprocessor, rule, or decorator section](#skip-preprocessor-rule-or-decorator).                                                                                                                              |
+| --use-titles-for-component-names        | boolean  | Build Schema component names from each schema's `title` field. See [Use titles for component names](#use-titles-for-component-names).                                                                                                                     |
```
RomanHotsiy: Let's change it to --component-names-strategy with options like "basename" (default) or "title"?

@tatomyr what do you think?

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -291,30 +294,97 @@ export function makeBundleVisitor({
...
+    const { node } = target;
+    const title = isPlainObject(node) && isString(node.title) ? node.title.trim() : '';
+    const key = toPascalCase(title);
+    const titleLocation = target.location.child('title');
+
+    if (title === '') {
+      return {
+        key,
+        problem: {
+          message: 'Schema must define a `title` to build a component name.',
```
RomanHotsiy: Need to clarify it is required only when using this option or someone may get confused:

```suggestion
          message: 'Schema must define a `title` when using --use-titles-for-component-names.',
```

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -291,30 +294,97 @@ export function makeBundleVisitor({
...
+          location: target.location,
+          forceSeverity: 'error',
+        },
+      };
+    }
+    if (!new RegExp(COMPONENT_NAME_PATTERN).test(key)) {
```
RomanHotsiy: do we need it? should we just replace any unsupported symbol ourselves with dashes?

docs/@v2/commands/bundle.md
tatomyr: Sounds even better to me. What about "filename" instead of "basename"?

packages/core/src/bundle/bundle-visitor.ts
kanoru3101: I'll change it to replace and add to docs about it

kanoru3101:
> Question — Should lint also evaluate the effective derived component name?
> 
> My understanding is that bundled inlined schemas already receive derived component names today, traditionally based on the referenced file basename or $ref pointer segment. This option adds another derivation strategy for schemas: when --use-titles-for-component-names is enabled, the component name can instead come from the schema’s title, with fallback to the existing basename/pointer logic when the title is missing or invalid.
> 
> Given that the bundle process already has to resolve an effective component name either way, would it make sense for the lint/rule phase to expose or reuse that same derived-name calculation?
> 
> Linting can certainly validate the presence or shape of title fields directly. However, that only covers one input to the naming strategy. What seems more useful is linting against the final effective name that bundle would produce, regardless of whether that name came from title or from the traditional basename / $ref pointer fallback.
> 
> That would allow rules such as component-name-unique to detect collisions before bundling, including collisions introduced by title-derived names as well as collisions that still occur through fallback naming. It would also let validity and naming/casing rules apply to the actual component key that will be emitted, with diagnostics reported against the original schema files and source locations.
> 
> This seems preferable to requiring a second lint pass against the generated bundled file, where diagnostics are tied to generated output and are less directly traceable to the originating schema.
> 
> I understand the lifecycle split: bundle runs preprocessors, rules, and decorators, while lint runs preprocessors and rules but not decorators. Is keeping effective component-name resolution bundle-only an intentional boundary, or would sharing the derived-name helper with lint be reasonable so these checks can happen before bundling?

Hi @drwicid 
Good question. If I read it right: you'd like lint to reuse bundle's effective-name calculation to catch collisions before bundling.

But before bundle the schemas live in separate files — there's no single combined namespace yet, so the real names (and their collisions) only appear once bundle merges everything. A pre-bundle rule like `component-name-unique` can only approximate them: it compares basename keys and ignores title. Matching bundle's exact names would mean re-running its stateful assembly (-N, title→fallback, dedup) in lint — a fair follow-up, but bigger than this flag.
 
 And collisions are caught at bundle anyway: bundle reports them, `--component-renaming-conflicts-severity=error` makes a collision fail the bundle, and a missing title already fails it — so nothing slips through.

docs/@v2/commands/bundle.md
```diff
@@ -147,3 +148,59 @@ You can adjust how the CLI handles these naming conflicts with the `--component-
...
+```
+
+The output uses `OrderModel` and `OrderRequest`.
+
+{% /tab  %}
+{% /tabs  %}
+
+{% admonition type="warning" name="Title sanitization" %}
+The OpenAPI and AsyncAPI specifications allow only ASCII letters (`a`–`z`, `A`–`Z`), digits, `.`, `-`, and `_` in a Components Object key.
+Those characters are kept; any other character — including non-ASCII letters such as `é` or `я` — is replaced with `-` (for example, `User & Group` becomes `User-Group`).
```
JLekawa:
```suggestion
All other characters, including non-ASCII letters such as `é` or `я`, are replaced with `-` (for example, `User & Group` becomes `User-Group`).
```

.changeset/bright-banks-wish.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Added the `--component-names-strategy` option to the `bundle` command.
+THis option allows a choice of how inline Schema components are named: `basename` (default) or `title` (from each schema's `title` field).
```
JLekawa:
```suggestion
This option allows a choice of how inline Schema components are named: `basename` (default) or `title` (from each schema's `title` field).
```

docs/@v2/commands/bundle.md
```diff
@@ -147,3 +148,59 @@ You can adjust how the CLI handles these naming conflicts with the `--component-
...
+A schema without a `title` can't be named this way, so bundling reports an error for that schema.
```
JLekawa:
```suggestion
Schemas without `title` can't be named using the `--component-names-strategy=title` strategy.
The bundling process reports an error for such schemas.
```

docs/@v2/commands/bundle.md
RomanHotsiy: I'm fine with either

RomanHotsiy (approved):

PR #2855, opened 2026-06-05
feat: add junit output format to the lint

### LINKED ISSUE #1989: Add Junit XML as a linting result format
Hi, I had been using Spectral CLI for linting and been looking to switch to Redocly's linting solution. At my workplace we have a CI Job which lints our OpenAPI specs, stores the test results and report them to CircleCI if any issues are recognized.

Now the issue is that CircleCI only supports uploading test results in Junit XML format which sadly Redocly does not support, hence I am writing this issue.

Adding a new option for the `--format` and writing a backend for it would be the expected solution. I can't comment on specifics.

An alternative I've been looking into is to convert from `checkstyle` to `junit` on the fly from within my CI Job before uploading the test results.

AlbinaBlazhko17:
## What/Why/How?
Added `junit` output format to the `lint` command.

## Reference
Resolves #1989

## Check yourself

## Security

---

> [!NOTE]
> **Low Risk**
> Additive reporting and CLI wiring with tests; no changes to lint rules or API validation behavior.
> 
> **Overview**
> Adds **`junit`** as a `--format` option for `redocly lint`, aimed at CI systems that ingest JUnit XML.
> 
> **Core (`formatProblems`)** emits a single `<testsuites>` document: one `<testsuite>` per linted file, each problem as a `<testcase>` with errors as `<error>` and warnings as `<failure>`, plus a detail block (rule, severity, location, pointer, message). **Empty successful runs** still output valid XML (same idea as `json`).
> 
> **CLI** treats `junit` like `checkstyle`: problems from multiple APIs are **aggregated** into one `formatProblems` call. **Config lint** output is **skipped** for `junit` (with `json` and `checkstyle`) so a second report does not break the single XML document.
> 
> Docs, changeset, CLI yargs choices, unit tests, and an e2e help snapshot are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de99eb56027f7861c5404fb6bdb8f5cce28dc848. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/@v2/commands/lint.md
+++ b/docs/@v2/commands/lint.md
@@ -283,15 +283,29 @@ The `lint` command supports the JUnit XML report format, which is consumed by ma
 Each linted API becomes a `<testsuite>`, and each problem becomes a `<testcase>`.
 Errors are reported as `<error>` elements and warnings as `<failure>` elements, so CI tools that render the two differently keep errors and warnings visually distinct.
 
+Each `<testcase>` body repeats the rule, severity, location, pointer, and message as a readable detail block, matching the [`scorecard-classic`](./scorecard-classic.md) JUnit output.
+
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
-<testsuites name="redocly lint" tests="3" errors="2" failures="1" skipped="0">
-<testsuite name="museum-with-errors.yaml" tests="3" errors="2" failures="1">
+<testsuites name="redocly lint" tests="2" errors="1" failures="1" skipped="0">
+<testsuite name="museum-with-errors.yaml" tests="2" errors="1" failures="1">
 <testcase classname="struct" name="struct - 19:7" file="museum-with-errors.yaml" line="19">
-<error message="Property `operationIds` is not expected here." type="struct">at #/paths/~1museum-hours/get</error>
+<error message="Property `operationIds` is not expected here." type="struct">Rule: struct
+Severity: error
+File: museum-with-errors.yaml
+Line: 19
+Column: 7
+Pointer: #/paths/~1museum-hours/get
+Message: Property `operationIds` is not expected here.</error>
 </testcase>
 <testcase classname="operation-operationId" name="operation-operationId - 16:5" file="museum-with-errors.yaml" line="16">
-<failure message="Operation object should contain `operationId` field." type="operation-operationId">at #/paths/~1museum-hours/get</failure>
+<failure message="Operation object should contain `operationId` field." type="operation-operationId">Rule: operation-operationId
+Severity: warn
+File: museum-with-errors.yaml
+Line: 16
+Column: 5
+Pointer: #/paths/~1museum-hours/get
+Message: Operation object should contain `operationId` field.</failure>
 </testcase>
 </testsuite>
 </testsuites>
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/lint.ts
```diff
@@ -143,8 +143,8 @@ export async function handleLintConfig(argv: Exact<CommandArgv>, version: string
     return;
   }
 
-  if (argv.format === 'json') {
-    // we can't print config lint results as it will break json output
+  if (argv.format === 'json' || argv.format === 'junit' || argv.format === 'checkstyle') {
```
AlbinaBlazhko17, line 146: Added `checkstyle`, because missed in prev PR. We avoid printing lint config results in this formats, because it will break them. e.g. two separate json outputs in one file is not valid json file. Added tests in `lint.test.ts` for this case.

packages/cli/src/__tests__/commands/lint.test.ts
```diff
@@ -288,3 +306,36 @@ describe('handleLint', () => {
     });
   });
 });
+
+describe('handleLintConfig', () => {
```
AlbinaBlazhko17, line 310: Added tests to cover this case: https://github.com/Redocly/redocly-cli/pull/2855/changes#diff-966ed90d020cffe70c22d110135ce19924a3a4450b4ab9df0893012519d4f3a9R146

JLekawa (approved):

docs/@v2/commands/lint.md
```diff
@@ -273,6 +273,46 @@ Run this command to use the following standard format output with your other too
...
+Line: 16
+Column: 5
+Pointer: #/paths/~1museum-hours/get
+Message: Operation object should contain `operationId` field.</failure>
+</testcase>
+</testsuite>
+</testsuites>
+```
+
+A valid (empty) report is produced even when no problems are found, so passing runs still populate the CI test results.
```
tatomyr, line 314: This looks redundant to me. The main goal for our docs is to explain how to use the tool and what options are, not explaining the format itself. The more concise docs are -- the easier they are to read.

packages/cli/src/commands/lint.ts
```diff
@@ -49,8 +49,8 @@ export async function handleLint({
 
   const totals: Totals = { errors: 0, warnings: 0, ignored: 0 };
   let totalIgnored = 0;
-  const isCheckstyle = argv.format === 'checkstyle';
-  const checkstyleResults: Awaited<ReturnType<typeof lint>> = [];
+  const isAggregatedXmlFormat = argv.format === 'checkstyle' || argv.format === 'junit';
```
tatomyr, line 52: Don't we consider every format aggregated? Why's the discrepancy with other formats?

docs/@v2/commands/lint.md
AlbinaBlazhko17: Every format has its own section with an example. I decided to add this section similar to the `checkstyle` format. Do i need to delete this section for `junit` or improve existing?

packages/cli/src/commands/lint.ts
AlbinaBlazhko17: It depends on the previous PR, where we fixed a behavior of the `checkstyle`, where we aggregate all the problems among all files in one xml output. Previously, we output the separate xml for each file, which was wrong and we received an issue.

docs/@v2/commands/lint.md
tatomyr: Let's discuss that offline.

packages/cli/src/commands/lint.ts
tatomyr: Yes, but that makes me think other formats are not aggregated. Is it correct? Why not aggregate all other formats as well?

AlbinaBlazhko17: Valid point. Other formats, such as `stylish` and `markdown`, are for user review and do not need to be output as one Markdown file with two specifications. We would like to see two separate Markdown files for each specification. That's why other formats are not aggregated.

tatomyr (approved):

PR #2937, opened 2026-07-06
feat: add `generate-spec` command

sobanieca-redocly:
## What/Why/How?

Added new `generate-spec` command which will infer OAD basing on traffic data. To improve accuracy, there is option to use AI agents to improve results.

## Screenshots (optional)

image

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Large new surface area and `--with-ai` sends traffic samples to external CLIs, but refinement is gated by validation and AI failure falls back to the deterministic baseline without failing the command.
> 
> **Overview**
> Introduces an experimental **`generate-spec`** CLI command that reads HAR/Kong/Nginx/Apache JSON/NDJSON traffic (same parsers as **`drift`**) and emits an **OpenAPI 3.2** document.
> 
> **Deterministic pipeline:** path templating for ID-like segments, merged request/response schemas (`oneOf`, nullable unions), repeated shapes lifted to **`components/schemas`**, and conservative string **format/enum** inference. Flags include **`--server`**, **`--title`**, and **`-o`**.
> 
> **Optional `--with-ai`:** refines one operation at a time via local **`claude`**, **`codex`**, or **`cursor`** CLIs (isolated temp cwd, configurable concurrency). Accepts only refinements that preserve path/method/status codes and pass the built-in **`spec`** ruleset; otherwise keeps the baseline and falls back on total failure.
> 
> Also adds v2 docs, a minor changeset, CODEOWNERS, broad unit tests, and e2e coverage with stub AI binaries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad8c15f8bcc2d7d7df7fb4556f33c0d69cb6a09a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

sobanieca-redocly:
@cursor review

packages/cli/src/commands/generate-spec/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      server: argv.server,
+    });
+
+    try {
+      const refined = await refineSpecWithAi({
+        provider,
+        model: argv['ai-model'],
+        baseline,
+        samplesByOperation,
+      });
```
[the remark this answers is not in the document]
sobanieca-redocly: This is expected. We don't want to run full lint because it will show errors for baseline document.

sobanieca-redocly:
@cursor review

.changeset/generate-spec-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': minor
+---
+
+Added an experimental `generate-spec` command that infers an OpenAPI description from recorded HTTP traffic (HAR, Kong, Nginx/Apache JSON, NDJSON) passed as a single file or a folder searched recursively. Alternative body shapes are preserved as `oneOf` variants and `null` observations become type unions. The description can optionally be refined with an AI provider (the `claude`, `codex`, or `cursor` CLI): the refinement runs one operation per prompt (up to `--ai-concurrency` operations in parallel, default 4) with progress reported per operation, so it scales to large traffic captures; the provider CLIs run isolated from the machine's configuration (no MCP servers, project instruction files, hooks, or plugins) to keep every call fast and free of unrelated context; refined output is only accepted when it keeps the observed operation and status codes and passes validation with the `spec` ruleset, and each rejected operation keeps its deterministic baseline.
```
JLekawa:
```suggestion
Added an experimental `generate-spec` command that infers an OpenAPI description from recorded HTTP traffic.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,97 @@
...
+- Object shapes that repeat across the document are extracted into `components/schemas` and referenced with `$ref`.
+  The same entity is recognized when it appears as a list item and as a single resource, with different `required` sets, or with near-identical properties (at least 75% shared, with compatible types).
+  Components are named from the path entity, the enclosing property name, or `Error` for error responses; a shape repeated only because its parent shape repeats stays inline.
+- Observed string values are analyzed conservatively: when every value of a property (or parameter) matches the same well-known pattern it gets a `format` (`uuid`, `date-time`, `date`, `email`, `uri`, `ipv4`); when a string only ever takes a small set of identifier-like values with enough repetition (at least 4 observations, at most 5 distinct values, each seen twice on average) it becomes an `enum`.
+  Enums apply to body properties and query parameters; path parameters and nullable unions get formats only.
+  Evidence is pooled across all operations a shared component was observed in.
+- Responses that were never received (status `0` in HAR captures) are ignored.
```
JLekawa:
```suggestion
- Identifier-like path segments (numeric, UUID, ULID, CUID, prefixed and opaque tokens) become named path parameters.
- Body schemas are merged across all observations.
  A property becomes optional as soon as one sample omits it.
- Alternative body shapes for the same operation are preserved as `oneOf` variants instead of being collapsed, and values observed as `null` produce type unions such as `["string", "null"]`.
- Object shapes that repeat across the document are extracted into `components/schemas` and referenced with `$ref`.
  The same entity is recognized when it appears as a list item and as a single resource, with different `required` sets, or with near-identical properties (at least 75% shared, with compatible types).
  Components are named from the path entity, the enclosing property name, or `Error` for error responses.
  A shape repeated only because its parent shape repeats stays inline.
- Observed string values are analyzed conservatively: when every value of a property (or parameter) matches the same well-known pattern it gets a `format` (`uuid`, `date-time`, `date`, `email`, `uri`, `ipv4`).
  When a string only ever takes a small set of identifier-like values with enough repetition (at least 4 observations, at most 5 distinct values, each seen twice on average) it becomes an `enum`.
  Enums apply to body properties and query parameters.
  Path parameters and nullable unions get formats only.
  Evidence is pooled across all operations a shared component was observed in.
- Responses that were never received (status `0` in HAR captures) are ignored.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,97 @@
...
+When two concurrent refinements touch the same shared component, the one merged last wins; the final whole-document lint still guards the result.
+Set `--ai-concurrency 1` to process operations strictly sequentially.
+Progress is reported per operation as refinements complete.
+
+The AI's answer is never trusted blindly.
+A refined operation is only accepted when it:
+
+- keeps the exact path template and method — the AI cannot invent, drop, or rename operations because only the requested operation is merged back;
+- keeps every response status code observed in the traffic;
+- passes validation with the `spec` ruleset (checked against the description's full component set).
```
JLekawa:
```suggestion
- keeps the exact path template and method — the AI cannot invent, drop, or rename operations because only the requested operation is merged back
- keeps every response status code observed in the traffic
- passes validation with the `spec` ruleset (checked against the description's full component set)
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,97 @@
...
+- Responses that were never received (status `0` in HAR captures) are ignored.
+
+## AI refinement
+
+The description is refined one operation at a time, so prompt and response stay small no matter how large the recorded traffic or the resulting description is.
+Each prompt carries a single operation from the baseline, the component schemas it references, the names of the other components (reserved against collisions), and a capped, shape-diverse sample of the real exchanges recorded for that operation (grouped by status and body shape, selected round-robin so every observed payload variant is represented).
+The AI is instructed to narrow types, add formats, enums, descriptions and examples, refine or add `components/schemas`, and model alternative payloads explicitly with `oneOf` (plus `discriminator`) and `allOf` composition.
+
+Up to `--ai-concurrency` operations (default 4) are refined in parallel, and every accepted refinement is merged back as it arrives, so operations prompted later see already-refined shared components.
+When two concurrent refinements touch the same shared component, the one merged last wins; the final whole-document lint still guards the result.
```
JLekawa:
```suggestion
When two concurrent refinements touch the same shared component, the one merged last wins.
  The final whole-document lint still guards the result.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## AI providers
+
+Every provider CLI is spawned in an empty temporary directory, so project context the CLIs normally auto-load from the working directory (`CLAUDE.md`, `AGENTS.md`, `.cursor/rules`) never enters the prompts — it would slow every call down and could steer the refinement.
+
+- **`claude`** — runs the local `claude` CLI in headless mode (`claude -p`), with built-in tools, MCP servers, settings (hooks, skills, plugins), and session persistence disabled so every call is a plain completion.
+  Because settings are not loaded, a model configured there does not apply: pass `--ai-model`, or the CLI's built-in default model is used.
+- **`codex`** — runs the local `codex` CLI in non-interactive mode (`codex exec`) with MCP servers and `AGENTS.md` discovery disabled and a read-only sandbox.
+  Other settings from `~/.codex/config.toml`, such as `model_reasoning_effort`, still apply.
+- **`cursor`** — runs the local Cursor CLI in print mode (`cursor-agent -p`; the renamed `agent` binary is tried as a fallback).
```
JLekawa:
```suggestion
- **`cursor`** — runs the local Cursor CLI in print mode: `cursor-agent -p`.
  The renamed `agent` binary is tried as a fallback).
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+All providers require the respective CLI to be installed and authenticated on the machine running the command.
+
+## Performance for large APIs
+
+Total time is roughly the per-operation AI time multiplied by the number of operations and divided by `--ai-concurrency`.
+For descriptions with many operations:
+
+- Raise `--ai-concurrency` (for example to `12`) — calls are network-bound and the provider CLIs retry rate-limit responses themselves.
+- Pick a faster model with `--ai-model`; smaller models roughly halve the per-operation time compared to the largest ones at some cost in refinement depth.
```
JLekawa:
```suggestion
- Pick a faster model with `--ai-model`.
  Smaller models roughly halve the per-operation time compared to the largest ones at some cost in refinement depth.
```

docs/@v2/commands/generate-spec.md
```diff
@@ -0,0 +1,119 @@
...
+| --server         | string  | Server URL the traffic was captured against (host, host + base path, or a path-only prefix like `/api`). Only requests under it are considered, the rest of their URL is treated as the API path, and it becomes the `servers` URL of the generated description. |
+| --title          | string  | Title for the generated API description. Default value is `Generated API`.                                                                                                                                                                                       |
+| --with-ai        | boolean | Refine the inferred description with an AI provider. Default value is `false`.                                                                                                                                                                                   |
+| --ai-provider    | string  | AI provider used with `--with-ai`; runs the corresponding CLI in non-interactive mode.<br/>**Possible values:** `claude`, `codex`, `cursor`. Default value is `claude`.                                                                                          |
```
JLekawa:
```suggestion
| --ai-provider    | string  | AI provider used with `--with-ai`. Runs the corresponding CLI in non-interactive mode.<br/>**Possible values:** `claude`, `codex`, `cursor`. Default value is `claude`.                                                                                          |
```

docs/@v2/commands/generate-spec.md
```diff
@@ -0,0 +1,119 @@
...
+
+A description inferred from traffic alone is a hypothesis: types are coarse and there are no descriptions, enums, or examples beyond what was observed.
+With `--with-ai`, the command sends each operation together with sample exchanges to an AI provider, which narrows types and adds formats, enums, descriptions, and examples.
+
+The AI's answer is never trusted blindly: a refined operation is only accepted when it keeps the operation's path, method, and observed response status codes, and passes validation with the `spec` ruleset.
+Rejected refinements keep their baseline version, and if refinement fails entirely the command falls back to the baseline description.
+The validation uses the built-in `spec` ruleset, not your project's `redocly.yaml` — lint the generated description with your own configuration afterward.
+
+{% admonition type="warning" name="Data sharing" %}
+`--with-ai` sends samples of the recorded traffic (URLs, query strings, request and response bodies) to the selected AI provider.
```
DmitryAnansky, line 38: What are those `samples` based on?

packages/cli/src/commands/generate-spec/generator.ts
```diff
@@ -0,0 +1,794 @@
...
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+const NUMERIC_RE = /^\d+$/;
+// Crockford base32 (ULID alphabet, no I/L/O/U), e.g. 01ARZ3NDEKTSV4RRFFQ69G5FAV.
+const ULID_RE = /^[0-9a-hjkmnp-tv-z]{26}$/i;
+// Type-prefixed opaque token with at least one digit, e.g. org_01ks7rnqsyy0g6h37rfmen7zv9.
+const PREFIXED_ID_RE = /^[a-z][a-z0-9]*_(?=[0-9a-z]*\d)[0-9a-z]{10,}$/i;
+// cuid ("cjld2cjxh0000qzrmn831i7rn") and cuid2 ("tz4a98xxat96iws9zmbrgj3a") as
+// generated by Prisma: 24-25 lowercase alphanumerics with at least one digit.
+const CUID_RE = /^[a-z](?=[a-z0-9]*\d)[a-z0-9]{23,24}$/;
+export const HTTP_METHODS = new Set([
```
DmitryAnansky: Can we use already defined constants `OPENAPI3_METHOD_NAMES` from core instead of redefining them?
Same for `drift` and `proxy` commands.
Its already redefined multiple times in 
- `packages/cli/src/commands/drift/openapi/loader.ts`
- `packages/cli/src/commands/proxy/server.ts`

docs/@v2/commands/generate-spec.md
```diff
@@ -0,0 +1,119 @@
...
+Refinement runs a locally installed AI CLI in non-interactive mode: `claude` (Claude Code), `codex` (Codex CLI), or `cursor` (Cursor CLI).
+The selected CLI must be installed and authenticated on the machine running the command.
+
+The provider runs in isolation: project context the CLIs normally load (such as `CLAUDE.md`, `AGENTS.md`, or `.cursor/rules`) and settings like a configured model do not apply.
+Use `--ai-model` to choose a model, or the provider's default is used.
+
+## Usage
+
+```bash
+redocly generate-spec <traffic>
```
DmitryAnansky, line 53: should we call it traffic-log or traffic-record to make it more clear?

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,104 @@
...
+| `--type`           | Target description type. Only `openapi` is supported.                                                 |
+| `--traffic-format` | Traffic input format (`auto`, `har`, `kong`, `nginx-json`, `apache-json`, `ndjson`).                  |
+| `--server`         | Server URL the traffic was captured against. Scopes which requests are considered and sets `servers`. |
+| `--title`          | Title for the generated description.                                                                  |
+| `--with-ai`        | Refine the inferred description with an AI provider.                                                  |
+| `--ai-provider`    | `claude`, `codex`, or `cursor` (default `claude`).                                                    |
+| `--ai-model`       | Model passed to the selected provider.                                                                |
+| `--ai-concurrency` | Number of operations refined in parallel (default `4`).                                               |
+| `-o, --output`     | Write the result to a file instead of stdout.                                                         |
```
DmitryAnansky, line 120: Should we also mentioned in README the limitation of producing only OAS v3.1.0 description?

packages/cli/src/commands/generate-spec/generator.ts
```diff
@@ -0,0 +1,794 @@
...
+  if (NUMERIC_RE.test(segment)) return true;
+  if (UUID_RE.test(segment)) return true;
+  if (ULID_RE.test(segment)) return true;
+  if (PREFIXED_ID_RE.test(segment)) return true;
+  if (CUID_RE.test(segment)) return true;
+  // long hex / opaque token
+  return /^[0-9a-f]{16,}$/i.test(segment);
+}
+
+export function singularize(value: string): string {
```
DmitryAnansky: We already have `pluralize` in our core. Could you please check if you can reuse it instead of implementing singularize ?

packages/cli/src/commands/generate-spec/generator.ts
```diff
@@ -0,0 +1,794 @@
...
+    if (index === -1) {
+      variants.push(candidate);
+    } else {
+      variants[index] = mergeVariantPair(variants[index], candidate);
+    }
+  }
+  return finalizeVariants(foldExcessObjectVariants(variants));
+}
+
+function sniffJson(bodyText: string | undefined): unknown {
```
DmitryAnansky, line 438: Can you just reuse isJson here?
We can  move it to core from respect if it feet your needs

```
export function isJSON(str: string): boolean {
  try {
    JSON.parse(str);
    return true;
  } catch {
    return false;
  }
}
```

packages/cli/src/commands/generate-spec/generator.ts
```diff
@@ -0,0 +1,794 @@
...
+        }
+        operation.responses[status] = response;
+      }
+    }
+
+    pathItem[accumulator.method] = operation;
+  }
+
+  return {
+    openapi: '3.1.0',
```
vadyvas: why not `3.2.0` ?

packages/cli/src/commands/generate-spec/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (argv.type !== 'openapi') {
+    return exitWithError(`Unsupported spec type "${argv.type}". Only "openapi" is supported.`);
+  }
+
+  const startedAt = Date.now();
+
+  const trafficPath = normalizeFsPath(argv.traffic);
+  const trafficFormat = argv['traffic-format'];
+
+  const baseline = applyValueInference(
```
vadyvas, line 53: should we lint generated spec without ai ?

packages/cli/src/commands/generate-spec/ai/refine.ts
```diff
@@ -0,0 +1,338 @@
...
+function applyFragment(
+  document: RefinedDocument,
+  path: string,
+  method: string,
+  fragment: OperationFragment
+): void {
+  document.paths[path][method] = fragment.operation;
+  if (Object.keys(fragment.components).length > 0) {
+    document.components ??= { schemas: {} };
+    Object.assign(document.components.schemas, fragment.components);
```
vadyvas, line 251: will it override already existed component ?

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
+import { stringifyYaml } from '@redocly/openapi-core';
+
+import type { TrafficSample } from '../samples.js';
+
+const SYSTEM_INSTRUCTIONS = `You are an expert API designer. You are given one operation from an OpenAPI 3.1 description that was inferred mechanically from recorded HTTP traffic, the component schemas it references, and a sample of the real recorded HTTP exchanges for that operation. The inferred description is only a hypothesis: types are coarse, every observed property is marked required, and descriptions are missing. Alternative body shapes observed in the traffic appear as "oneOf" variants or type unions. Common string formats (uuid, date-time, date, email, uri, ipv4) and small closed enums are detected conservatively from the observed values.
```
vadyvas: why only 3.1?

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
...
+- Relax "required" for properties that are plausibly optional (e.g. absent in some samples, or clearly nullable). Express nullable values as type unions such as ["string", "null"].
+- When a request or response body has alternative shapes — the baseline shows "oneOf" variants, or the samples show clearly different payloads — model them explicitly with "oneOf". Name each variant in components/schemas, and add a "discriminator" when a property (such as "type" or "kind") selects the variant.
+- Factor structure shared between variants or schemas with "allOf": extract the common base into components/schemas and compose each variant as "allOf" of the base plus its specific properties.
+- Keep the $ref references to the component schemas you were given. You may refine the definition of a referenced component when the samples justify it, and you may add new components, but never rename an existing component and never reuse a reserved component name listed in the input.
+- Add concise, useful "summary" and "description" fields to the operation, its parameters and important schema properties.
+- Add example values drawn from the real samples where helpful.
```
vadyvas: email, tokens etc ?

packages/cli/src/commands/generate-spec/ai/providers.ts
```diff
@@ -0,0 +1,213 @@
...
+  text: string;
+}
+
+// All provider CLIs pull project context (CLAUDE.md, AGENTS.md, .cursor/rules)
+// from the working directory into every prompt; running them from an empty
+// directory keeps that context out and each invocation fast.
+let emptyCwdPromise: Promise<string> | undefined;
+
+function getEmptyCwd(): Promise<string> {
+  emptyCwdPromise ??= mkdtemp(path.join(tmpdir(), 'redocly-generate-spec-'));
```
vadyvas, line 28: should we remove created dirs?

packages/cli/src/commands/generate-spec/ai/refine.ts
```diff
@@ -0,0 +1,338 @@
...
+        completed += 1;
+        inFlight.delete(label);
+        finishProgress(spinner);
+        logger.info(
+          `[${completed}/${operations.length}] ${label} — refined (${Math.round(
+            (Date.now() - startedAt) / 1000
+          )}s)\n`
+        );
+      } catch (error) {
+        completed += 1;
```
vadyvas, line 330: can we allow ai to fix the problems ?
generate -> validate -> fix -> validate -> accept ?

packages/cli/src/commands/generate-spec/samples.ts
```diff
@@ -0,0 +1,181 @@
...
+ * result is filled round-robin across groups, so every observed payload
+ * variant is represented before any group gets a second sample. Keying the
+ * result by operation keeps each AI prompt small no matter how large the
+ * recorded traffic is.
+ */
+export async function collectTrafficSamples(
+  options: CollectSamplesOptions
+): Promise<Map<string, TrafficSample[]>> {
+  const perGroup = options.perGroup ?? 2;
+  const perOperation = options.perOperation ?? 8;
```
vadyvas: why `8` by defaut?

packages/cli/src/commands/generate-spec/generator.ts
```diff
@@ -0,0 +1,794 @@
...
+
+/**
+ * Generate an OpenAPI 3.1 document by observing recorded traffic. Returns the
+ * document object; the caller decides whether to print it, write it, or hand it
+ * to an AI provider for refinement.
+ */
+export async function generateSpecFromTraffic(
+  options: GenerateSpecOptions
+): Promise<GeneratedDocument> {
+  const trafficFiles = await listFilesRecursively(options.trafficPath);
```
vadyvas, line 562: `listFilesRecursively` calls twice (samples.ts:106), can it be optimized?

packages/cli/src/commands/generate-spec/ai/refine.ts
```diff
@@ -0,0 +1,338 @@
...
+
+async function refineOperation(options: RefineOperationOptions): Promise<OperationFragment> {
+  const schemas = options.document.components?.schemas ?? {};
+  const referenced = transitivelyReferencedComponents(options.operation, schemas);
+  const components: Record<string, unknown> = {};
+  for (const name of referenced) {
+    components[name] = schemas[name];
+  }
+
+  const { system, user } = buildOperationPrompt({
```
vadyvas, line 201: for each operation - new call to AI ? Won't it be expensive?
just asking

packages/cli/src/commands/generate-spec/ai/providers.ts
```diff
@@ -0,0 +1,213 @@
...
+    '-c',
+    'mcp_servers={}',
+    '-c',
+    'project_doc_max_bytes=0',
+  ];
+  if (request.model) {
+    args.push('--model', request.model);
+  }
+  args.push('-');
+  const prompt = `${request.system}\n\n${request.user}`;
```
vadyvas, line 144: can it create potential prompt injection?

tests/e2e/generate-spec/generate-spec.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    expect(output).toContain('No HTTP exchanges in the traffic matched the server');
+  });
+
+  test('fails when the traffic path does not exist', () => {
+    const { code } = runGenerateSpec(['does-not-exist.ndjson']);
+    expect(code).toBe(1);
+  });
+});
+
+describe('generate-spec - AI refinement', () => {
```
vadyvas, line 124: can we test the prompt somehow? to avoid regression while editing

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
...
+
+Refine the operation into the most accurate OpenAPI 3.1 definition you can justify from the evidence, while staying conservative:
+- Keep the path template and the HTTP method exactly as given. Do not rename path parameters, do not move parameters to the path item level, and keep every response status code. The result is programmatically checked and rejected otherwise.
+- Keep the operationId unless it is clearly wrong.
+- Narrow property types where the samples clearly support it (formats such as date-time, uuid, email, uri; integer vs number; enums when a field only ever takes a small fixed set of values). Verify the formats and enums already detected against the samples; correct or extend them where justified.
```
vadyvas: > Verify the formats and enums already detected against the samples

will it remove what was found in the samples but was not passed to ai?

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
+import { stringifyYaml } from '@redocly/openapi-core';
+
+import type { TrafficSample } from '../samples.js';
+
+const SYSTEM_INSTRUCTIONS = `You are an expert API designer. You are given one operation from an OpenAPI 3.1 description that was inferred mechanically from recorded HTTP traffic, the component schemas it references, and a sample of the real recorded HTTP exchanges for that operation. The inferred description is only a hypothesis: types are coarse, every observed property is marked required, and descriptions are missing. Alternative body shapes observed in the traffic appear as "oneOf" variants or type unions. Common string formats (uuid, date-time, date, email, uri, ipv4) and small closed enums are detected conservatively from the observed values.
+
+Refine the operation into the most accurate OpenAPI 3.1 definition you can justify from the evidence, while staying conservative:
+- Keep the path template and the HTTP method exactly as given. Do not rename path parameters, do not move parameters to the path item level, and keep every response status code. The result is programmatically checked and rejected otherwise.
+- Keep the operationId unless it is clearly wrong.
```
vadyvas, line 12: > Keep the operationId unless it is clearly wrong

can it remove operationId ?

packages/cli/src/commands/generate-spec/samples.ts
```diff
@@ -0,0 +1,181 @@
+import { isPlainObject } from '@redocly/openapi-core';
+
+import { selectTrafficParser } from '../drift/log-formats/registry.js';
+import type { NormalizedHttpMessage, TrafficFormat } from '../drift/types/index.js';
+import { listFilesRecursively } from '../drift/utils/files.js';
+import { getPathWithoutTrailingSlash, normalizeContentType } from '../drift/utils/http.js';
+import { normalizeServerPrefix, resolvePathForServer } from '../drift/utils/server.js';
+import { HTTP_METHODS, templatizePath } from './generator.js';
+
+export interface TrafficSample {
```
vadyvas, line 10: can we add another parameters (e.g. headers) ?

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
...
+- Relax "required" for properties that are plausibly optional (e.g. absent in some samples, or clearly nullable). Express nullable values as type unions such as ["string", "null"].
+- When a request or response body has alternative shapes — the baseline shows "oneOf" variants, or the samples show clearly different payloads — model them explicitly with "oneOf". Name each variant in components/schemas, and add a "discriminator" when a property (such as "type" or "kind") selects the variant.
+- Factor structure shared between variants or schemas with "allOf": extract the common base into components/schemas and compose each variant as "allOf" of the base plus its specific properties.
+- Keep the $ref references to the component schemas you were given. You may refine the definition of a referenced component when the samples justify it, and you may add new components, but never rename an existing component and never reuse a reserved component name listed in the input.
```
vadyvas, line 17: can concurrency break this instruction?

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: Thanks. We need singularize, but in this case I've just added an export in core with singularize. Hope it's fine.

docs/@v2/commands/generate-spec.md
sobanieca-redocly: The description already says "traffic log file or folder", I would keep it short.

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+```bash
+# Deterministic inference only
+redocly generate-spec ./traffic.har
+
+# Refine the hypothesis with AI
+redocly generate-spec ./traffic.har --with-ai --ai-provider claude -o openapi.yaml
+```
+
+The `<traffic>` argument is a file or folder of recorded traffic.
+Supported formats: HAR, Kong, Nginx/Apache JSON, and NDJSON (auto-detected, or forced with `--traffic-format`).
```
JLekawa:
```suggestion
The supported formats are: 

- HAR
- Kong
- Nginx/Apache JSON
- NDJSON

Formats are auto-detected, or can be forced with `--traffic-format`.

```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+The command generates OpenAPI 3.1 descriptions only.
+
+## Deterministic inference
+
+The baseline is derived from every exchange in the traffic:
+
+- Identifier-like path segments (numeric, UUID, ULID, CUID, prefixed and opaque tokens) become named path parameters.
+- Body schemas are merged across all observations.
+  A property becomes optional as soon as one sample omits it.
+- Alternative body shapes for the same operation are preserved as `oneOf` variants instead of being collapsed, and values observed as `null` produce type unions such as `["string", "null"]`.
```
JLekawa:
```suggestion
- Alternative body shapes for the same operation are preserved as `oneOf` variants instead of being collapsed.
- Values observed as `null` produce type unions such as `["string", "null"]`.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Object shapes that repeat across the document are extracted into `components/schemas` and referenced with `$ref`.
+  The same entity is recognized when it appears as a list item and as a single resource, with different `required` sets, or with near-identical properties (at least 75% shared, with compatible types).
+  Components are named from the path entity, the enclosing property name, or `Error` for error responses.
+  A shape repeated only because its parent shape repeats stays inline.
```
JLekawa:
```suggestion
  - The same entity is recognized when it appears as a list item and as a single resource, with different `required` sets, or with near-identical properties (at least 75% shared, with compatible types).
  - Components are named from the path entity, the enclosing property name, or `Error` for error responses.
  - A shape repeated only because its parent shape repeats stays inline.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Observed string values are analyzed conservatively: when every value of a property (or parameter) matches the same well-known pattern it gets a `format` (`uuid`, `date-time`, `date`, `email`, `uri`, `ipv4`).
+  When a string only ever takes a small set of identifier-like values with enough repetition (at least 4 observations, at most 5 distinct values, each seen twice on average) it becomes an `enum`.
+  Enums apply to body properties and query parameters.
+  Path parameters and nullable unions get formats only.
+  Evidence is pooled across all operations a shared component was observed in.
```
JLekawa:
```suggestion
  - When a string only ever takes a small set of identifier-like values with enough repetition it becomes an `enum`.
  To trigger this conversion, the command needs at least 4 observations, at most 5 distinct values, each seen twice on average.
  - Enums apply to body properties and query parameters.
  - Path parameters and nullable unions get formats only.
  - Evidence is pooled across all operations a shared component was observed in.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+- Responses that were never received (status `0` in HAR captures) are ignored.
+
+## AI refinement
+
+The description is refined one operation at a time, so prompt and response stay small no matter how large the recorded traffic or the resulting description is.
```
JLekawa:
```suggestion
The description is refined one operation at a time.
The prompt and response stay small regardless of the volume of recorded traffic or the resulting description's size.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+Each prompt carries a single operation from the baseline, the component schemas it references, the names of the other components (reserved against collisions), and a capped, shape-diverse sample of the real exchanges recorded for that operation (grouped by status and body shape, selected round-robin so every observed payload variant is represented).
+The AI is instructed to narrow types, add formats, enums, descriptions and examples, refine or add `components/schemas`, and model alternative payloads explicitly with `oneOf` (plus `discriminator`) and `allOf` composition.
```
JLekawa:
```suggestion
Each prompt carries:
- a single operation from the baseline
- referenced component schemas
- names of the other components (reserved against collisions)
- a capped, shape-diverse sample of real exchanges recorded for that operation 

The sample of exchanges is grouped by status and body shape, selected round-robin to represent every observed payload variant.

The AI:
- narrows types
- adds formats, enums, descriptions and examples
- refines or adds `components/schemas`
- models alternative payloads explicitly with `oneOf` (plus `discriminator`) and `allOf` composition
```

packages/cli/src/commands/generate-spec/generator.ts
DmitryAnansky: Please check how it work, pluralize can produce both results.

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Up to `--ai-concurrency` operations (default 4) are refined in parallel, and every accepted refinement is merged back as it arrives, so operations prompted later see already-refined shared components.
```
JLekawa:
```suggestion
The number of operations refined in parallel is set in `--ai-concurrency`, 4 by default.
Every accepted refinement is merged back as it arrives.
Operations prompted later see already-refined shared components.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+When two concurrent refinements touch the same shared component, the one merged last wins.
+  The final whole-document lint still guards the result.
```
JLekawa:
```suggestion

When two concurrent refinements touch the same shared component, the one merged last wins.
The final whole-document lint still guards the result.

```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,105 @@
...
+Set `--ai-concurrency 1` to process operations strictly sequentially.
```
JLekawa:
```suggestion
To process operations strictly sequentially, set `--ai-concurrency 1`.
```

packages/cli/src/commands/generate-spec/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      logger.info(
+        `AI refinement complete: ${refined.refined} of ${refined.total} operation(s) refined (${provider}).\n`
+      );
+    } catch (error) {
+      logger.warn(
+        `AI refinement failed, falling back to the baseline description: ${
+          error instanceof Error ? error.message : String(error)
+        }\n`
+      );
+    }
```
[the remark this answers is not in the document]
sobanieca-redocly: Intentional. If the final whole-document lint fails, the merged document is invalid, and emitting an invalid description would be worse than falling back to the deterministic baseline — the only output we can guarantee. Each fragment already passed lint against the full component set at merge time, so a final-lint failure can only come from concurrent merge interactions; the warning tells the user why the fallback happened.

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: Done — `HTTP_METHODS` now reuses `OPENAPI3_METHOD_NAMES` (it lives in the split command, `packages/cli/src/commands/split/oas/constants.ts`, rather than core). It became a clean fit once the command switched to generating OpenAPI 3.2, where `query` is a valid operation — before that we had to exclude it. The drift and proxy sets differ slightly (proxy needs uppercase names for Node's HTTP handling, drift predates 3.2), so I'd consolidate those in a separate cleanup PR rather than grow this one.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Added — the intro now states "The command generates OpenAPI 3.2 descriptions only." (The command moved from 3.1 to 3.2 output based on the other review threads.)

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: You're right — `pluralize(word, 1)` returns the singular form. Reverted the core `singularize` export (and its changeset) and reused the existing `pluralize` from core.

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: `sniffJson` needs the parsed value, not a boolean — the schema is inferred from the result. Reusing `isJSON` would parse every body twice, and the `{`/`[` prefix check skips `JSON.parse` entirely for plain-text bodies. So I'd keep it as is.

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: Good push — switched to 3.2.0. Core already fully supports 3.2 (the acceptance lint runs the `oas3_2` spec rules), and it let us accept the QUERY HTTP method as a real operation — `HTTP_METHODS` now reuses `OPENAPI3_METHOD_NAMES`, which includes `query`. Prompt, docs, and snapshots updated to match.

packages/cli/src/commands/generate-spec/index.ts
sobanieca-redocly: The baseline generator is deterministic and its exact output is pinned by unit and e2e snapshots, so linting on every run would re-check what tests already guarantee — and if it warned, we'd emit the document anyway. The AI path lints because model output is untrusted. The README recommends linting the generated file with your own ruleset afterward. Can add a baseline lint pass if you feel strongly.

packages/cli/src/commands/generate-spec/ai/refine.ts
sobanieca-redocly: Yes, intentionally — the prompt allows the AI to refine the definition of a referenced component when the samples justify it, so the refined definition replaces the baseline one on merge. Renaming or reusing reserved names is forbidden (now also enforced in code), each fragment is linted against the full component set before merging, unused components are pruned, and the final document is linted again. With concurrency, last-merged wins (documented in the `refineSpecWithAi` doc comment and README).

packages/cli/src/commands/generate-spec/ai/prompt.ts
sobanieca-redocly: The prompt targets what the command emits, and both now say OpenAPI 3.2 (see the version thread on the generator).

packages/cli/src/commands/generate-spec/ai/prompt.ts
sobanieca-redocly: Good point — extended the instruction: examples must never copy values that look like secrets or personal data (tokens, API keys, passwords, session identifiers, email addresses, personal names); the AI is told to use clearly fictional placeholders instead. Note the traffic samples themselves still reach the provider — that's the documented `--with-ai` warning.

packages/cli/src/commands/generate-spec/ai/providers.ts
sobanieca-redocly: It's a single empty directory per run under the OS temp dir, shared by all provider calls. Deleting it while concurrent provider processes still use it as their cwd would be racy, and an exit hook felt like more machinery than the leak justifies — OS temp cleanup reclaims it. Happy to add cleanup if you think it's worth it.

packages/cli/src/commands/generate-spec/ai/refine.ts
sobanieca-redocly: Considered it, but kept single-shot per operation: a repair loop multiplies cost and latency on exactly the operations where the model is already struggling, and a rejected refinement isn't lost work — the operation keeps its valid deterministic baseline. If real usage shows frequent rejections, a retry-with-lint-feedback loop is a natural follow-up.

packages/cli/src/commands/generate-spec/samples.ts
sobanieca-redocly: It was a budget balancing evidence against prompt size — samples are grouped by status + body shape and picked round-robin, so every observed payload variant is represented before any group gets a second sample. Lowered the default to 4 to be conservative; it's an internal option, so we can tune it or expose a flag later if refinement quality asks for more.

packages/cli/src/commands/generate-spec/generator.ts
sobanieca-redocly: It could be shared, but I kept the two passes independent on purpose: sample collection only runs with `--with-ai`, and both the directory listing and the streaming re-parse cost milliseconds against the minutes of AI calls that follow. Threading the file list (or buffering all exchanges) through would couple `generateSpecFromTraffic` and `collectTrafficSamples` for no measurable gain.

packages/cli/src/commands/generate-spec/ai/refine.ts
sobanieca-redocly: It's a deliberate trade-off. One call per operation keeps each prompt small and independently verifiable (per-operation acceptance and fallback), and it scales to traffic of any size, where a whole-document prompt would blow the context window and one bad answer would reject everything. Cost scales with operation count, not traffic size — samples per operation are capped (now at 4) and bodies truncated. Subscription-based provider CLIs keep the marginal call cost low, and `--ai-concurrency` addresses the latency.

packages/cli/src/commands/generate-spec/ai/providers.ts
sobanieca-redocly: The concatenation itself doesn't add risk — traffic bodies are attacker-influenced text inside the prompt for every provider, so injection is possible regardless of the system/user split (`codex exec` just has no separate system channel). The design assumes the response may be adversarial: providers run with tools disabled or a read-only sandbox in an empty temp dir, and a response is only merged if it keeps the path/method/statuses/operationId, parses, and passes the spec lint. Worst case is a corrupted refinement for that one operation, which the checks reject back to the baseline.

tests/e2e/generate-spec/generate-spec.test.ts
sobanieca-redocly: Added `__tests__/prompt.test.ts` — it snapshots the full system instructions and a built user prompt, so any prompt edit produces a reviewable snapshot diff (it already caught the 3.1 → 3.2 wording change).

packages/cli/src/commands/generate-spec/ai/prompt.ts
sobanieca-redocly: Good catch — it could have. The prompt now states that the detected formats and enums were derived from the full recorded traffic while the AI sees only a subset, and forbids removing a detected format or enum merely because the shown samples don't cover it — only when a sample contradicts it. That matters even more now that the per-operation sample cap is 4.

packages/cli/src/commands/generate-spec/ai/prompt.ts
sobanieca-redocly: It could before — now it's enforced in code: a refinement that drops the `operationId` is rejected and the operation keeps its baseline. Added a unit test and an acceptance bullet in the README.

packages/cli/src/commands/generate-spec/samples.ts
sobanieca-redocly: Left out deliberately: headers are where credentials concentrate (`Authorization`, `Cookie`, API keys), so forwarding them to the provider multiplies the leak risk the `--with-ai` warning covers — and the baseline generator doesn't infer header parameters, so the AI couldn't merge them back into anything it's allowed to change. If we later infer header parameters in the baseline, adding redacted headers to the samples would make sense.

packages/cli/src/commands/generate-spec/ai/prompt.ts
sobanieca-redocly: In a narrow window, yes: reserved names are computed when a prompt is dispatched, so two in-flight refinements can each introduce the same new component name. Merges themselves are serialized on the event loop and the last one wins — the same documented last-write-wins rule as for shared components — and the final whole-document lint still guards the merged result. Redefining an already-reserved name is now rejected outright, and `--ai-concurrency 1` removes the race entirely.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied, thanks.

packages/cli/src/commands/generate-spec/ai/prompt.ts
```diff
@@ -0,0 +1,112 @@
...
+      }
+      if (sample.responseContentType) {
+        lines.push(`Response Content-Type: ${sample.responseContentType}`);
+      }
+      if (sample.responseBody) {
+        lines.push(`Response body: ${sample.responseBody}`);
+      }
+      return lines.join('\n');
+    })
+    .join('\n\n');
```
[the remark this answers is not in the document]
sobanieca-redocly: True in principle, but delimiters or escaping can't prevent injection — the model reads the text either way. The defense is downstream: providers run with tools disabled or in a read-only sandbox from an empty temp directory, and a response is only merged if it keeps the path/method/statuses/operationId, parses as the expected fragment, does not redefine reserved components, and passes the spec lint. An injected instruction can at most corrupt the refinement of its own operation, which those checks reject back to the deterministic baseline. The README also warns that `--with-ai` sends traffic samples to the provider.

packages/cli/src/commands/generate-spec/ai/refine.ts
```diff
@@ -0,0 +1,349 @@
...
+  document: RefinedDocument,
+  path: string,
+  method: string,
+  fragment: OperationFragment
+): void {
+  document.paths[path][method] = fragment.operation;
+  if (Object.keys(fragment.components).length > 0) {
+    document.components ??= { schemas: {} };
+    Object.assign(document.components.schemas, fragment.components);
+  }
```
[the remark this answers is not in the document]
sobanieca-redocly: Good catch — enforced now: a fragment that redefines any reserved component name (the ones the prompt lists as off-limits) is rejected and the operation keeps its baseline. Covered by a unit test and added to the README acceptance list.

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+- keeps every response status code observed in the traffic
+- keeps an `operationId`
+- does not redefine reserved components (those used only by other operations)
+- passes validation with the `spec` ruleset (checked against the description's full component set)
+
+An operation whose refinement is rejected keeps its deterministic baseline (reported with the reason), components no operation references anymore are pruned, and the final document is linted again as a whole.
+If refinement fails for every operation, the command falls back to the deterministic baseline.
+Each provider call times out after 5 minutes.
+
+The acceptance lint is intentionally pinned to the `spec` ruleset and does not use the project's `redocly.yaml`: a stricter governance config would reject refinements for problems the baseline itself has, and a looser one could let structurally broken output through.
```
JLekawa:
```suggestion
The acceptance lint is intentionally pinned to the `spec` ruleset and does not use the project's `redocly.yaml`.
A stricter governance config would reject refinements for problems the baseline itself has.
A looser one could let structurally broken output through.
```

packages/cli/src/commands/generate-spec/components.ts
```diff
@@ -0,0 +1,486 @@
...
+    result.splice(bestB, 1);
+  }
+  return result;
+}
+
+function compareStrings(a: string, b: string): number {
+  return a < b ? -1 : a > b ? 1 : 0;
+}
+
+function pascalCase(value: string): string {
```
DmitryAnansky, line 338: Can we reuse existing `toPascalCase` ?
packages/core/src/utils/to-pascal-case.ts

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+- **`codex`** — runs the local `codex` CLI in non-interactive mode (`codex exec`) with MCP servers and `AGENTS.md` discovery disabled and a read-only sandbox.
+  Other settings from `~/.codex/config.toml`, such as `model_reasoning_effort`, still apply.
+- **`cursor`** — runs the local Cursor CLI in print mode: `cursor-agent -p`.
+  The renamed `agent` binary is tried as a fallback).
+
+All providers require the respective CLI to be installed and authenticated on the machine running the command.
+
+## Performance for large APIs
+
+Total time is roughly the per-operation AI time multiplied by the number of operations and divided by `--ai-concurrency`.
```
JLekawa:
```suggestion
The total time is roughly the per-operation AI time multiplied by the number of operations and divided by `--ai-concurrency`.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `--title`          | Title for the generated description.                                                                  |
+| `--with-ai`        | Refine the inferred description with an AI provider.                                                  |
+| `--ai-provider`    | `claude`, `codex`, or `cursor` (default `claude`).                                                    |
+| `--ai-model`       | Model passed to the selected provider.                                                                |
+| `--ai-concurrency` | Number of operations refined in parallel (default `4`).                                               |
+| `-o, --output`     | Write the result to a file instead of stdout.                                                         |
+
+## AI providers
+
+Every provider CLI is spawned in an empty temporary directory, so project context the CLIs normally auto-load from the working directory (`CLAUDE.md`, `AGENTS.md`, `.cursor/rules`) never enters the prompts — it would slow every call down and could steer the refinement.
```
JLekawa:
```suggestion
Every provider CLI is spawned in an empty temporary directory.
The project context the CLIs normally auto-load from the working directory (`CLAUDE.md`, `AGENTS.md`, `.cursor/rules`) never enters the prompts.
Otherwise, it would slow every call down and could steer the refinement.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+| ------------------ | ----------------------------------------------------------------------------------------------------- |
+| `--type`           | Target description type. Only `openapi` is supported.                                                 |
+| `--traffic-format` | Traffic input format (`auto`, `har`, `kong`, `nginx-json`, `apache-json`, `ndjson`).                  |
+| `--server`         | Server URL the traffic was captured against. Scopes which requests are considered and sets `servers`. |
+| `--title`          | Title for the generated description.                                                                  |
+| `--with-ai`        | Refine the inferred description with an AI provider.                                                  |
+| `--ai-provider`    | `claude`, `codex`, or `cursor` (default `claude`).                                                    |
+| `--ai-model`       | Model passed to the selected provider.                                                                |
+| `--ai-concurrency` | Number of operations refined in parallel (default `4`).                                               |
```
JLekawa:
```suggestion
| `--ai-concurrency` | Number of operations refined in parallel. Default: `4`.                                               |
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+An operation whose refinement is rejected keeps its deterministic baseline (reported with the reason), components no operation references anymore are pruned, and the final document is linted again as a whole.
+If refinement fails for every operation, the command falls back to the deterministic baseline.
+Each provider call times out after 5 minutes.
+
+The acceptance lint is intentionally pinned to the `spec` ruleset and does not use the project's `redocly.yaml`: a stricter governance config would reject refinements for problems the baseline itself has, and a looser one could let structurally broken output through.
+Lint the generated description with your own config afterward.
+
+> **Warning:** `--with-ai` sends samples of the recorded traffic (URLs, query strings, request and response bodies) to the selected AI provider.
+> Make sure the traffic contains no secrets or personal data you are not allowed to share.
```
JLekawa, line 104: Should this be wrapped in a Markdoc admonition tag?

packages/cli/src/commands/generate-spec/ai/refine.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  // Lint the refined operation against the description's full component set,
+  // so $refs to unchanged shared components still resolve.
+  const fragmentDocument = {
+    openapi: options.document.openapi,
+    info: options.document.info,
+    paths: { [options.path]: { [options.method]: fragment.operation } },
+    components: { schemas: { ...schemas, ...fragment.components } },
+  };
+  await lintDocumentSource(stringifyYaml(fragmentDocument), options.config);
```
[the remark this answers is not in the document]
sobanieca-redocly: Mostly a restatement of the documented concurrency trade-off, with one correction: the per-operation lint isn't stale — `schemas` is a live reference and the fragment document is built after the provider call, so lint sees all merges up to that point. The reserved-name list is intentionally fixed at prompt time because it must match what the model was told it may not touch. What remains is the documented last-merge-wins rule for shared components: merges are serialized on the event loop, the final whole-document lint guards the merged result, and `--ai-concurrency 1` gives strictly sequential behavior. Making acceptance transactional would add locking/re-validation for an edge the final lint already covers.

packages/cli/src/commands/generate-spec/components.ts
sobanieca-redocly: Checked it — core's `toPascalCase` splits on whitespace only and isn't exported from the package. The hints here are path segments and property names, so the separators are hyphens and underscores (`payment-methods`, `created_at`), which it would keep as-is (`Payment-methods`). The local helper splits on any non-alphanumeric run, which is what component naming needs. Reusing the core util would mean exporting it and pre-replacing separators with spaces — more churn than the six local lines. Happy to go that way instead if you prefer.

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+  When a string only ever takes a small set of identifier-like values with enough repetition (at least 4 observations, at most 5 distinct values, each seen twice on average) it becomes an `enum`.
+  Enums apply to body properties and query parameters.
+  Path parameters and nullable unions get formats only.
+  Evidence is pooled across all operations a shared component was observed in.
+- Responses that were never received (status `0` in HAR captures) are ignored.
+
+## AI refinement
+
+The description is refined one operation at a time, so prompt and response stay small no matter how large the recorded traffic or the resulting description is.
+Each prompt carries a single operation from the baseline, the component schemas it references, the names of the other components (reserved against collisions), and a capped, shape-diverse sample of the real exchanges recorded for that operation (grouped by status and body shape, selected round-robin so every observed payload variant is represented).
```
JLekawa:
```suggestion
Each prompt carries:
- a single operation from the baseline
- component schemas it references
- names of other components (reserved against collisions)
- a capped, shape-diverse sample of the real exchanges recorded for that operation  

The sample of exchanges is grouped by status and body shape, selected round-robin to represent every observed payload variant.

```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+The AI is instructed to narrow types, add formats, enums, descriptions and examples, refine or add `components/schemas`, and model alternative payloads explicitly with `oneOf` (plus `discriminator`) and `allOf` composition.
```
JLekawa:
```suggestion
The AI is instructed to:
- narrow types
- add formats, enums, descriptions and examples
- refine or add `components/schemas`
- model alternative payloads explicitly with `oneOf` (plus `discriminator`) and `allOf` composition.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+Up to `--ai-concurrency` operations (default 4) are refined in parallel, and every accepted refinement is merged back as it arrives, so operations prompted later see already-refined shared components.
```
JLekawa:
```suggestion
You can set the number of operations refined in parallel using `--ai-concurrency`.
The default value is `4`.
Every accepted refinement is merged back as it arrives: operations prompted later see already-refined shared components.
```

packages/cli/src/commands/generate-spec/README.md
```diff
@@ -0,0 +1,114 @@
...
+The AI's answer is never trusted blindly.
+A refined operation is only accepted when it:
+
+- keeps the exact path template and method — the AI cannot invent, drop, or rename operations because only the requested operation is merged back
+- keeps every response status code observed in the traffic
+- keeps an `operationId`
+- does not redefine reserved components (those used only by other operations)
+- passes validation with the `spec` ruleset (checked against the description's full component set)
+
+An operation whose refinement is rejected keeps its deterministic baseline (reported with the reason), components no operation references anymore are pruned, and the final document is linted again as a whole.
```
JLekawa:
```suggestion
An operation whose refinement is rejected keeps its deterministic baseline and the reason for rejection is reported).
Components that no operation references are pruned.
The final document is linted again as a whole.

```

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied via your later split suggestions for these paragraphs (the "Each prompt carries:" and "The AI is instructed to:" lists) in dcb7ff62.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied via your later version of this paragraph in dcb7ff62.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: This README ships with the package on GitHub/npm rather than the docs site, so a Markdoc admonition tag would render as literal text there. Keeping the plain blockquote for now — when the command graduates from experimental and gets a docs/@v2 page, that page should use the admonition tag.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62, thanks.

packages/cli/src/commands/generate-spec/README.md
sobanieca-redocly: Applied in dcb7ff62 (dropped the stray closing parenthesis after "reported"), thanks.

JLekawa (approved):

DmitryAnansky (approved):

packages/cli/src/commands/generate-spec/ai/providers.ts
vadyvas: i think we can leave it as is

packages/cli/src/commands/generate-spec/ai/refine.ts
vadyvas: ok 👌

packages/cli/src/commands/generate-spec/samples.ts
vadyvas: 8 was fine, I was just curious how you arrived at that number.. based on status codes?

packages/cli/src/commands/generate-spec/ai/providers.ts
vadyvas: sounds good 👍

tests/e2e/generate-spec/generate-spec.test.ts
vadyvas: i was thinking more about a prompt quality check - eval harness or some kind of .md file with test cases

eval harness seems like overkill for a first iteration, but at least the test cases you leaned on when testing the agents' responses after the modification would be good to have

vadyvas (approved):
left a few comments, but overall it looks good, thanks 🔥

packages/cli/src/commands/generate-spec/samples.ts
sobanieca-redocly: Not status codes — honestly, 8 was an arbitrary pick: it came from my machine's 8 cores more than from any property of the traffic. Since there was no principled derivation behind it, let's keep the conservative 4 for now: the round-robin grouping still represents every observed status + body shape before any group gets a second sample, and it stays an internal option we can raise or expose as a flag if refinement quality shows the model needs more evidence.

tests/e2e/generate-spec/generate-spec.test.ts
sobanieca-redocly: Good idea — added `ai/PROMPT_TESTING.md` next to the prompt: a manual quality checklist with copy-paste runs against the existing e2e traffic fixtures and concrete per-run checks (detected formats and enums kept, `oneOf` + `discriminator` modeled, secrets never copied into examples), plus guidance on judging nondeterministic output (rerun once, then compare against `main`). `prompt.ts` and the command README now point to it. Agreed the eval harness is overkill for now — if prompt edits become frequent, it's the natural next step.

packages/cli/src/commands/generate-spec/ai/providers.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+  // The instructions go in as the prompt argument; the piped stdin is
+  // attached as context, mirroring how the claude provider splits the two.
+  args.push(request.system);
+
+  // The Cursor CLI originally installed a "cursor-agent" binary and later
+  // renamed it to "agent"; accept either.
+  let result;
+  try {
+    result = await runCommand('cursor-agent', args, request.user);
```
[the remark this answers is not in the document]
sobanieca-redocly: I will apply fixes for cursor CLI in next PR.

PR #2940, opened 2026-07-07
feat(cli): add an experimental diff command

vadyvas:
## What/Why/How?

Adds an **experimental** `redocly diff <base> <revision>` command that compares two API descriptions and reports what was **added, removed, and changed**.

- **Structural diff for every supported specification** (OpenAPI 2/3.0/3.1/3.2, AsyncAPI 2/3, Arazzo, Overlay, OpenRPC) by reusing the `openapi-core` type trees — no hand-written traversal per type.
- **Breaking-change verdicts for OpenAPI 3.x and AsyncAPI 3**, from a lint-style rule registry keyed by node type. A rule is a small `{ id, description, visit(change, ctx) }` object; every rule runs, every verdict is kept, and the worst one decides the change. The scale is binary — `breaking` / `non-breaking` — because a verdict either breaks a consumer or it does not. What cannot be judged (a `$ref` that now points elsewhere) is reported as breaking.
- **Pipeline:** bundle each side → walk it with `walkDocument` into a flat map keyed by a *stable* pointer → compare the two maps in two passes → classify. Array indexes become identity keys (a parameter by `in` + `name`, a server by `url`), so reordering is not a change. A `$ref` is recorded as an attribute and never followed, so a shared component is compared once, at its own path.
- **Direction decides the verdict.** Narrowing what a request accepts breaks the clients that send the old value; widening what a response returns breaks the clients that read it. The direction comes from the node types along the ancestor chain, never from pointer text. A component takes it from the sites that reference it, `callbacks` and `webhooks` invert it, and AsyncAPI 3 declares it outright with `action: send|receive` — which is why an AsyncAPI payload is judged by the same schema rules as an OpenAPI one.
- **Output formats:** `stylish`, `json`, `markdown`, `html`, plus the `lint` formats (`codeframe`, `checkstyle`, `codeclimate`, `summary`, `github-actions`, `junit`) for the breaking changes, so existing CI integrations accept the report as it is.
- **CI gate:** `--fail-on breaking|none` (default `breaking`) sets the exit code.

```bash
redocly diff v1/openapi.yaml v2/openapi.yaml
redocly diff https://example.com/openapi.yaml ./openapi.yaml --format=json
redocly diff main-openapi.yaml pr-openapi.yaml --format=github-actions
redocly diff v1.yaml v2.yaml --format=html -o diff-report.html
```

**Nothing is added to `openapi-core`.** The diff of this branch against `main` touches `packages/cli`, `tests/e2e/diff`, `docs`, and the changeset only. The command reuses what core already exports — `bundle`, `walkDocument`, `normalizeVisitors`, `getTypes`, `detectSpec`, `getLineColLocation`, `formatProblems`, `dequal` — and keeps everything of its own under `packages/cli/src/commands/diff`.

Comparison by URL or path is supported; comparison of two Git revisions is not wired up yet.

## Reference

Documented in `docs/@v2/commands/diff.md`: the options, how the direction is decided, the rule catalog for both specifications, and the limitations (component renames, positional matching inside `allOf`/`oneOf`/`anyOf`, and the AsyncAPI features that have no rules yet).

The rule catalog in the docs is checked against the registries by a unit test, so a rule cannot ship without a row, and a row cannot outlive its rule.

## Testing

**Unit** — `packages/cli/src/commands/diff/__tests__`. The engine invariants that the terminal output cannot show: stable pointers and identity keys, subtree collapse in the comparison, path-rename matching and its ambiguous fallback, the direction for both specification families (including a payload that references itself), verdict aggregation, and the rendered reports as snapshots.

**e2e** — `tests/e2e/diff`. One fixture per rule, named `oas3-*` or `async3-*`, each a `base.yaml`/`revision.yaml` pair differing only in what it exercises. Each test reads the verdict from the machine-readable report and snapshots the terminal one. A rule is registered per node type, so only a real run catches a rule wired to a type the walker never reports — two such gaps in the parameter rules were found this way. Mirror fixtures cover the edits that must stay non-breaking, so a rule that stops reading the direction fails there instead of passing everywhere.

## Screenshots (optional)

image

## Check yourself

## Security


vadyvas:
@RomanHotsiy 
short overview of how the `diff` command works and what it looks like. interested in your feedback on the design

```mermaid
flowchart LR
    A["base + revision"] --> C["collect ×2<br/>(walkDocument → flat maps)"] --> D["compare<br/>(set diff)"] --> E["classify<br/>(polarity + rules)"] --> F["report<br/>(stylish/json/md/html)"]
    C -. "$ref edges" .-> U[UsageIndex] -. polarity .-> E
```

Inputs for now are files or urls - resolved through the existing `BaseResolver`; comparing against git revisions isn't supported

1. **Collect** — each side (base and revision) is bundled and walked with the existing `walkDocument` + type trees into a flat `Map<stablePointer, NodeEntry>`. Pointers are *stable*: list items are keyed by identity (e.g. params by `in`+`name`), not array index, so reordering isn't a change; `$ref` are kept as attributes, so a shared schema is diffed once at its component path.
2. **Compare** — dumb two-pass set-diff over the union of keys: only in base → `removed`, only in revision → `added`, in both → shallow property diff. A removed/added subtree collapses into one change.
3. **Classify** — verdicts are binary: `breaking` / `non-breaking`. The key concept is **polarity**: the same change flips meaning by direction (a property becoming `required` breaks *requests* but is safe in *responses*). Polarity is derived from the pointer; for shared components it comes from the `UsageIndex`, where the component is referenced (request/response), transitively.
Rules are tiny lint-style `{id, description, visit()}` objects in a registry (granular, since the rule set will likely grow)
4. **Report** — `stylish`, `json`, `markdown`, `html`.

```bash
redocly diff base.yaml revision.yaml 
redocly diff base.yaml revision.yaml --fail-on=breaking --format=json
```

```
GET /pets
  ✖ breaking      changed  parameters/{query:limit} · required
      Parameter became required. (parameter-became-required)

3 breaking, 1 non-breaking.
```

vadyvas:
@cursor review

PR #2957, opened 2026-07-22
fix: validate discriminated oneOf schemas against only the branch selected by the discriminator in the drift command

sobanieca-redocly:
## What/Why/How?

Fix for handling `oneOf` schemas with a `discriminator` inside `drift` command. Extended list of well known headers from 3rd party providers.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes drift JSON schema validation semantics for discriminated oneOf (fewer false positives, different errors for invalid discriminators); header ignore changes only affect undocumented-header findings.
> 
> **Overview**
> Fixes **false positives** in the `drift` command’s **schema-consistency** rule for **`oneOf` + `discriminator`** payloads, and cuts noise from common **AWS / GitHub** headers in **undocumented-header** checks.
> 
> **Schema validation:** `SchemaValidator` now compiles with Ajv’s **`discriminator: true`**, so traffic is checked against the branch implied by the discriminator value instead of every `oneOf` variant. If Ajv cannot compile the discriminator (loose tag property), compilation **falls back** to a second Ajv instance with discriminators disabled—same behavior as before.
> 
> **Headers:** Built-in ignores add **`x-amzn-*`** and **`x-github-*`** prefixes, plus **`x-hub-signature`** / **`x-hub-signature-256`**; **`x-amzn-trace-id`** moves from the exact-name list into the prefix rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b58b54c5c2063c02f7daa79391fdf80fe1c157fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sobanieca-redocly:
@cursor review

.changeset/drift-infra-header-ignores.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+Extended the `drift` command's built-in undocumented-header ignore list with well-known infrastructure headers: the `x-amzn-` (AWS load balancer and API Gateway) and `x-github-` (GitHub webhook delivery metadata) prefixes, and the `x-hub-signature` / `x-hub-signature-256` webhook signature headers.
```
JLekawa:
```suggestion
Extended the `drift` command's built-in undocumented-header ignore list with `x-amzn-` and `x-github-` prefixes, and the `x-hub-signature` / `x-hub-signature-256` webhook signature headers.
```

.changeset/drift-discriminator-validation.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/cli': patch
+---
+
+Fixed an issue where the `drift` command's `schema-consistency` rule reported false-positive findings for `oneOf` schemas with a `discriminator`.
+Payloads are now validated only against the branch selected by the discriminator value instead of every `oneOf` branch, so a mismatch no longer produces noise from the non-matching branches.
```
JLekawa:
```suggestion
Payloads are validated only against the branch selected by the discriminator value instead of every `oneOf` branch.
```

JLekawa (approved):

packages/cli/src/commands/drift/engine/schema-validator.ts
```diff
@@ -141,9 +143,16 @@ export class SchemaValidator {
...
         enum: true,
       },
-    });
+    };
+
+    this.ajv = new AjvConstructor(ajvOptions);
+    // Ajv rejects loosely defined discriminators at compile time (for example
+    // when the tag property has no `const`/`enum` or is not required); such
+    // schemas fall back to being validated against every oneOf branch.
+    this.fallbackAjv = new AjvConstructor({ ...ajvOptions, discriminator: false });
```
vadyvas, line 152: How does the lint handle this case? i cant find any fallback

sobanieca-redocly: Lint doesn't have a fallback — it just reports that Ajv couldn't compile the schema (the try/catch is in validateExample in packages/core/src/rules/utils.ts), which is fair there, since it points to an issue in the OAD itself. Drift is a different story. Lint only compiles a schema when it has an example attached, so even a spec with no lint issues can contain a loose discriminator. Without the fallback, drift would then report "Schema compilation failed" for every exchange hitting that schema — errors about perfectly valid traffic. The fallback is used in compileSchema() at the bottom of this file: if the strict instance throws, the schema is recompiled with discriminator: false, which validates against every oneOf branch — same as drift did before this PR.

vadyvas: Looks good, the const/enum + required constraints come from Ajv, not the spec ([docs](https://ajv.js.org/json-schema.html#discriminator)), so the fallback is ok

vadyvas (approved):

packages/cli/src/commands/drift/engine/schema-validator.ts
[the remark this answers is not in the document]
sobanieca-redocly: False positive: @redocly/ajv's discriminator keyword defines no metaSchema, so validateSchema never gates the discriminator object's own properties — defaultMapping and x- extensions compile fine either way.

zalewskigrzegorz (commented):
Fix itself looks right. Four notes below, none blocking: the fallback is invisible when it triggers, and the AWS prefix list still leaves `x-amz-apigw-id` flagged.

packages/cli/src/commands/drift/engine/schema-validator.ts
```diff
@@ -195,8 +204,16 @@ export class SchemaValidator {
...
-    const compiled = this.ajv.compile(schema as AnySchema);
+    const compiled = this.compileSchema(schema as AnySchema);
     this.scalarSchemaCache.set(cacheKey, compiled);
     return compiled;
   }
+
+  private compileSchema(schema: AnySchema): ValidateFunction {
+    try {
+      return this.ajv.compile(schema);
+    } catch {
```
zalewskigrzegorz, line 215: This catch swallows every compile error, not just discriminator ones. A schema that fails for an unrelated reason (bad `pattern`, unresolvable `$ref`, `$id` already registered) gets silently recompiled on `fallbackAjv`, and the message that reaches `validate()` comes from the second instance instead of the real cause.

```suggestion
    } catch (error) {
      if (!/^discriminator:/.test((error as Error).message ?? '')) {
        throw error;
      }
      return this.fallbackAjv.compile(schema);
    }
```

packages/cli/src/commands/drift/engine/schema-validator.ts
```diff
@@ -129,10 +130,11 @@ export class SchemaValidator {
   private readonly scalarSchemaCache = new Map<string, ValidateFunction>();
 
   public constructor(options?: { coerceTypes?: boolean }) {
-    this.ajv = new AjvConstructor({
+    const ajvOptions: Options = {
       strict: false,
       allErrors: true,
       allowUnionTypes: true,
+      discriminator: true,
```
zalewskigrzegorz, line 137: `relaxRequiredForTarget` strips `readOnly` properties from `required` on the request copy of the schema. When the discriminator property is `readOnly` (a server-assigned `type` or `kind` field, which is common), the tag is no longer required in any branch, Ajv refuses to compile, and request validation silently keeps the old every-branch behavior while responses get the fix. Worth either preserving the discriminator property in `required` during relaxation, or logging the fallback so this is visible.

packages/cli/src/commands/drift/utils/http.ts
```diff
@@ -67,9 +68,11 @@ const IGNORED_UNDOCUMENTED_HEADER_PREFIXES = [
   'cf-',
   'sec-ch-',
   'sec-fetch-',
+  'x-amzn-',
```
zalewskigrzegorz, line 72: `x-amzn-` misses the `x-amz-` family, which is where most AWS response headers live: `x-amz-apigw-id` from API Gateway, `x-amz-cf-id` and `x-amz-cf-pop` from CloudFront, `x-amz-request-id` and `x-amz-id-2` from S3. An API behind API Gateway emits both prefixes on every call, so the finding this is meant to remove still fires for `x-amz-apigw-id`.

```suggestion
  'x-amz-',
  'x-amzn-',
```

packages/cli/src/__tests__/commands/drift/schema-validator.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  it('reports only the selected branch errors instead of noise from every oneOf branch', () => {
+    const paymentWithInvalidId = { ...validBankTransferPayment, id: 'PAY-123' };
+
+    const result = validator.validate(paymentSchema, paymentWithInvalidId, 'response');
+
+    expect(result.valid).toBe(false);
+    expect(result.errors).toHaveLength(1);
+    expect(result.errors[0]).toMatchObject({ keyword: 'pattern', instancePath: '/id' });
+  });
```
zalewskigrzegorz, line 61: Missing the case that produces the loudest noise today: a discriminator value that matches no branch (`method: 'CHECK'`). With `discriminator: true` Ajv reports one error on the tag property instead of a failure per branch, and nothing here locks that in.

zalewskigrzegorz (approved):

packages/cli/src/commands/drift/engine/schema-validator.ts
sobanieca-redocly: Not taking this one. The two Ajv instances differ only in the `discriminator` flag, so a compile error from an unrelated cause reproduces identically on both and its real message is what reaches `validate()`. Checked the cases you named:

```
bad pattern → "Schema compilation failed: Invalid regular expression: /([a-/u: Unterminated character class"
bad $ref    → "Schema compilation failed: can't resolve reference #/nope from id #"
```

Both are the true cause, not a message from the second instance.

More importantly, `/^discriminator:/` couples the fallback to `@redocly/ajv`'s internal error wording. Every throw site in its discriminator vocabulary happens to be prefixed that way today, but if a fork bump reflows one of those strings the fallback silently stops firing and we are back to the every-branch false positives this PR exists to remove — a silent regression in the fix itself, traded for a double compile on an already-failing path.

You are right about one thing, though not for the reason given: a second schema object reusing an `$id` compiles quietly on `fallbackAjv` and validates, so discriminator support is lost without a word, and only a third occurrence surfaces `schema with key or id ... already exists`. Real, but it needs duplicate `$id`s across distinct schema objects, and the outcome is a validation that is looser rather than wrong. Not worth guarding against here.

packages/cli/src/commands/drift/engine/schema-validator.ts
sobanieca-redocly: The mechanism is right — confirmed on a schema whose tag is `readOnly`:

```
response → 1 error:  /cardNumber:type                                    ← discriminator routing works
request  → 4 errors: /cardNumber:type, :required, /method:enum, :oneOf   ← every-branch noise
```

But the fallback is the correct outcome here, not a bug to fix. If the discriminator property is `readOnly`, a request body legitimately may not carry it, so there is nothing to key on and validating every branch is the only sound option. Preserving the tag in `required` on the request copy would make drift report a missing required property for requests that correctly omit a server-assigned field — trading silent looseness for loud false positives, which is the opposite of what this PR is for.

Not logging it either: a per-schema warning on a code path that is behaving correctly is exactly the noise `drift` is trying to reduce.

Worth saying that a `readOnly` discriminator on a schema used for request bodies is a spec smell in the first place — the client has to send the tag for the union to mean anything.

packages/cli/src/__tests__/commands/drift/schema-validator.test.ts
sobanieca-redocly: Done. Thanks.

packages/cli/src/commands/drift/utils/http.ts
sobanieca-redocly: Added. Thanks.

PR #2965, opened 2026-07-24
fix: keep sibling keywords when a referenced schema has a root-level $ref

### LINKED ISSUE #2964: bundle drops sibling keywords when an external schema file has a root-level `$ref`
## Describe the bug

When an external schema file uses OpenAPI 3.1 root-level `$ref` with sibling keywords (`type`, `title`, `properties`, `required`, etc.), `redocly bundle` resolves that file to only the `$ref` target and discards the siblings.

References to that schema are then rewritten to the target schema. Intermediate schemas effectively disappear from the useful bundled model (or remain as a pure `$ref` alias with no siblings).

`redocly lint` does not report a problem on the source; the loss happens at bundle time.

### OpenAPI 3.1 context (Reference Object vs Schema Object)

This report is specifically about **Schema Objects** (data models), not Reference Objects elsewhere in the document.

In OpenAPI 3.1+:

- **Reference Objects** outside Schema Objects (for example path item, parameter, response, or request body refs) are still Reference Objects. Sibling keywords next to `$ref` are restricted (notably `summary` and `description`).
- **Schema Objects** adopt **JSON Schema Draft 2020-12**. In that model, `$ref` may appear alongside other schema keywords, and those siblings are applied together with the referenced schema (composition), not ignored.

So a schema like:

```yaml
title: Bad request
type: object
$ref: ./BaseProblem.yaml
properties:
  status:
    type: integer
required:
  - status
```

is a valid OAS 3.1 Schema Object / JSON Schema 2020-12 schema. Bundling should preserve that composition, not treat the root `$ref` as a full replacement the way a non-schema Reference Object might be inlined.

## Repro

```yaml
# openapi.yaml
openapi: 3.1.0
info:
  title: file $ref sibling repro
  version: 1.0.0
paths:
  /demo:
    get:
      operationId: getDemo
      responses:
        '400':
          description: Bad request
          content:
            application/json:
              schema:
                $ref: ./schemas/BadRequest.yaml
components:
  schemas:
    BadRequest:
      $ref: ./schemas/BadRequest.yaml
    BaseProblem:
      $ref: ./schemas/BaseProblem.yaml
```

```yaml
# schemas/BaseProblem.yaml
type: object
properties:
  title:
    type: string
```

```yaml
# schemas/BadRequest.yaml
title: Bad request
type: object
$ref: ./BaseProblem.yaml
properties:
  status:
    type: integer
    minimum: 400
    maximum: 400
required:
  - status
```

```bash
npx @redocly/cli@2.38.0 bundle openapi.yaml --ext json
# also reproduced with 2.39.0 and 2.40.0
```

### Actual

```json
"BadRequest": { "$ref": "#/components/schemas/BaseProblem" }
```

Path/response `400` schema also points at `BaseProblem`. Sibling keywords from `BadRequest.yaml` (`title`, `type`, `properties.status`, `required`) are gone.

### Expected

Bundled `BadRequest` keeps sibling keywords composed with `BaseProblem` (or an equivalent merged/`allOf` form), and references keep pointing at `BadRequest`.

## Control

The same API expressed with `allOf` instead of root `$ref` + siblings bundles correctly (`BadRequest` retained; refs point at `BadRequest`).

Property-level `$ref` + siblings (for example `writeOnly` next to a property `$ref`) appear to bundle correctly in our larger spec. The failure we hit is specific to **root-level `$ref` + siblings on an external schema file**.

## Independent validation (not Redocly)

We validated the same `$ref` + siblings composition with popular JSON Schema Draft 2020-12 implementations. Sibling keywords are applied, not ignored:

Schema under test (equivalent form):

- `$ref` to `BaseProblem`
- sibling `properties.status` (`minimum`/`maximum` 400)
- sibling `required: ["status"]`

| Instance | Ajv (2020-12) | `@hyperjump/json-schema` | Python `jsonschema` (`Draft202012Validator`) |
|----------|---------------|--------------------------|-----------------------------------------------|
| `{ "title": "Bad request", "status": 400 }` | PASS | PASS | PASS |
| `{ "title": "Bad request" }` (missing `status`) | FAIL | FAIL | FAIL |
| `{ "title": "Bad request", "status": 500 }` | FAIL | FAIL | FAIL |

Ajv also compiles the schema successfully against the Draft 2020-12 meta-schema.

If the schema is collapsed to Redocly’s bundled form (`{ "$ref": "..." }` only), Ajv accepts the missing-`status` instance — i.e. the sibling constraints are lost. That matches the buggy bundle output and shows the bundler is dropping JSON Schema semantics that other validators enforce.

Separately, `@apidevtools/swagger-parser` accepts the OpenAPI 3.1 document; after dereference, `BadRequest` still retains `properties.status`.

## Environment

- `@redocly/cli` 2.38.0, 2.39.0, 2.40.0
- OpenAPI 3.1.0 / 3.1.1

## Notes

Possibly related to sibling traversal work in #2768 / #2772 / #1907 / #2362, but those focus on resolving refs *inside* siblings. Here sibling keywords on an external schema file root are dropped entirely when that root `$ref` is resolved.

RomanHotsiy:
## What/Why/How?

`bundle` dropped the sibling keywords of a root-level `$ref` when the referenced schema itself started with a `$ref` — for example an external schema file that composes another file (`$ref: ./BaseProblem.yaml` plus `title`/`type`/`properties`/`required`).
The bundled output collapsed such schemas to a bare `$ref` pointing at the end of the chain, silently losing the sibling constraints.
In OpenAPI 3.1, Schema Objects follow JSON Schema draft 2020-12, where `$ref` with siblings is a composition, not a transparent alias.

Root cause: `followRef` in `resolve.ts` chased every chained `$ref` to the end of the chain, discarding the intermediate nodes together with their sibling keys.

The fix — resolution results now carry the composition as an additive `chain` field, so `ctx.resolve()` keeps returning the chain-end node exactly as before and nothing changes for existing rules, decorators, or plugins:

- `resolve.ts` — `followRef` still chases to the end of the chain, but records every composed `$ref` it passes through as a `chain` hop (`ResolvedRefChainHop`) on the resolve result.
- `walk.ts` — the walker walks each chain hop as a ref node from its own location, so hop siblings and their inner `$ref`s are processed; `ResolveResult` exposes the chain (with `Location`s) to visitors.
- `bundle-visitor.ts` — a chained ref is rewritten to point at the first composed hop (saved as a component with siblings intact) instead of the chain end; the hop's own `$ref` is rewritten by the walker pass. Discriminator mappings and component-name dedup use the same effective target.
- `no-required-schema-properties-undefined.ts` — the one rule that needs the composition: it validates `required` on chain hops (they are never visited as Schema nodes) and follows chains and sibling `properties`/`allOf`/`anyOf`/`oneOf` when looking up required properties.
- `rules/utils.ts` — `resolveSchema` passes the chain through for chain-aware rules.

Because `resolve()` behavior is unchanged for consumers, the rule regressions found on the earlier revision of this PR (`array-parameter-serialization`, `struct`, discriminator checks) cannot occur — verified explicitly for each.
The `remove-unused-components` decorators are chain-safe by construction (they build their usage graph from raw `$ref` pointers), covered by a new test.

Composed schemas are now preserved as components with their sibling keywords intact, references keep pointing at the composed schema, and `--dereferenced` merges the composition in chain order.

## Reference

Fixes #2964

## Testing

- Unit tests (written to fail first on the broken behavior): bundle tests for external-file and internal component chains plus `--remove-unused-components`, a resolver test asserting the recorded chain, and regression tests for `struct`, `no-required-schema-properties-undefined` (sibling-defined, chain-defined, allOf-sibling-defined, and genuinely-missing required properties), and `no-unused-components`.
- Two e2e bundle tests: `sibling-refs-root-external-file` (fails on unfixed code) and `sibling-refs-root-in-file`.
- Full unit suite (2031 tests) and e2e suite pass; no existing snapshots changed. Lint warning count unchanged.
- Verified by hand: `bundle`, `bundle --dereferenced`, `lint` on the issue repro, and the rule-regression repros from the review.

## Check yourself

## Security

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches core ref resolution, document walking, and bundling—widely used paths—but behavior for plain `$ref` aliases is preserved via chain-end resolution; risk is mainly regressions in edge cases (multi-hop chains, discriminators, lint on composed schemas).
> 
> **Overview**
> Fixes **`bundle`** dropping schema keywords (`title`, `properties`, `required`, etc.) next to a `$ref` when the target schema itself starts with another `$ref` (common in external files that compose a base schema). Bundled output used to collapse to a bare pointer at the chain end.
> 
> **Resolution** now records each **composed** `$ref` (ref plus siblings) as a **`chain`** hop on resolve results while still returning the chain-end node from `ctx.resolve()`, so most consumers stay unchanged.
> 
> **Bundling** uses the first chain hop as the effective component target so composed schemas keep siblings; discriminator mappings and component dedup follow the same rule. **`walk`** visits chain hops and sibling keys on refs once, and passes **`chain`** into ref visitors.
> 
> **Lint rules** gain chain-aware behavior where needed: **`no-required-schema-properties-undefined`** validates `required` on composed refs and chain hops; **`schemaHasProperty`** considers sibling and chain-hop keywords. **`replaceRef`** merges object targets with sibling precedence and handles non-object targets via the parent.
> 
> Unit, rule, and e2e tests cover external root refs, collisions, discriminators, unused components, and dereferenced output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d78bb2094d88771cf32a4301e0c71bb119f92d98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/root-ref-sibling-keywords.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Fixed the `bundle` command losing schema keywords (such as `title`, `properties`, or `required`) written next to a `$ref` when the referenced schema starts with its own `$ref`.
+These keywords are now kept in the bundled output, and the `lint` command validates them instead of skipping them.
```
RomanHotsiy:
```suggestion
```

RomanHotsiy: Applied in 113c97d5f — the trailing sentence was removed, so the changeset is now a single line.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -33,6 +35,17 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+            // a JSON Schema 2020-12 $ref can carry sibling keywords, so check them too
+            const siblingProperties = (schemaOrRef as AnySchema).properties;
+            if (
+              isPlainObject(siblingProperties) &&
+              getOwn(siblingProperties, propertyName) !== undefined
+            ) {
+              return true;
+            }
+          }
+
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed in 420e2e9de. Verified both cases false-positive on this branch (a required property defined further down a composed `$ref` chain, and one defined in an `allOf`/`anyOf`/`oneOf` sibling of a referenced composed `$ref`) and pass on origin/main — a regression from the resolver now stopping at composed refs. `hasProperty` now checks composition keywords on the composed node's siblings and follows the resolved schema's own `$ref` chain. Added 3 regression tests (two false-positive cases + one that confirms genuinely-undefined props are still reported).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

bencagri (approved):
lgtm

.changeset/root-ref-sibling-keywords.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Fixed the `bundle` command losing schema keywords (such as `title`, `properties`, or `required`) written next to a `$ref` when the referenced schema starts with its own `$ref`.
```
JLekawa:
```suggestion
Fixed the `bundle` command losing schema keywords (such as `title`, `properties`, or `required`) written next to a `$ref` when the referenced schemas started with their own `$ref`.
```

JLekawa (approved):

.changeset/root-ref-sibling-keywords.md
RomanHotsiy: Applied in 232e3e304 — the changeset now reads "referenced schemas started with their own `$ref`".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -27,41 +28,68 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          resolveFrom?: string
+        ): boolean => {
+          // A JSON Schema 2020-12 $ref can carry sibling keywords that compose the target,
+          // so check them before the $ref is resolved away.
+          if (
+            isRef(schemaOrRef) &&
+            definesProperty(schemaOrRef, propertyName, visited, resolveFrom)
+          ) {
+            return true;
+          }
```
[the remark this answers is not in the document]
RomanHotsiy: Investigated but not changing — the `visited` guard is reached on every cycle-forming path, so this doesn't overflow. A composition cycle can only form through a `$ref`, and every `$ref` traversal goes through `resolveSchema` + the `visited` check inside `hasProperty`. The top `definesProperty(schemaOrRef)` only reads the current node's own `allOf`/`anyOf`/`oneOf` arrays; each element is handed back to `hasProperty` (which guards), and `visited` is shared through the `allOf` `.some()` path so it accumulates and catches every back-edge. Verified with 6 cyclic schemas, including `allOf`/`anyOf` loop elements that themselves carry keywords — all terminate normally (no `Maximum call stack size exceeded`), and a cycle whose property is reachable resolves correctly. Adding a redundant guard would be defensive noise the repo's code-quality rules discourage.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/resolve.ts
```diff
@@ -476,7 +476,9 @@ export async function resolveDocument(opts: {
       resolvedRef.node = target;
       resolvedRef.document = targetDoc;
       const refId = makeRefId(document.source.absoluteRef, ref.$ref);
-      if (resolvedRef.document && isRef(target)) {
+      // A $ref with sibling keys is a composition, not a transparent alias — resolution stops
+      // there so the siblings survive; only pure refs are followed further.
+      if (resolvedRef.document && isRef(target) && Object.keys(target).length === 1) {
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed by reworking the design in 2cb68571b. Verified the regression was real (`array-parameter-serialization` stopped reporting on a composed chain that reports on main). Instead of stopping resolution at a composed `$ref`, `followRef` now chases to the chain end exactly like before and records the composed hops in an additive `chain` field on the resolve result — so `ctx.resolve()` returns the same node as on main for every rule and this regression class can't occur. Only the walker, the bundler, and `no-required-schema-properties-undefined` (which validates `required` on chain hops) consume the chain. Release bumped to minor since `ResolveResult` gains a public field.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -18,7 +20,122 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          validatedChainHops.add(chainHop.node);
+          reportUndefinedRequired(
+            chainHop.node as AnySchema,
+            chainHop.location,
+            ctx,
+            chainHop.location.source.absoluteRef
+          );
+        }
+      },
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed in 1c03c85fa — `ref.leave` now validates the ref node itself as well as the chain hops, so an inline `$ref` with sibling `required` is checked. Digging into this also surfaced a crash the same code path had: a composed *parameter* `$ref` with `required: true` (a boolean) reached `.entries()` and crashed the linter — now gated on Schema-typed refs plus an `isNotEmptyArray` guard. Both covered by new regression tests.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -308,20 +346,24 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
+              );
+            }
+            // properties on the same level as $ref
+            props.push(...refSiblingProps.filter((k) => !props.includes(k)));
+          } else {
+            // the resolved node was already visited, but this ref's sibling keys still need walking
+            props = refSiblingProps;
           }
 
           for (const propName of props) {
```
vadyvas, line 384: problem when resolved $ref properties have same names as siblings (same on main)

```yaml
openapi: 3.1.0
info:
  title: Example
  version: '0.0.1'
servers:
  - url: 'http://localhost:8080'
paths:
  /sample:
    get:
      operationId: example1
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/test'
                properties:
                  second:
                    $ref: reference.yaml
components:
  schemas:
    test:
      type: object
      title: test schema
      properties:
        first:
          type: integer
```

packages/core/src/walk.ts
```diff
@@ -369,6 +411,15 @@ export function walkDocument<T extends BaseVisitor>(opts: {
         }
       }
 
+      if (nodeIsRef && resolvedChain?.length) {
+        // walk the composed $ref hop from its own location so its siblings and inner $ref
+        // are processed; the hop's own resolution carries the rest of the chain
+        const chainHop = resolvedChain[0];
+        if (!walkedComposedRefs.has(`${type.name}::${chainHop.location.absolutePointer}`)) {
```
vadyvas: A design question about the chain approach

resolution follows the chain to the end (as on main), composed hops are recorded in `chain`, and the walker does an extra `walkNode(chain[0])` jump from the hop's own location - a traversal path that exists next to the normal tree walk

Did you consider a version without the `chain` field at all? `followRef` would still resolve the whole chain (so every hop and external file still ends up in the map), but it would return the first composed node instead of the last one. Then the walker doesn't need the extra jump -  it just walks what it got: a ref with siblings. It walks the siblings, resolves the `$ref`, gets the next hop, and repeats until the end of the chain. Each hop is walked from its own file with its own keys. And an inline `$ref` with siblings written directly in the document would go through the same path instead of being a special case.

packages/core/src/walk.ts
RomanHotsiy: Fixed in 094ff48af — the props loop preferred the resolved node's value and only fell back to the sibling when the key was absent on the target, so a colliding sibling subtree was never walked (your repro left `$ref: reference.yaml` unrewritten in the bundle). The walker now walks both values when a `$ref` sibling key collides with a resolved key — the sibling from its original location. This also subsumes the old scalar-ref special case, which was a partial fix for the same hole limited to scalar props. Covered by a new bundle test with your repro shape; verified your exact YAML now bundles with `second` rewritten to `#/components/schemas/reference`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Yes — the earlier revision of this PR was essentially that design: resolution returned the first composed node instead of the last. It broke down in two places. First, `ctx.resolve()`'s return value is consumed by a dozen-plus call sites (rules like `array-parameter-serialization`, `spec-discriminator-defaultMapping`, ajv-based example validation, respect-core) that all read keywords off the result expecting the effective schema — returning the first hop made `array-parameter-serialization` go silent on composed chains and `no-required-schema-properties-undefined` false-positive (both verified with repros, and Bugbot flagged the same class). Each consumer would need to learn to loop the chain itself. Second, "the walker just walks what it got and repeats" still needs a mechanism for the hop's own `$ref`: `walkNode` resolves one hop per node, and the resolved node's `$ref` key isn't in any type's properties — the repeat step was exactly the same extra jump, just triggered by `isRef(resolvedNode)` instead of `chain`, plus Schema visitors then receive `$ref`-bearing nodes (struct needed guards for `$ref`-not-expected and required-field checks; assertions and plugins would see a node shape they've never seen). The `chain` field keeps main's `resolve()` contract for every consumer and confines the new traversal to one keyed jump. Inline `$ref`s with siblings do go through the normal path today — their siblings are walked as children of the ref node; the chain only covers hops that are otherwise invisible. Happy to discuss further if you see a way around the `resolve()` consumers issue.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tests/e2e/bundle/sibling-refs-root-in-file/openapi.yaml
vadyvas, line 1: can you add test with `dereference` ?

packages/core/src/walk.ts
```diff
@@ -165,22 +179,40 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
+      node: resolvedNode,
+      location: resolvedLocation,
+      error,
+      chain: resolvedChain,
+    } = resolve(node);
     const enteredContexts: Set<VisitorLevelContext> = new Set();
 
+    if (nodeIsRef && Object.keys(node).length > 1) {
+      // composed $refs can also be reached as chain hops — record this walk to avoid a repeat
+      walkedComposedRefs.add(`${type.name}::${location.absolutePointer}`);
```
vadyvas: maybe it worth to create helper for `${type.name}::${location.absolutePointer}` or reuse `makeRefId` ? 

```ts
const composedRefWalkId = (type: NormalizedNodeType, location: Location) =>
  `${type.name}::${location.absolutePointer}`;
```

packages/core/src/walk.ts
```diff
@@ -282,12 +314,18 @@ export function walkDocument<T extends BaseVisitor>(opts: {
         }
       }
 
-      if (visitedBySome || !isNodeSeen) {
-        seenNodesPerType[type.name] = seenNodesPerType[type.name] || new Set();
-        seenNodesPerType[type.name].add(resolvedNode);
+      const shouldWalkChildren = visitedBySome || !isNodeSeen;
+      const refSiblingProps =
+        nodeIsRef && isPlainObject(node) ? Object.keys(node).filter((k) => k !== '$ref') : [];
```
vadyvas: `isRef` already have `isPlainObject(node)`

packages/core/src/resolve.ts
```diff
@@ -477,7 +494,20 @@ export async function resolveDocument(opts: {
       resolvedRef.document = targetDoc;
       const refId = makeRefId(document.source.absoluteRef, ref.$ref);
       if (resolvedRef.document && isRef(target)) {
+        // A $ref with sibling keys is a composition (JSON Schema 2020-12), not a transparent
+        // alias — record it as a chain hop so the composition survives for consumers that need it.
+        const chainHop: ResolvedRefChainHop | undefined =
```
vadyvas:
```suggestion
        const chainHop: ResolvedRefChainHop | undefined =
          Object.keys(target).length > 1
            ? {
                node: target,
                location: new Location(resolvedRef.document.source, resolvedRef.nodePointer!),
              }
            : undefined;
```
wdyt ?

packages/core/src/walk.ts
```diff
@@ -165,22 +179,40 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
         : error instanceof YamlParseError
           ? new Location(error.source, '')
           : undefined;
 
-      return { location: newLocation, node, error };
+      return {
+        location: newLocation,
+        node,
+        error,
+        chain: chain?.map((chainHop) => ({
```
vadyvas: can we avoid if https://github.com/Redocly/redocly-cli/pull/2965/changes#r3659009998 ?

packages/core/src/resolve.ts
```diff
@@ -477,7 +494,20 @@ export async function resolveDocument(opts: {
...
+          Object.keys(target).length > 1
+            ? {
+                node: target,
+                document: resolvedRef.document,
+                nodePointer: resolvedRef.nodePointer!,
+              }
+            : undefined;
         resolvedRef = await followRef(resolvedRef.document, target, pushRef(refStack, target));
+        if (chainHop && resolvedRef.resolved) {
+          resolvedRef = { ...resolvedRef, chain: [chainHop, ...(resolvedRef.chain ?? [])] };
```
tatomyr, line 513: Wouldn't this read more cleanly?


```suggestion
          resolvedRef.chain.unshift(chainHop);
```

RomanHotsiy: Kept the non-mutating version, but it deserved a better explanation than it had — `unshift` isn't equivalent here. `followRef` returns `{ ...resolvedRef }`, a shallow copy, so the returned `chain` array is the same array object the inner recursion just stored in its own map entry. `unshift` would prepend the outer hop into the inner entry's chain (e.g. resolving `#/second` would list `#/first` in its chain). I tried your suggestion against a 3-level chain test and it happened to pass — but only because each node's own `followRef` re-walk later overwrites the corrupted entry, which makes correctness depend on promise interleaving order rather than being guaranteed. Added a comment on that line stating the shared-array constraint, plus a multi-hop test pinning that each level's chain stays independent (71b7bc5f8).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -173,14 +180,14 @@ export function makeBundleVisitor({
...
           if (dereference) {
-            saveComponent(componentType, resolved, ctx);
-            replaceRef(node, resolved, ctx);
+            saveComponent(componentType, target, ctx);
+            replaceRef(node, target, ctx);
           } else {
-            node.$ref = saveComponent(componentType, resolved, ctx);
-            resolveBundledComponent(node, resolved, ctx);
+            node.$ref = saveComponent(componentType, target, ctx);
+            resolveBundledComponent(node, target, ctx);
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed in 7801c2b21 — `resolveBundledComponent` now re-registers the original resolution (chain-end node plus its composed hops) under the rewritten pointer instead of storing the first hop with no chain, so `ctx.resolve()` keeps the same contract for rewritten refs as for any other ref during the bundle walk.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -173,14 +180,14 @@ export function makeBundleVisitor({
...
         const componentType = mapTypeToComponent(ctx.type.name, version);
         if (!componentType) {
-          replaceRef(node, resolved, ctx);
+          replaceRef(node, target, ctx);
         } else {
           if (dereference) {
-            saveComponent(componentType, resolved, ctx);
-            replaceRef(node, resolved, ctx);
+            saveComponent(componentType, target, ctx);
+            replaceRef(node, target, ctx);
```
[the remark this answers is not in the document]
RomanHotsiy: Not changing this one — the shallow sibling-wins merge is the existing `replaceRef` semantics that `--dereferenced` has always applied to single-level `$ref` + siblings on main; this PR extends it consistently along the chain (before the PR, chained sibling keywords were dropped entirely). True JSON Schema 2020-12 composition can't be expressed as a plain merge at all (two `pattern`s or disjoint `enum`s have no merged form) — a faithful dereference would need to rewrite compositions to `allOf`, which changes the output shape for all existing `--dereferenced` users and is a separate feature decision. This limitation is called out in the PR description's reviewer note.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tests/e2e/bundle/sibling-refs-root-in-file/openapi.yaml
RomanHotsiy: Added in 5d8b082c5 — a `--dereferenced` e2e test on this fixture with its own `snapshot_2.txt`, locking in the composed merge output.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Added the `composedRefWalkId` helper as suggested (5d8b082c5). Kept it separate from `makeRefId` — that one keys on source + $ref string, while this keys on type + resolved location, so reusing it would conflate two different namespaces.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Removed the redundant `isPlainObject(node)` — `isRef` already guarantees it (5d8b082c5).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/resolve.ts
RomanHotsiy: Applied with one deviation (5d8b082c5): the hop now carries a ready `Location` as you suggested, but keeps `document` alongside it — the resolver itself needs the parsed document to walk hop subtrees for refs inside sibling keywords (`followRef` reads `document.parsed`), and a `Location` only carries the `Source`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Yes — with the hop carrying a `Location` (see the other thread), the per-resolve mapping is gone; the walker now passes `chain` straight through (5d8b082c5).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/rules/common/no-required-schema-properties-undefined.ts
```diff
@@ -18,49 +20,130 @@ export const NoRequiredSchemaPropertiesUndefined:
...
+          chainHop.node as AnySchema,
+          propertyName,
+          visited,
+          ctx,
+          chainHop.location.source.absoluteRef
+        )
+      ) {
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed in 7f9f30103 — reproduced the false positive with an `allOf` whose first branch visits the leaf directly and second branch reaches it through a composed hop defining the required property. `hasProperty` now gates the chain-end schema and each chain hop on `visited` independently instead of returning before the hop check; hops are added to `visited` themselves so cycle termination is preserved (re-verified against the six cyclic-schema repros). Regression test added.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.changeset/root-ref-sibling-keywords.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Fixed the `bundle` command losing schema keywords (such as `title`, `properties`, or `required`) written next to a `$ref` when the referenced schemas started with their own `$ref`.
+Ref resolution now exposes such composed `$ref`s as a `chain` field on the resolve result, so custom rules and decorators can inspect them.
```
RomanHotsiy: This is not needed in changelog
```suggestion
```

tatomyr:
image


Nice performance gain. Do you know the reason?

RomanHotsiy:
Verified locally and bisected it — the gain is real and comes from one commit: 094ff48af, the sibling-collision fix in `walk.ts`. Bundling Rebilly's api-definitions: main 645.9ms ± 3.8 → after the chain rework (2cb68571b) 651.0ms ± 5.1 (no change) → after the props-loop restructure (094ff48af) 578.1ms ± 3.1, HEAD 573.7ms ± 2.9 (~11%, matching the 1.10x here). The mechanism: the old loop ran the scalar-sibling special case — `isRef(node[propName])` plus a fallback property read — for every property of every object node in the document. The restructured loop gates all `$ref`-sibling handling behind `nodeIsRef`, which is false for almost every node, so those per-property checks dropped off the hot path entirely. So the correctness fix for colliding sibling keys happened to also delete the walker's most-executed redundant check.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -355,20 +392,42 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
         }
       }
 
+      if (nodeIsRef && resolvedChain?.length) {
+        // walk the composed $ref hop from its own location so its siblings and inner $ref
+        // are processed; the hop's own resolution carries the rest of the chain
+        const chainHop = resolvedChain[0];
+        if (!walkedComposedRefs.has(composedRefWalkId(type, chainHop.location))) {
+          walkNode(chainHop.node, type, chainHop.location, parent, key);
+        }
```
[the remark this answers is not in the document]
RomanHotsiy: Fixed in 64e1a0f6d as part of a broader walker simplification — chain hops are now walked with `parent: undefined` (they have no parent at their own location), and `replaceRef` dereferences a parentless hop in place by dropping the `$ref` and keeping the sibling keys, so a hop resolving to a non-object (e.g. an OAS 3.1 boolean schema) can no longer overwrite the outer ref site. Verified with a composed-over-boolean repro in both `--dereferenced` (`{title}` output, no corruption, no phantom errors) and plain bundle mode (boolean saved as its own component).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -29,9 +35,22 @@ export type NonUndefined =
   | object
   | Record<string, any>;
 
+// A composed $ref (with sibling keys) the resolution chased through on the way to `node`.
+export type ResolvedChainHop = ResolvedRefChainHop;
```
RomanHotsiy: Why do we need new type? Let's get rid of it.

packages/core/src/walk.ts
```diff
@@ -320,51 +384,30 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
             }
+          }
+        }
+      }
 
-            if (!isNamedType(propType) || (propType.name === 'scalar' && !isRef(value))) {
-              continue;
-            }
+      if (nodeIsRef && resolvedChain?.length) {
+        // the target composes another $ref: the hop is walked as a ref node from its own
```
RomanHotsiy: make those comments simpler to understand

packages/core/src/walk.ts
```diff
@@ -320,51 +384,30 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
+        // location (it has no parent there); its resolution carries the rest of the chain
+        const chainHop = resolvedChain[0];
+        if (!walkedComposedRefs.has(composedRefWalkId(type, chainHop.location))) {
+          walkNode(chainHop.node, type, chainHop.location, undefined, key);
+        }
+      }
 
-            walkNode(value, propType, loc.child([propName]), resolvedNode, propName);
+      if (nodeIsRef) {
+        // $ref sibling keys belong to the ref node itself and walk from its own location;
```
RomanHotsiy: and this too

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -255,22 +266,35 @@ export function makeBundleVisitor({
               return;
             }
 
-            mapping[name] = saveComponent(schemaComponentType, resolved, ctx);
+            mapping[name] = saveComponent(
+              schemaComponentType,
+              resolved.chain?.[0] ?? resolved,
+              ctx
+            );
```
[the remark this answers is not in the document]
RomanHotsiy: Investigated and refuted — the hop does get walked. Discriminator mapping values (and 3.2 `defaultMapping`) are typed `directResolveAs: 'Schema'`, so the walker wraps each mapping string in a synthetic ref node and walks it like any tree ref, which triggers the composed-hop walk from the hop's own location before `DiscriminatorMapping.leave` saves it. Verified with a repro: a mapping pointing at a composed external file whose sibling `properties` contains a further relative `$ref` bundles cleanly — the mapping string, the hop's inner `$ref`, and the sibling's relative ref are all rewritten to component pointers with zero unresolved errors. Since this interaction had no coverage, added it as a regression test in 6fbd859fa.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -320,51 +384,30 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
+        // the resolved node's (or hops') keys are walked in their own frames
+        for (const propName of Object.keys(node)) {
+          if (propName !== '$ref' && rawNode[propName] !== resolvedNode[propName]) {
+            walkProp(propName, rawNode[propName], location, node);
```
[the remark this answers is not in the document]
RomanHotsiy: Confirmed and fixed in b704b0ba9 — reproduced the crash (`Cannot read properties of null (reading 'description')`) with a `$ref` + sibling resolving to a YAML `null`; a regression from moving the sibling walk out of the `isPlainObject` branch. The comparison now uses optional chaining (`resolvedNode?.[propName]`), so null targets walk their siblings instead of aborting. Regression test added.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

RomanHotsiy:
Okay, now it's ready to reveiw.

packages/core/src/bundle/bundle-visitor.ts
```diff
@@ -290,12 +314,12 @@ export function makeBundleVisitor({
...
   function isEqualOrEqualRef(node: unknown, target: ComponentTarget, ctx: UserContext) {
-    if (
-      isRef(node) &&
-      ctx.resolve(node, rootLocation.absolutePointer).location?.absolutePointer ===
-        target.location.absolutePointer
-    ) {
-      return true;
+    if (isRef(node)) {
+      const resolved = ctx.resolve(node, rootLocation.absolutePointer);
+      const effectiveLocation = resolved.chain?.[0]?.location ?? resolved.location;
```
vadyvas: i think creating a utility or helper could improve readability in several places

```ts
const effectiveTarget = (resolved: ResolveResult<NonUndefined>): ComponentTarget =>
  resolved.chain?.[0] ?? resolved;
```

packages/core/src/walk.ts
```diff
@@ -165,22 +187,32 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
-    const { node: resolvedNode, location: resolvedLocation, error } = resolve(node);
+    const {
+      node: resolvedNode,
+      location: resolvedLocation,
+      error,
+      chain: resolvedChain,
+    } = resolve(node);
     const enteredContexts: Set<VisitorLevelContext> = new Set();
 
+    if (nodeIsRef && Object.keys(node).length > 1) {
```
vadyvas: i think creating a util in `ref-utils.ts` could improve readability

```ts
export function isRefWithSiblings(node: unknown): node is OasRef & Record<string, unknown> {
  return isRef(node) && Object.keys(node).length > 1;
}
```

packages/core/src/__tests__/bundle.test.ts
```diff
@@ -976,6 +976,140 @@ describe('sibling $ref resolution by spec', () => {
...
   });
+
+  it('should keep sibling keywords when an external schema file has a root-level $ref', async () => {
+    const { bundle: res, problems } = await bundle({
+      config: await createConfig({}),
+      ref: path.join(__dirname, 'fixtures/sibling-refs/openapi-root-ref-siblings.yaml'),
+    });
+
+    expect(problems).toHaveLength(0);
+    expect((res.parsed as any).components.schemas.BadRequest).toEqual({
```
vadyvas: wouldn't it be better to use snapshot check?

packages/core/src/ref-utils.ts
```diff
@@ -130,7 +130,15 @@ export function isAnchor(ref: string) {
...
   if (!isPlainObject(resolved.node)) {
-    ctx.parent[ctx.key] = resolved.node;
+    if (ctx.parent) {
+      ctx.parent[ctx.key] = resolved.node;
+    } else {
+      // a composed chain hop has no parent to replace, so dereference it in place
+      // by dropping the $ref and keeping the sibling keys
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      delete ref.$ref;
```
vadyvas: wdyt
```ts
export function replaceRef(ref: OasRef, resolved: ResolveResult<any>, ctx: UserContext) {
  if (!isPlainObject(resolved.node) && ctx.parent) {
    ctx.parent[ctx.key] = resolved.node;
    return;
  }

  delete (ref as Partial<OasRef>).$ref;

  if (isPlainObject(resolved.node)) {
    Object.assign(ref, { ...resolved.node, ...ref });
  }
}
```

packages/core/src/walk.ts
vadyvas: agree (to trigger the update)

vadyvas (commented):
One stylistic thing, non-blocking: the new code more comments than these files usually are, and a few of the comments are hard to read in themselves

packages/core/src/ref-utils.ts
RomanHotsiy: nice

packages/core/src/walk.ts
RomanHotsiy: Removed in d1b3b5595 — `ResolveResult` and `resolveSchema` now use `ResolvedRefChainHop` from `resolve.ts` directly; the alias is gone.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Simplified in d1b3b5595 — the three walker comments are now single plain-English sentences.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/bundle/bundle-visitor.ts
RomanHotsiy: Added in d1b3b5595 — `effectiveTarget` is now used in `ref.leave`, both discriminator handlers, and `isEqualOrEqualRef`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Added `isRefWithSiblings` to `ref-utils.ts` in d1b3b5595, used in both the walker's composed-ref marker and `followRef`'s hop check.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/__tests__/bundle.test.ts
RomanHotsiy: Converted to a snapshot check in d1b3b5595 — the explicit assertions were originally there to watch the test fail before the fix, but the snapshot is the better long-term guard.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/ref-utils.ts
RomanHotsiy: Applied in d1b3b5595 — verified it's behaviorally identical to the previous version (including the parentless-hop path keeping sibling keys) and the `Partial<OasRef>` delete drops both ts-ignores.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -165,22 +184,32 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
-    const { node: resolvedNode, location: resolvedLocation, error } = resolve(node);
+    const {
+      node: resolvedNode,
+      location: resolvedLocation,
+      error,
+      chain: resolvedChain,
+    } = resolve(node);
     const enteredContexts: Set<VisitorLevelContext> = new Set();
 
+    if (isRefWithSiblings(node)) {
```
vadyvas, line 212: found issue with node order in api description: 
`ref` visitors fire twice on a composed `$ref`, and whether it's once or twice depends on key order.

To reproduce:

```js
// plugin.cjs
module.exports = () => ({
  id: 'probe',
  rules: {
    oas3: {
      'count-ref-visits': () => ({
        ref(node, ctx) {
          if (typeof node.$ref === 'string' && Object.keys(node).length > 1) {
            ctx.report({ message: `ref visit: ${ctx.location.pointer}` });
          }
        },
      }),
    },
  },
});
```

```yaml
# redocly.yaml
plugins: [./plugin.cjs]
extends: []
rules:
  probe/count-ref-visits: error
```

```yaml
# openapi.yaml
openapi: 3.1.0
info: { title: t, version: 1.0.0 }
paths: {}
components:
  schemas:
    Usage:
      $ref: '#/components/schemas/Composed'
    Composed:
      $ref: '#/components/schemas/Base'
      title: Composed
    Base:
      type: object
```

```yaml
openapi: 3.1.0
info: { title: t, version: 1.0.0 }
paths: {}
components:
  schemas:
    # Same three schemas, only reordered: the tree walk reaches Composed first.
    Composed:
      $ref: '#/components/schemas/Base'
      title: Composed
    Base:
      type: object
    Usage:
      $ref: '#/components/schemas/Composed'
```

`lint` reports `Composed` twice. Move `Usage` below `Base` - reports once. On main it's always once

packages/core/src/walk.ts
```diff
@@ -211,6 +240,38 @@ export function walkDocument<T extends BaseVisitor>(opts: {
     }
 
     if (resolvedNode !== undefined && resolvedLocation && type.name !== 'scalar') {
+      // the isRef check narrows `node` to OasRef, but sibling keys are read from it too
+      const rawNode = node as Record<string, unknown>;
```
vadyvas: do we still need it after adding `isRefWithSiblings` ?

packages/core/src/walk.ts
RomanHotsiy: Confirmed with your repro and fixed in db869048c — the marker previously only gated the chain-hop walk, not a later tree walk of the same composed `$ref`. `walkNode` now checks the marker before firing: on a repeat visit of the same (type, location) it skips the `ref` visitors and the sibling-key walk (both fired identically on the first visit) while still running node-level visitors and children, which nested per-position visitors like `NamedSchemas.Schema` rely on. Added a walk test asserting exactly one `ref` visit for both key orders.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
RomanHotsiy: Gone in db869048c — the sibling walk is now gated on `isRefWithSiblings(node)`, whose type predicate narrows `node` for indexing, so the `rawNode` cast and its comment are removed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/core/src/walk.ts
```diff
@@ -320,51 +381,29 @@ export function walkDocument<T extends BaseVisitor>(opts: {
...
-            if (!isNamedType(propType) || (propType.name === 'scalar' && !isRef(value))) {
-              continue;
-            }
+      if (nodeIsRef && resolvedChain?.length) {
+        // walk the composed $ref the resolution passed through; the rest of the chain
+        // is walked recursively the same way
+        const chainHop = resolvedChain[0];
+        if (!walkedComposedRefs.has(composedRefWalkId(type, chainHop.location))) {
+          walkNode(chainHop.node, type, chainHop.location, undefined, key);
+        }
```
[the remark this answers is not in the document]
RomanHotsiy: Fixed in db869048c together with the sibling thread from @vadyvas — same root cause. Repeat visits of a composed `$ref` at the same (type, location) now skip `ref` visitors and the sibling walk, so plugins see exactly one callback per composed ref in both key orders; covered by a new walk test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

RomanHotsiy:
On the comment density note — fair point; trimmed in db869048c: the multi-line explanations in `resolve.ts`, `walk.ts`, and `bundle-visitor.ts` are now single plain sentences, and the `rawNode` cast comment is gone entirely along with the cast. What remains states constraints the code can't show (the shared chain array, the composed-ref revisit skip).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

vadyvas (approved):

PR #2987, opened 2026-07-30
feat: add skipPluginEval option to loadConfig

RomanHotsiy:
## What/Why/How?

Adds a `skipPluginEval` option to `loadConfig` (and `resolveConfig`/`resolvePlugins`) that resolves plugin paths without importing or executing any plugin code.
In some cases the consumer only needs to know *which* plugin files a config references — evaluating arbitrary plugin code there is unwanted.

With `loadConfig({ configPath, skipPluginEval: true })`:

- `config.plugins` contains `{ absolutePath }` references only (plus the built-in default plugin), so no rules/decorators/presets are loaded.
- Relative plugin paths, including the default project plugin (`@theme/plugin.js`), are resolved to absolute paths via `preResolvePluginPath`, which only uses `fs.existsSync`/`require.resolve` — resolution, never execution.
- Duplicate plugin references are deduplicated by absolute path, matching the eval path's behavior.

Known limitation (intentional): `extends` entries that reference a plugin preset (`my-plugin/config-name`) still throw in this mode, because presets live in plugin code that is never loaded. Built-in presets (`recommended`, etc.) resolve as usual. This matches the existing browser-mode behavior where string plugins are not loaded either.

## Testing

Unit tests in `resolve-plugins.test.ts` use fixture plugins whose module body is `throw new Error(...)` — the tests hard-fail if plugin code is ever evaluated. Covered:

- explicitly listed plugin: only its absolute path is returned;
- default project plugin (`@theme/plugin.js`) discovered by convention: path resolved to absolute, code not loaded.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes core config/plugin resolution; skip mode avoids arbitrary code execution but yields incomplete plugin objects and can break `extends` that reference plugin presets, so callers must opt in explicitly.
> 
> **Overview**
> Adds a **`skipPluginEval`** flag on `loadConfig` (threaded through `resolveConfig` and `resolvePlugins`) for callers that only need to know which plugin files a config references.
> 
> When enabled, string plugins are resolved via **`preResolvePluginPath`** (filesystem / `require.resolve` only) and returned as **`{ absolutePath }`** stubs—no `loadPluginModule`, so rules, decorators, and presets from those files are not loaded. The built-in default plugin is still appended, and the usual default project plugin (`@theme/plugin.js`) is discovered and path-resolved the same way. Duplicate entries are deduplicated by absolute path, matching the normal load path.
> 
> Tests use fixtures whose module bodies throw on evaluation to ensure the skip path never runs plugin code (explicit `plugins` entries and convention-based default plugin).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1a58572c4e1d72c6ca9f15b0166a4600c01b125. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/config/load.ts
+++ b/packages/core/src/config/load.ts
@@ -59,9 +59,10 @@ export async function loadConfig(
     configPath?: string;
     customExtends?: string[];
     externalRefResolver?: BaseResolver;
+    skipPluginEval?: boolean;
   } = {}
 ): Promise<Config> {
-  const { configPath = findConfig(), customExtends, externalRefResolver } = options;
+  const { configPath = findConfig(), customExtends, externalRefResolver, skipPluginEval } = options;
 
   const resolver = externalRefResolver ?? new BaseResolver();
 
@@ -78,6 +79,7 @@ export async function loadConfig(
     customExtends,
     configPath,
     externalRefResolver,
+    skipPluginEval,
   });
 
   const ignore = await loadIgnoreConfig(configPath, resolver);
--- a/.changeset/skip-plugin-eval.md
+++ b/.changeset/skip-plugin-eval.md
@@ -1,0 +1,5 @@
+---
+'@redocly/openapi-core': minor
+---
+
+Added a `skipPluginEval` option to `loadConfig` that resolves plugin paths without importing or executing plugin code — the returned plugins contain only their `absolutePath`.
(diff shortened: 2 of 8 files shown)


vadyvas (approved):

packages/core/src/config/load.ts
```diff
@@ -59,9 +59,10 @@ export async function loadConfig(
     configPath?: string;
     customExtends?: string[];
     externalRefResolver?: BaseResolver;
+    skipPluginEval?: boolean;
```
tatomyr, line 62: Is it worth adding a note of this option to the public README?

RomanHotsiy: We can do as a follow-up PR plz.

PR #2994, opened 2026-07-30
fix(drift): match path segments that mix literal text with parameters

ariesclark:
## What/Why/How?

> [!NOTE]
> This was heavily assisted by Anthropic's Opus 5. I've reviewed the code to the best of my ability, but if there's any obvious issues I didn't catch, let me know.

A path template whose segment mixes literal text with parameters never matched a request. `compileOpenApiPath` tested each segment with `/^\{([^}]+)\}$/`, which recognizes only a segment that is entirely one parameter. Anything else, such as two parameters or a parameter beside literal text, fell through to `escapeRegex` and became a literal. So `/instances/{worldId}:{instanceId}` matched only a URL containing that raw text, which no request carries. `drift` reported those requests as undocumented and `coverage` left them out of its figures.

The compiler now scans each segment for `\{([^}]+)\}`, escapes the literal runs between matches, and emits `([^/]+?)` per parameter. Non-greedy is deliberate: with two parameters around a separator, the first must stop at the first occurrence rather than swallow to the last.

Specificity scoring gains a middle tier. A pure literal still scores `+2` and a bare parameter `0`, with `+1` for a segment holding both. The counter reads a per-segment `paramsBefore` snapshot, because `params` accumulates across the whole template.

## Reference

None.

## Testing

`compile-openapi-path.test.ts` covers single-parameter segments, the `/` boundary, two parameters in one segment, a multi-parameter segment with a following suffix, the first-separator split when the trailing value holds another separator, score ordering, and pure literals.

Against a 140-exchange capture, `drift`'s false "undocumented endpoint" reports went from 3 to 0 and `coverage` rose from 87 to 90 operations. `drift` now validates the three newly matched responses.

## Check yourself

## Security

The change only widens which documented paths a recorded request can match, and adds no new input handling.

---

> [!NOTE]
> **Low Risk**
> Scoped to OpenAPI path matching for drift/coverage; only broadens which documented paths match captured traffic, with added tests including ReDoS timing.
> 
> **Overview**
> **`compileOpenApiPath`** no longer treats a path segment as either a single `{param}` or a fixed literal. It now finds every `{param}` in a segment, escapes literal runs, and builds capture groups that honor separators between parameters (including adjacent params and trailing literals like `{name}.json`). That fixes **`drift`** and **`coverage`** missing documented routes such as `/instances/{worldId}:{instanceId}`.
> 
> Specificity **scoring** adds a middle tier (`+1`) for segments that mix literals and parameters, between pure literals (`+2`) and bare-parameter segments (`0`).
> 
> A new **`compile-openapi-path.test.ts`** suite covers multi-parameter segments, separator splitting, score ordering, suffix segments, and a performance guard against regex blow-up on long non-matching paths. A patch **changeset** documents the fix for `@redocly/cli`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7ec3042a8447d85bb3f18d126be8e2a3d31e939. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/drift-multi-parameter-path-segments.md
+++ b/.changeset/drift-multi-parameter-path-segments.md
@@ -1,0 +1,7 @@
+---
+'@redocly/cli': patch
+---
+
+Fixed `drift` and `coverage` failing to match a path template whose segment mixes literal text with parameters, such as `/instances/{worldId}:{instanceId}`.
+Only a segment that was entirely one parameter was recognized, so these templates were compiled as literal text and never matched any request.
+Affected requests were reported as undocumented by `drift` and left out of the `coverage` figures.
--- a/.changeset/drift-multi-parameter-path-segments.md
+++ b/.changeset/drift-multi-parameter-path-segments.md
@@ -3,5 +3,3 @@
 ---
 
 Fixed `drift` and `coverage` failing to match a path template whose segment mixes literal text with parameters, such as `/instances/{worldId}:{instanceId}`.
-Only a segment that was entirely one parameter was recognized, so these templates were compiled as literal text and never matched any request.
-Affected requests were reported as undocumented by `drift` and left out of the `coverage` figures.
--- a/packages/cli/src/__tests__/commands/drift/compile-openapi-path.test.ts
+++ b/packages/cli/src/__tests__/commands/drift/compile-openapi-path.test.ts
@@ -1,0 +1,58 @@
+import { compileOpenApiPath } from '../../../commands/drift/utils/http.js';
+
+describe('compileOpenApiPath', () => {
+  it('compiles a segment that is a single parameter', () => {
+    const { regex, params } = compileOpenApiPath('/users/{userId}');
+
+    expect(params).toEqual(['userId']);
+    expect(regex.exec('/users/usr_abc')?.[1]).toBe('usr_abc');
+  });
+
+  it('does not let a parameter span a path separator', () => {
+    const { regex } = compileOpenApiPath('/users/{userId}');
+
+    expect(regex.exec('/users/usr_abc/friends')).toBeNull();
+  });
+
+  it('compiles two parameters separated by a literal inside one segment', () => {
+    const { regex, params } = compileOpenApiPath('/instances/{worldId}:{instanceId}');
+
+    expect(params).toEqual(['worldId', 'instanceId']);
+
+    const match = regex.exec(
+      '/instances/wrld_a:85981~group(grp_b)~groupAccessType(public)~region(us)'
+    );
+    expect(match?.[1]).toBe('wrld_a');
+    expect(match?.[2]).toBe('85981~group(grp_b)~groupAccessType(public)~region(us)');
+  });
+
+  it('splits on the first separator when the trailing value holds another', () => {
+    const { regex } = compileOpenApiPath('/instances/{worldId}:{instanceId}');
+    const match = regex.exec('/instances/wrld_a:12345~region(us):extra');
+
+    expect(match?.[1]).toBe('wrld_a');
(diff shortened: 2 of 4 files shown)


packages/cli/src/__tests__/commands/drift/compile-openapi-path.test.ts
```diff
@@ -0,0 +1,58 @@
+import { compileOpenApiPath } from '../../../commands/drift/utils/http.js';
```
tatomyr: Please move the test inside the commands/drift folder.

ariesclark: Good catch, done.

packages/cli/src/commands/drift/__tests__/compile-openapi-path.test.ts
```diff
@@ -0,0 +1,58 @@
+import { compileOpenApiPath } from '../utils/http.js';
+
+describe('compileOpenApiPath', () => {
```
sobanieca-redocly, line 3: Many thanks for the fix — the matching logic is correct, but the generated regexes need a performance improvement before this can merge.

Each parameter compiles to a lazy ([^/]+?) group, so when a segment holds several parameters, the regex engine tries every possible way to split a non-matching value between them — super-linear backtracking. Measured on this branch, a single exec call:

- {a}:{b}:{c}.json against a 5 KB non-matching path: ~2 s
- {a}:{b}:{c}:{d}:{e}.json against the same input: >120 s (had to kill the process)

Since exec is synchronous, this blocks the event loop — drift just hangs, no error, no timeout. Captures routinely contain long junk paths (scanner traffic, malformed URLs), so this is reachable in normal use, and main isn't affected today because it never emits more than one group per segment.

Could you rework the compilation so matching stays linear no matter how many parameters a segment holds? Here's a test that pins the requirement — it currently fails at ~2,000 ms:

```
it('rejects a long non-matching value without backtracking blow-up', () => {
  const { regex } = compileOpenApiPath('/v1/{a}:{b}:{c}.json');
  const longPath = '/v1/' + 'a:'.repeat(2500) + 'a';

  const startedAt = performance.now();
  expect(regex.exec(longPath)).toBeNull();
  expect(performance.now() - startedAt).toBeLessThan(100);
});
```

Please also add this test which ensures that we don't introduce solution that may break trailing literal handling:

```
it('lets a parameter before a trailing literal contain the literal separator char', () => {
  const { regex } = compileOpenApiPath('/files/{name}.json');

  expect(regex.exec('/files/report.v1.json')?.[1]).toBe('report.v1');
});
```

Matching behavior shouldn't change — e.g. /files/{name}.json must still match /files/report.v1.json with name = report.v1, so simply excluding the separator characters from the parameter's character class isn't enough.

sobanieca-redocly (approved):

.changeset/drift-multi-parameter-path-segments.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/cli': patch
+---
+
+Fixed `drift` and `coverage` failing to match a path template whose segment mixes literal text with parameters, such as `/instances/{worldId}:{instanceId}`.
+Only a segment that was entirely one parameter was recognized, so these templates were compiled as literal text and never matched any request.
+Affected requests were reported as undocumented by `drift` and left out of the `coverage` figures.
```
JLekawa:
```suggestion
```

JLekawa (approved):

tatomyr:
Thank you @ariesclark for the contribution! Your changes will land in the next release.

PR #3002, opened 2026-07-31
feat(client-generator): expose response headers via opt-in throw-mode envelope

### LINKED ISSUE #3001: feat(client-generator): expose response headers on successful calls (e.g. pagination totals)
## Summary

`generate-client` throw mode returns only the parsed success body. APIs that put important metadata in **response headers** (pagination totals, rate limits, etc.) cannot read those headers without switching to `errorMode: 'result'`.

We would like a first-class, throw-mode-friendly way to access the `Response` (and ideally typed header values from the OpenAPI description) on successful calls.

Happy to open a PR once there is agreement on the shape.

## Motivation

Many collection APIs (including Rebilly core API) paginate with `limit` / `offset` query params, return a **JSON array** body, and put totals in headers, for example:

- `Pagination-Total`
- `Pagination-Limit`
- `Pagination-Offset`

Those headers are already documented on the operation’s `200` response in OpenAPI.

We use `Pagination-Total` in product UI (e.g. tab badges like “Customers (128)”, “Transactions (6.7k)”). Offset auto-pagination via `client.pagination` works well for iterating items; the gap is **reading the total on a normal list call** in throw mode.

### What works today

| Mode | Success return | Headers available? |
| --- | --- | --- |
| `errorMode: 'throw'` (default) | parsed body only | No |
| `errorMode: 'result'` | `{ data, error, response }` | Yes — `response.headers` |
| `link` pagination | (internal) | Runtime already reads `Link` |

`errorMode: 'result'` is a workable workaround for internal apps, but it changes every call site and is a poor default for a published SDK that otherwise wants throw-mode DX.

## Proposed solutions

Two options; either would unblock us. **A** is smaller; **B** is nicer long-term for typed SDKs.

### Option A — Opt-in `includeResponse` on the call

Add a per-call flag on `RequestOptions` (name flexible):

```ts
await client.listCustomers(
  { params: { limit: 1, offset: 0 } },
  { includeResponse: true }
);
// → { data: Customer[], response: Response }
```

**Behavior**

- Default unchanged: throw mode still returns `TData` only.
- When `includeResponse: true`, throw mode returns `{ data: TData; response: Response }` (and still throws `ApiError` on non-2xx).
- Result mode can ignore the flag (it already has `response`) or treat it as a no-op.

**Pros**

- Small, non-breaking.
- Same escape hatch for any header (pagination, rate limits, `Location`, …).
- Mirrors how callers already think about `fetch`.

**Cons**

- Header names/types stay untyped (`headers.get('Pagination-Total')`).
- Return type becomes a union depending on the flag (needs careful typing, e.g. overloads).

### Option B — Envelope return with typed response headers

For operations that declare response headers in OpenAPI, support an envelope (generate-time and/or per-call) that includes a **typed** header object derived from `responses.*.headers`:

```ts
const { data, headers } = await client.listCustomers(
  { params: { limit: 1 } },
  { envelope: true } // or a client-level / generate-time knob
);

headers.paginationTotal; // number | undefined — from Pagination-Total schema
data;                   // Customer[]
```

**Behavior (sketch)**

- Generate a per-operation (or shared) headers type from documented success-response headers (normalize names to camelCase in TS).
- Envelope shape roughly: `{ data: TData; headers: THeaders; response?: Response }`.
- Knobs could be:
  - per-call (`envelope: true`), and/or
  - generate-time (`client.returnStyle: 'body' | 'envelope'`) for publishers who want envelope by default.

**Pros**

- Best DX for SDKs: totals and similar metadata are typed and discoverable.
- Aligns with OpenAPI as source of truth (we already document these headers).
- Still non-breaking if default remains “body only.”

**Cons**

- More design/implementation surface (naming, which status’ headers, optional vs required headers, multi-status).
- Larger PR than A.

## Out of scope (for this issue)

- Changing offset pagination to stop early using `Pagination-Total` (avoiding a final empty page). Nice later; not required for our UI use case.
- Nested facade / resource-style clients.

## Prior art in this package

- Result mode already threads `response` through on every call.
- Link-style pagination already inspects response headers (`Link`) inside the runtime.
- An opt-in success envelope would reuse the same `Response` the runtime already has in `execute` before discarding it in throw mode.

## Ask

1. Preference between **A** (`includeResponse`) and **B** (typed header envelope), or a sequenced “A now, B later”?
2. If you are open to a contribution, we can prepare a PR against `@redocly/client-generator` (runtime + emitter typings + e2e) once the shape is agreed.

## Environment

- `@redocly/cli` / `@redocly/client-generator` 2.43.x / 0.2.x (experimental `generate-client`)
- OpenAPI 3.1 description with collection `200` responses that document `Pagination-*` headers and a JSON array body

adamaltman:
## What/Why/How?

**What:** Generated throw-mode clients can now return declared response headers. Passing `{ envelope: true }` on a call returns `{ data, headers, response }` instead of the parsed body alone.

**Why:** Throw mode (the default) returns only the body, so OpenAPI-declared success-response headers — `Pagination-Total`, rate limits, `Location` — were unreachable without switching the whole client to `--error-mode result`.

**How:**
- **IR:** operations carry `successResponseHeaders` built from the selected success response (including bodyless responses like `201` + `Location`); referenced scalar header schemas are dereferenced.
- **Runtime:** `execute` returns the envelope when `envelope === true`; header values are coerced to `number`/`boolean` per schema (blank/unparsable values are omitted, never `0`). Pagination iterators strip the flag so page pointers keep resolving against the body.
- **Types:** the `EnvelopeResult` conditional type keeps default call sites body-only (non-breaking) and narrows to `Envelope<…>` only for a literal `envelope: true`; a statically unknowable flag widens to the union. Result-mode clients ignore the option (they already return `response`).
- **Emitters:** header names become identifier-safe camelCase keys (`Pagination-Total` → `paginationTotal`, `3D-Secure` → `_3dSecure`) with deterministic collision suffixes; `<Op>ResponseHeaders` aliases are emitted; required headers are required properties (the type trusts the description, like body types).
- **Wrappers:** TanStack Query and SWR inits are `Omit<RequestOptions, 'envelope'>` and the forwarded calls strip the flag at runtime, so cached data is always the plain body. This also fixes a latent bug where grouped-style no-input SWR hooks passed `init` in the args slot.

## Reference

Closes #3001

## Testing

- New unit coverage across IR build, descriptor/alias/wrapper emitters, runtime coercion, and `EnvelopeResult` typing (993 client-generator tests).
- New e2e suite (`envelope.test.ts`) generates a client from a response-headers fixture and strict-tsc-checks envelope and default call sites; the TanStack e2e pins `useQuery(...).data` to the plain body type.
- Full suite: compile, typecheck, unit, e2e, lint, format all pass locally (one pre-existing `generator-contract` e2e failure in the local environment only; passes on CI from main).

## Check yourself

## Security

Made with [Cursor](https://cursor.com)


docs/@v2/guides/use-generated-client.md
```diff
@@ -282,6 +282,32 @@ It changes the runtime reader only, not the static return type.
 
 An operation whose success response declares no content is typed `void`, but if the server sends a JSON body anyway (a gap in the API description), the runtime still parses and returns it rather than silently dropping real data — reach it with a cast while the description catches up.
```
JLekawa:
```suggestion
An operation whose success response declares no content is typed `void`.
However, if the server sends a JSON body anyway (a gap in the API description), the runtime still parses and returns it rather than silently dropping real data.
Reach it with a cast while the description catches up.
```

docs/@v2/guides/use-generated-client.md
```diff
@@ -282,6 +282,32 @@ It changes the runtime reader only, not the static return type.
...
+headers.paginationTotal; // number — required Pagination-Total in the description
+headers.xFlag; // boolean | undefined — optional X-Flag
+response.headers.get('X-Undocumented'); // anything not declared in OpenAPI
+
+// Grouped args / instance client: trailing init is always separate.
+const envelope = await client.listCustomers({ params: { limit: 1 } }, { envelope: true });
+```
+
+- `headers` is a safe camelCase object of headers declared on the operation's success response.
+  String, number, and boolean schemas drive the TypeScript type and number/boolean coercion; complex header schemas remain strings because HTTP exposes header values as text.
```
JLekawa:
```suggestion
  String, number, and boolean schemas drive the TypeScript type and number/boolean coercion.
  Complex header schemas remain strings because HTTP exposes header values as text.
```

docs/@v2/guides/use-generated-client.md
```diff
@@ -282,6 +282,32 @@ It changes the runtime reader only, not the static return type.
...
+  Required response headers are required properties — the type trusts the API description, the same way response body types do; colliding normalized names get a deterministic numeric suffix.
```
JLekawa:
```suggestion
  Required response headers are required properties — the type trusts the API description, the same way response body types do.
  Colliding normalized names get a deterministic numeric suffix.
```

docs/@v2/guides/use-generated-client.md
```diff
@@ -282,6 +282,32 @@ It changes the runtime reader only, not the static return type.
...
+- `response` is the raw `Response` — use it for undocumented headers.
+- Non-2xx responses still throw `ApiError`.
+- Default call sites stay body-only (non-breaking), including calls that pass other options (`headers`, `signal`, `parseAs`, a retry override).
+- In `--error-mode result` the flag is ignored; that mode already returns `response`.
+- The TanStack Query and SWR wrappers don't accept `envelope` — it's excluded from their options and stripped from the forwarded call, so cached data is always the plain body; call the sdk function directly when you need headers.
```
JLekawa:
```suggestion
- The TanStack Query and SWR wrappers don't accept `envelope`.
It's excluded from their options and stripped from the forwarded call, so cached data is always the plain body.
Call the sdk function directly when you need headers.
```

JLekawa (approved):

Marshevskyy (approved):
Good job

.changeset/client-envelope-response-headers.md
```diff
@@ -0,0 +1,7 @@
+---
+'@redocly/client-generator': minor
+---
+
+Added an opt-in success envelope to throw-mode calls: pass `{ envelope: true }` to get `{ data, headers, response }` instead of the body alone.
```
RomanHotsiy: I think this changelog entry is too detailed

packages/client-generator/src/runtime/create-client.ts
```diff
@@ -210,6 +211,38 @@ async function prepareRequest(
...
+  type: ResponseHeaderSpec['type']
+): string | number | boolean | undefined {
+  if (raw === null) return undefined;
+  if (type === 'number') {
+    if (raw.trim() === '') return undefined;
+    const value = Number(raw);
+    return Number.isFinite(value) ? value : undefined;
+  }
+  if (type === 'boolean') {
+    if (raw === 'true') return true;
```
DmitryAnansky: @adamaltman 
Should we also handle cases when raw response contain `True/TRUE/False/FALSE`?

DmitryAnansky: Should we try to do coercion ?

```
if (type === 'boolean') {
  const value = raw.trim().toLowerCase();
  if (value === 'true') return true;
  if (value === 'false') return false;
  return undefined;
}
```

packages/client-generator/src/emitters/response-headers.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const { factory } = ts;
+
+type PlannedResponseHeader = ResponseHeaderModel & {
+  key: string;
+  type: ResponseHeaderSpec['type'];
+};
+
+/** Runtime coerce hint from a header schema (complex schemas fall back to string). */
+export function headerParseType(schema: SchemaModel): ResponseHeaderSpec['type'] {
```
DmitryAnansky:
```suggestion
export function headerParseType(schema: SchemaModel): ResponseHeaderSpec['type'] {
   if (schema.kind === 'union') {
      const members = schema.members.filter((member) => member.kind !== 'null');
      if (members.length === 1) return headerParseType(members[0]);
      return 'string';
    }
```

.changeset/client-envelope-response-headers.md
adamaltman:
```suggestion
Added an opt-in success envelope to throw-mode calls.
```
How about this?

adamaltman: Shortened to just that one-liner in faad4664.

packages/client-generator/src/runtime/create-client.ts
adamaltman: Good call — done in faad4664. Boolean header values are now trimmed and lowercased before matching (`TRUE` / ` False ` → `true` / `false`).

packages/client-generator/src/emitters/response-headers.ts
adamaltman: Accepted in faad4664. Nullable wrappers (`boolean | null`, OpenAPI 3.0 `nullable`) now unwrap to the inner coerce type instead of falling through to `string`.

packages/client-generator/src/runtime/types.ts
```diff
@@ -283,6 +318,53 @@ type Paginated<Entry extends OpsShape[string]> = 'item' extends keyof Entry
...
+  TInit extends RequestOptions | undefined,
+> = TInit extends undefined
+  ? TData
+  : 'envelope' extends keyof TInit
+    ? [TInit['envelope' & keyof TInit]] extends [true]
+      ? Envelope<TData, THeaders>
+      : [TInit['envelope' & keyof TInit]] extends [false | undefined]
+        ? TData
+        : TData | Envelope<TData, THeaders>
+    : TData;
```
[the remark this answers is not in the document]
adamaltman: Confirmed and fixed in the latest commit.

`EnvelopeResult<T, RequestOptions>` was widening to `T | Envelope<T>`. Pre-envelope package-mode flat sugar annotates every `init` as `RequestOptions`, so after a runtime upgrade without regenerating, *every* call site became the union — a real break of the non-breaking claim.

Exact `RequestOptions` now resolves to the plain body again. Literal `{ envelope: true }` still narrows to the envelope; a widened `{ envelope: boolean }` still unions. To get the envelope type from a variable, narrow it (`as const` / `{ envelope: true as const }`).

packages/client-generator/src/emitters/response-headers.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (schema.kind === 'literal') {
+    if (typeof schema.value === 'number') return 'number';
+    if (typeof schema.value === 'boolean') return 'boolean';
+  }
+  if (schema.kind === 'enum') {
+    if (schema.scalar === 'integer' || schema.scalar === 'number') return 'number';
+    if (schema.scalar === 'boolean') return 'boolean';
+  }
+  return 'string';
+}
```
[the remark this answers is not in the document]
adamaltman: Already fixed in 1e53ca95 (before the latest Bugbot pass). `headerParseType` now resolves `kind: 'ref'` through named schemas and peels metadata-only `allOf` intersections, so integer/boolean component schemas keep number/boolean coercion.

DmitryAnansky (approved):

PR #3021, opened 2026-08-05
feat: add vendor extensions metric with per-extension counts

n0rahh:
## What/Why/How?
Adds a **Vendor Extensions** metric to the `stats` command: how many distinct `x-` extensions a document uses and how often each one occurs, shown in the `stylish`, `json`, and `markdown` output, for OpenAPI and AsyncAPI.

The stats rule counts through a `SpecExtension` visitor plus one hook per extension that is declared with its own type (`x-codeSamples`, `x-webhooks`, `x-query`, `x-logo`, `x-tagGroups`, `x-servers`, `x-examples`, `x-enumDescriptions`, `x-usePkce`). A completeness test derives the required hooks from the type definitions, so a future declared extension without a hook fails the suite. The walker and type changes that enable it:

- `extensionsPrefix: 'x-'` is now declared on every object type whose spec allows extensions (audited against Swagger 2.0, OAS 3.0/3.1/3.2, and AsyncAPI 2.6/3.0 spec texts; includes a new oas2 `Scopes` type — the one map the spec allows `^x-` in).
- An `x-` key never falls into `additionalProperties`: extensions are not additional properties, so undeclared ones dispatch as `SpecExtension` even inside catch-all maps (OAS callbacks, oas2 scopes, AsyncAPI `Message`).
- Extensions declared with a named type keep their typed walk and lint validation and are counted at their own typed hooks; extensions declared with a plain schema (`x-hideTryItPanel` and similar) have no typed walk, so they are now visited as `SpecExtension` instead of being skipped.
- Primitive nodes dedupe by location instead of by value — previously `x-internal: true` on many operations was visited only once (guarded by the walk test "should visit every occurrence of extensions with equal scalar values").
- Fixed a pre-existing double walk of declared `x-` properties: a declared extension key (`x-query`) is already in the type's property list, and the extensions scan pushed it a second time, so its subtree was walked twice and nested visitors fired twice — inflating the `operations` and `webhooks` metrics. Now deduped with `!props.includes(k)`, matching the `additionalProperties` branch above it (guarded by the walk test "should walk a declared extension subtree once for nested visitors").
- Keys written next to a `$ref` are **not** counted — the Reference Object spec says extra properties are ignored (pinned as a negative case in the counting test).

**Note**: two intentional behavior changes beyond stats. `SpecExtension` visitors and configurable rules targeting `SpecExtension` now fire for every occurrence of an untyped extension (and for declared plain-schema extensions), and `$ref`s inside AsyncAPI extensions are now resolved, linted, and bundled the same way as in OpenAPI. As a side effect of the dispatch fix, `x-` keys inside callbacks and oas2 scopes no longer produce false-positive struct errors (previously reported as "Expected type `PathItem` (object) but got `boolean`").

Also fixed the `stats` command always reporting `Parameters: 0` for AsyncAPI 2.x and 3.x — channel parameters are keyed by name rather than carrying a `name` property, so none were counted.
## Testing
Covered with unit and e2e tests.
Published snapshot and tested `cli` commands in terminal.

## Screenshots (optional)
image
image
image

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes touch shared document walking and ref resolution used by lint and bundling, not only stats; behavior shifts for extension visitation and metrics may affect custom rules and CI snapshots.
> 
> **Overview**
> Adds a **Vendor Extensions** metric to `redocly stats` for OpenAPI and AsyncAPI: distinct `x-` extension names as the total, with per-name occurrence counts in stylish, JSON (`counts`), and Markdown (breakdown table) output.
> 
> Core **stats** logic gains `xExtensions` accumulation via `SpecExtension` and typed-extension hooks, shared `finalizeStats` for set-based metrics, and skips counting extensions on `$ref` siblings. **AsyncAPI 2/3** parameter totals now use channel parameter map keys instead of a missing `name` field.
> 
> **Walker** and **ref resolution** treat `x-` keys separately from `additionalProperties`, dispatch undeclared extensions as `SpecExtension`, dedupe declared extension keys to avoid double walks, and dedupe primitive extension visits by location. Type maps gain broader `extensionsPrefix: 'x-'` coverage (including oas2 `Scopes`, callbacks, bindings maps).
> 
> Beyond stats, `SpecExtension` rules see more occurrences; extension `$ref`s resolve more consistently; some false struct errors on extensions in callbacks/scopes go away. Docs, changesets, unit/e2e tests, and snapshots cover the new metric and parameter fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b775d9cf9e6571f8198364334321f0500bccccd5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/seven-waves-create.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Add a Vendor Extensions metric to the `stats` command that reports how many distinct `x-` extensions a spec uses and how often each one occurs.
```
JLekawa:
```suggestion
Added a Vendor Extensions metric to the `stats` command that reports how many distinct `x-` extensions a descrtption file uses and how often each one occurs.
```

JLekawa (approved):

packages/core/src/rules/other/spec-extensions.ts
```diff
@@ -0,0 +1,122 @@
...
+const MASKED = '<masked>';
+
+// Keys that suggest a credential or personal data — their values are never sampled.
+const SENSITIVE_KEY_REGEX = /key|token|secret|password|credential|session|bearer|email|auth/i;
+// Value shapes masked regardless of the key: opaque token-like blobs, emails, URLs with a scheme.
+const TOKEN_LIKE_REGEX = /^(?=.*\d)[A-Za-z0-9+/=_-]{16,}$/;
+const EMAIL_REGEX = /\S@\S+\.\S/;
+const URL_SCHEME_REGEX = /:\/\//;
+
+export const StatsSpecExtensions = (accumulator: SpecVendorExtensionsAccumulator) => {
```
DmitryAnansky: @n0rahh 
Why did you decide to create a separate rule to collect extensions instead of modifying the existing one?

DmitryAnansky: This duplicates the existing codebase and only adds one additional piece of functionality (collecting x-extensions).
Could you please consider extending the existing stats rule instead of introducing a separate one?

n0rahh: @DmitryAnansky The existing stats rules are per-spec (StatsOAS, StatsAsync2, StatsAsync3), so putting extension collection there would mean duplicating it in all three — as a separate spec-agnostic visitor it's written once and just composed with whichever stats rule runs. But if you prefer to put it into existing rules - let me know, I can change it

DmitryAnansky: Stats command is spec oriented, as might content different data, but why would extensions be mixed in this case?
I guess you can reuse existing accumulator and reuse code and not duplicate for all 3 specs.

cc: @tatomyr

.changeset/seven-waves-create.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': minor
+'@redocly/cli': minor
+---
+
+Added a Vendor Extensions metric to the `stats` command that reports how many distinct `x-` extensions a descrtption file uses and how often each one occurs.
```
DmitryAnansky:
```suggestion
Added a Vendor Extensions metric to the `stats` command that reports how many distinct `x-` extensions a description file uses and how often each one occurs.
```

docs/@v2/commands/stats.md
DmitryAnansky, line 121: Should we update docs example as well?

DmitryAnansky, line 121: Should we update docs example as well?

tests/e2e/stats/stats-async2-json/snapshot.txt
```diff
@@ -26,6 +26,11 @@
   "tags": {
     "metric": "🔖 Tags",
     "total": 2
+  },
+  "xExtensions": {
+    "metric": "🧩 Vendor Extensions",
+    "total": 0,
+    "counts": {}
```
DmitryAnansky, line 33: Could you please check if it is expected? ` "counts": {}`

n0rahh: Yes — that fixture has no x- keys at all, so total: 0 / counts: {} is the correct output

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,119 @@
...
+  addBounded((props[prop] ??= new Set()), value);
+}
+
+function addBounded(set: Set<string>, value: string) {
+  if (set.has(value) || set.has(TRUNCATED)) return;
+  set.add(set.size >= MAX_VALUES_PER_PROP ? TRUNCATED : value);
+}
+
+export function applySpecExtensionsStats(
+  accumulator: SpecVendorExtensionsAccumulator,
```
tatomyr:
```suggestion
  collectedExtensions: SpecVendorExtensionsAccumulator,
```

?

packages/core/src/rules/other/stats.ts
```diff
@@ -8,17 +12,27 @@ import type {
...
 } from '../../typings/openapi.js';
 import type { Oas2Parameter } from '../../typings/swagger.js';
+import { applySpecExtensionsStats, collectSpecExtensions } from '../../utils/spec-extensions.js';
+import type { UserContext } from '../../walk.js';
 
 export const StatsOAS = (statsAccumulator: OASStatsAccumulator) => {
+  const extensions: SpecVendorExtensionsAccumulator = {};
+
   return {
+    any: {
```
tatomyr: Have you tried this?

```suggestion
    SpecExtension: {
```

This would be a single entrypoint. 
If it doesn't help, I'd like to make it work first.

packages/core/src/typings/common.ts
```diff
@@ -3,6 +3,8 @@ export interface StatsRow {
   total: number;
   color: 'red' | 'yellow' | 'green' | 'white' | 'magenta' | 'cyan';
   items?: Set<string>;
+  counts?: Record<string, number>;
+  details?: SpecVendorExtensionsAccumulator;
```
tatomyr: Can we just put everything into `details` and use what we need? Why creating 2 fields for the same purpose. If you need separate fields, you can put them inside the `details` placeholder, like so:

```
details:
  extensionCounts: ...
  someExtraStats: ...
```

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function recordExtensions(
+  accumulator: SpecVendorExtensionsAccumulator,
+  node: unknown,
+  type: NormalizedNodeType
+) {
+  if (!isPlainObject(node)) return;
+  for (const [key, value] of Object.entries(node)) {
+    if (!key.startsWith(EXTENSION_PREFIX)) continue;
```
tatomyr: This looks wrong since it will collect any `x-` properties defined in Schemas which are not  extensions.

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (isMapEntryKey(type, key, value)) continue;
+    recordExtension(accumulator, key, value);
+  }
+}
+
+// An x- key is not an extension when the type resolves it to a named map entry (a schema name, a channel address).
+function isMapEntryKey(type: NormalizedNodeType, key: string, value: unknown): boolean {
+  if (getOwn(type.properties, key) !== undefined) return false;
+  const entryType =
+    typeof type.additionalProperties === 'function'
```
tatomyr: This does the walker's job. Let's leverage the visitor pattern to get to the node you need.

packages/core/src/utils/__tests__/spec-extensions.test.ts
```diff
@@ -0,0 +1,494 @@
...
+    schemas: { metric: 'Schemas', total: 0, color: 'white' },
+    parameters: { metric: 'Parameters', total: 0, color: 'yellow', items: new Set() },
+    channels: { metric: 'Channels', total: 0, color: 'green' },
+    operations: { metric: 'Operations', total: 0, color: 'yellow' },
+    tags: { metric: 'Tags', total: 0, color: 'white', items: new Set() },
+    xExtensions: { metric: 'Vendor Extensions', total: 0, color: 'cyan' },
+  };
+}
+
+async function collect(yaml: string): Promise<SpecVendorExtensionsAccumulator> {
```
tatomyr: This is too complicated as a test helper. If you cannot setup unit tests in a simpler manner, please use e2e tests instead.

adamaltman:
I like this idea

adamaltman:
I found things I didn't know existed in there. Thank you.

```
Document: ./website/api-definitions/openapi/openapi.yaml stats:

🚗 References: 972
📦 External Documents: 0
📈 Schemas: 838
👉 Parameters: 61
🔗 Links: 0
🔀 Path Items: 382
🎣 Webhooks: 148
👷 Operations: 613
🔖 Tags: 88
🧩 Vendor Extensions: 16
  - x-badges: 151
  - x-basic: 92
  - x-codeSamples: 261
  - x-currency-field: 16
  - x-enumDescriptions: 79
  - x-is-free-form: 1
  - x-label: 26
  - x-logo: 1
  - x-multiline: 31
  - x-products: 761
  - x-sdk-operation-name: 613
  - x-sdk-resource: 43
  - x-sdk-skip: 5
  - x-sortable: 155
  - x-tagGroups: 1
  - x-type: 138

./website/api-definitions/openapi/openapi.yaml: stats processed in 143ms
```

docs/@v2/commands/stats.md
```diff
@@ -123,6 +127,11 @@ Document: museum.yaml stats:
 🎣 Webhooks: 0
 👷 Operations: 8
 🔖 Tags: 3
+🧩 Vendor Extensions: 4
+  - x-badges: 1
```
RomanHotsiy: one more space is missing here

packages/cli/src/commands/stats/visitor-and-accumulator-resolver.ts
```diff
@@ -20,6 +20,7 @@ export function resolveStatsVisitorAndAccumulator(specVersion: SpecVersion) {
     webhooks: { metric: '🎣 Webhooks', total: 0, color: 'green' },
     operations: { metric: '👷 Operations', total: 0, color: 'yellow' },
     tags: { metric: '🔖 Tags', total: 0, color: 'white', items: new Set() },
+    xExtensions: { metric: '🧩 Vendor Extensions', total: 0, color: 'cyan' },
```
tatomyr: Just saw we have `items` inside the `tags` accumulator (one line above). What if we arrange extensions similarly to tags? Is there something else we are going to show in the output except of the extension items and their count?

packages/core/src/rules/other/stats.ts
```diff
@@ -43,6 +78,13 @@ export const StatsOAS = (statsAccumulator: OASStatsAccumulator) => {
         },
       },
     },
+    Operation: {
+      enter(operation: Oas3Operation, ctx: UserContext) {
+        if (ctx.key === 'x-query') {
+          collectSpecExtension(extensions, 'x-query', operation);
```
tatomyr: ~Why do we need to treat `x-query` differently? Isn't it already a `SpecExtension`?~

I see, it has its own type.

tests/e2e/stats/stats-extensions/openapi.yaml
```diff
@@ -0,0 +1,81 @@
...
+      name: p
+      in: query
+      x-hideReplay: true
+      schema:
+        type: string
+  responses:
+    SharedResponse:
+      description: ok
+  schemas:
+    x-MySchema:
```
tatomyr, line 83: This one (as well as `x-trace-id`) shouldn't be counted as an extension, right?

packages/core/src/walk.ts
```diff
@@ -273,7 +273,8 @@ export function walkDocument<T extends BaseVisitor>(opts: {
       };
 
       currentLocation = resolvedLocation;
-      const isNodeSeen = seenNodesPerType[type.name]?.has?.(resolvedNode);
+      const seenKey = type === SpecExtension ? location.absolutePointer : resolvedNode;
```
tatomyr: What is this change for?

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,95 @@
...
+const URL_SCHEME_REGEX = /:\/\//;
+
+// Kept typed so other metrics still traverse their subtrees; the stats visitors count them explicitly.
+const STRUCTURAL_EXTENSIONS = new Set(['x-webhooks', 'x-query']);
+
+// Makes the walker dispatch every x- key as SpecExtension, including natively-typed ones (x-codeSamples and others).
+export function ensureSpecExtensionDispatch(types: Record<string, NormalizedNodeType>) {
+  for (const type of Object.values(types)) {
+    if (type === SpecExtension) continue;
+    type.extensionsPrefix ??= EXTENSION_PREFIX;
```
tatomyr: Not sure this is still needed after using the SpecExtension visitor.

packages/core/src/rules/other/stats.ts
```diff
@@ -32,6 +62,11 @@ export const StatsOAS = (statsAccumulator: OASStatsAccumulator) => {
       },
     },
     WebhooksMap: {
+      enter(node: unknown, ctx: UserContext) {
```
tatomyr: We also have other known extensions, like the following:

```ts
    'x-servers': 'XServerList',
    'x-tagGroups': 'TagGroups',
    'x-ignoredHeaderParameters': { type: 'array', items: { type: 'string' } },
```
Have you covered them?

packages/core/src/rules/other/stats.ts
```diff
@@ -8,9 +13,34 @@ import type {
...
+  statsAccumulator: StatsAccumulator,
+  extensions: SpecVendorExtensionsAccumulator
+) {
+  for (const row of Object.values(statsAccumulator)) {
+    if (row.items) {
+      row.total = row.items.size;
+    }
+  }
+  const extensionNames = Object.keys(extensions).sort();
+  statsAccumulator.xExtensions.total = extensionNames.length;
```
tatomyr: Why not use the same approach with Set like in other rows?

tatomyr: Moreover, maybe we can simply calculate the items' size in place instead of assigning the total?

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,95 @@
...
+    for (const propName of Object.keys(type.properties)) {
+      if (propName.startsWith(EXTENSION_PREFIX) && !STRUCTURAL_EXTENSIONS.has(propName)) {
+        delete type.properties[propName];
+      }
+    }
+    const entryType = type.additionalProperties;
+    // An untyped catch-all (`additionalProperties: {}`) swallows x- keys before the extensions fallback.
+    if (isPlainObject(entryType) && !isNamedType(entryType) && entryType.type === undefined) {
+      type.additionalProperties = (_value, key: string) =>
```
tatomyr: I don't think it's a good idea to modify types. What are you're trying to achieve?

.changeset/olive-donkeys-shave.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Fixed the `stats` command always reporting `Parameters: 0` for AsyncAPI 2.x and 3.x descriptions.
```
tatomyr:
```suggestion
Fixed the `stats` command reporting wrong parameter count for AsyncAPI descriptions.
```

tests/e2e/stats/stats-extensions/openapi.yaml
n0rahh: Yes, it just counts as a schema name

packages/core/src/walk.ts
n0rahh: SpecExtension nodes are often equal scalars (true, "internal"), and deduping them by node value made the walker visit only the first occurrence document-wide — deduping by location visits every occurrence exactly once (while a $ref-shared extension still counts once), which the stats need to count extensions correctly.

packages/core/src/utils/spec-extensions.ts
n0rahh: It is needed: the walker only dispatches an x- key to SpecExtension when the type declares extensionsPrefix, and most AsyncAPI types (everything except SecurityScheme) plus Paths in OAS never declare it. Removing this line drops the AsyncAPI fixture's extension count — only the additionalProperties: {} path survives. 

Adding extensionsPrefix to the AsyncAPI type definitions themselves would make extension values walked (and their $refs bundled/linted) for every consumer, which is a behavior change beyond stats — could be a follow-up if we want it.

packages/core/src/walk.ts
vadyvas: @n0rahh  
why scope the fix to SpecExtension only? can other primitive nodes hit the same problem? 
If they can, using resolvedLocation.absolutePointer as a key instead of checking type === SpecExtension would cover them all in one place

packages/core/src/rules/other/stats.ts
n0rahh: Yes, they are covered. Added e2e tests to make it clear

packages/core/src/utils/spec-extensions.ts
n0rahh: This modifies only the stats command's own copy of the types — lint keeps the original ones where the declarations matter for validation. Without it, typed extensions (x-codeSamples, x-logo) and most AsyncAPI nodes never reach the SpecExtension visitor, since the walker dispatches them to their declared types instead. The only alternative is going back to key-scanning in the any hook — the approach you wanted to replace with SpecExtension

packages/cli/src/commands/stats/visitor-and-accumulator-resolver.ts
n0rahh: We'd need a Map (a Set can't do keyed count++ per occurrence), which turns items into Set<string> | Map<string, {count, props}> and pushes type-narrowing into every consumer — plus a Map isn't JSON-serializable for the JSON printer and the portal collector. Keeping items as "distinct names, size = total" and putting the per-extension payload in details seemed cleaner for me

packages/core/src/rules/other/stats.ts
n0rahh: On Set question already answered above.

Four consumers (three printers and the portal collector) read .total uniformly for every row, so deriving the size at read time would push an items ? items.size : total check into each of them. This one-time assignment at Root.leave also predates the PR — the loop just replaces the four per-row copies that main already had.

packages/core/src/walk.ts
n0rahh: Only SpecExtension holds arbitrary user data, so equal scalar values like true are its normal case — other node types hold objects, which never collide as keys. Using the location as the key for all types isn't safe either: a YAML anchor puts the same object in many places, and every rule would then visit and report it once per place instead of once

packages/core/src/utils/spec-extensions.ts
tatomyr: So we can modify types to declare the extensions where needed. The purpose of having right visitors tree is exactly to avoid manipulations like this.

packages/core/src/utils/spec-extensions.ts
tatomyr: Then we have wrong types or wrong walker logic. Making different types for different commands makes no sense to me.

packages/cli/src/commands/stats/visitor-and-accumulator-resolver.ts
tatomyr: No need in Map. It could be a simple object, or an array with duplicates so you can easily calculate the entries. What data are you trying to collect?

packages/core/src/walk.ts
tatomyr: There are plenty of other scalar nodes. It looks like you've found a but that has to be fixed instead of masking.

packages/core/src/utils/spec-extensions.ts
vadyvas: Agree, most of this can be fixed at the source instead of remapping types per command

1. Declare extensionsPrefix: 'x-' in the type definitions where the specs allow extensions
2. In the walker, check extensionsPrefix before falling back to additionalProperties

@tatomyr  wdyt?

tatomyr: Agree. I believe it makes sense to do that in this PR as it directly touches the spec extensions functionality and it'd be easier to spot uncovered places.

packages/core/src/utils/spec-extensions.ts
```diff
@@ -0,0 +1,95 @@
...
+        key.startsWith(EXTENSION_PREFIX) ? SpecExtension : entryType;
+    }
+  }
+}
+
+export function collectSpecExtension(
```
vadyvas: Nothing in this repo reads props, the printers only use count, and CLI telemetry never sends it

So could the sampling (props, describe, these regexes) move to the portal collector instead?

packages/core/src/resolve.ts
```diff
@@ -321,16 +321,14 @@ export async function resolveDocument(opts: {
       for (const propName of Object.keys(node)) {
         let propValue = node[propName];
         let propType = getOwn(type.properties, propName);
-        if (propType === undefined) propType = type.additionalProperties;
+        if (propType === undefined) {
```
vadyvas: what is the difference?

packages/core/src/rules/other/stats.ts
```diff
@@ -8,9 +14,75 @@ import type {
   OasRef,
 } from '../../typings/openapi.js';
 import type { Oas2Parameter } from '../../typings/swagger.js';
+import type { UserContext } from '../../walk.js';
+
+function countExtension(row: StatsRow, ctx: UserContext) {
+  const extensionName = ctx.key.toString();
+  if (!extensionName.startsWith('x-') || isRef(ctx.parent)) return;
```
tatomyr: It should always start with `x-` as we only go inside if it's a SpectExtension, don't we?

tatomyr: Oh, I see, it could come from [Operation](https://github.com/Redocly/redocly-cli/pull/3021/changes#diff-16df724ed35e34e45e4be3d9d645a9c08a45d174df1f3658871fc8fe8a2da4a4R81) or something similar. I guess, it's better to tackle it inside some wrapper in the exact visitor we expect to receive the extension key. Something like the following:


```ts
    Operation: {
      enter(_node: unknown, ctx: UserContext) {
        if (ctx.key === 'x-query') countExtension(statsAccumulator.xExtensions, ctx);
      },
    },

```

tatomyr: Same for the webhooks map and others.

n0rahh: We use this not only in SpecExtension, but in other hooks also, for example in Operation we dont want to count `get`, `put`, `post` etc

packages/core/src/types/oas3.ts
```diff
@@ -213,6 +213,7 @@ const License: NodeType = {
 };
 
 const Paths: NodeType = {
+  extensionsPrefix: 'x-',
```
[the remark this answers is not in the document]
n0rahh: the inconsistency is real, but the fix aligns OAS2 with OAS3 rather than the reverse, because content inside a vendor extension is the extension's value, not a separate extension — matching how undeclared extensions already behave

packages/core/src/types/index.ts
```diff
@@ -68,7 +68,7 @@ export function listOf(
 
 export function mapOf(
   typeName: string,
-  opts: { description?: string; documentationLink?: string } = {}
+  opts: { description?: string; documentationLink?: string; extensionsPrefix?: string } = {}
```
tatomyr, line 71: What is this for?

packages/core/src/types/oas2.ts
```diff
@@ -49,7 +49,6 @@ const Logo: NodeType = {
     backgroundColor: { type: 'string' },
     href: { type: 'string' },
   },
-  extensionsPrefix: 'x-',
```
tatomyr, line 52: Don't we need it anymore?

packages/core/src/rules/other/stats.ts
```diff
@@ -8,9 +14,79 @@ import type {
...
+      },
+    },
+    XUsePkce: {
+      enter(_node: unknown, ctx: UserContext) {
+        countExtension(statsAccumulator.xExtensions, ctx);
+      },
+    },
+    Operation: {
+      enter(_node: unknown, ctx: UserContext) {
+        if (ctx.key.toString().startsWith('x-')) {
```
tatomyr: It could be even more concrete, like `ctx.key.toString() === 'x-query'`, so it clarifies the intention. Same for x-webhooks.

packages/core/src/types/index.ts
n0rahh: Lets a map type declare that its spec allows x- keys, so the walker treats them as extensions instead of map entries — needed for Callback, where x- keys are extensions, not path items

packages/core/src/types/oas2.ts
n0rahh: Removed deliberately. Logo is the value of x-logo, which is itself a vendor extension — the OAS spec defines extension points only on spec objects, so an x- key inside an extension's value is just a property of that extension's own schema, not a new spec extension

vadyvas (approved):
lgtm

tatomyr (approved):

tatomyr:
@n0rahh please make sure to test the latest snapshot with out products.

tatomyr:
BTW, is this real perf. gain or a fluctuation? 
image

n0rahh:
> BTW, is this real perf. gain or a fluctuation? image

Looks like fluctuation, I benchmarked locally - no measurable difference in lint or bundle

PR #3028, opened 2026-08-06
feat: configurable operation-4xx-response (excludeMethods); add tests…

jeremyfiel:
## What/Why/How?

What:
- Run the operation-4xx-response validation at the Operation level so the HTTP method is available.
- Add an excludeMethods rule option (case-insensitive array) to skip 4XX validation for specified HTTP methods.
- Default excludeMethods to ["get", "head", "options"].
- Mirror the same behavior for webhooks (honor existing validateWebhooks flag).
- Add unit tests, documentation, and a changeset.

Why:
- Safe/read-only methods commonly do not validate request bodies and often don't need a documented 4XX response. Requiring a 4XX response for every operation is noisy for many APIs. Making the excluded methods configurable lets projects relax the rule where appropriate without removing it globally.

How:
- Move the 4XX check from the Responses visitor to Operation.leave so the HTTP verb (ctx.key) can be inspected.
- If the operation method is in excludeMethods, skip validation.
- Preserve reporting location to point at the operation.responses map (keeps tests/docs stable).
- Provide a config example in docs to override excludeMethods.

Files changed (high level)
- packages/core/src/rules/common/operation-4xx-response.ts — implementation and option parsing
- packages/core/src/rules/common/__tests__/operation-4xx-response.exclude.test.ts — new unit tests for excluded/non-excluded methods
- docs/@v2/rules/oas/operation-4xx-response.md — documented excludeMethods option and examples
- .changeset/skip-4xx-safe-methods.md — changeset entry

## Reference

- Rule docs updated: docs/@v2/rules/oas/operation-4xx-response.md
- Implementation: packages/core/src/rules/common/operation-4xx-response.ts
- Branch: jeremyfiel/fix/operation-4xx-skip-safe-methods

## Testing

Local verification steps (repo root):
1. npm install
2. npm run compile
3. npm run unit
   - or to run only the new tests:
     npm run unit -- packages/core/src/rules/common/__tests__/operation-4xx-response.exclude.test.ts
4. npm run lint

CI:
- The PR will run the repository GitHub Actions (compile, unit tests, lint). The PR includes a changeset.

## Screenshots (optional)

n/a

## Check yourself

- [x] This PR follows the contributing guide: https://github.com/Redocly/redocly-cli/blob/main/CONTRIBUTING.md#pull-request-guidelines
- [x] All new/updated code is covered by tests (unit tests added)
- [x] Documentation update has been considered and included (docs/@v2 updated)
- [x] Changeset included (.changeset/skip-4xx-safe-methods.md)

## Security

- [x] The security impact of the change has been considered. This is a linter/rules change and does not execute untrusted code or change runtime behavior of published packages.
- [x] Code follows company security practices and guidelines (no new dependencies added).

---

> [!NOTE]
> **Medium Risk**
> Default lint behavior changes for a common rule: GET/HEAD/OPTIONS no longer require documented 4XX responses unless users override `excludeMethods`, which may surprise teams on strict presets.
> 
> **Overview**
> The **`operation-4xx-response`** rule now skips 4XX checks for configurable HTTP methods, with **`get`**, **`head`**, and **`options`** excluded by default. Validation runs at the **Operation** level (instead of on the responses map alone) so the verb is available; webhook operations follow the same rules when **`validateWebhooks`** is enabled. Lint locations still point at **`operation.responses`**.
> 
> Projects can extend exclusions via **`excludeMethods`** (documented with an example) or restore strict checks for every method by overriding that list (e.g. an empty array). Unit tests cover default exclusion for GET and enforcement for POST.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17c84be22f5baff464a2ccb7ade47b1c47e745b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/skip-4xx-safe-methods.md
+++ b/.changeset/skip-4xx-safe-methods.md
@@ -1,0 +1,9 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
+---
+
+Make the `operation-4xx-response` rule configurable to exclude safe HTTP
+methods by default (get, head, options). This allows projects to avoid
+requiring 4XX responses for read-only operations while keeping the default
+behavior conservative.
--- a/docs/@v2/rules/oas/operation-4xx-response.md
+++ b/docs/@v2/rules/oas/operation-4xx-response.md
@@ -23,10 +23,11 @@ While this thinking has mostly changed (for the better in our opinion), it does 
 
 ## Configuration
 
-| Option           | Type    | Description                                                                               |
-| ---------------- | ------- | ----------------------------------------------------------------------------------------- |
-| severity         | string  | Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
-| validateWebhooks | boolean | Determines if responses inside webhooks are validated. Default `false`.                   |
+| Option           | Type    | Description                                                                                                 |
+| ---------------- | ------- | ----------------------------------------------------------------------------------------------------------- |
+| severity         | string  | Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration).                   |
+| validateWebhooks | boolean | Determines if responses inside webhooks are validated. Default `false`.                                     |
+| excludeMethods   | array   | List of HTTP methods (case-insensitive) to exclude from 4XX validation. Default: `['get','head','options']` |
 
 An example configuration:
 
@@ -42,6 +43,15 @@ rules:
   operation-4xx-response:
     severity: error
     validateWebhooks: true
+```
+
+To exclude additional methods from 4XX validation:
+
+```yaml
+rules:
+  operation-4xx-response:
+    severity: error
+    excludeMethods: [get, head, options, trace]
 ```
 
 ## Examples
--- a/packages/core/src/rules/common/operation-4xx-response.ts
+++ b/packages/core/src/rules/common/operation-4xx-response.ts
@@ -2,32 +2,55 @@ import type { Oas3Rule, Oas2Rule } from '../../visitors.js';
 import type { UserContext } from '../../walk.js';
 import { validateResponseCodes } from '../utils.js';
 
-export const Operation4xxResponse: Oas3Rule | Oas2Rule = ({ validateWebhooks }) => {
+export const Operation4xxResponse: Oas3Rule | Oas2Rule = (opts: any = {}) => {
+  const { validateWebhooks, excludeMethods: rawExcludeMethods } = opts || {};
+  const defaultExcluded = ['get', 'head', 'options'];
+  const excludeMethods = Array.isArray(rawExcludeMethods)
+    ? rawExcludeMethods.map((m: string) => String(m).toLowerCase())
+    : defaultExcluded;
+
   return {
     Paths: {
-      Responses(responses: Record<string, object>, { report }: UserContext) {
-        const codes = Object.keys(responses || {});
+      Operation: {
+        leave(operation: Record<string, any>, { report, key, location }: UserContext) {
+          const method = String(key).toLowerCase();
+          if (excludeMethods.includes(method)) return;
 
-        validateResponseCodes({
-          responseCodes: codes,
-          codeRange: '4XX',
-          report: report as UserContext['report'],
-          reference: 'https://redocly.com/docs/cli/rules/oas/operation-4xx-response',
-        });
+          const codes = Object.keys((operation.responses as Record<string, object>) || {});
+
+          // keep the reported location consistent with previous implementation
+          const childReport: UserContext['report'] = (problem) =>
+            report({ ...problem, location: location.child(['responses']).key() });
+
+          validateResponseCodes({
+            responseCodes: codes,
+            codeRange: '4XX',
+            report: childReport,
+            reference: 'https://redocly.com/docs/cli/rules/oas/operation-4xx-response',
+          });
+        },
       },
     },
     WebhooksMap: {
-      Responses(responses: Record<string, object>, { report }: UserContext) {
-        if (!validateWebhooks) return;
+      Operation: {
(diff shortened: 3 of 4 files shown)


docs/@v2/rules/oas/operation-4xx-response.md
```diff
@@ -23,10 +23,11 @@ While this thinking has mostly changed (for the better in our opinion), it does
...
-| ---------------- | ------- | ----------------------------------------------------------------------------------------- |
-| severity         | string  | Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration). |
-| validateWebhooks | boolean | Determines if responses inside webhooks are validated. Default `false`.                   |
+| Option           | Type    | Description                                                                                                 |
+| ---------------- | ------- | ----------------------------------------------------------------------------------------------------------- |
+| severity         | string  | Possible values: `off`, `warn`, `error`. Default `warn` (in `recommended` configuration).                   |
+| validateWebhooks | boolean | Determines if responses inside webhooks are validated. Default `false`.                                     |
+| excludeMethods   | array   | List of HTTP methods (case-insensitive) to exclude from 4XX validation. Default: `['get','head','options']` |
```
adamaltman, line 30: I'm not sure if this should be the default. GET requests can still have 4xx responses like 401, 403, 429 which could be documented and the documentation could help consumers and code gen.

packages/core/src/rules/common/operation-4xx-response.ts
```diff
@@ -2,32 +2,55 @@ import type { Oas3Rule, Oas2Rule } from '../../visitors.js';
 import type { UserContext } from '../../walk.js';
 import { validateResponseCodes } from '../utils.js';
 
-export const Operation4xxResponse: Oas3Rule | Oas2Rule = ({ validateWebhooks }) => {
+export const Operation4xxResponse: Oas3Rule | Oas2Rule = (opts: any = {}) => {
+  const { validateWebhooks, excludeMethods: rawExcludeMethods } = opts || {};
+  const defaultExcluded = ['get', 'head', 'options'];
```
adamaltman, line 7:
```suggestion
  const defaultExcluded = ['options'];
```
I think I would start with this possibly.

Also, even this is a breaking change.

docs/@v2/rules/oas/operation-4xx-response.md
jeremyfiel: good call. 
i wonder if it makes more sense to target the operations and codes directly.  the way it stands, it's fairly noisy but to your point, there are definitely use cases where it's helpful for those other non validation responses

how do you feel about this? 

```yaml
rules:
  operation-4xx-response:
  severity: warn
  exclusions:
    get: 
    - 400
    head:
    - 400
```

adamaltman: I'm not sure I follow what does the exclusion `get` `400` mean?

I would read it to mean that if the `get` request has a `400` response, then exclude that 400 response from being evaluated here but it doesn't make sense because it would pass this rule then if it has a 400 response. Or am I interpreting it wrong?

I think `excludeMethods` makes more sense, but just that the default no methods are excluded (for BC).

.changeset/skip-4xx-safe-methods.md
```diff
@@ -0,0 +1,9 @@
+---
+'@redocly/openapi-core': patch
+'@redocly/cli': patch
```
kanoru3101, line 3:
```suggestion
'@redocly/openapi-core': minor
'@redocly/cli': minor
```
Since this is the addition of a new feature, it makes more sense to classify it as a minor change.

packages/core/src/rules/common/operation-4xx-response.ts
```diff
@@ -2,32 +2,55 @@ import type { Oas3Rule, Oas2Rule } from '../../visitors.js';
...
+            reference: 'https://redocly.com/docs/cli/rules/oas/operation-4xx-response',
+          });
+        },
       },
     },
     WebhooksMap: {
-      Responses(responses: Record<string, object>, { report }: UserContext) {
-        if (!validateWebhooks) return;
+      Operation: {
+        leave(operation: Record<string, any>, { report, key, location }: UserContext) {
```
kanoru3101, line 36: This body is the same as `Paths.Operation.leave` above except for the `validateWebhooks`. Please use  one small function in both places

packages/core/src/rules/common/operation-4xx-response.ts
```diff
@@ -2,32 +2,55 @@ import type { Oas3Rule, Oas2Rule } from '../../visitors.js';
 import type { UserContext } from '../../walk.js';
 import { validateResponseCodes } from '../utils.js';
 
-export const Operation4xxResponse: Oas3Rule | Oas2Rule = ({ validateWebhooks }) => {
+export const Operation4xxResponse: Oas3Rule | Oas2Rule = (opts: any = {}) => {
+  const { validateWebhooks, excludeMethods: rawExcludeMethods } = opts || {};
+  const defaultExcluded = ['get', 'head', 'options'];
+  const excludeMethods = Array.isArray(rawExcludeMethods)
+    ? rawExcludeMethods.map((m: string) => String(m).toLowerCase())
+    : defaultExcluded;
```
kanoru3101, line 10:
```suggestion
export const Operation4xxResponse: Oas3Rule | Oas2Rule = ({
  validateWebhooks,
  excludeMethods: rawExcludeMethods = [],
}) => {
  const excludeMethods = rawExcludeMethods.map((method: string) => method.toLowerCase());
```

What if we simplify this and take excludeMethods straight from the rule options, like other rules do? Also sets the default to [] per the discussion above.

kanoru3101:
Hi @jeremyfiel  !
Just checking in on this PR. If you’re not planning to continue with it, I’d be happy to take it over and keep working on it. Let me know if that works for you!

PR #3038, opened 2026-08-11
fix: do not run nested visitors with a stale parent on shared $refs

### LINKED ISSUE #2829: Linter reports duplicate parameters when multi-referencing path items
**Describe the bug**

The linter reports
```
Operations must have unique `name` + `in` parameters. Repeats of `in:header` + `name:x-duplicate`.
```
even though there is no conflict

**To Reproduce**

Given this yaml
```yaml
openapi: 3.1.0
info:
  title: Sample API
  version: 0.0.1
  contact:
    email: onavratil@monetplus.cz
  description: mock
  license:
    name: Proprietary
    identifier: proprietary
servers:
  - url: '{address}'
    description: | 
      Local development server
    variables:
      address:
        default: 'http://localhost:8080'
        description: 'Server address'
security: []
tags:
  - name: test
    description: test.
paths:
  /sample:
    get:
      summary: sample
      description: Sample desc
      operationId: opId
      tags:
        - test
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  sample:
                    type: string
        4XX:
          description: FAIL
          content:
            application/json:
              schema:
                type: object
                properties:
                  error:
                    type: string
      callbacks:
        test:
          'some-uri':
            $ref: '#/components/pathItems/test-pi'
        test2:
          'some-uri-2':
            $ref: '#/components/pathItems/test-pi'
components:
  pathItems:
    test-pi:
      post:
        summary: test
        description: XYZ
        operationId: test23
        parameters: 
          - name: x-duplicate
            in: header
            required: true
            schema:
              type: string
```
Note that there are multiple references of the pathItem in callback. I have a common "pathItem component for notification callbacks" that report some event and i use it in multiple /paths in the callback. In the example above, i put two webhooks under one /path just for simplicity, but the behavior is the same.

The linter evaluates this as duplicate parameter, although there are in separate webhooks:
```bash
$ redocly lint api.yaml 
No configurations were provided -- using built in recommended configuration by default.

validating api.yaml...
[1] api.yaml:26:7 at #/paths/~1sample/get/parameters/0

Operations must have unique `name` + `in` parameters. Repeats of `in:header` + `name:x-duplicate`.

24 |   /sample:
25 |     get:
26 |       summary: sample
27 |       description: Sample desc
 … |       < 28 more lines >
56 |             $ref: '#/components/pathItems/test-pi'
57 | components:
58 |   pathItems:

Error was generated by the operation-parameters-unique rule.


api.yaml: validated in 37ms

run `redocly lint --generate-ignore-file` to add all problems to the ignore file.
```

**Expected behavior**

Api passes linter

**Redocly version(s)**

```
$ redocly --version
2.31.3
```

**Node.js version(s)**

```
$ node --version
v25.5.0
```

**Additional context**

I guess it relates to this file https://github.com/Redocly/redocly-cli/blob/cf30dae730a20644575da70e4400cc8941e2f40b/packages/core/src/rules/common/operation-parameters-unique.ts#L4

Note that I tried to reproduce this without using path item references, but failed. The bug must somehow related to the path item being referenced - perhaps the walk function then does not invoke `enter` on multiple occurances of the same ref

vadyvas:
## What/Why/How?

`operation-parameters-unique` reported a duplicate parameter that is not there. It happens when the same path item is reused through `$ref`, for example from two callbacks, and the reported location points at an operation that has no `parameters` at all.

The reason is in the walker. Two `$ref`s to the same path item resolve to one object, so the walker enters it twice. On the second entry it does not call the rule's `enter` hooks again, because it already called them for that node — but it still calls the rule's nested `Parameter` visitor. So the visitor runs with the state and the parent location left over from the first entry, and reports a parameter it had already counted.

Added a guard: a nested visitor now runs only if the node its parent hook sits on is the node the walker is really inside. On the second entry the `Parameter` visitor stays quiet, and the rule reports only what it actually saw. `operation-parameters-unique` is not the only rule with this shape — `spec-querystring-parameters` had the same problem.

## Reference

Closes https://github.com/Redocly/redocly-cli/issues/2829

## Testing

- https://github.com/Redocly/redocly/actions/runs/33093171908/job/98591935934?pr=26673 ✅

## Screenshots (optional)

`x-signature` is declared once, in one list:

```yaml
paths:
  /sample:
    get:                                  # no parameters here
      callbacks:
        onEventA:
          'https://callbacks.example.com/a':
            $ref: '#/components/pathItems/notify'
        onEventB:
          'https://callbacks.example.com/b':
            $ref: '#/components/pathItems/notify'
components:
  pathItems:
    notify:
      post:
        parameters:
          - in: header
            name: x-signature            # declared once, used in both callbacks
```

Before: 
```sh
validating openapi.yaml using lint rules for api 'main'...
[1] openapi.yaml:9:7 at #/paths/~1sample/get/parameters/0

Operations must have unique `name` + `in` parameters. Repeats of `in:header` + `name:x-signature`.

 7 |   /sample:
 8 |     get:
 9 |       operationId: getSample
   |       ^^^^^^^^^^^^^^^^^^^^^^
10 |       summary: example summary
   |       ^^^^^^^^^^^^^^^^^^^^^^^^

Error was generated by the operation-parameters-unique rule.

```

After: 
```sh
validating openapi.yaml using lint rules for api 'main'...
openapi.yaml: validated in 4ms

Woohoo! Your API description is valid. 🎉
```

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Changes core lint walking behavior for all nested visitors and shared `$ref` re-entries; low security impact but could affect rule results across the ecosystem.
> 
> **Overview**
> Fixes **false duplicate-parameter lint** when the same path item is reached through multiple `$ref`s (e.g. two callbacks or a path plus callbacks). Rules like `operation-parameters-unique` could fire on an operation that has no `parameters` because nested visitors still ran after a second walk into the same resolved object, using parent context from the first visit.
> 
> The OpenAPI document walker in `walk.ts` now tracks the **active node per type** (`walkingNodeByType`) for the current descent. Nested rule visitors activate only when the parent hook’s node matches that active node, so a re-entry into a shared `$ref` target does not re-run nested visitors with leftover parent state. Sibling keys beside `$ref` keep working as before.
> 
> Adds unit coverage for callback + shared path item refs and composed `$ref` siblings, plus an e2e lint fixture that expects a clean pass for reused `components/pathItems`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9158410070637c9d851ef7d4dd944b7e407fb36a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/quiet-bears-allow.md
+++ b/.changeset/quiet-bears-allow.md
@@ -3,4 +3,4 @@
 '@redocly/openapi-core': patch
 ---
 
-Fixed an issue where rule reported a duplicate parameter when two or more `$ref`s point to the same path item.
+Fixed an issue where rule incorrectly reported a duplicate parameter when two or more `$ref`s pointed to the same path item.
--- a/.changeset/quiet-bears-allow.md
+++ b/.changeset/quiet-bears-allow.md
@@ -1,0 +1,6 @@
+---
+'@redocly/cli': patch
+'@redocly/openapi-core': patch
+---
+
+Fixed an issue where rule reported a duplicate parameter when two or more `$ref`s point to the same path item.
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -152,6 +152,7 @@ export function walkDocument<T extends BaseVisitor>(opts: {
   const composedRefWalkId = (type: NormalizedNodeType, location: Location) =>
     `${type.name}::${location.absolutePointer}`;
   const ignoredNodes = new Set<string>();
+  const walkingNodeByType: Record<string, unknown> = {};
 
   // Pre-compute combined enter/leave arrays per type to avoid per-node array allocations
   const anyEnter = normalizedVisitors.any.enter as VisitorNode<any>[];
@@ -300,6 +301,8 @@ export function walkDocument<T extends BaseVisitor>(opts: {
           if (
             (context.parent && // if nested
               context.parent.activatedOn &&
+              context.parent.activatedOn.value.node ===
+                walkingNodeByType[context.parent.type.name] &&
               context.activatedOn?.value.withParentNode !== context.parent.activatedOn.value.node &&
               // do not enter if visited by parent children (it works thanks because deeper visitors are sorted before)
               context.parent.activatedOn.value.nextLevelTypeActivated?.value !== type) ||
@@ -342,6 +345,9 @@ export function walkDocument<T extends BaseVisitor>(opts: {
           }
         }
       }
+
+      const prevWalkingNode = walkingNodeByType[type.name];
+      walkingNodeByType[type.name] = resolvedNode;
 
       if (visitedBySome || !isNodeSeen) {
         seenNodesPerType[type.name] = seenNodesPerType[type.name] || new Set();
@@ -406,6 +412,8 @@ export function walkDocument<T extends BaseVisitor>(opts: {
           }
         }
       }
+
+      walkingNodeByType[type.name] = prevWalkingNode;
 
       const currentLeaveVisitors =
         combinedLeave[type.name] || (normalizedVisitors[type.name]?.leave || []).concat(anyLeave);
--- a/packages/core/src/walk.ts
+++ b/packages/core/src/walk.ts
@@ -152,6 +152,7 @@ export function walkDocument<T extends BaseVisitor>(opts: {
   const composedRefWalkId = (type: NormalizedNodeType, location: Location) =>
     `${type.name}::${location.absolutePointer}`;
   const ignoredNodes = new Set<string>();
+  const walkingNodeByType: Record<string, unknown> = {};
 
   // Pre-compute combined enter/leave arrays per type to avoid per-node array allocations
   const anyEnter = normalizedVisitors.any.enter as VisitorNode<any>[];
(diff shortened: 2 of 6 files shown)


vadyvas:
@cursor review

.changeset/quiet-bears-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@redocly/cli': patch
+'@redocly/openapi-core': patch
+---
+
+Fixed an issue where rule reported a duplicate parameter when two or more `$ref`s point to the same path item.
```
JLekawa:
```suggestion
Fixed an issue where rule incorrectly reported a duplicate parameter when two or more `$ref`s pointed to the same path item.
```

JLekawa (approved):

packages/core/src/walk.ts
```diff
@@ -407,6 +413,8 @@ export function walkDocument<T extends BaseVisitor>(opts: {
         }
       }
 
+      walkingNodeByType[type.name] = prevWalkingNode;
```
tatomyr, line 426: Why reassigning it back?

tatomyr:
Please make sure to test the snapshot: 
image

packages/core/src/walk.ts
vadyvas: this keeps track of the node we are currently inside
without the restore, after we leave a child and move to a sibling, the map would still say we are inside that child and the check below would wrongly skip visitors

tatomyr (approved):
LGTM. Make sure to test in with internal products.

vadyvas:
https://github.com/Redocly/redocly/actions/runs/33093171908/job/98591935934?pr=26673

PR #3040, opened 2026-08-17
fix: name the rule when a configurable rule is missing its assertions block

### LINKED ISSUE #2185: Configurable rule fails with unclear error message
**Describe the bug**

**To Reproduce**

```yaml

rules:
  rule/test:
    subject:
      type: Operation
      property: summary


```

3. And any OpenAPI file(
4. Run this command with these arguments... `redocly lint`
5. See error

Image

**Expected behavior**

The error should reflect that the  `assertsion` section is missing.

luantaraschi:
## What/Why/How?

A configurable rule declared without its `assertions` block dies with a message that says nothing about the config:

```
Something went wrong when processing openapi.yaml:

  - Cannot read properties of undefined (reading 'pattern')
```

`getAssertsToApply` reads `assertion.assertions[assertName]` for the first key of `asserts`, which is `pattern`, so that is where the `undefined` surfaces.

The config schema already declares the block as mandatory. `ConfigurableRule` and `Where` in `packages/core/src/types/redocly-yaml.ts` both carry `required: ['subject', 'assertions']`, so no valid configuration is being turned away here. The runtime simply was not checking what the schema promises, and `redocly lint` still ran the rule after the config struct warning.

The change adds that check next to the `'type' (String) is required` guards that were already there, in the four places a configurable rule definition gets built:

- `rules/common/assertions/index.ts`, the rule itself
- `rules/common/assertions/utils.ts`, each `where` entry
- `graphql/assertions.ts`, both of the same two spots

The GraphQL path keeps its own `if (!kind) continue`, so a rule with no subject type is still skipped rather than rejected. Only the missing `assertions` block now raises, which is what the OpenAPI path already did.

After:

```
Something went wrong when processing openapi.yaml:

  - rule/test: 'assertions' (Object) is required
```

## Reference

Closes #2185.

## Testing

Two unit tests in `packages/core/src/rules/common/assertions/__tests__/index.test.ts` for the rule and its `where` entry, and two in `packages/core/src/rules/graphql/__tests__/assertions.test.ts` that go through `lintFromString` against a schema. All four fail on `main` with the `Cannot read properties of undefined` message.

```
VITEST_SUITE=unit npx vitest run packages/core/src/rules/common/assertions packages/core/src/rules/graphql
```

The whole `packages/core` suite was run on `main` and on this branch back to back, on Windows with Node 22.20. `main` gives 36 failed and 1015 passed, the branch gives 36 failed and 1019 passed, so the only difference is the four tests added here. Those 36 are pre-existing path and snapshot mismatches on Windows and none of them touch assertions.

## Check yourself

- [x] Documentation update has been considered  The docs already describe `assertions` as part of the rule shape, so nothing there changes.

## Security

- [x] The security impact of the change has been considered  The change only turns an unhandled `TypeError` into a named error before any document is read.

---

> [!NOTE]
> **Low Risk**
> Validation-only change at rule-build time; no lint behavior change for valid configs, only clearer failures for invalid ones.
> 
> **Overview**
> Configurable rules without an `assertions` object no longer crash with `Cannot read properties of undefined (reading 'pattern')` during lint setup. **Runtime checks** now mirror the config schema and throw named errors like `rule/test: 'assertions' (Object) is required`.
> 
> The same validation is applied for **top-level rules** and for each **`where`** entry, on both the **OpenAPI** path (`rules/common/assertions/index.ts`, `utils.ts`) and the **GraphQL** path (`graphql/assertions.ts`). GraphQL still skips rules with no subject `type`; only a missing `assertions` block is rejected.
> 
> Unit and GraphQL lint tests cover the rule-level and `where`-level cases. A patch changeset documents the improved error message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb017e04cd3f1498f56f009da13f0bef06c6b533. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/configurable-rule-missing-assertions.md
+++ b/.changeset/configurable-rule-missing-assertions.md
@@ -1,0 +1,5 @@
+---
+'@redocly/openapi-core': patch
+---
+
+Fixed the unclear error raised when a configurable rule, or one of its `where` entries, is missing the `assertions` block. The message now names the rule instead of failing with `Cannot read properties of undefined (reading 'pattern')`. This covers both the OpenAPI and the GraphQL paths.
--- a/.changeset/configurable-rule-missing-assertions.md
+++ b/.changeset/configurable-rule-missing-assertions.md
@@ -2,4 +2,4 @@
 '@redocly/openapi-core': patch
 ---
 
-Fixed the unclear error raised when a configurable rule, or one of its `where` entries, is missing the `assertions` block. The message now names the rule instead of failing with `Cannot read properties of undefined (reading 'pattern')`. This covers both the OpenAPI and the GraphQL paths.
+Improved the error message raised when a configurable rule is missing the `assertions` block.
--- a/packages/core/src/graphql/assertions.ts
+++ b/packages/core/src/graphql/assertions.ts
@@ -51,6 +51,10 @@ export const GraphqlAssertions: GraphqlRule = (
     const kind = assertion.subject?.type as GraphqlNodeKind | undefined;
     if (!kind) continue;
 
+    if (!isPlainObject(assertion.assertions)) {
+      throw new Error(`${assertion.assertionId}: 'assertions' (Object) is required`);
+    }
+
     const assertsToApply = getGraphqlAssertsToApply(assertion);
     const whereMatchers = buildWhereMatchers(assertion);
 
@@ -72,6 +76,11 @@ function buildWhereMatchers(assertion: Assertion): WhereMatcher[] {
     if (!isString(definition.subject?.type)) {
       throw new Error(
         `${assertion.assertionId} -> where -> [${index}]: 'type' (String) is required`
+      );
+    }
+    if (!isPlainObject(definition.assertions)) {
+      throw new Error(
+        `${assertion.assertionId} -> where -> [${index}]: 'assertions' (Object) is required`
       );
     }
     return {
--- a/packages/core/src/rules/common/assertions/__tests__/index.test.ts
+++ b/packages/core/src/rules/common/assertions/__tests__/index.test.ts
@@ -97,4 +97,32 @@ describe('Oas3 assertions', () => {
       ]
     `);
   });
+
+  it('should throw a named error when the assertions block is missing', () => {
+    const rule = {
+      'rule/no-assertions': {
+        assertionId: 'rule/no-assertions',
+        subject: { type: 'Operation', property: 'summary' },
+      },
+    };
+
+    expect(() => Assertions(rule as any)).toThrow(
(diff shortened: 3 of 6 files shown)


tatomyr:
Thanks for the contribution, @luantaraschi. Frankly, I'd prefer to silently skip incorrectly configured rules rather than throw an error, since there's already a lint check for the configuration itself that should flag the problem to the user. What do you think?

luantaraschi:
That works for me on the CLI side, where `lintConfig` runs first and already reports `The field 'assertions' must be present on this level`.

The case I would flag is the programmatic one. `createConfig` on its own neither throws nor reports anything, and `lintFromString` does not run the config check, so with a silent skip a rule missing `assertions` would simply do nothing and the caller would get no signal at all.

There is also the sibling guard a few lines above: a missing `subject.type` throws `'type' (String) is required` rather than skipping. If misconfigured rules should be skipped, that one probably wants the same treatment, otherwise the same class of mistake behaves in two different ways.

I am happy to switch this to skipping either way, with or without the `type` guard following. Just say which you prefer.

.changeset/configurable-rule-missing-assertions.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/openapi-core': patch
+---
+
+Fixed the unclear error raised when a configurable rule, or one of its `where` entries, is missing the `assertions` block. The message now names the rule instead of failing with `Cannot read properties of undefined (reading 'pattern')`. This covers both the OpenAPI and the GraphQL paths.
```
tatomyr:
```suggestion
Fixed the unclear error raised when a configurable rule is missing the `assertions` block.
```

tatomyr:
You're probably right. Let's stick with the current approach for now and see how it goes.

luantaraschi:
Sounds good, thanks. Keeping the throw also leaves the `subject.type` guard right above it consistent, so there is no follow-up needed on that one.

The PR is unchanged from what you reviewed and CI is green. Let me know if you want anything adjusted before it goes in.

tatomyr:
@luantaraschi Just waiting for approval from @Redocly/technical-writers, and we're good to go.

.changeset/configurable-rule-missing-assertions.md
JLekawa:
```suggestion
Improved the error message raised when a configurable rule is missing the `assertions` block.
```

JLekawa (approved):

tatomyr (approved):

tatomyr:
@luantaraschi  Merged. Your changes will be available in the next release. Thank you again for taking you time!

PR #3071, opened 2026-09-01
docs: clarify 101 responses and WebSocket support

dimitropoulos:
## What/Why/How?

We had someone (by misunderstanding) try to describe a websockets connection upgrade using OpenAPI and got flagged by this rule.  So, in this PR I clarify why an operation with only a `101 Switching Protocols` response fails the `operation-2xx-response` rule. The v1 and v2 rule docs now state that 101 is informational rather than successful and that WebSocket APIs should be described with AsyncAPI because OpenAPI does not support them.

## Reference

https://redocly.com/docs/cli/rules/oas/operation-2xx-response

## Testing

- `npx oxfmt --check docs/@v1/rules/oas/operation-2xx-response.md docs/@v2/rules/oas/operation-2xx-response.md`
- Pre-commit lint and formatting hooks passed

## Screenshots (optional)

Not applicable.

## Check yourself

- [x] All new/updated code is covered by tests (documentation-only change)
- [x] Core code changed? - Not applicable
- [x] New package installed? - Not applicable

## Security

---

> [!NOTE]
> **Low Risk**
> Documentation-only edits to two markdown rule pages; no runtime or lint rule behavior changes.
> 
> **Overview**
> Updates the **operation-2xx-response** rule documentation for CLI v1 and v2 with guidance that was missing from the API design principles section.
> 
> The docs now explain that an operation whose only declared response is **`101 Switching Protocols`** still fails the rule, because **101 is an informational (1xx) status**, not a successful **2xx** response. They also note that **WebSocket APIs are outside OpenAPI’s scope** and should be described with **AsyncAPI** instead.
> 
> No rule implementation or configuration behavior changes—only user-facing explanation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f948be0068b4cce47fd660a5a4e10b2bec4043b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/@v2/rules/oas/operation-2xx-response.md
+++ b/docs/@v2/rules/oas/operation-2xx-response.md
@@ -22,6 +22,9 @@ If it doesn't, what is the purpose of the operation?
 Even if there is no response content (204), it can still return a successful response with no content.
 You can greatly improve the developer and user experience of your APIs by making it a standard to provide this information.
 
+An operation with only a `101 Switching Protocols` response fails this rule because 101 is an informational response, not a successful response.
+WebSocket APIs are not supported by OpenAPI; use AsyncAPI to describe them instead.
+
 ## Configuration
 
 | Option           | Type    | Description                                                                               |
--- a/docs/@v1/rules/oas/operation-2xx-response.md
+++ b/docs/@v1/rules/oas/operation-2xx-response.md
@@ -21,6 +21,9 @@ If it doesn't, what is the purpose of the operation?
 Even if there is no response content (204), it can still return a successful response with no content.
 You can greatly improve the developer and user experience of your APIs by making it a standard to provide this information.
 
+An operation with only a `101 Switching Protocols` response fails this rule because 101 is an informational response, not a successful response.
+WebSocket APIs are not supported by OpenAPI; use AsyncAPI to describe them instead.
+
 ## Configuration
 
 | Option           | Type    | Description                                                                               |


docs/@v2/rules/oas/operation-2xx-response.md
```diff
@@ -22,6 +22,9 @@ If it doesn't, what is the purpose of the operation?
 Even if there is no response content (204), it can still return a successful response with no content.
 You can greatly improve the developer and user experience of your APIs by making it a standard to provide this information.
 
+An operation with only a `101 Switching Protocols` response fails this rule because 101 is an informational response, not a successful response.
+WebSocket APIs are not supported by OpenAPI; use AsyncAPI to describe them instead.
```
tatomyr, line 26: I'm not fully convinced this addition is necessary. From the rule's own description it should already be clear that `1xx` responses aren't `2xx` responses, and that the rule applies only to OAS, not AsyncAPI.

dimitropoulos: Right right - yeah, that's precisely the confusion though.  The situation we hit here at Cloudflare is we turned this rule on and of our 3,500 operations there were 4 that were getting this error because they returned a `1xx`.  So, not realizing that those operations just shouldn't have ever been added in the first place, those people chimed in to say "hey, the rule is broken, it doesn't say anything about WebSockets where you don't have 2xx".  Of course, at that point, like you're saying, I then shared that that's actually by design.

So, with that context, I thought it would make a better overall experience to make explicit note of that for people that may not understand all the details - which seems very much on-mission for a linter to me.

thoughts?

tatomyr: Okay, that makes sense. But please adjust the wording so it reflects the case. Otherwise your addition looks like it's taken out of context to me. 

cc @Redocly/technical-writers

dimitropoulos: ok, how would you feel about:

> Note that WebSocket APIs are not supported by OpenAPI (use AsyncAPI to describe them instead).  If you're getting this error because you have a `101 Switching Protocols` response, the correct action is to remove this operation.

tatomyr: > the correct action is to remove this operation.

Sounds too restrictive. Another option could be to ignore it -- it's up to the author to decide what to do with such a violation.

dimitropoulos: sure - I said that because it's literally invalid OpenAPI, but I know what you mean - some people may go "oh welllll"

how about

> If you're getting this error for an operation that has a `101 Switching Protocols` response, note that WebSockets are not supported by OpenAPI (use AsyncAPI to describe them instead).

PR #3080, opened 2026-09-03
feat(recheck): add the recheck engine package

adamaltman:
## What/Why/How?

Recheck Phase 2 relocates the prose-linting engine from the Redocly monorepo into this repository (design: Redocly/redocly#26970).
This PR adds `packages/recheck` as an engine-only package in the fixed version group.
It adds no command; `redocly recheck` follows in the next PR.

What moved: parser, scopes, rules, runner, auto-fix, directives, config validation, and presets.
Also moved: metrics, reporters, the preset provenance files, examples, the example generator, and the two agent skills.
What did not move: the standalone `cli.ts` and yargs command files, and the `recheck.yaml` discovery.
It also left behind the markdownlint parity harness and a drift test that needs the Realm theme source.

The command files became plain actions (`runLint`, `generateBaseline`, `runReadability`, `generateMarkdocSchema`).
Each takes a resolved config and a `Logger`, and the reporters print through the same `Logger`, so the engine writes nothing to the console.
A new `resolveRecheckConfig` turns a `redocly.yaml` `recheck` block plus the `recheck/*` names from the root `extends` into normalized rules.
The block nests rules under `rules` and accepts a severity shorthand.

The optional spell-check peers (`nspell`, `dictionary-en`) stay optional peer dependencies.
The CLI bundle marks them external in the next PR.

`.oxlintrc.json` extends its existing `sonarjs/cognitive-complexity` threshold-82 override to cover `packages/recheck/**`.
The engine's tests type-check under the root `npm run typecheck`.
`RecheckConfig` is a type with the engine's declared keys plus a template-literal index signature for slashed rule names.
`.oxfmtrc.json` excludes the lint fixtures under `packages/recheck/src/config/__tests__/fixtures/` from formatting.
A `minor` on `@redocly/recheck` bumps `@redocly/cli`, `@redocly/openapi-core`, and `@redocly/respect-core` to the same version.

## Reference

- Design: Redocly/redocly#26970 (private: relocate the engine and add the command).
- Tracker: Redocly/redocly#26711 (private: the Recheck implementation tracker for Redocly CLI).

## Testing

- The engine's unit tests run under the root vitest config: 131 files, 2,282 tests (2 skipped).
- New tests: `resolve.test.ts` covers ten cases: preset composition, object overrides, shorthand overrides, baseline path, markdoc, and apiDescriptions pass-through.
- It also covers a rejected block `extends`, a rejected non-object `rules` block, engine errors, and the warn callback.
[...]

PR #3082, opened 2026-09-03
feat(cli): add the redocly recheck command

adamaltman:
## What/Why/How?

This PR adds the `redocly recheck` command.
It lints Markdown prose and structure with the recheck engine from `packages/recheck`.
Configuration lives in the `recheck` block of `redocly.yaml`.
You name presets in the root `extends`, for example `extends: [recommended, recheck/markdown]`.

**Core.**
Core sets `recheck/*` entries in the root `extends` aside as `ResolvedConfig.recheckExtends`.
The API preset resolver skips them, so `redocly lint` ignores them.
Core takes no dependency on `@redocly/recheck`.

**Engine.**
The engine `Logger` gained an `output` channel for report payloads.
The table, JSON, SARIF, GitHub Actions, readability, and `--stats` payloads write to it.
Progress, warnings, and errors stay on `log`, `warn`, and `error`.
The readability action lost its `quiet` flag, because the channel split makes it unnecessary.
Library callers of `runReadability` now receive progress lines through `log` in JSON mode too.

**CLI.**
`packages/cli` bundles the engine.
The CLI bundles the spell-check packages, `nspell` and `dictionary-en`, and the build copies the dictionary data files next to their chunk.
The handler maps the engine's channels onto the CLI logger: progress to stderr, payloads to stdout.
So `--format json`, `sarif`, and `github-actions` keep stdout machine-readable.

**Command.**
`redocly recheck [paths..]` lints by default.
Action flags select one other action: `--readability`, `--generate-baseline`, or `--generate-markdoc-schema` (with `--from`, `--out`, `--check`).
Runs pick up a `.redocly.recheck-baseline.yaml` next to `redocly.yaml` by presence; `baseline` in the `recheck` block sets another path.
`--fix` applies to linting only.

Output flags: `--format` (`table` default), `--output-path`, `--stats`, `--max-problems`, `--summary`, `--summary-path`.
`--output-path` applies to `json` and `sarif`; with other formats the command warns and writes the report to stdout.
Selection flags: `--severity` (`warn` or `error`), `--tags`, `--rule`, `--skip-rule`.
To scope a run to changed files, pass their paths.

With no `redocly.yaml`, the command uses `recheck/markdown` and says so on stderr.
With a `redocly.yaml` that has no `recheck/*` preset and no `recheck` block, it checks nothing and says so, the same as `lint`.
The command skips an API description path with a stderr notice.
That path arrives in a later PR.

**Shared-code changes.**
`commandWrapper` keeps its original behavior; the handler throws `AbortFlowError` on a failed run, the same path `lint` uses.
Core reserves the plugin ids `recheck`, `redocly`, `redoc`, `realm`, and `reunite`; a plugin that declares one of them fails to load.
`handleLintConfig` gained a type guard, because the recheck `--format` values widened the shared `CommandArgv` union.
`handleLintConfig` also skips config linting for `--format sarif`, as it does for `json`, `junit`, and `checkstyle`, so SARIF stdout stays a single document.

**Config schema.**
Core pins `@redocly/config` 0.56.0, which carries the `recheck` root key.
`check-config` accepts the block, and the config lint no longer warns on it.
A `check-config` e2e fixture covers a valid block.

This PR stacks on #3080, which relocates the engine.
Its base changes to `main` after #3080 merges.

## Reference

- Redocly/redocly#26970: the Phase 2 design.
  It chose approach A: presets in the root `extends`, rules in a `recheck` block, no engine dependency in core.
- Redocly/redocly#26711: the recheck-to-CLI tracker.
  Phase 1 shipped the standalone engine; Phase 2 moves it into the CLI.
- #3080: PR-A, the engine relocation this PR builds on.

## Testing

Unit tests cover the `recheck/` namespace in core, the reserved plugin ids, and `selectAction`.
End-to-end tests live in `tests/e2e/recheck/` with twelve fixtures and snapshots.
They cover lint, JSON and SARIF stdout, readability, and baseline generation.
They also cover the Markdoc schema, the fallback notice, the no-config notice, config errors, conflicting flags, and the output-path warning.
The test normalizes timing lines before it compares snapshots.

Gates: `npm run compile`, `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run unit`, and the recheck e2e suite.
In `npm run unit`, the three pre-existing oas3 timeouts are the only failures, the same as on `main`.
Manual: `node packages/cli/lib/index.js recheck --help`, and a run against an invalid `recheck` block to confirm `--generate-markdoc-schema` no longer depends on it.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Touches shared config resolution and CLI bundling (spell-check assets) and introduces a new user-facing command, but API lint behavior is isolated by filtering recheck presets and changes are covered by tests.
> 
> **Overview**
> Adds **`redocly recheck`**, a CLI command that lints Markdown prose and structure via the bundled **`@redocly/recheck`** engine. Configuration uses a **`recheck`** block in `redocly.yaml` and presets such as **`recheck/markdown`** in the root **`extends`**; **`openapi-core`** strips those presets from API lint resolution into **`recheckExtends`** without depending on recheck, **`check-config`** validates the block, and reserved plugin ids (`recheck`, `redocly`, etc.) are rejected at load time.
> 
> The handler supports default lint (with **`--fix`**, formats **`table`/`json`/`sarif`/`github-actions`**), plus alternate actions **`--readability`**, **`--generate-baseline`** (default **`.redocly.recheck-baseline.yaml`**, auto-discovered when present), and **`--generate-markdoc-schema`**. Progress goes to stderr and report payloads to stdout through a new **`Logger.output`** channel; the CLI build also copies Hunspell data for bundled spell-checking.
> 
> Docs, changesets, unit/e2e coverage, and small **`lint`** tweaks (skip config lint for **`sarif`**, output-format type guard) round out the integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0cdd84c52e8b1c2a0fedf6b7c3e76df41ceaf00f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,137 @@
...
+
+GitHub reads this format and places each finding next to its line in the pull request diff.
+
+<pre>
+::error title=recheck/single-h1,file=docs/index.md,line=2,endLine=2,col=1,endColumn=1::Multiple top-level headings in the same document
+</pre>
+
+### Generate and apply a baseline
+
+Record every current error, then gate only new ones.
```
RomanHotsiy: very claudi-sh

docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,137 @@
+# `recheck`
+
+## Introduction
+
+The `recheck` command lints Markdown prose and structure.
+Configuration lives in the [`recheck` block](../configuration/reference/recheck.md) of `redocly.yaml`.
+Name presets in the root `extends`, for example `recheck/markdown`.
```
RomanHotsiy: this is confusing

docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,137 @@
...
+Without a `recheck` block or a `recheck/*` preset, the command falls back to `recheck/markdown`.
+
+Progress messages go to stderr.
+The report goes to stdout.
+For `--format json` and `sarif`, use `--output-path` to write the report to a file instead.
+
+A later release adds linting for API descriptions.
+This release skips them and prints a notice for each skipped file.
```
RomanHotsiy: should be removed

RomanHotsiy (commented):
Re-read the docs plz

packages/cli/src/commands/recheck/cli-logger.ts
```diff
@@ -0,0 +1,13 @@
+import { logger } from '@redocly/openapi-core';
+import type { Logger } from '@redocly/recheck';
+
+// Progress and diagnostics go to stderr; report payloads go to stdout, so
+// `--format json` and `github-actions` stay machine-readable.
+export function createCliLogger(): Logger {
+  return {
+    log: (line) => void logger.info(`${line}\n`),
```
tatomyr: Please do not introduce needless wrappers, just use the existing logger where you need it.

docs/@v2/commands/recheck.md
adamaltman: Rewrote the introduction in 2c3fe6d4. It now says where presets go and what the block does, in that order.

docs/@v2/commands/recheck.md
adamaltman: Removed in 2c3fe6d4, together with the other release notes on the page.

docs/@v2/commands/recheck.md
adamaltman: Rewrote the baseline example in 2c3fe6d4 in plain terms.

adamaltman:
Re-read both docs pages and rewrote them in 2c3fe6d4: plain introductions, no release notes, examples in the same shape as the `lint` page, and the not-yet-active `apiDescriptions` option removed from the reference.

adamaltman:
Rebased onto the current #3080 head and added the `@redocly/config` 0.56.0 pin in 03cf1afc.
`check-config` now accepts the `recheck` block, the skipped config-lint test is on, a `check-config` e2e fixture covers a valid block, and the two snapshots lost the transient warning line.
Nothing is deferred any more; the PR body reflects the final state.

packages/cli/src/commands/recheck/args.ts
```diff
@@ -0,0 +1,49 @@
+import type { VerifyConfigOptions } from '../../types.js';
+
+export type RecheckFormat = 'table' | 'json' | 'sarif' | 'github-actions';
+
+export type RecheckArgv = {
+  paths?: string[];
+  format: RecheckFormat;
+  'output-path'?: string;
+  severity?: 'off' | 'info' | 'warn' | 'warning' | 'error';
```
tatomyr: What do we need `info` and `warning` levels for?

packages/cli/src/index.ts
```diff
@@ -373,6 +375,90 @@ yargs(hideBin(process.argv))
...
+          severity: {
+            description: 'Minimum severity to run.',
+            choices: ['off', 'info', 'warn', 'warning', 'error'] as ReadonlyArray<
+              'off' | 'info' | 'warn' | 'warning' | 'error'
+            >,
+            type: 'string',
+          },
+          tags: { description: 'Run only rules with these tags.', array: true, type: 'string' },
+          rule: { description: 'Run only these rules.', alias: 'r', array: true, type: 'string' },
+          'exclude-rule': { description: 'Skip these rules.', array: true, type: 'string' },
```
tatomyr: We already have a similar option in lint, `skip-rule`. What if we use the existing name?

packages/cli/src/index.ts
```diff
@@ -373,6 +375,90 @@ yargs(hideBin(process.argv))
...
+          },
+          severity: {
+            description: 'Minimum severity to run.',
+            choices: ['off', 'info', 'warn', 'warning', 'error'] as ReadonlyArray<
+              'off' | 'info' | 'warn' | 'warning' | 'error'
+            >,
+            type: 'string',
+          },
+          tags: { description: 'Run only rules with these tags.', array: true, type: 'string' },
+          rule: { description: 'Run only these rules.', alias: 'r', array: true, type: 'string' },
```
tatomyr, line 462: We don't have this CLI option in lint, only through config file. Do you think it's a good idea to introduce a different interface for recheck?

packages/cli/src/index.ts
```diff
@@ -373,6 +375,90 @@ yargs(hideBin(process.argv))
...
+          'exclude-rule': { description: 'Skip these rules.', array: true, type: 'string' },
+          stats: { description: 'Print rule statistics.', alias: 's', type: 'boolean' },
+          fix: { description: 'Apply fixes to Markdown files.', alias: 'f', type: 'boolean' },
+          'annotations-limit': {
```
tatomyr: Again, in the lint command the same option is named `max-problems`.

packages/cli/src/index.ts
```diff
@@ -373,6 +375,90 @@ yargs(hideBin(process.argv))
...
+            description: 'Print a run summary.',
+            choices: ['json', 'text'] as ReadonlyArray<'json' | 'text'>,
+            type: 'string',
+          },
+          'summary-path': { description: 'Write the summary to a file.', type: 'string' },
+          'changed-only': {
+            description: 'Lint only files listed by --changed-list.',
+            type: 'boolean',
+          },
+          'changed-list': {
```
tatomyr: Why do we need the changed files list? If we want to scope the check to what has been changed, we can simply pass the changed `paths`, cannot we?

packages/cli/src/commands/recheck/index.ts
```diff
@@ -0,0 +1,133 @@
...
+      { from: argv.from ?? [], out: argv.out ?? '', check: argv.check },
+      engineLogger
+    );
+    if (exitCode !== 0) process.exitCode = exitCode;
+    return;
+  }
+
+  const block = config.resolvedConfig.recheck;
+  let presets = config.resolvedConfig.recheckExtends ?? [];
+  if (block == null && presets.length === 0) {
```
tatomyr, line 74: Wouldn't this logic contradict the existing configuration behaviour? For example, if there's no `redocly.yaml`, the lint command applies a default configuration, but if the config file exists and there's no `rules`/`extends` section, it doesn't apply anything. 

Moreover, it seems that with the proposed approach, it would be difficult to switch the checks off if they are not needed.

packages/core/src/config/config-resolvers.ts
```diff
@@ -115,13 +116,21 @@ export async function resolveConfig({
...
+
   const bundledConfig = bundleConfig(
     rootDocument,
     deepCloneMapWithJSON(resolvedRefMap),
     resolvedPlugins,
     skipPluginEval
   );
 
+  if (recheckExtends.length > 0) {
+    bundledConfig.recheckExtends = recheckExtends;
```
tatomyr: Please avoid modifying openapi-core if possible (especially, mutating the bundled config seems suspicious). Could this be filtered on the `cli` package's side?

tsconfig.json
```diff
@@ -25,7 +25,8 @@
       "@redocly/client-generator/contracts/*": ["./packages/client-generator/src/contracts/*.ts"],
       "@redocly/client-generator/runtime-sources": [
         "./packages/client-generator/src/runtime-sources.ts"
-      ]
+      ],
+      "@redocly/recheck": ["./packages/recheck/src/index.ts"]
```
tatomyr, line 23: Probably has to be adjusted in `tsconfig.build.json` too. Oh, and the new package should be included into the `local-pack.sh` script.

packages/cli/src/commands/recheck/__tests__/cli-logger.test.ts
```diff
@@ -0,0 +1,25 @@
+import { logger } from '@redocly/openapi-core';
+import { describe, expect, it, vi } from 'vitest';
+
+import { createCliLogger } from '../cli-logger.js';
+
+describe('createCliLogger', () => {
```
tatomyr: Testing logger doesn't have too much sense, especially if you reuse the existing one.

packages/core/src/config/__tests__/recheck-namespace.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  extends:
+    - recommended
+    - recheck/markdown
+  recheck:
+    rules:
+      recheck/line-length: off
+    baseline: ./recheck-baseline.yaml
+`;
+
+async function load(yaml: string) {
```
tatomyr: Please avoid creating needless wrappers.

packages/core/src/config/utils.ts
```diff
@@ -12,6 +12,11 @@ import { assignOnlyExistingConfig, assignConfig } from '../utils/assign-config.j
 import { isPlainObject } from '../utils/is-plain-object.js';
 import type { ImportedPlugin, ResolvedGovernanceConfig, Plugin, PluginCreator } from './types.js';
 
+// `recheck/*` presets belong to the prose engine, not to API rule plugins.
```
tatomyr: The code is self-explanatory, the comment only distracts here.

packages/cli/src/commands/lint.ts
```diff
@@ -141,12 +141,33 @@ export async function handleLint({
   }
 }
 
+// formatProblems only understands these values; other commands' own report
+// formats (e.g. recheck's `table`, `sarif`) are not config-lint formats.
```
tatomyr: Please avoid comments that either blatantly duplicate what is stated in the code itself or are negations of what the code doesn't do.

tatomyr (commented):
Had a quick look. My main concern is that the recheck interface/behaviour diverges from lint for no obvious reason. Please consider aligning them more closely.

packages/cli/src/commands/recheck/cli-logger.ts
adamaltman: Removed in 04fcfc58.
The handler now builds the engine's logger object inline from the existing `logger`.
The engine needs the four-channel shape, so that object literal is all that remains.

packages/cli/src/commands/recheck/args.ts
adamaltman: Dropped in 04fcfc58.
`--severity` now accepts `warn` and `error` only.
`warning` was an alias, and `info` as a filter meant "run everything", which is the default.

packages/cli/src/index.ts
adamaltman: Renamed to `--skip-rule` in 04fcfc58.

packages/cli/src/index.ts
adamaltman: Renamed to `--max-problems` in 04fcfc58.

packages/cli/src/index.ts
adamaltman: Agreed; both `--changed-only` and `--changed-list` are gone in 04fcfc58.
Passing the changed paths as arguments covers the case.

packages/cli/src/commands/recheck/index.ts
adamaltman: Changed in 04fcfc58 to follow `lint`.
With no `redocly.yaml`, the command uses `recheck/markdown`.
With a `redocly.yaml` that has no recheck configuration, it checks nothing and says so.

packages/core/src/config/config-resolvers.ts
adamaltman: The mutation is gone in 04fcfc58; the resolver now returns a new object.
The `extends` filter itself has to stay in core.
Without it, `lint` fails on `recheck/markdown` in `extends` before the CLI ever sees the config.
That split is the design Roman approved in Redocly/redocly#26970: core knows the `recheck/` prefix and nothing else about the engine.

tsconfig.json
adamaltman: `tsconfig.build.json` already lists `packages/recheck` as a project reference and carries no `paths` map.
`scripts/local-pack.sh` now packs the engine and rewires the CLI to the tarball, in 04fcfc58.
The CLI-side tarball rewrite came out again in bc166df2, because the CLI publish manifest has no dependencies.

packages/cli/src/commands/recheck/__tests__/cli-logger.test.ts
adamaltman: Removed with the wrapper in 04fcfc58.

packages/core/src/config/__tests__/recheck-namespace.test.ts
adamaltman: Removed in 04fcfc58; the tests call `createConfig` directly.

packages/core/src/config/utils.ts
adamaltman: Removed in 04fcfc58.

packages/cli/src/commands/lint.ts
adamaltman: Removed in 04fcfc58.

packages/cli/src/index.ts
adamaltman: Kept.
Recheck rules are many and cheap to run one at a time, so `--rule` is the quickest way to try one on a folder.
Adam decided to keep it.

tsconfig.json
tatomyr: Aha, I see. Missed that in another PR.

packages/recheck/src/actions/logger.ts
```diff
@@ -1,33 +1,40 @@
...
+  const outputs: string[] = [];
   return {
     lines,
     warnings,
     errors,
+    outputs,
     log: (line) => void lines.push(line),
     warn: (line) => void warnings.push(line),
     error: (line) => void errors.push(line),
+    output: (line) => void outputs.push(line),
```
tatomyr, line 38: To me, the new output lands perfectly on the existing log. Is there any difference I'm missing?

packages/cli/src/wrapper.ts
```diff
@@ -116,7 +117,9 @@ export function commandWrapper<T extends CommandArgv>(
       if (typeof commandHandler === 'function') {
         await commandHandler({ argv, config, version, collectSpecData, collectResults });
       }
-      code = 0;
+      // A handler that returns instead of throwing signals its own exit code
+      // through `process.exitCode` (e.g. recheck); default to success.
+      code = process.exitCode === 1 || process.exitCode === 2 ? process.exitCode : 0;
```
tatomyr: Please do not change the existing behaviour. It's deliberate.

packages/core/src/config/utils.ts
```diff
@@ -12,6 +12,10 @@ import { assignOnlyExistingConfig, assignConfig } from '../utils/assign-config.j
 import { isPlainObject } from '../utils/is-plain-object.js';
 import type { ImportedPlugin, ResolvedGovernanceConfig, Plugin, PluginCreator } from './types.js';
 
+export function isRecheckPreset(name: string): boolean {
+  return name.startsWith('recheck/');
```
tatomyr, line 16: One concern here: if someone accidentally names their ruleset `recheck/...`, this would interfere with the recheck extension. It's an edge case, but still no guarantee that it won't happen. We have to mention this quirk in the `lint` docs to prevent confusion.

packages/recheck/skills/recheck-config/SKILL.md
```diff
@@ -53,14 +55,20 @@ Configure recheck for a project the way its maintainers would: measure first, de
 When the corpus has too many errors to fix at once, record them and gate only new ones:
 
 ```bash
-npx recheck --generate-baseline
+redocly recheck --generate-baseline
+```
+
+Commit `recheck-baseline.yaml` and add it to the `recheck` block:
```
tatomyr: I'd prefer it to be a hidden file, for it is an analogue of `.redocly.lint-ignore.yaml`. And the `redocly` prefix too.

packages/recheck/skills/recheck-lint/SKILL.md
```diff
@@ -10,21 +10,22 @@ Your markdown must pass it, the same as code must pass its tests.
 
 ## When this applies
 
-A `recheck.yaml` (or `recheck.yml`) exists in the project root or above the files you touched.
-If none exists, this skill does not apply.
+Look for a `redocly.yaml` (or `redocly.yml`) in the project root or above the files you touched.
```
tatomyr: We agreed some time ago that the only default config file name must be `redocly.yaml` for consistency with Realm. So `redocly.yml` is just a random YAML file unless passed via the `--config` option.

packages/core/src/config/config-resolvers.ts
tatomyr: I'd prefer to introduce the ability to make the `extends` general and command agnostic instead of adding yet another branching in the config resolver, but I have to play a bit with the config architecture myself first. Will try to do that this week.

adamaltman: Agreed, we wait for your exploration.
The current filter stays until then, so the branch keeps working.

packages/recheck/src/actions/logger.ts
adamaltman: The difference is the destination.
The CLI sends `output` to stdout and the other three channels to stderr.
That keeps `--format json > report.json` free of progress lines.

packages/cli/src/wrapper.ts
adamaltman: Reverted in ebc910f8.
The handler now throws `AbortFlowError` on a failed run, the same path `lint` uses.

packages/core/src/config/utils.ts
adamaltman: Fixed in ebc910f8.
Core now reserves the plugin ids `recheck`, `redocly`, `redoc`, `realm`, and `reunite`.
A plugin that declares one of them fails to load with a clear message.
The `extends` and custom plugins pages say so.

packages/recheck/skills/recheck-config/SKILL.md
adamaltman: Changed in ebc910f8.
The default baseline is now `.recheck-baseline.yaml`, a hidden file next to `redocly.yaml`.
We left out the `.redocly.` prefix; it dates from the time the config file was `.redocly.yaml`.

packages/recheck/skills/recheck-lint/SKILL.md
adamaltman: Fixed in ebc910f8.
The skill now names `redocly.yaml` only.

packages/cli/src/commands/recheck/args.ts
```diff
@@ -0,0 +1,47 @@
+import type { VerifyConfigOptions } from '../../types.js';
```
AlbinaBlazhko17, line 1: Could you please refactor the structure a bit? 
Let's move types inside the separate `types.ts` file and `selectAction` function inside the separate file named `select-action.ts` and then you can rename test file to `select-action.test.ts`.

packages/cli/src/commands/recheck/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  try {
+    if (!statSync(path).isFile()) return false;
+    detectSpec(parseYaml(readFileSync(path, 'utf8')));
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function lintOptions(argv: RecheckArgv): LintOptions {
```
AlbinaBlazhko17, line 34: I'd prefer `toLintOptions`.

packages/cli/src/index.ts
```diff
@@ -424,6 +426,80 @@ yargs(hideBin(process.argv))
...
+            type: 'string',
+          },
+          out: { description: 'Output file for the generated schema.', type: 'string' },
+          check: {
+            description: 'Fail if the generated schema differs from --out.',
+            type: 'boolean',
+          },
+        }),
+    (argv) => {
+      commandWrapper(handleRecheck)(argv);
```
AlbinaBlazhko17, line 500: Could you please add lazy import, because now we have performance degradation in `check-config` command: https://github.com/AlbinaBlazhko17/redocly-cli/pull/7#issuecomment-5636982289. 
```ts
const { handleRecheck } = await import('./commands/recheck/index.js');
      commandWrapper(handleRecheck)(argv);
```

packages/cli/src/commands/recheck/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    stats: argv.stats,
+    fix: argv.fix,
+    annotationsLimit: argv['max-problems'],
+    summary: argv.summary,
+    summaryPath: argv['summary-path'],
+  };
+}
+
+export async function handleRecheck({ argv, config }: CommandArgs<RecheckArgv>): Promise<void> {
+  const engineLogger: Logger = {
```
AlbinaBlazhko17, line 51: I was thinking how we can improve logger logic inside the recheck and better solution for this is to keep `recheck` package only for the main computations and leave for CLI the logging output. The `recheck` will just return data which CLI will print. In this case we can remove this wrapper, `isOutputFormat` function and checks inside the `lint.ts` file, remove `RecheckArgv` from the general `types.ts` and move `reporter/formats` to the `packages/cli/src/commands/recheck/formatters` with the `index.ts` file and `table.ts`,  `json.ts`, etc (the same as `scorecard-classic` did). @adamaltman @tatomyr wdyt?

packages/cli/scripts/build.mjs
```diff
@@ -17,6 +17,8 @@ const result = await build({
   platform: 'node',
   format: 'esm',
   target: 'node20.19',
+  // The engine imports these spell-check packages dynamically; a user installs them on demand.
+  external: ['nspell', 'dictionary-en'],
```
AlbinaBlazhko17, line 21: @adamaltman could you please explain why you decided to keep those two packages as a peer dependency?
I tested the case when we use those packages inside the deps and it has no performance impact: https://github.com/AlbinaBlazhko17/redocly-cli/pull/7/changes#diff-0ba29bdf689d78abba7ff7ecf6f2d5520d8ba5f3e0d3a3c842be64e265533847.

packages/core/src/config/utils.ts
tatomyr: Yes, but this check is still only about 'recheck/' and `isReservedPluginId` is a separate thing.
Also, what do you need to reserve other words for? I bet a couple of our users name their plugin folders `redocly` 🙂, so it could be a breaking change for some.

Taking into account [this comment](https://github.com/Redocly/redocly-cli/pull/3082/changes#r4017393851), I suggest the recheck presets should be of the same structure as our current presets (which are basically plugins), with the extra `recheck` property exported instead of existing `rules`, `decorators`, etc..

packages/recheck/src/actions/logger.ts
tatomyr: It creates divergence from our current approach. Please follow our convention described in CONTRIBUTING.md (section `Logging`) otherwise it will confuse our existing users. You can check how the lint command handles JSON formatting. It's the same concept in essence.

packages/recheck/README.md
```diff
@@ -14,11 +12,11 @@ extends:
 recheck:
   rules:
     recheck/line-length: off
-  baseline: ./recheck-baseline.yaml
+  baseline: ./.recheck-baseline.yaml
```
tatomyr: Why do we need to specify the baseline file in the config? Isn't it an autogenerated file similar to `redocly.lint-ignore.yaml`?

docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,139 @@
...
+```bash
+redocly recheck docs --generate-baseline
+```
+
+The command writes `.recheck-baseline.yaml` next to `redocly.yaml`.
+Add the file to the `recheck` block:
+
+```yaml
+recheck:
+  baseline: ./.recheck-baseline.yaml
```
tatomyr: Why add it manually? The lint ignore file is picked up merely by virtue of being in the project. I suggest we keep the same approach for baseline.

docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,139 @@
...
+### Use a baseline
+
+A baseline records the current errors.
+Later runs report only errors that the baseline does not list.
+
+```bash
+redocly recheck docs --generate-baseline
+```
+
+The command writes `.recheck-baseline.yaml` next to `redocly.yaml`.
```
tatomyr:
```suggestion
The command writes `.redocly.recheck-baseline.yaml` next to `redocly.yaml`.
```

So it becomes similar to `.redocly.lint-ignore.yaml`

packages/cli/src/commands/recheck/__tests__/handle-recheck.test.ts
```diff
@@ -0,0 +1,189 @@
+import { logger, type Config } from '@redocly/openapi-core';
+import { existsSync, mkdtempSync, writeFileSync, mkdirSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest';
+
+import { AbortFlowError } from '../../../utils/error.js';
+import { handleRecheck } from '../index.js';
+
+function fakeConfig(
```
tatomyr: These tests don't work well as unit tests, since it's hard to tell what they're actually testing. Converting them to e2e tests would make the use cases clearer.

packages/recheck/README.md
```diff
@@ -14,11 +12,11 @@ extends:
 recheck:
   rules:
```
tatomyr, line 13: Haven't we agreed [here](https://github.com/Redocly/redocly/issues/26423) that the rules should go directly inside the `recheck` section? It'd be similar to the current approach (for instance, all GraphQL rules go under the top-level `graphqlRules`, AyncAPI rules under `async3Rules`, and so on) and easier to integrate with the current architecture.

packages/core/src/config/config-resolvers.ts
```diff
@@ -115,18 +118,25 @@ export async function resolveConfig({
     resolvedPlugins = [...plugins, defaultPlugin];
   }
 
+  // Read before bundling: the bundler deletes `extends` from the root node in place.
+  const rootExtends = isPlainObject<RawUniversalConfig>(config) ? (config.extends ?? []) : [];
```
tatomyr, line 122: Instead of changing the config resolver, it most likely should be adjusted in `mergeExtends` (packages/core/src/config/utils.ts) and in the `Config` class itself (perhaps by adding a new global `recheck` section along with `rules`/`decorators`/`preprocessors`).

packages/cli/src/commands/recheck/args.ts
adamaltman: Done in 7e1138f5.
Types live in `types.ts`, `selectAction` in `select-action.ts`, and the test is `select-action.test.ts`.

packages/cli/src/commands/recheck/index.ts
adamaltman: Renamed to `toLintOptions` in 7e1138f5.

packages/cli/src/index.ts
adamaltman: Done in 7e1138f5.
The recheck handler loads through `await import`, the same way `introspect-mcp` does.

packages/cli/scripts/build.mjs
adamaltman: Agreed, and bundled in 32add50b.
They were externals so the engine's optional spell-check stayed on demand, but the CLI manifest carries no runtime dependencies, so a user could never get them.
Bundling needed one addition: `dictionary-en` reads its data files relative to its own module URL, so the build copies `index.aff` and `index.dic` next to the chunk esbuild picks.
The bundle grows by about 0.6 MB and the spell-check chunk stays lazy.

docs/@v2/commands/recheck.md
adamaltman: Done in 6d6bc972.
A `.redocly.recheck-baseline.yaml` next to `redocly.yaml` is picked up by presence.
The `baseline` key stays as an override for another path.

docs/@v2/commands/recheck.md
adamaltman: Renamed in 6d6bc972; the docs now use `.redocly.recheck-baseline.yaml`.

packages/cli/src/commands/recheck/__tests__/handle-recheck.test.ts
adamaltman: Agreed.
47772137 replaces the handler unit tests with e2e fixtures under `tests/e2e/recheck/`, one directory per use case with a snapshot.
The stacked API-description PR gets the same treatment.

packages/recheck/src/actions/logger.ts
adamaltman: The CLI maps the engine channels onto the core logger: `output` goes to `logger.output`, the rest to `logger.info`, `logger.warn`, and `logger.error`.
If the ask is that the engine imports the core logger directly, that makes `@redocly/recheck` depend on `@redocly/openapi-core`.
We weigh that together with the config structure question below.

packages/core/src/config/utils.ts
adamaltman: Fair point on `redocly` as an existing plugin id.
A reservation of `recheck` only, folded into one check, is on the table.
We settle it with the config structure question.

packages/cli/src/commands/recheck/index.ts
adamaltman: Thank you, that is a clean split.
It moves the report formats and the engine logger out of `@redocly/recheck`, which also serves the LSP and the monorepo until the cutover.
Adam decides where the formats live, and the answer follows here.

packages/recheck/README.md
adamaltman: Do you mean `recheck` as a flat rule map, like `graphqlRules`, or a new top-level key such as `recheckRules`?
The block also holds `excludes`, `baseline`, `markdoc`, and `apiDescriptions`, and `@redocly/config` 0.56 already publishes this shape.

packages/core/src/config/config-resolvers.ts
adamaltman: This changes the split from Redocly/redocly#26970, where core knows the `recheck/` prefix and nothing else about the engine.
Adam, Roman, and you should settle it in one place, and I follow the outcome.

adamaltman:
The red `e2e` check is two `respect` tests that call the live Cafe API (`cafe.redocly.com`).
They fail the same way on a build of `main` at 93c155fa, so the cause is upstream: the request now gets a 400 with a new required field, and the `accept` header lost `application/problem+json`.
This branch changes nothing under `respect`.

packages/recheck/skills/recheck-config/SKILL.md
tatomyr: Not necessarily.  The hidden file here makes sense because users aren't supposed to interact with it as it's an autogenerated file whilst `redocly.yaml` is a visible file because users have to configure it manually. Also, leaving the same `.redocly.` prefix makes all three files tied to the same tool -- Redocly CLI without users scratching their heads and having to look into the docs.

packages/recheck/README.md
tatomyr: I'm not certain on that, but for consistency with the existing interfaces, it sounds like it should be a flat rule map (either `recheck` or `recheckRules`).
I'd avoid changing the `@redocly/config` package structure until we settle the interface down here. You can just  add the needed fields here in `redocly-yaml.ts`, and later we'll sync the `@redocly/config`.

@RomanHotsiy WDYT, do we need a nested structure here? 

```yaml
rules:
  some-lint-rule: error
decorators:
  some-decorator: on
recheck: 
  some-recheck-rule: warn
```

VS

```yaml
rules:
  some-lint-rule: error
decorators:
  some-decorator: on
recheck: 
  rules:
    some-recheck-rule: warn
```

The latter looks a bit off compared to the existing config.

packages/core/src/config/config-resolvers.ts
tatomyr: Okay, we've explored the config extension options on another PR ([adding the diff command](https://github.com/Redocly/redocly-cli/pull/2940)). It's currently in draft state, but the config changes are close to the final version. 

What should be done for Recheck (similar to the `diff` PR):
- In `packages/core/src/config/builtIn.ts`: add the Recheck preset and a default plugin section (without the nested specification version sections as that doesn't make sense for markdown).
- In `packages/core/src/config/config-resolvers.ts`: changes depend on the need for custom plugins support -- if we don't need them -- just skip the file changes (could be added later).
- In `packages/core/src/config/config.ts`: add the `recheck` section to the Config class (again, without the spec. version differentiations, thus without the need to override the base `recheck` section). Maybe add something similar to `getRulesSettings` but for Recheck (or reuse it).
- In `packages/core/src/config/constants.ts`: enhance  the default config with the Recheck preset (the name doesn't matter, no need for the `recheck/` prefix or reserved keywords anymore), and add the corresponding preset file similar to `packages/core/src/config/diff-recommended.ts` (probably could be imported from the `recheck` package itself).
- In `packages/core/src/config/types.ts`: adjust types.
- In `packages/core/src/config/utils.ts`: only apply the `assignConfig` in `mergeExtends` for the root recheck section.

This way the config extension would be consistent with the current approach.

docs/@v2/commands/recheck.md
tatomyr: Ideally there should be no way to override the ignore file, otherwise it creates divergence from the current approach in the lint command.

docs/@v2/commands/recheck.md
```diff
@@ -0,0 +1,142 @@
...
+| --max-problems            | number   | Maximum number of annotations in the report.                                                                                      |
+| --out                     | string   | Output file for the generated schema.                                                                                             |
+| --output-path             | string   | Write the report to this file instead of stdout. Applies to `--format json` and `sarif`.                                          |
+| --readability             | boolean  | Report readability scores instead of lint findings.                                                                               |
+| --rule                    | [string] | Run only these rules. Alias: `-r`.                                                                                                |
+| --severity                | string   | Run only rules at this severity or higher.<br />**Possible values:** `warn`, `error`.                                             |
```
tatomyr: Also it could be `off` (same as for `lint`).

packages/cli/src/commands/recheck/index.ts
```diff
@@ -0,0 +1,140 @@
...
+    log: (line) => void logger.info(`${line}\n`),
```
tatomyr, line 53: Just use `info` instead of `log` in your code, and this will eliminate the whole wrapper.

packages/cli/src/commands/recheck/cli-logger.ts
tatomyr: The wrapper still exists (see the [link](https://github.com/Redocly/redocly-cli/pull/3082/changes#r4073197763)). Let's get rid of it.

packages/cli/src/commands/recheck/types.ts
```diff
@@ -0,0 +1,26 @@
+import type { VerifyConfigOptions } from '../../types.js';
+
+export type RecheckFormat = 'table' | 'json' | 'sarif' | 'github-actions';
+
+export type RecheckArgv = {
+  paths?: string[];
+  format: RecheckFormat;
+  'output-path'?: string;
+  severity?: 'warn' | 'error';
```
tatomyr, line 9: What do we need `severity` on this level for? This is against the current `lint` behaviour.

adamaltman: `--severity error` runs the error rules only; the docs CI gate uses it to keep warnings out of the gate step.
`lint` has no such flag, and the exit code already fails on errors alone, so dropping it is on the table.
Adam decides with the other interface questions.

packages/cli/src/commands/recheck/index.ts
adamaltman: Understood.
The engine's channel is line based and the core logger is stream based, so the rename alone leaves the newline handling to settle.
It goes with the logger decision above; the adapter object is what remains until then.

docs/@v2/commands/recheck.md
adamaltman: Noted.
Dropping the `baseline` key means discovery only, the same as the ignore file, and it removes a key that `@redocly/config` 0.56 already publishes.
It joins the config-shape decision, so one change covers both.

packages/cli/src/commands/recheck/cli-logger.ts
adamaltman: Same answer as on the `info` thread: the four-line adapter stays until the logger decision, then it goes with it.

packages/core/src/config/config-resolvers.ts
adamaltman: Thank you for the plan; it is the concrete option now.
Adam decides between it and the Redocly/redocly#26970 split, and I follow the outcome here.

packages/recheck/skills/recheck-config/SKILL.md
adamaltman: Agreed, and in place since 6d6bc972: the default is `.redocly.recheck-baseline.yaml`, hidden and picked up by presence.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/recheck/README.md
adamaltman: Waiting for Roman's answer on the shape; the `@redocly/config` sync follows it.

adamaltman:
Folded into #3080 at 22c7c571; review continues there.

packages/core/src/config/config-resolvers.ts
adamaltman: Adam is fine with your plan if Roman agrees, since it reverses the split from Redocly/redocly#26970.
Roman, can you settle this and the rules shape below in one go?
The work then lands on #3080 after the logger rework.

packages/recheck/README.md
adamaltman: Adam leaves the shape to Roman.
One input: the block also holds `excludes`, `markdoc`, and `apiDescriptions`, so a flat map needs a second key for those, while the nested form keeps one block.

packages/cli/src/commands/recheck/index.ts
adamaltman: Adam picked this split.
The engine returns data, and the CLI owns the formatters and the logging; it lands on #3080 next.

packages/cli/src/commands/recheck/types.ts
adamaltman: Adam agreed; `--severity` goes, on #3080.

docs/@v2/commands/recheck.md
adamaltman: Adam agreed; the `baseline` key goes and only discovery stays, on #3080.

packages/cli/src/commands/recheck/index.ts
adamaltman: Landed on #3080 (969febe6..1c00cf77).
The engine's four actions return typed results, the report formats live in `packages/cli/src/commands/recheck/formatters/`, and `print.ts` prints through the core logger; the engine `Logger` is gone and no e2e snapshot changed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/recheck/src/actions/logger.ts
adamaltman: Done on #3080: the engine no longer logs at all.
Its actions return data, and the CLI prints through the core logger per CONTRIBUTING.md.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/cli/src/commands/recheck/index.ts
adamaltman: Gone on #3080, together with the engine logger it adapted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/cli/src/commands/recheck/cli-logger.ts
adamaltman: Gone on #3080; the engine has no logger left to wrap.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #3118, opened 2026-09-14
fix: resolve file paths from a $ref-ed config relative to that file

kanoru3101:
## What/Why/How?

Relative file paths written in a config file that is pulled in with `$ref` were resolved against the root `redocly.yaml` instead of the file they are written in.

```yaml
# redocly.yaml
$ref: ./docs/redocly.yaml
```

```yaml
# docs/redocly.yaml
apis:
  main:
    root: ./apis/openapi.yaml
```

`redocly lint main` reported `apis/openapi.yaml does not exist or is invalid`, because the path was looked up next to the root config. The documented split-configuration form, `apis: { $ref: ./config/apis.yaml }`, broke the same way.

- Node types now declare which of their fields hold a file path (`isFilePath`), so the bundler needs no list of field names and a new path field only has to be marked on its type.
- The config bundler rebases those fields when a node comes from a referenced file, reusing the source-awareness that `plugins` and `extends` already relied on.
- Covered: `root`, `output`, and `clientOutput` in `apis`, plus `setup` and `cliOutput` in `client`. `openapi.htmlTemplate` is declared in `@redocly/config`, so it needs the marker there and is not covered yet.
- Paths stay untouched when they are absolute or remote, and resolve to an absolute path when the config was loaded without a path of its own.

## Check yourself

## Security


kanoru3101:
@cursor review

packages/core/src/config/visitors.ts
```diff
@@ -93,6 +119,12 @@ export const configBundlerVisitor = normalizeVisitors(
             replaceRef(node, resolved, ctx);
           },
         },
+        // any node type can declare a file path, so each node is asked for its own type instead of listing types here
+        any: {
+          leave(node: unknown, ctx: UserContext) {
+            rebaseFilePaths(node, ctx);
```
tatomyr, line 123: So it works with $refs but not with other fields that could reference local files...

packages/core/src/ref-utils.ts
```diff
@@ -116,6 +116,21 @@ export function resolvePath(base: string, relative: string): string {
...
+
+export function rebaseFilePath(filePath: string, fromRef: string, toRef: string): string {
+  if (!filePath || isAbsoluteUrl(filePath) || path.isAbsolute(filePath)) {
+    return filePath;
+  }
+  const absolutePath = resolvePath(parentDir(fromRef), filePath);
+  if (isAbsoluteUrl(absolutePath) || !toRef) {
+    return absolutePath;
+  }
+  return path.relative(parentDir(toRef), absolutePath);
```
tatomyr, line 131: Shouldn't we check if the file exists? It only makes sense for a real files, not an example (theoretically, some property could be just an examplary file reference).

tests/e2e/lint/config-ref-relative-paths/docs/redocly.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+  no-empty-servers: error
+
+apis:
+  main:
```
tatomyr, line 5: Could you try more levels of $refs nesting?

packages/core/src/types/index.ts
```diff
@@ -7,6 +7,7 @@ export type ScalarSchema = {
   resolvable?: boolean;
   directResolveAs?: string;
   minimum?: number;
+  isFilePath?: boolean;
```
tatomyr, line 10: Is it possible to avoid changing the types?

packages/core/src/ref-utils.ts
tatomyr: ~Maybe we can just enlist all the potential properties that could bear the file names in config and replace only those?~ Ah, I see, you're doing that already.

packages/core/src/types/redocly-yaml.ts
```diff
@@ -382,7 +399,7 @@ const Client: NodeType = {
     runtime: { enum: ['inline', 'module'] },
     importExt: { enum: ['js', 'ts'] },
     goPackage: { type: 'string' },
-    cliOutput: { type: 'string' },
+    cliOutput: { type: 'string', isFilePath: true },
```
tatomyr, line 402: What about other fields that only live in the config package?

packages/core/src/config/visitors.ts
kanoru3101: Rebasing applies to every node that came in through a `$ref`, for the fields its type marks with `isFilePath`. Covered now: `root`, `output`, `clientOutput` in `apis`, and `setup`, `cliOutput` in `client`.

`openapi.htmlTemplate` is not covered: it is declared in `@redocly/config` inside a nested named type, and marking it from core would mean depending on generated type names. That one needs the marker in the schema itself (follow-up in the monorepo).

Fields written in the root `redocly.yaml` never needed rebasing — they are already relative to it.

packages/core/src/ref-utils.ts
kanoru3101: An existence check would break the outputs: `output`, `clientOutput` and `cliOutput` point at files that do not exist yet. It would also need `fs`, and this code runs in the browser too.

The `isFilePath` marker on the type is what replaces the check — only fields declared as paths are touched.

tests/e2e/lint/config-ref-relative-paths/docs/redocly.yaml
kanoru3101: Checked three directories deep, a `$ref` → `$ref` chain across directories, and a root config that is only a `$ref` to the nested one — all resolve correctly. They are covered in the unit fixture (`fixtures/resolve-refs-in-config/file-paths`); the e2e case stays minimal.

packages/core/src/types/index.ts
kanoru3101: Only by hardcoding a list of type name + field pairs in the visitor, which is what this replaces. A list breaks silently when a field is renamed, cannot see the types inside a `$ref`-ed subtree, and has to be extended by hand for every new path field.

Without a marker there is no general way to tell a path from any other string: `directResolveAs` loads the target (and `output` does not exist yet), the consumer side loses the origin once `forAlias` merges root and api config, and heuristics on `./` misfire.

`isFilePath` is the same kind of flag as the existing `isExample`, `resolvable` and `directResolveAs` on `ScalarSchema` — the type is where the walker already learns what a field is.

packages/core/src/types/redocly-yaml.ts
kanoru3101: `root` and `output` also live in `@redocly/config`, but they sit on the `apis` entry type that core already rebuilds, so `asFilePaths` marks them here.

Fields inside schema-derived nested types — `openapi.htmlTemplate` at both levels — cannot be marked from core without depending on generated type names. They need `isFilePath: true` in the schema itself; the adapter already passes unknown keys through, so once the schema has it no core change is needed. Planned as a separate PR in the monorepo.

packages/core/src/ref-utils.ts
kanoru3101: No: `output`, `clientOutput` and `cliOutput` point at files that don't exist yet, and core also runs in the browser without `fs`. The `isFilePath` marker is what limits this to real path fields.

packages/core/src/config/visitors.ts
kanoru3101: Any field marked `isFilePath` on its node type is handled, wherever the node sits: today `root`, `output`, `clientOutput`, `client.setup`, `client.cliOutput`, plus fields written next to a `$ref` in a referenced file. `extends` and `plugins` keep their own resolution. Fields declared in `@redocly/config` (`htmlTemplate`, `logo.*`) get the same marker there in a follow-up PR and need no core change.

packages/core/src/ref-utils.ts
kanoru3101: No: `output` and `clientOutput` point to files that do not exist yet, and a missing `root` should fail in the command with the usual error rather than keep a path that is wrong relative to the root config. The type marker decides, not the file system. After the latest push this file is no longer changed.

packages/core/src/types/index.ts
kanoru3101: I tried a field-name list in the bundler, JSON-schema `format`, and reusing `resolvable`/`directResolveAs`. Only the node type can say which strings are paths without hardcoding names, so `isFilePath` sits next to `isExample`/`resolvable`. It is the only type change now; the walker context is untouched.

packages/core/src/types/redocly-yaml.ts
kanoru3101: They get `isFilePath: true` in `@redocly/config` (`root`, `output`, `htmlTemplate`, `logo.image`, `logo.favicon`). The adapter already passes the key through, and the bundler rebases a shared node only once even when the walker reaches it under two generated type names, so no core change is needed. Once the CLI bumps that version, `asFilePaths` here goes away.

PR #3088, opened 2026-09-03
feat: add `@redocly/cli/api` entry point for `push` and `push-status`

vadyvas:
## What/Why/How?

adds the `@redocly/cli/api` entry point. It exports the `push` and `pushStatus` functions and their TypeScript types

- `src/api.ts` is a second esbuild entry. It declares every public type and the two functions. An import of this file does not start the CLI. It does not read `redocly.yaml` or send telemetry.
- `push.ts` and `push-status.ts` each have two functions now. `handlePush` and `handlePushStatus` take the yargs arguments (`CommandArgs<…>`). `push` and `pushStatus` take an options object and contain the logic.
- `build.mjs` creates one file, `lib/api.d.ts`, from `src/api.ts` with `ts.transpileDeclaration`. The build fails if this file contains an import.
- `package.json` gets an `exports` map with `./api` and `./package.json`. `prepare-publish-dir.mjs` copies the map to the published package.
- The response types that only the API client uses are now local to `api-client.ts`. `reunite/api/types.ts` is removed.

## Reference

- Closes #3087
- Redocly/reunite-push-action#138

## Check yourself

- [x] This PR follows the [contributing guide](https://github.com/Redocly/redocly-cli/blob/main/CONTRIBUTING.md#pull-request-guidelines) 
- [x] All new/updated code is covered by tests
- [ ] Core code changed? - Tested with other Redocly products (internal contributions only)
- [ ] New package installed? - Tested in different environments (browser/node)
- [x] Documentation update has been considered 

## Security

- [x] The security impact of the change has been considered 
- [x] Code follows company security practices and guidelines

---

> [!NOTE]
> **Medium Risk**
> Refactors deploy/push command internals and introduces a new public API surface for uploads and deployment polling; behavior should stay aligned with the CLI but programmatic callers depend on stable types and error handling.
> 
> **Overview**
> Adds a **programmatic** `@redocly/cli/api` export so CI and other Node consumers can call Reunite **push** and **push-status** without starting the CLI binary.
> 
> `package.json` maps `./api` to `lib/api.js` and `lib/api.d.ts`. The build bundles `src/api.ts` as a second esbuild entry and emits declarations from that file alone via `ts.transpileDeclaration`. Publish prep copies the `exports` field into the published package.
> 
[...]

PR #2447, opened 2025-12-05
fix: env `HTTP_PROXY` in `push` command

kanoru3101:
## What/Why/How?

Replaced HttpsProxyAgent (incompatible with native fetch) with `ProxyAgent` from `undici` to restore proxy support for `HTTP_PROXY` environment variables.

## Testing
Tested via local proxy and via online`HTTP_PROXY=http://user-viktor_d0fnq-country-US:SpwUXccQ7fJv_GA@dc.oxylabs.io:8000`
## Check yourself

- [ ] Code changed? - Tested with Redoc/Realm/Reunite (internal)
- [ ] All new/updated code is covered by tests
- [ ] New package installed? - Tested in different environments (browser/node)
- [ ] Documentation update considered

## Security

- [x] The security impact of the change has been considered
- [x] Code follows company security practices and guidelines

PR #3119, opened 2026-09-15
chore(reunite-integration): move reunite features to a separate package

### LINKED ISSUE #3087: Expose Reunite push and push-status as a public programmatic API with types

Since `@redocly/cli` 2.34.0 the published package is a minified esbuild bundle: `bin/cli.js`, `lib/index.js`, and hashed chunks. It has no `exports` field, no `lib/reunite/commands/*` modules, and no `.d.ts` files. `handlePush`, `handlePushStatus`, and the Reunite response types (`PushResponse`, `DeploymentStatus`, and so on) are internal to the bundle.

[reunite-push-action](https://github.com/Redocly/reunite-push-action) used to import these directly. Bumping the CLI broke it (Redocly/reunite-push-action#129, #131, #135). As a bridge, Redocly/reunite-push-action#138 now copies the CLI into the action's `dist/`, spawns `redocly push`, parses `Push ID:` from the text output, and re-implements `push-status --wait` against the Reunite API to keep GitHub commit statuses working. That means the action owns an endpoint contract the CLI is supposed to own, and it will drift.

Expose the Reunite commands as a public programmatic entry point again, for example a second esbuild entry published as `@redocly/cli/api`, with type declarations:

```ts
import { handlePush, handlePushStatus } from '@redocly/cli/api';
import type { PushResponse, PushStatusSummary, DeploymentStatus } from '@redocly/cli/api';
```

Requirements for the action:

- `handlePush` returns the push id.
- `handlePushStatus` keeps the `onRetry` callback (or equivalent) that reports the intermediate `commit.statuses`, because the action mirrors them to GitHub while a deployment runs.
- The entry is bundleable by esbuild without reading `package.json` at runtime (the version must stay inlined, as it is in the bin bundle today).

- `--format json` for `push` and `push-status`. Then the action becomes a thin wrapper around the binary and stops calling the API itself. Also acceptable, and simpler to keep stable, but the action still has to vendor or install the binary.
- Keep the current bridge in the action. Rejected as the long-term state: duplicated push-status logic and an API contract owned outside the CLI.

**Additional context**

- CLI 2.34.0 changelog: "Improved CLI install speed by bundling the CLI into a dependency-free package."
- Bridge implementation: Redocly/reunite-push-action#138 (`src/redocly-cli.ts`, `src/push-status.ts`, `src/types.ts`). These files should be deleted once the entry point ships.

tatomyr:
## What/Why/How?

- Added the `@redocly/reunite-integration` package that exposes the Reunite platform related code and types.
- Moved `push`, `pushStatus`, `login`, `logout`, `scorecardClassic` command handlers to the new package.
- Fixed the openBrowser method to validate the url being opened.
- Removed `exitWithError` (replaced with `HandledError` explicitly for clarity), `getProxyAgent`, `https-proxy-agent` (both became a dead code in https://github.com/Redocly/redocly-cli/pull/2447; @kanoru3101 please correct me if I'm wrong).

## Reference

Resolves https://github.com/Redocly/redocly-cli/issues/3087.

Supersedes https://github.com/Redocly/redocly-cli/pull/3088.

## Check yourself

## Security

---

> [!NOTE]
> **Medium Risk**
> Touches authentication, encrypted credential storage, and Reunite push/deploy flows, plus a broad CLI error-path refactor; behavior should stay equivalent via `commandWrapper`, but regressions in login/push or exit codes are the main risk.
> 
> **Overview**
> Introduces **`@redocly/reunite-integration`** as a published workspace package for Reunite API access, OAuth/credentials, and the handlers behind **`push`**, **`push-status`**, **`login`**, **`logout`**, and **`scorecard-classic`**. The CLI keeps yargs registration and thin wiring (e.g. resolving API entrypoints for scorecard-classic) while importing those handlers and auth helpers from the new package so GitHub Actions and other consumers can use them directly.
> 
> **Error handling** is aligned across CLI and the new package: local **`exitWithError`** is removed in favor of throwing **`HandledError`** (and **`AbortFlowError`** is exported from **`@redocly/openapi-core`**). Reunite handlers take a caller **`version`** for User-Agent instead of reading CLI package metadata internally.
> 
> **Security/maintenance**: OAuth device-flow browser open no longer shells out with a configurable URL—it validates **`https?`** and uses **`spawn`** with argument arrays. Unused **`https-proxy-agent`** / **`getProxyAgent`** are dropped from the CLI; HTTP in reunite-integration uses **undici** proxy support.
> 
> Release tooling (changesets fixed group, snapshot publish, **`tsconfig.build`**, vitest coverage) and docs are updated for the fifth package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f470743514cd87b80a1e0f46e26c4e79c76e6a0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/cli/src/commands/scorecard-classic.ts
```diff
@@ -0,0 +1,30 @@
...
+  const { path, alias } = apis[0];
+  const matchedAlias = getAliasOrPath(config, path).alias || alias;
+
+  await runScorecardClassic({
+    argv,
+    config,
+    version,
+    api: { path, alias: matchedAlias },
+    collectSpecData,
+  });
```
tatomyr: Check if printExecutionTime should be in this handler. See packages/reunite-integration/src/utils/execution-time.ts.

tatomyr:
@cursor review

packages/cli/src/commands/push.ts
```diff
@@ -0,0 +1,125 @@
+import { logger, pluralize, type OutputFormat } from '@redocly/openapi-core';
+import { collectFilesToPush, getApiKeys, getDomain, pushFiles } from '@redocly/reunite-integration';
```
tatomyr, line 2: Let's await imports from reunite-integration.

packages/reunite-integration/src/utils/__tests__/retry-until-condition-met.test.ts
```diff
@@ -11,48 +11,44 @@ describe('retryUntilConditionMet()', () => {
...
+  it('fails when the condition is not met within the timeout', async () => {
     const operation = vi.fn().mockResolvedValue({ status: 'pending' });
 
     await expect(
       retryUntilConditionMet({
         operation,
         condition: (result: any) => result?.status === 'done',
-        retryIntervalMs: 100,
-        retryTimeoutMs: 1000,
+        retryIntervalMs: 10,
```
tatomyr: Why the timing was changed?

packages/reunite-integration/src/scorecard-classic/__tests__/fetch-scorecard.test.ts
```diff
@@ -241,41 +219,4 @@ describe('fetchRemoteScorecardAndPlugins', () => {
       })
     );
   });
-
-  it('should handle verbose flag and fetch plugins successfully', async () => {
```
tatomyr, line 245: WHy it was removed?

packages/reunite-integration/src/scorecard-classic/__tests__/plugin-evaluator.test.ts
```diff
@@ -52,27 +52,8 @@ describe('evaluatePluginsFromCode', () => {
       export default [() => ({ id: 'test', dirname })];
     `;
 
-    const result = await evaluatePluginsFromCode(pluginCodeWithDirname, false, '/some/config/dir');
+    const result = await evaluatePluginsFromCode(pluginCodeWithDirname, '/some/config/dir');
     expect(result).toHaveLength(1);
     expect((result[0] as any).dirname).toBe('file:///some/config/dir/');
   });
-
-  it('should handle verbose flag', async () => {
```
tatomyr, line 60: Same here.

packages/reunite-integration/src/utils/proxy-env.ts
```diff
@@ -0,0 +1,35 @@
+// Copied from @redocly/cli.
```
tatomyr, line 1: Please check if we can handle the proxy part on the CLI side, so we don't need to copypaste it here. Maybe passing some parameters down could help with that.

packages/cli/src/commands/scorecard-classic/index.ts
```diff
@@ -108,30 +119,106 @@ export async function handleScorecardClassic({
...
+      logger.info(
+        `Found ${levelProblems.length} ${pluralize('problem', levelProblems.length)} for level "${level.name}".\n`
+      );
+    }
+  }
+
+  reportResults({ path, result, targetLevel, format: argv.format, version, startedAt });
+}
+
+async function fetchScorecard({
```
tatomyr, line 166: ~Looks like it's better to move this to the new package.~
Ah, okay, I see it's using a function from there.

packages/reunite-integration/src/scorecard-classic/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { ScorecardConfig } from '@redocly/config';
+import type { NormalizedProblem, ResolvedConfig } from '@redocly/openapi-core';
+
+export type ScorecardProblem = NormalizedProblem & { scorecardLevel?: string };
+
+export type RemoteScorecardAndPlugins = {
+  scorecard: ScorecardConfig;
+  plugins: string | undefined;
+  pluginsUrl: string | undefined;
```
tatomyr, line 9: There were no `pluginsUrl` in the original type: https://github.com/Redocly/redocly-cli/pull/3119/changes?w=1&show-viewed-files=false#diff-51f52728f54bfd589dbc22955f41ddf9c3ab4e9ebed14d9632a36b1a52b0f626L19. Could you check what's changed?

packages/cli/src/commands/scorecard-classic/types.ts
```diff
@@ -1,28 +1,13 @@
...
+import type { OutputFormat } from '@redocly/openapi-core';
 
 export type ScorecardClassicOutputFormat =
   | Extract<OutputFormat, 'stylish' | 'json' | 'checkstyle'>
   | 'junit';
 
 export type ScorecardClassicArgv = {
-  api: string;
-  config: string;
+  api?: string;
```
tatomyr, line 8: Maybe make it required in yargs?

packages/reunite-integration/src/utils/__tests__/retry-until-condition-met.test.ts
DmitryAnansky: Those values are just test timing knobs, I will bring old value back it does not really effect the outcome.

packages/reunite-integration/src/scorecard-classic/__tests__/fetch-scorecard.test.ts
DmitryAnansky: The test was dropped because its one distinguishing input, `verbose: true` is no longer exists on the package function.
Everything else it checked is already covered by another test in the same file.

packages/reunite-integration/src/scorecard-classic/__tests__/plugin-evaluator.test.ts
DmitryAnansky: Same as above, but let me actually add tests to cover verbose flag, just updated version ...

packages/reunite-integration/src/utils/proxy-env.ts
DmitryAnansky: The CLI keeps its copy whatever we do, because most of its users are not Reunite calls. scorecard-classic uses plain fetch with no proxy on both main and this branch.

Why not pass a dispatcher down. The package exists so the GitHub action and other programs can call Reunite without the CLI. On main the action got proxy support by spawning the CLI. If the package stops reading the proxy environment, the action has to grow its own copy of this code, or silently lose proxy support on self-hosted runners. The change would also add an undici type to five public signatures, pushFiles, getPushStatus, waitForDeployment, the OAuth client and the device flow, for a concern every caller wants handled the same way. And only the 40 lines of environment parsing would leave. The connect-timeout half of the wrapper stays in the package regardless.

The clean removal exists, but it needs undici 7, and we had some issues with integrations, but can return to it in a separate pr.

DmitryAnansky: I would keep it as it is for now.

packages/reunite-integration/src/scorecard-classic/types.ts
DmitryAnansky: The field is new on purpose. 
As we removed logger from the package part.
It carries one piece of information the CLI lost when the printing moved out of the package.

It needs to tell three cases apart:

plugins configured and fetched, printed as "Successfully fetched plugins from <url>"
plugins configured but the fetch failed, printed as "No plugins were loaded from <url>"
no plugins configured, printed as "No custom plugins configured for this scorecard"

On main the same value was read from project.config.pluginsUrl inside the function.

packages/cli/src/commands/scorecard-classic/types.ts
DmitryAnansky: This change to `scorecard-classic <api>` would break redocly scorecard-classic with no argument for anyone who has apis in their config and relies on the fallback. That flow works on main today.

On `main` the yargs command is already scorecard-classic [api], square brackets, so the positional is optional there too.
The handler on main passes `argv.api ? [argv.api] : []` into `getFallbackApisOrExit`, which falls back to the apis defined in `redocly.yaml` and takes the first entry. 
So `api: string` on main was wrong; `argv.api` is `undefined` whenever the user relies on the config.
The branch changed only the type, not the behavior.

kanoru3101:
@tatomyr Yes, `getProxyAgent`, and  `https-proxy-agent` are dead code

packages/cli/src/commands/push.ts
```diff
@@ -0,0 +1,125 @@
...
+          'file',
+          files.length
+        )} uploaded to organization ${organization}, project ${project}. Push ID: ${pushId}.`
+      );
+    }
+
+    return { pushId };
+  } catch (err) {
+    handleReuniteError('✗ File upload failed.', err);
+  }
```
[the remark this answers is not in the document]
tatomyr: @DmitryAnansky  could you take a look ☝️?

RomanHotsiy (approved):

.changeset/login-open-browser.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+Fixed the `login` command so it no longer waits for the browser launcher to exit and no longer passes the authorization URL through a shell.
```
JLekawa:
```suggestion
Updated the `login` command to prevent it from passing the authorization URL through a shell.

The `login` command no longer waits for the browser launcher to exit.
```

packages/cli/src/commands/push.ts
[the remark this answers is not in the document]
DmitryAnansky: Fixed, issue was there before current changes.

LukassF:
No conflicts with https://github.com/orgs/Redocly/projects/35/views/1?pane=issue&itemId=223481357&issue=Redocly%7Credocly%7C25764. Reference in [thread](https://redoc-ly.slack.com/archives/C060SJ1TSSG/p1789734452626789).

JLekawa (approved):

thunderquake (approved):

sobanieca-redocly (approved):
Cyberpunks ownership looks good.

packages/reunite-integration/src/auth/oauth-client.ts
```diff
@@ -19,7 +20,7 @@ export class RedoclyOAuthClient {
   private readonly key: Buffer;
   private readonly iv: Buffer;
 
-  constructor() {
+  constructor(private readonly version: string = DEFAULT_CLI_VERSION) {
```
AlbinaBlazhko17: Do we need to set in every constructor `version` to default or only inside the `ReuniteApi` class?

packages/reunite-integration/src/auth/device-flow.ts
AlbinaBlazhko17: Looks like you miss to remove these logger messages. And `logger` appeared in different places, could we totally remove all of them?
Image

packages/cli/src/utils/telemetry.ts
```diff
@@ -84,8 +83,8 @@ export async function sendTelemetry({
       $0: _,
       ...args
     } = argv as Exact<Arguments<CommandArgv>>;
-    const { RedoclyOAuthClient } = await import('../auth/oauth-client.js');
-    const oauthClient = new RedoclyOAuthClient();
+    const { RedoclyOAuthClient, getReuniteUrl } = await import('@redocly/reunite-integration');
```
AlbinaBlazhko17, line 86: Could cause the performance degradation, let's release a snapshot and i will test this.

.changeset/sunset-warning-once.md
```diff
@@ -0,0 +1,5 @@
+---
+'@redocly/cli': patch
+---
+
+Fixed the `push-status --wait` and `push --wait-for-deployment` commands printing the same Reunite API sunset warning more than once.
```
AlbinaBlazhko17: Are you sure about this statement? Because: 
`handlePush` still calls `printSunsetWarning('push', …)` after `handlePushStatus` already printed `printSunsetWarning('push-status', …)`. With `--wait-for-deployment` you can still get two warnings, labelled differently.

packages/reunite-integration/src/scorecard-classic/remote/fetch-scorecard.ts
```diff
@@ -0,0 +1,108 @@
...
+  isApiKey = false,
+}: FetchRemoteScorecardAndPluginsParams): Promise<RemoteScorecardAndPlugins> {
+  const parsedProjectUrl = parseProjectUrl(projectUrl);
+
+  if (!parsedProjectUrl) {
+    throw new HandledError(`Invalid project URL format: ${projectUrl}`);
+  }
+
+  try {
+    const project = await fetchProjectConfigBySlugs({ ...parsedProjectUrl, auth, isApiKey });
```
AlbinaBlazhko17: There are a bunch of missed verbose logs:
Project fetch response status, Authentication failed with status N, Fetching plugins from, Plugins fetch response status.
Could you please add all of them if possible? Because i built it to be able to step by step debug.

packages/reunite-integration/src/scorecard-classic/validation/plugin-evaluator.ts
```diff
@@ -0,0 +1,30 @@
...
+  basePath?: string
+): Promise<Plugin[]> {
+  if (!pluginsCode) {
+    return [];
+  }
+
+  try {
+    const dirname = pathToFileURL(basePath ?? process.cwd()).href + '/';
+    const pluginsCodeWithDirname = pluginsCode.replaceAll('__redocly_dirname', `"${dirname}"`);
+    const dataUri = `data:text/javascript;base64,${btoa(pluginsCodeWithDirname)}`;
```
AlbinaBlazhko17: Same missed verbose logs:
Starting plugin evaluation…, Successfully evaluated N plugins + the Plugin i: id list, and the ❌ Failed to evaluate plugins with stack trace.

packages/reunite-integration/src/scorecard-classic/remote/fetch-scorecard.ts
```diff
@@ -0,0 +1,108 @@
...
+    const scorecard = project.config.scorecardClassic || project.config.scorecard;
+
+    if (!scorecard) {
+      throw new Error('No scorecard configuration found.');
+    }
+
+    const pluginsUrl = project.config.pluginsUrl;
+    const plugins = pluginsUrl ? await fetchPlugins(pluginsUrl) : undefined;
```
AlbinaBlazhko17: The same miss of the verbose logs:
Validating level "X", Using N plugins for this level, Linting document against level rules…

packages/cli/src/utils/telemetry.ts
DmitryAnansky: We can't release snapshot for now, as there is a new package added.
Testing here with tgz https://github.com/Redocly/reunite-push-action/pull/141

packages/reunite-integration/src/auth/oauth-client.ts
DmitryAnansky: Only the innermost one needs it. The other three defaults are redundant, will update ...

packages/cli/src/utils/telemetry.ts
AlbinaBlazhko17: You are right

.changeset/sunset-warning-once.md
kanoru3101: Confirmed on the published tarball against a local Reunite mock that returns a `Sunset` header:

- `push-status --wait` on a main-branch push → **1** warning, so that half of the changeset holds.
- `push --wait-for-deployment` → **2** warnings: `"push"` from `handlePush` (`push.ts:99`, printed before the wait starts) and `"push-status"` from the `finally` in `handlePushStatus` (`push-status.ts:116`).

Suggest skipping `printSunsetWarning('push', …)` when `--wait-for-deployment` is set (or collecting both and printing once at the end), and adjusting the changeset wording until then.

kanoru3101 (approved):
LGTM

AlbinaBlazhko17 (approved):

Daryna-del (approved):

